PR #1, opened 2024-06-05
encapsulate pdf viewer styles

artem-harbour:
Check comment before merging:
https://linear.app/harbour/issue/HAR-6968/black-overlay-when-modal-appears-in-ae#comment-df499f53


harbournick (approved):
LGTM - great save here!

PR #7, opened 2024-06-18
HAR-6913 - super-editor: strikethrough and underline

mattConnHarbour:
Thanks for teeing this up :)
This branch comes from Artem's branch, so I'm using his as the trunk.

fyi: I had some issues with this line in `packages/super-editor/src/extensions/Comments/comments.js` :
`import { v4 as uuidv4 } from 'uuid';`
I think we might need to change it to `uuidv4`? But I see this is imported as `uuid` everywhere else so maybe it's just a problem for me?


mattConnHarbour:
using this: https://github.com/Harbour-Enterprises/superdoc/pull/8

PR #9, opened 2024-06-19
minimal core infra for existing functionality

artem-harbour:


artem-harbour:
@harbournick - ready to merge!

PR #10, opened 2024-06-20
extend Editor class and add basic structure

artem-harbour:


packages/super-editor/src/core/utilities/createStyleTag.js
```diff
@@ -0,0 +1,15 @@
+export function createStyleTag(style, suffix) {
+  const hrbrStyleTag = document.querySelector(`style[data-hrbr-style${suffix ? `-${suffix}` : ''}]`);
```
harbournick: Since this is a public / general component we shouldn't have any Harbour-specific style to look for here. Can we provide a way to pass in styles to be injected for any user?

packages/super-editor/src/core/Editor.js
```diff
@@ -66,65 +123,149 @@ export class Editor extends EventEmitter {
     this.view.updateState(this.state);
   }
 
+  setEditable(editable, emitUpdate = true) {
```
harbournick, line 126: Nice

packages/super-editor/src/core/utilities/createStyleTag.js
artem-harbour: This style is not specific to Harbour.

It's just a helper method that is used to add a minimum recommended style for prosemirror view.
https://github.com/ProseMirror/prosemirror-view/blob/master/style/prosemirror.css
https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/style.ts

The date attribute `data-hrbr-style` is used to identify style element. I can change it for example to something like `data-supereditor-style`.

The rest of the styles will be added in Vue components.

harbournick:
Got it- sounds good.

PR #12, opened 2024-06-24
HAR-7073 - Updates to superdoc PDF for handling extra wide documents

harbournick:


harbournick:
@edoversb would you mind doing a quick review of this one? This is already in the webapp package + canary, just needs a review + merge so main is up to date with that package.

PR #13, opened 2024-07-01
HAR-7163 - extend core with attribute support

artem-harbour:
Linear:
https://linear.app/harbour/issue/HAR-7163/core-editor-infrastructure

- Created `Attribute` class that is a space for methods of working with attributes.
- Moved related methods from `Schema` class.
- Added additional methods that add support for attributes in extensions.

### DIFF
--- a/packages/super-editor/src/core/Editor.js
+++ b/packages/super-editor/src/core/Editor.js
@@ -1,6 +1,5 @@
 import { EditorState } from 'prosemirror-state';
 import { EditorView } from 'prosemirror-view';
-import { history } from 'prosemirror-history';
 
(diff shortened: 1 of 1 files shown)


mattConnHarbour:
Cool, I was just working with attributes on text nodes

packages/super-editor/src/core/Attribute.js
```diff
@@ -0,0 +1,237 @@
+import { getExtensionConfigField } from './helpers/getExtensionConfigField.js';
+
+/**
+ * Attribute class is used to work with attributes. 
+ */
+export class Attribute {
```
harbournick, line 6: This is great!

PR #15, opened 2024-07-01
HAR-7163 - history extension, editor focus events

artem-harbour:
- Implemented "history" properly. Since everything is now an extension, to add "history" plugin we need to create an extension.
- Added focus events to editor with the core extension.


artem-harbour:
fyi @mattConnHarbour

PR #18, opened 2024-07-05
Restore general attributes key in node attrs

harbournick:
- Restore a generic 'attributes' key in any node in the schema 
- This key stores various docx related attributes like sectionProperties, paragraphProperties etc.

### DIFF
--- a/packages/super-editor/src/extensions/body/body.js
+++ b/packages/super-editor/src/extensions/body/body.js
@@ -11,7 +11,9 @@ export const Body = Node.create({
 
   addAttributes() {
     return {
-      attributes: {},
+      attributes: {
+        rendered: false,
+      },
     };
   },
 


harbournick:
@artem-harbour Since infra upgrades we lost the generic 'attributes' key. I've been using this one to pass through things like "sectionProperties" and "paragraphProperties", basically anything coming out of the docx, that way it is retained on export as well. 

LMK if you have any issues with this approach.

artem-harbour:
> @artem-harbour Since infra upgrades we lost the generic 'attributes' key. I've been using this one to pass through things like "sectionProperties" and "paragraphProperties", basically anything coming out of the docx, that way it is retained on export as well.
> 
> LMK if you have any issues with this approach.

Should this generic attribute be renderable?

If not, then add the following property:
```javascript
addAttributes() {
    return {
      attributes: {
        rendered: false,
      },
    };
},
```

harbournick:
> > @artem-harbour Since infra upgrades we lost the generic 'attributes' key. I've been using this one to pass through things like "sectionProperties" and "paragraphProperties", basically anything coming out of the docx, that way it is retained on export as well.
> > LMK if you have any issues with this approach.
> 
> Should this generic attribute be renderable?
> 
> If not, then add the following property:
> 
> ```js
> addAttributes() {
>     return {
>       attributes: {
>         rendered: false,
>       },
>     };
> },
> ```

SG - Added

PR #19, opened 2024-07-08
HAR-7232 - remove fontawesome styles from superdoc

artem-harbour:
Linear:
https://linear.app/harbour/issue/HAR-7232/special-characters-appearing-throughout-canary-dashboard

Replaced font icons with SVG.

Note:
The icon is not replaced in `CommentGroup.vue` component because I did not find its usage and could not verify it.


artem-harbour:
@harbournick, could you please test this? And if everything is ok, then update the version of superdoc in the webapp for main and canary too.

PR #27, opened 2024-07-30
Toolbar functionality - latest

mattConnHarbour:
Remaining functionality - indent left/right, search


packages/super-editor/src/blankDoc.js
```diff
@@ -0,0 +1 @@
...wlsRYsxsyGAVxdU5BWwIYBNJ4KQifaX4SSDa++NliLRXvmDlsFF16Ti6WS2yN3ApkeABtih3AedjllqmsK9tni6YAjIVRBtEL47wUkGOykMbmh03ggTEdg3gI3rvbAkx8aI9N7is6/dTd7CZ8tPOBnxVadu54UcvPzJYUcUFbkfDYwA7uhnBJPVqde2qL7O/C7k9b0Mr5IvlvOKvtO+vhhNPT4X/gEAAP//AwBQSwECLQAUAAYACAAAACEA36TSbFoBAAAgBQAAEwAAAAAAAAAAAAAAAAAAAAAAW0NvbnRlbnRfVHlwZXNdLnhtbFBLAQItABQABgAIAAAAIQAekRq37wAAAE4CAAALAAAAAAAAAAAAAAAAAJMDAABfcmVscy8ucmVsc1BLAQItABQABgAIAAAAIQDeTB1RyAIAAPkKAAARAAAAAAAAAAAAAAAAALMGAAB3b3JkL2RvY3VtZW50LnhtbFBLAQItABQABgAIAAAAIQDWZLNR9AAAADEDAAAcAAAAAAAAAAAAAAAAAKoJAAB3b3JkL19yZWxzL2RvY3VtZW50LnhtbC5yZWxzUEsBAi0AFAAGAAgAAAAhAJSpUobUBgAAxyAAABUAAAAAAAAAAAAAAAAA4AsAAHdvcmQvdGhlbWUvdGhlbWUxLnhtbFBLAQItABQABgAIAAAAIQCLd5YPwQMAAAALAAARAAAAAAAAAAAAAAAAAOcSAAB3b3JkL3NldHRpbmdzLnhtbFBLAQItABQABgAIAAAAIQAmte967A8AAFmmAAAPAAAAAAAAAAAAAAAAANcWAAB3b3JkL3N0eWxlcy54bWxQSwECLQAUAAYACAAAACEA7wopTk4BAAB+AwAAFAAAAAAAAAAAAAAAAADwJgAAd29yZC93ZWJTZXR0aW5ncy54bWxQSwECLQAUAAYACAAAACEAh5N0R+8BAAB4BgAAEgAAAAAAAAAAAAAAAABwKAAAd29yZC9mb250VGFibGUueG1sUEsBAi0AFAAGAAgAAAAhAMjNs0N0AQAA9wIAABEAAAAAAAAAAAAAAAAAjyoAAGRvY1Byb3BzL2NvcmUueG1sUEsBAi0AFAAGAAgAAAAhAFt9D+5rAQAAxQIAABAAAAAAAAAAAAAAAAAAOi0AAGRvY1Byb3BzL2FwcC54bWxQSwUGAAAAAAsACwDBAgAA2y8AAAAA"
```
harbournick: Do we need this? If for local testing let's remove it or let's find a better way to do it. We could place a blank docx inside the super-converter/tests/fixtures area and load from there, for instance

mattConnHarbour: It was quicker for testing and google docs starts with a blank page, so I'd like to keep it if you don't mind. I can move it to fixtures.

packages/super-editor/src/extensions/bullet-list/bullet-list.js
```diff
@@ -2,7 +2,7 @@ import { Node } from '@core/index.js';
 import { Attribute } from '@core/index.js';
 
 export const BulletList = Node.create({
-  name: 'bulletList',
+  name: 'bulletedList',
```
artem-harbour: I would be more careful with renaming node names since the scheme is based on this. Is there any good reason for this? If so, please make sure it is renamed in all places.

packages/super-editor/src/core/schema/DocxSchema.js
```diff
@@ -0,0 +1,236 @@
+import { Schema } from "prosemirror-model"
+
+const olDOM = ["ol", 0], ulDOM = ["ul", 0], liDOM = ["li", 0];
+
+/**
+ * Custom schema for docx files with prose mirror
+ * Reference: https://github.com/ProseMirror/prosemirror-schema-basic/blob/master/src/schema-basic.ts
+ */
+const DocxSchema = new Schema({
```
artem-harbour: This file is no longer used and should be removed to avoid confusion.

packages/super-editor/src/components/toolbar/ToolbarItem.js
```diff
@@ -0,0 +1,45 @@
+
+export default (options) => {
```
artem-harbour: Avoid default exports, named ones are always better. It's also always better to name functions as it makes debugging easier.

For factory functions usually - `createToolbarItem`.

packages/super-editor/src/components/toolbar/ToolbarItem.js
```diff
@@ -0,0 +1,45 @@
...
+    const types = ['button', 'options', 'separator'];
+    if (!types.includes(options.type)) {
+        console.error('Invalid toolbar item type:',options.type);
```
artem-harbour: `throw new Error` instead. Since it's obvious that incorrect use is unacceptable.

packages/super-editor/src/components/toolbar/ToolbarItem.js
```diff
@@ -0,0 +1,45 @@
...
+        return;
+    }
+    return {
+        type: options.type,
+        name: options.name,
+
+        // top-level style
+        style: options.style || null,
```
artem-harbour: I think this operator is better suited for checking null/undefined options.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

packages/super-editor/src/components/toolbar/Toolbar.vue
```diff
@@ -1,52 +1,259 @@
...
-})
+
 const toolbarItems = ref([
-  // font
-  toolbarItem({
-    type: 'dropdown',
-    name: 'fontFamily',
-    command: 'toggleFont',
-    dropdownOptions: [
+  zoom,
```
artem-harbour, line 552: It seems to me that all items should be created with `ToolbarItem` factory func in the toolbar component, and not in separate files. 

For example, so that we can customize toolbar items using  `itemOptions` props that need to be added to the toolbar.

```
const myItemsOptions = {
  alignmentOptions: {
    preCommand: () => {
      console.log("Here is my behavior for preCommand");
    },
  }
};

<Toolbar itemsOptions="myItemsOptions" />
```

Otherwise it looks like all callbacks and item options are hardcoded and there is no way to add your own behavior from the outside.

packages/super-editor/src/components/toolbar/Toolbar.vue
```diff
@@ -72,307 +279,200 @@ const toolbarItems = ref([
...
-  if (!item) return;
-  item.active = active;
+  if (command) {
+    console.log("Executing command", command, argument)
+    emit('command', {command, argument});
+  }
+
+  if (postCommand) {
+    console.log("Calling postcommand", item, argument)
+    postCommand(item, argument);
```
artem-harbour: This is another question that I would think about and maybe it even makes sense to redesign it (not sure yet). Since in essence there is no guarantee that `postCommand` will be executed after the full execution of the command. 

Maybe it makes sense to look at command chinaning or something like that (need to use it more to understand).

packages/super-editor/src/components/toolbar/items/alignment.js
```diff
@@ -0,0 +1,41 @@
+import ToolbarItem from "../ToolbarItem";
+
+const button = ToolbarItem({
+    type: 'button',
+    name: 'textAlign',
+    tooltip: "Alignment",
+    icon: "fa-align-left",
+    hasCaret: true,
+    preCommand(self) {
```
artem-harbour: This is the issue that I already flagged.

This looks more like a behavior for a button that should be there by default and hardcoded, and the functionality itself suggests that it can be customized using a callback (although it can't be done now).

packages/super-editor/src/components/toolbar/items/fontButton.js
```diff
@@ -0,0 +1,44 @@
...
+const fontButton = ToolbarItem({
+    type: 'button',
+    name,
+    tooltip: "Font",
+    overflowIcon: 'fa-font',
+    label: "Arial",
+    hasCaret: true,
+    isWide: true,
+    style: {width: '120px'},
+    preCommand(self) {
```
artem-harbour: And one more question. It seems to me that there should be an opportunity to get here in the options, for example, an editor instance and maybe some basic things for writing custom logic related to PM.

packages/super-editor/src/components/toolbar/Toolbar.vue
artem-harbour: This comment and the comments below are basically the most important ones.

fyi @mattConnHarbour @harbournick

packages/super-editor/src/extensions/bullet-list/bullet-list.js
mattConnHarbour: Changing it back

packages/super-editor/src/components/toolbar/ToolbarItem.js
mattConnHarbour: Later on I would like to turn this function into a proper class, as well.

packages/super-editor/src/components/toolbar/Toolbar.vue
mattConnHarbour: Since I haven't had a use for postCommand yet, I'll just take it out for now and use chaining in the next PR.

packages/super-editor/src/components/toolbar/items/fontButton.js
mattConnHarbour: good idea!

mattConnHarbour: I'm going to pass the active editor to the toolbar as a prop

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -436,7 +436,7 @@ class SuperConverter {
     }
 
     return {
-      type: overallListType || 'bulletList',
+      type: overallListType || 'bulletedList',
```
harbournick: Why is this being changed? Please revert since the name was reverted in the node definition

packages/super-editor/src/core/super-converter/numbering.js
```diff
@@ -79,7 +79,7 @@ function getNodeNumberingDefinition(attributes, level) {
 
   // Get style for this list level
   let listType;
-  if (unorderedListTypes.includes(listTypeDef.toLowerCase())) listType = 'bulletList';
+  if (unorderedListTypes.includes(listTypeDef.toLowerCase())) listType = 'bulletedList';
```
harbournick: Please revert

packages/super-editor/src/dev/components/DeveloperPlayground.vue
```diff
@@ -6,21 +6,25 @@
 -->
 
 <script setup>
-import { ref, reactive, computed } from 'vue';
+import { ref, reactive, watch, onMounted } from 'vue';
 import BasicUpload from './BasicUpload.vue';
+import blankDocBase64 from '../../blankDoc';
```
harbournick: Would still prefer this moved either to fixtures or to /src/dev/ 
It should not be in /src since its only for testing in the dev playground

mattConnHarbour: moving

PR #44, opened 2024-08-13
Add import to DOCX comments, render in editor

harbournick:


packages/super-editor/src/core/ExtensionService.js
```diff
@@ -134,6 +136,11 @@ export class ExtensionService {
       return plugins;
     }).flat();
 
+    const commentsPlugin = CommentsPlugin();
+    const decorationClick = DecorationClick(editor);
+    const customPlugins = [commentsPlugin, decorationClick];
+    allPlugins.push(...customPlugins);
```
artem-harbour: FYI - Plugins should not be created directly and imported/added to the service manually.

To add a new plugin - need to create a new extension inside `packages/super-editor/src/extensions` and use `addPmPlugins` method to add PM plugins.

After importing the extension in the corresponding `index.js` files, it will be automatically installed.

An example can be found in the core extension:
`packages/super-editor/src/core/extensions/editorFocus.js`

packages/super-editor/src/plugins/decorationClick/decoration-click.js
```diff
@@ -0,0 +1,18 @@
+import { Plugin } from 'prosemirror-state';
```
artem-harbour: Example:
`packages/super-editor/src/extensions`

```javascript
import { Plugin, PluginKey } from 'prosemirror-state';
import { Extension } from '../Extension.js';

export const DecorationClick = Extension.create({
  name: 'decorationClick',

  addPmPlugins() {
    const editor = this.editor; // available via bound context to the method

    const decorationClickPlugin = new Plugin({
      key: new PluginKey('decorationClick'),
      props: {
        handleClick(view, pos, event) {
          const target = event.target;
          const threadId = target.getAttribute('data-thread-id');
          if (target && threadId) {
            const conversation = editor.getComment(threadId);
            editor.emit('commentClick', { conversation });
          }
          return false;
        },
      },
    });

    return [decorationClickPlugin];
  },
});
```

packages/super-editor/src/core/ExtensionService.js
harbournick: Ah cool! Thank you - was going to ask about this at the deep dive but this makes sense 🙌🏽

PR #49, opened 2024-08-15
WIP: UI Tweaks, format painter 

harbournick:


packages/super-editor/src/extensions/format-commands/format-commands.js
```diff
@@ -27,6 +27,118 @@ export const FormatCommands = Extension.create({
...
+        }
+      
+        let start = from;
+        let end = to;
+        if (empty) state.doc.nodesBetween(start, end + 1, (node) => getAttrs(node))
+        else state.doc.nodesBetween(from, to, (node) => getAttrs(node));
+        return attributes;
+      },
+
+      copyFormat: () => ({ chain }) => {
```
harbournick: @artem-harbour  - I'm trying to create the format painter button here and it requires storing the active marks when a user first clicks the button, then if we have a stored state, then apply those marks. 

Any suggestions on how to accomplish the storage here? I put it in the main Editor.js for the moment so I could test it out but this is clearly the wrong place. Wanted to get your thoughts here.

artem-harbour: Try this inside extension:
```javascript
addStorage() {
  return {
    foo: 0,
    bar: 1,
    storedStyle: null,
  };
},
```

Then in another place of the extension (command, plugin)
```javascript
this.storage.foo += 1;
this.storage.storedStyle = 'my-style';
```

packages/super-editor/src/extensions/format-commands/format-commands.js
```diff
@@ -27,6 +27,118 @@ export const FormatCommands = Extension.create({
...
+        } else {
+          state.doc.nodesBetween(from, to, node => {
+            marks = marks.concat(node.marks);
+          });
+        }
+      
+        return marks;
+      },
+    
+      getActiveAttributes: () => ({ chain }) => {
```
artem-harbour: Also I think that methods like this (for example, get or find attributes, marks, etc.) are not suitable for commands, they should be helpers.
`packages/super-editor/src/core/helpers`

Commands should perform an editing action and always return `true` or `false` depending on whether the action was processed by the command or not.

https://prosemirror.net/docs/guide/#commands
"When a command isn't applicable, it should return false and do nothing. When it is, it should dispatch a transaction and return true."

packages/super-editor/src/extensions/format-commands/format-commands.js
harbournick: ah! amazing - thank you

PR #55, opened 2024-08-19
HAR-7413 - add annotations

artem-harbour:
Linear:
https://linear.app/harbour/issue/HAR-7413/superdoc-launch-interactive-annotations-in-demo-site


packages/super-editor/src/core/helpers/posToDOMRect.js
```diff
@@ -0,0 +1,33 @@
+import { minMax } from '../utilities/minMax.js'
+
+export function posToDOMRect(view, from, to) {
```
harbournick, line 3: nice

packages/super-editor/src/extensions/field-annotation/FieldAnnotationView.js
```diff
@@ -0,0 +1,173 @@
+import { Attribute } from '@core/index.js';
+
+export class FieldAnnotationView {
```
harbournick, line 3: Very cool. I'll need to spend some time really understanding this!

PR #67, opened 2024-08-26
Add 'editing' and 'viewing' document modes (no suggesting until tracked changes)

harbournick:


packages/super-editor/src/core/Editor.js
```diff
@@ -146,6 +158,38 @@ export class Editor extends EventEmitter {
     this.options.onFocus({ editor, event });
   }
 
+  /**
+   * Update the editor config with new plugins/extensions
+   * 
+   * @param {Object} options 
+   */
+  #updateEditor(options) {
```
harbournick: @artem-harbour This is what I figured out for updating the editor config (where we need to add/remove plugins from outside of the class) for toggling between 'viewing' (read only + no comments + eventually no tracked changes) and 'editing' (normal). Not sure if there's an easier way.

packages/super-editor/src/core/Editor.js
```diff
@@ -29,13 +29,20 @@ export class Editor extends EventEmitter {
 
   view;
 
-  isFocused = false;
+  id;
```
artem-harbour: What does this "id" mean?

packages/super-editor/src/core/Editor.js
```diff
@@ -146,6 +157,38 @@ export class Editor extends EventEmitter {
...
+    removedExtensions.forEach((e) => this.unregisterPlugin(e.name));
+
+    this.setOptions(options);
+    this.#setDocumentMode(options);
+    this.#createExtensionService();
+    this.#createCommandService();
+
+    addedExtensions.forEach((e) => {
+      const plugin = this.extensionService.plugins.find((p) => p.key.startsWith(e.name));
+      this.registerPlugin(plugin)
```
artem-harbour: It seems to me that in this case need to completely rely on the `extensionService` (since you initialize it again).

```javascript
const newState = this.state.reconfigure({
  plugins: this.extensionService.plugins,
});

this.view.updateState(newState);
```

packages/super-editor/src/core/Editor.js
artem-harbour: Is there a reason why you decided to completely remove the extension instead of just disabling the plugin outside with `unregisterPlugin` method?

packages/super-editor/src/core/Editor.js
```diff
@@ -146,6 +157,38 @@ export class Editor extends EventEmitter {
...
+
+    const addedExtensions = options.extensions.filter((extension) => {
+      return !this.options.extensions.some((e) => e.name === extension.name);
+    });
+  
+    removedExtensions.forEach((e) => this.unregisterPlugin(e.name));
+
+    this.setOptions(options);
+    this.#setDocumentMode(options);
+    this.#createExtensionService();
```
artem-harbour: I don't really like that there is a complete reinitialization of services here, but if we are completely removing extensions (and not just disabling plugins), then I think it is needed.

packages/superdoc/src/index.js
```diff
@@ -105,13 +118,62 @@ class Superdoc extends EventEmitter {
...
+    }
+  }
+
+  setDocumentMode(type) {
+    if (!type) return;
+
+    type = type.toLowerCase();
+    this.config.documentMode = type;
+
+    const types = {
```
artem-harbour, line 139: I like this pattern!

artem-harbour:
@harbournick, I've added my comments please check them out. Overall, I think it looks good.

packages/super-editor/src/core/Editor.js
harbournick: Yeah, for some reason the editor instance I was getting in superdoc was losing public methods like registerPlugin and unregisterPlugin - turns out I need to use a shallowRef to store it properly and now it works without all the extra nonsense! much much simpler

packages/super-editor/src/core/Editor.js
harbournick: Ended up removing all of this and only using registerPlugin and unregisterPlugin

packages/super-editor/src/core/Editor.js
harbournick: Removed this - it was just the document ID but clearly not needed - oops.

packages/super-editor/src/core/Editor.js
harbournick: removed!

packages/super-editor/src/index.js
```diff
@@ -1,13 +1,14 @@
+import './style.css';
```
artem-harbour, line 1: Don't think it's a good idea to import styles in `index.js` file. This file is mainly used for re-export things.

And importing styles here is essentially a side effect.

packages/super-editor/src/vue-3/Editor.js
```diff
@@ -72,4 +72,8 @@ export class Editor extends CoreEditor {
     super.unregisterPlugin(nameOrPluginKey);
     this.reactiveState.value = this.view.state;
   }
+
+  setDocumentMode(mode) {
```
artem-harbour, line 76: Is this method not inherited from the prototype? Or is it just for clarity?

packages/super-editor/src/index.js
harbournick: SG - fixing

PR #70, opened 2024-08-28
HAR-7516 - сorrect handling of packages

artem-harbour:
Since we work with packages inside SuperDoc, I think we need to manage them properly, or at least learn how to do it.

- Repalced `superdoc-monorepo` name to the correct one - `harbour-enterprises`.
- Changed `super-editor` name to `@harbour-enterprises/super-editor` for the package.
- Fixed imports from packages. 

Note - it's incorrect to use Vite shortcuts to point to another package.

- Fixed package dependencies.

Wrong:
`"super-editor": "file:../super-editor"`

Correct:
`"@harbour-enterprises/super-editor": "0.0.1-alpha.0"`

- Created `@harbour-enterprises/common` package inside `shared` workspace. Note - I think this package should be simple and contain only js files, since no bundler is used here.

Notes:
- We also need to convert the toolbar (`packages/super-toolbar`) into the correct package. It's not entirely clear here, because the toolbar essentially belongs to `@harbour-enterprises/super-editor` package. 

`packages/super-toolbar` basically imports the toolbar functionality from the package `@harbour-enterprises/super-editor` and then `@harbour-enterprises/super-editor` package imports the SuperToolbar back which is quite strange.

I think the correct solution would be to completely move the toolbar functionality into a separate package if possible.

- I also think it would be a good idea to review dependency management there.

FYI @harbournick, @mattConnHarbour


artem-harbour:
@harbournick, @mattConnHarbour - please check my notes about the toolbar.

PR #83, opened 2024-09-09
handle edit and view mode for field annotations

artem-harbour:
- Handled double click for field annotation.
- Added checks if the editor is not editable (to avoid emitting events and annotation selection).
- Minor refactoring, removing duplicate code.
- Documenting functions.

Note:
If the editor is not editable, these annotation commands will still work and ignore the flag. I'm not sure it's a good idea to add a check there (my opinion may change though).
`addFieldAnnotation`, `updateFieldAnnotations`, `deleteFieldAnnotations`

But it seems to me that the commands should be independent of these checks, and we should prevent the execution of the commands from the outside (in AE code for example).

I found that other commands don't rely on this flag either. Also, I found confirmation that this can be useful.


harbournick (approved):
LGTM

Let's discuss your findings on the editor editable mode, and interactions with fields etc. further. Sounds interesting.

PR #84, opened 2024-09-10
Standard preview modal

mattConnHarbour:
## Webapp companion PR
https://github.com/Harbour-Enterprises/Harbour-WebApp/pull/4469

### DIFF
--- a/packages/superdoc/package.json
+++ b/packages/superdoc/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@harbour-enterprises/superdoc",
   "type": "module",
-  "version": "1.0.0-alpha.53",
+  "version": "1.0.0-alpha.54",
   "files": [
     "dist"
   ],
--- a/packages/superdoc/src/Superdoc.vue
+++ b/packages/superdoc/src/Superdoc.vue
@@ -150,7 +150,7 @@ const onEditorSelectionChange = ({ editor, transaction }) => {
 
   handleSelectionChange(selection);
 
-  // TODO: Figure out why the selection is being undone here andwe need the delay
+  // TODO: Figure out why the selection is being undone here and we need the delay
   setTimeout(() => {
     const { activeThreadId } = transaction.getMeta('commentsPluginState') || {};
     const document = getDocument(documentId);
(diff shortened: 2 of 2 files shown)


packages/superdoc/src/components/HtmlViewer/HtmlViewer.vue
```diff
@@ -1,66 +1,58 @@
...
+    };
+    reader.onerror = (e) => reject(e);
+    reader.readAsText(fileSource);
+  });
 }
 
-onMounted(() => {
-    loadDocument(props.fileSource);
+onMounted(async () => {
+  try {
```
artem-harbour, line 41: It would be good to move all the code inside hooks into functions. Because over time a lot of code accumulates there that is not connected to each other in any way.

It also seems to me that in case of an error it would be good to emit an error event to the top and somehow handle it.

packages/superdoc/src/components/HtmlViewer/HtmlViewer.vue
```diff
@@ -1,66 +1,58 @@
...
-    });
-    console.log("found fields", fields.value)
+const emit = defineEmits([
+  'ready',
+  'selection-change'
+]);
+
+const handleSelectionChange = () => {
+  const selection = window.getSelection();
+  console.log('selection from html viewer', selection);
```
artem-harbour: These `console.log`s need to be deleted or converted to `console.debug` so as not to pollute the console.

packages/superdoc/src/components/HtmlViewer/HtmlViewer.vue
```diff
@@ -1,66 +1,58 @@
...
 
-onMounted(() => {
-    loadDocument(props.fileSource);
+onMounted(async () => {
+  try {
+    const documentHtml = await getDocumentHtml(props.fileSource);
+    documentContent.value = documentHtml;
+    emit('ready', props.documentId);
+  } catch (error) {
+    console.error('Error loading document', error);
```
harbournick, line 47: In addition to @artem-harbour's error handling comment, let's make sure that a sentry error is raised here if something does go wrong, we should know.

PR #87, opened 2024-09-10
Node type change

tg44:
ongoing PR for node type changes

### DIFF
--- a/packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
+++ b/packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
@@ -4,6 +4,8 @@ import {twipsToInches} from "../../helpers.js";
 import {tableNodeHandlerEntity} from "./tableImporter.js";
 import {drawingNodeHandlerEntity} from "./imageImporter.js";
 import {
+    handleParagraphNodeChange,
+    nodeTypeChangeAttributeChanger,
     trackChangeNodeHandlerEntity
 } from "./trackChangesImporter.js";
 import {hyperlinkNodeHandlerEntity} from "./hyperlinkImporter.js";
@@ -96,6 +98,8 @@ const createNodeListHandler = (nodeHandlers) => {
      */
     const nodeListHandlerFn = (elements, docx, insideTrackChange) => {
         if (!elements || !elements.length) return [];
+
+        const handlerObj = {handler: nodeListHandlerFn, handlerEntities: nodeHandlers}
         const processedElements = [];
 
         for (let index = 0; index < elements.length; index++) {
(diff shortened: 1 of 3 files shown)


packages/super-editor/src/extensions/track-changes/track-changes-base.js
```diff
@@ -290,6 +328,42 @@ const applyTrackChanges = (action, state, tr, from, to) => {
...
+                }
+            }
+            if (blockDelete) {
+                if (action === "accept") {
+                    tr.setNodeMarkup(pos + offset, null, {
+                        ...node.attrs,
+                        track: node.attrs.track.filter(track => track.type !== TrackDeleteMarkName)
+                    });
+                } else if (action === "revert") {
+                    //TODO: we should create all of the marks again
```
tg44, line 353: Here we should import the node recreation based of the marks.

packages/super-editor/src/core/super-converter/v2/importer/trackChangesImporter.js
```diff
@@ -0,0 +1,104 @@
...
+            newElements.push(changeProps)
+            const nodeWithChange = {
+                type: xmlNode.type,
+                name: xmlNode.name,
+                attributes: carbonCopy(xmlNode.attributes),
+                elements: newElements,
+            }
+            const res = nodeListHandler.handler([nodeWithChange], docx, false)
+            if(res.length > 0) {
+                //TODO: here we should diff the old node and a new node, and add the attributes
```
tg44, line 96: Here we have the new node (as the parsed), and the old node (as res), and we should find out how we could add the correct marks to the "new node".

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * @typedef {import()} XmlNode
+ * @typedef {{type: string, content: *, attrs: {}}} PmNodeJson
+ * @typedef {{type: string, attrs: {}}} PmMarkJson
+ *
+ * @typedef {(nodes: XmlNode[], docx: ParsedDocx, insideTrackCahange: boolean) => PmNodeJson[]} NodeListHandlerFn
+ * @typedef {{handler: NodeListHandlerFn, handlerEntities: NodeHandlerEntry[]}} NodeListHandler
+ *
+ * @typedef {(nodes: XmlNode[], docx: ParsedDocx, nodeListHandler: NodeListHandler, insideTrackCahange: boolean) => PmNodeJson[]} NodeHandler
```
tg44, line 28: This could be refactored to `node: XmlNode` bcs we removed the `n to n` mapping, and we did an `1 to n` which is way easier to think about. (This is really just refactoring the array to a single element.)

PR #88, opened 2024-09-12
HAR-7572 - update ordered list style when sink or lift list item

artem-harbour:
Linear:
https://linear.app/harbour/issue/HAR-7572/sub-lists-in-an-ordered-list-use-specific-rules-for-ordering-type

When sink list item:
Screenshot 2024-09-12 at 22 11 41

When lift list item:
Screenshot 2024-09-12 at 22 14 54


harbournick (approved):
Awesome!! Great work. LGTM

PR #95, opened 2024-09-20
HAR-7635 - add basic support for html/paragraph field

artem-harbour:
So we support the following field types:
- text - text fields, default type.
- image, signature - if there is a value in the `imageSrc` attribute, then renders the image instead of the `displayLabel`.
- checkbox - works exactly the same as the `text` type, but we assign `checkbox` type in case the behavior needs to be extended later.
- html - renders raw html  if there is a value in the `rawHtml`attribute or `displayLabel` instead (to support paragraph field).

Note:
- There are two more field types in the webapp that I'm not entirely sure about, but they seem to work just like `text` fields -`YESNOINPUT`, `DROPDOWNINPUT`.

If needed we can add separate field types (in `field-annotation`) for them in case we need to extend their behavior later or handle them differently. For example - `yesno` and `dropdown` types.


harbournick (approved):
LGTM

PR #101, opened 2024-09-26
HAR-7831 - placeholders for empty fields

mattConnHarbour:
companion to webapp PR: https://github.com/Harbour-Enterprises/Harbour-WebApp/pull/4657

Should also solve:
- HAR-7833 https://linear.app/harbour/issue/HAR-7833/as-a-signer-i-am-unable-to-see-my-fields-on-the-pdf-document-on-the
- HAR-7835 https://linear.app/harbour/issue/HAR-7835/superdoc-previews-pdf-sign-date-annotation-missing


packages/superdoc/src/components/HtmlViewer/HtmlViewer.vue
```diff
@@ -1,66 +1,63 @@
...
+    emit('ready', props.documentId);
+  } catch (error) {
+    emit('error', error);
+    console.error('Error loading document', error);
+  }
 }
 
-onMounted(() => {
-    loadDocument(props.fileSource);
+onMounted(async () => {
```
harbournick: why do we need async here if we aren't awaiting anything?

packages/superdoc/src/components/HtmlViewer/HtmlViewer.vue
```diff
@@ -1,66 +1,63 @@
 <script setup>
-import { defineProps, onMounted, ref } from 'vue';
+import { defineProps, defineEmits, onMounted, ref } from 'vue';
```
harbournick: You don't need to explicitly import defineProps or defineEmits - in fact they'll generate a warning. Let's remove them.

packages/superdoc/src/Superdoc.vue
```diff
@@ -434,7 +434,10 @@ const handlePdfClick = (e) => {
             v-if="doc.type === PDF"
             :document-data="doc"
             @selection-change="handleSelectionChange"
-            @ready="handlePdfReady" 
+            @selection-drag="handleSelectionDrag"
```
harbournick: Did you mean to re-add these emits here? I removed them yesterday. If not, can you remove? If yes, we should figure out why - the selection is now handled in Superdoc.vue now in the viewer.

harbournick (changes requested):
Couple minor tweaks

packages/superdoc/src/components/HtmlViewer/HtmlViewer.vue
mattConnHarbour: removing

harbournick (approved):
LGTM

PR #107, opened 2024-10-07
HAR-7592 Responsive toolbar

VladaHarbour:
Hi @artem-harbour @harbournick. I have one concern about data update process on window resize. Since toolbar items are SuperToolbar class property I'm not really sure how to properly make it reactive for Toolbar component. Here is the way I implemented it
Screenshot 2024-10-07 at 14 59 06
but for some reason I had to pass it as raw object to make everything downstream work
Screenshot 2024-10-07 at 14 59 39
Happy to hear your thoughts about best practices here

### DIFF
--- a/packages/super-editor/src/components/toolbar/Toolbar.vue
+++ b/packages/super-editor/src/components/toolbar/Toolbar.vue
@@ -1,39 +1,29 @@
 <script setup>
-import {ref, getCurrentInstance, onMounted, onDeactivated, toRaw, computed} from 'vue';
+import {ref, getCurrentInstance, onMounted, onDeactivated} from 'vue';
+import {throttle} from './helpers.js';
 import ButtonGroup from './ButtonGroup.vue';
 
 const { proxy } = getCurrentInstance();
 const emit = defineEmits(['command', 'toggle', 'select']);
 
-const leftItems = ref([]);
-const centerItems = ref([]);
-const rightItems = ref([]);
-const overflowItems = ref([]);
+let toolbarKey = ref(1);
 
-const showLeftSide = computed(() => proxy.$toolbar.config?.toolbarGroups?.includes('left'));
-const showRightSide = computed(() => proxy.$toolbar.config?.toolbarGroups?.includes('right'));
-
+const showLeftSide = proxy.$toolbar.config?.toolbarGroups?.includes('left');
+const showRightSide = proxy.$toolbar.config?.toolbarGroups?.includes('right');
 
 onMounted(() => {
-  window.addEventListener('resize', onWindowResized);
-  getToolbarItems();
+  window.addEventListener('resize', onResizeThrottled);
 });
 
 onDeactivated(() => {
-  window.removeEventListener('resize', onWindowResized);
+  window.removeEventListener('resize', onResizeThrottled);
 });
 
 const onWindowResized = async () => {
   await proxy.$toolbar.onToolbarResize();
-  getToolbarItems();
+  toolbarKey.value += 1;
 };
-
-const getToolbarItems = () => {
-  leftItems.value = proxy.$toolbar.getToolbarItemByGroup('left');
-  centerItems.value = proxy.$toolbar.getToolbarItemByGroup('center');
-  rightItems.value = proxy.$toolbar.getToolbarItemByGroup('right');
-  overflowItems.value = proxy.$toolbar.overflowItems;
-};
+const onResizeThrottled = throttle(onWindowResized, 300);
 
 const handleCommand = ({ item, argument }) => {
   proxy.$toolbar.emitCommand({ item, argument });
@@ -43,23 +33,24 @@ const handleCommand = ({ item, argument }) => {
 <template>
   <div 
       class="superdoc-toolbar"
+      :key="toolbarKey"
   >
     <ButtonGroup 
       v-if="showLeftSide"
-      :toolbar-items="toRaw(leftItems)" 
+      :toolbar-items="proxy.$toolbar.getToolbarItemByGroup('left')" 
       position="left" 
       @command="handleCommand" 
       class="superdoc-toolbar-group-side" 
     />
     <ButtonGroup 
-      :toolbar-items="toRaw(centerItems)" 
-      :overflow-items="toRaw(overflowItems)"
+      :toolbar-items="proxy.$toolbar.getToolbarItemByGroup('center')" 
+      :overflow-items="proxy.$toolbar.overflowItems"
       position="center" 
       @command="handleCommand" 
     />
     <ButtonGroup 
       v-if="showRightSide"
-      :toolbar-items="toRaw(rightItems)" 
+      :toolbar-items="proxy.$toolbar.getToolbarItemByGroup('right')" 
       position="right" 
       @command="handleCommand" 
       class="superdoc-toolbar-group-side"
@@ -79,7 +70,7 @@ const handleCommand = ({ item, argument }) => {
     min-width: auto !important;
   }
 }
-@media (max-width: 600px)  {
+@media (max-width: 768px)  {
   .superdoc-toolbar {
     padding: 4px 10px;
   }
--- a/packages/super-editor/src/components/toolbar/defaultItems.js
+++ b/packages/super-editor/src/components/toolbar/defaultItems.js
@@ -624,7 +624,7 @@ export const makeDefaultItems = (superToolbar, isDev = false, windowWidth) => {
 
   // Responsive toolbar calculations
   const itemsToHide = ['zoom', 'fontFamily', 'fontSize', 'redo'];
-  const mobileWidth = 600;
+  const hideWideItemsEndpoint = 600;
   const toolbarPadding = 32;
   const stickyItemsWidth = 120;
   
@@ -660,7 +660,7 @@ export const makeDefaultItems = (superToolbar, isDev = false, windowWidth) => {
   ];
   
   // Hide separators on small screens
-  if (windowWidth <= mobileWidth) {
+  if (windowWidth <= hideWideItemsEndpoint) {
     toolbarItems = toolbarItems.filter(item => item.type !== 'separator');
   }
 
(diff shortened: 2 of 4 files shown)


packages/super-editor/src/components/toolbar/defaultItems.js
```diff
@@ -1,69 +1,69 @@
-import { undoDepth, redoDepth } from "prosemirror-history";
-import { h, onDeactivated } from "vue";
+import { undoDepth, redoDepth } from 'prosemirror-history';
+import { h, onDeactivated } from 'vue';
```
artem-harbour, line 2: For the future - please do not use Prettier (or whatever you use) and exclude such changes from the PR, because we don't have standardized configs for formatting and lingintg here yet. That is, such changes must correspond to the repo configs (if any) or they should be avoided.

I also think that formatting fixing should be done by a separate PR, otherwise it makes it difficult to read and track current changes.

FYI @harbournick - we will need to create configs for this later.

packages/super-editor/src/components/toolbar/defaultItems.js
```diff
@@ -739,16 +611,26 @@ export const makeDefaultItems = (superToolbar, isDev = false) => {
...
+    ['documentMode', 45],
+    ['zoom', 70],
+    ['fontSize', 56],
+    ['fontFamily', 72],
+    ['default', 32]
+  ]);
+
+  // Responsive toolbar calculations
+  const itemsToHide = ['zoom', 'fontFamily', 'fontSize', 'redo'];
+  const mobileWidth = 600;
```
artem-harbour: It seems to me `768px` checkpoint is more standardized for mobile, although it depends on the requirements.

VladaHarbour: I defined it based on what our toolbar can fit. I can name this variable in different way just not to make it confusing

packages/super-editor/src/components/toolbar/Toolbar.vue
```diff
@@ -45,10 +71,18 @@ const handleCommand = ({ item, argument }) => {
...
+  padding: 4px 16px;
 }
-.superdoc-toolbar-group-side {
-  width: 200px;
+@media (max-width: 1120px) {
+  .superdoc-toolbar-group-side {
+    min-width: auto !important;
+  }
+}
+@media (max-width: 600px)  {
```
artem-harbour: `768px` checkpoint is more standardized for mobile.

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -104,9 +110,16 @@ export class SuperToolbar extends EventEmitter {
       } else {
         item.deactivate();
       }
-
     });
   }
+  
+  onToolbarResize = async () => {
+    const timeout = (ms) => {
```
artem-harbour: Remove this hack with a promised timeout.

Instead, create a throttled version of the function inside the component that you call on the resize event.

Take the implementation from here for example:
`Harbour-WebApp/src/utils/throttle.js`

Here you can find an example of usage:
`Harbour-WebApp/src/MainApp.vue`

Example
```javascript
const onResize = () => {};
const onResizeThrottled = throttle(onResize, 300);
window.addEventListener('resize', onResizeThrottled);
```

packages/super-editor/src/components/toolbar/Toolbar.vue
```diff
@@ -1,41 +1,67 @@
...
+  <div 
+      class="superdoc-toolbar"
+  >
+    <ButtonGroup 
       v-if="showLeftSide"
-      :toolbar-items="leftItems"
-      position="left"
-      @command="handleCommand"
-      class="superdoc-toolbar-group-side"
+      :toolbar-items="toRaw(leftItems)"
```
artem-harbour: I'm not sure why it needs to be wrapped up in `toRaw`, it needs to be explored further. But I think we need to avoid it here.

packages/super-editor/src/components/toolbar/Toolbar.vue
```diff
@@ -1,41 +1,67 @@
...
-const centerItems = proxy.$toolbar.toolbarItems.filter((item) => item.group.value === "center");
-const rightItems = proxy.$toolbar.toolbarItems.filter((item) => item.group.value === "right");
+const leftItems = ref([]);
+const centerItems = ref([]);
+const rightItems = ref([]);
+const overflowItems = ref([]);
 
-const showLeftSide = computed(() => proxy.$toolbar.config?.toolbarGroups?.includes("left"));
-const showRightSide = computed(() => proxy.$toolbar.config?.toolbarGroups?.includes("right"));
+const showLeftSide = computed(() => proxy.$toolbar.config?.toolbarGroups?.includes('left'));
```
artem-harbour: Do these computed properties work? I noticed that the `toolbarGroups` property is not reactive.

packages/super-editor/src/components/toolbar/Toolbar.vue
VladaHarbour: I tried to explain this issue in the PR overview. The problem was that originally everything which is downstream expected raw object since class prop is the raw object. The way we use ToolbarItem composable props in child components just didn't work with ref passed.
So I asked about your thoughts here https://github.com/Harbour-Enterprises/SuperDoc/pull/107#issue-2570243766

packages/super-editor/src/components/toolbar/Toolbar.vue
```diff
@@ -1,41 +1,67 @@
...
+import ButtonGroup from './ButtonGroup.vue';
 
 const { proxy } = getCurrentInstance();
-const emit = defineEmits(["command", "toggle", "select"]);
+const emit = defineEmits(['command', 'toggle', 'select']);
 
-const leftItems = proxy.$toolbar.toolbarItems.filter((item) => item.group.value === "left");
-const centerItems = proxy.$toolbar.toolbarItems.filter((item) => item.group.value === "center");
-const rightItems = proxy.$toolbar.toolbarItems.filter((item) => item.group.value === "right");
+const leftItems = ref([]);
```
artem-harbour: It seems to me that you can use `shallowref` here since you don't need reactivity for every property of the object, right?
https://vuejs.org/api/reactivity-advanced#shallowref

However, this depends on the further implementation of reactivity, I will leave a comment about this later.

packages/super-editor/src/components/toolbar/Toolbar.vue
VladaHarbour: Maybe you know the best way how to make class properties reactive

packages/super-editor/src/components/toolbar/Toolbar.vue
VladaHarbour: Every object is composable returned by useToolbarItem function. As I can see, almost every prop is a ref by itself. I have a feeling that it doesn't work right all together. There are toolbar item state mutations happen when User interacts with the item. Do you think this mutations will be reflected on the UI when using  shallowref
We can have a quick call and I will show you my conserns

packages/super-editor/src/components/toolbar/Toolbar.vue
VladaHarbour: It was added before my PR created so I will check

packages/super-editor/src/components/toolbar/Toolbar.vue
artem-harbour: Yeah, I understand the current issue with reactivity, since you need to update the state of components on resize. 

Basically, we could just use reactive properties inside the class, but I'd rather avoid that and not mix Core code with Vue reactivity.

Here are a few options that could potentially solve the reactivity issue:
1. Your implementation, but I think we should try to avoid using `toRaw` call here if possible. Because refs should be revealed automatically in templates.

And try maybe using `shallowref` instead `ref`. 
https://vuejs.org/api/reactivity-advanced#shallowref

2. `forceupdate` on resize event. But I would avoid this as it is not the best practice.
https://vuejs.org/api/component-instance.html#forceupdate

3. Use raw objects and don't wrap them in ref. And add a reactive key to the toolbar that you will update on resize. A pretty good option when you need to update the state of a component.

Example 1:
```javascript
let toolbarKey = ref(1);
const onResize = () => toolbarKey.value += 1;
window.addEventListener('resize', onResize);
```
`<div class="superdoc-toolbar" :key="toolbarKey">`

Example 2:
```javascript
let toolbarKey = ref(true);
const onResize = () => toolbarKey.value = !toolbarKey.value ;
window.addEventListener('resize', onResize);
```
`<div class="superdoc-toolbar" :key="toolbarKey">`

packages/super-editor/src/components/toolbar/Toolbar.vue
artem-harbour: Yes, I see it. Then I think we need to avoid changes that could potentially break other parts.

Please check my version with the key, I think it might be a good option in your case (trigger reactivity when changing key on resize).
https://github.com/Harbour-Enterprises/SuperDoc/pull/107#discussion_r1791777957

packages/super-editor/src/components/toolbar/Toolbar.vue
```diff
@@ -1,41 +1,58 @@
...
+
+onMounted(() => {
+  window.addEventListener('resize', onResizeThrottled);
+});
+
+onDeactivated(() => {
+  window.removeEventListener('resize', onResizeThrottled);
+});
+
+const onWindowResized = async () => {
```
artem-harbour, line 22: Remove `async` and `await` here.

packages/super-editor/src/components/toolbar/helpers.js
```diff
@@ -9,4 +9,44 @@ const sanitizeNumber = (value, defaultNumber) => {
     return sanitized;
 };
 
-export { sanitizeNumber };
\ No newline at end of file
+const throttle = (func, wait, options) => {
```
artem-harbour, line 12: Move this to `@harbour-enterprises/common` package and import from there.

PR #108, opened 2024-10-07
HAR-7814 - fix setFontSize command in toolbar

artem-harbour:
Linear:
https://linear.app/harbour/issue/HAR-7814/font-size-defaults-to-12-even-though-user-didnt-select-that-size


harbournick (approved):
LGTM - intercepted commands, much better name

PR #134, opened 2024-10-18
HAR-8016 - create styled list marker plugin

artem-harbour:
Linear:
https://linear.app/harbour/issue/HAR-8016/custom-list-markers-font-doesnt-match

**Merge this PR first** - https://github.com/Harbour-Enterprises/SuperDoc/pull/133

For each `listItem` we find all `textStyle` marks and apply the first found font-family and font-size (`listItem` may contain different font-size and font-family in the text) to the marker if defined there.

Note:
- Color is excluded from this list because it would lead to undesirable behavior.

Screenshot:
Screenshot 2024-10-18 at 14 46 03

### DIFF
--- a/packages/super-editor/src/core/config/style.js
+++ b/packages/super-editor/src/core/config/style.js
@@ -36,6 +36,19 @@ export const style = `.ProseMirror {
   content: '';
 }
 
+.ProseMirror li::marker {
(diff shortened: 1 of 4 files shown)


harbournick (approved):
LGTM

PR #136, opened 2024-10-22
HAR-8052 - add nostyle check for pdf annotation

artem-harbour:
Linear:
https://linear.app/harbour/issue/HAR-8052/viewing-as-next-signer-shows-first-signers-fields-highlighted-as


artem-harbour:
@harbournick - can you publish a new minor version of superdoc?

PR #141, opened 2024-10-25
HAR-7988 Tab plugin for general tab usage

VladaHarbour:
This plugin defines behaviour for general tab usage (pressing tab key). There are a couple of uncertainties like selecting tab nodes to support copy and paste and fonts inconsistency which may lead to incorrect tab width calculation comparing with Word

I'm on my way to add support for tab stops with defined width

File to test 
[tabs (3).docx](https://github.com/user-attachments/files/17521663/tabs.3.docx)


harbournick:
This is working really nicely! Nice work!
@artem-harbour I'll wait for your review before merging this one

packages/super-editor/src/core/commands/insertTabNode.js
```diff
@@ -0,0 +1,7 @@
+export const insertTabNode = () => ({ tr, state }) => {
+    let newPos = tr.mapping.map(tr.selection.from);
```
artem-harbour: Use just `tr.selection.from` instead.

packages/super-editor/src/extensions/index.js
```diff
@@ -47,6 +47,7 @@ import { Placeholder } from './placeholder/index.js';
 import { PopoverPlugin } from './popover-plugin/index.js';
 import {TrackChangesBase} from "./track-changes/track-changes-base.js";
 import {TrackMarks} from "./track-changes/track-marks.js";
+import { TabPlugin } from './tab-plugin/index.js';
```
artem-harbour: Use a more abstract extension name as it may contain more than just a plugin. Like `TabExtension`.

Although I think it makes sense to move this plugin to `tab.js` extension and add it there, since it is related functionality.

You can add `addPmPlugins` directly in this extension.
`packages/super-editor/src/extensions/tab/tab.js`

packages/super-editor/src/style.css
```diff
@@ -67,3 +67,14 @@
   margin: 0;
   padding: 0;
 } */
+
+.tab {
```
artem-harbour: These are global styles that can create issues for all elements on the page except the editor.

I would recommend using something like `.ProseMirror .tab` this and placing styles related to elements inside the editor here, although this approach will need to be standardized.
`packages/super-editor/src/core/config/style.js`

packages/super-editor/src/extensions/tab/tab.js
```diff
@@ -2,15 +2,26 @@ import { Node, Attribute } from '@core/index.js';
```
artem-harbour, line 4: Move tab plugin to this extension.

packages/super-editor/src/extensions/tab/tab.js
```diff
@@ -2,15 +2,26 @@ import { Node, Attribute } from '@core/index.js';
...
-
+  selectable: false,
+  atom: true,
+  
   addOptions() {
     return {
-      htmlAttributes: {},
+      htmlAttributes: {
+        class: 'tab',
+        contentEditable: false
```
artem-harbour, line 21: Unused properties.

packages/super-editor/src/extensions/tab/tab.js
```diff
@@ -20,7 +31,7 @@ export const TabNode = Node.create({
       tabSize: {
```
artem-harbour, line 36: Don't we need to update this attribute? Because decorations only add visual changes.

packages/super-editor/src/core/extensions/keymap.js
```diff
@@ -38,6 +38,7 @@ export const Keymap = Extension.create({
       'Delete': handleDelete,
       'Mod-Delete': handleDelete,
       'Mod-a': () => this.editor.commands.selectAll(),
+      'Tab': () => this.editor.commands.insertTabNode(),
```
artem-harbour, line 41: I would disallow this behavior inside lists (and maybe elsewhere), but the call is yours. @harbournick

packages/super-editor/src/extensions/tab-plugin/tab-plugin.js
```diff
@@ -0,0 +1,83 @@
...
+  if (domNode) {
+    const range = document.createRange();
+    range.selectNodeContents(domNode);
+    return range.getBoundingClientRect().width;
+  }
+  return 0;
+}
+
+function calcTabHeight(node) {
+  const parentNode = node?.parentNode;
```
artem-harbour: Try to avoid using `node?.` style of writing frequently because it will gloss over PM errors.

Instead, use properties from the PM and more explicit checks.

Example: isBlock
https://prosemirror.net/docs/ref/#model.Node.isBlock

packages/super-editor/src/extensions/tab-plugin/tab-plugin.js
```diff
@@ -0,0 +1,83 @@
...
+  if (!node || !parentNode || !node.classList?.contains('tab')) return '0px';
+  
+  // getting line-height to support multi-lined text
+  return window.getComputedStyle(node.parentElement, null).getPropertyValue('line-height');
```
artem-harbour: If there are any issues with the performance, `getComputedStyle` could be a potential reason. FYI @harbournick

packages/super-editor/src/extensions/tab/tab.js
VladaHarbour: I need this class to apply general styles. contentEditable was added intentionally after I faced this problem: when Node content is not defined Prosemirror adds separator after every custom tab inserted into the document. This separator makes it impossible to type any text after. So I use content: 'inline*' for Prosemirror to not treat this tab as empty block + contentEditable: false to make it impossible insert text into the tab node

packages/super-editor/src/core/extensions/keymap.js
VladaHarbour: Since lists have their own command defined it will have higher priority than this global one. The first tab you click will work as it is required for lists, the second tab will add tab node after the text in list item

artem-harbour: I mean the case when a tab doesn't work for list item, but the next tab from the main keymap works. Try pressing tab a couple of times.

packages/super-editor/src/extensions/tab/tab.js
artem-harbour: Ahh, I see. That's right.
`Attribute.mergeAttributes(this.options.htmlAttributes, htmlAttributes)`

packages/super-editor/src/extensions/tab/tab.js
VladaHarbour: Do you mean changing this attribute while adding decorations?

artem-harbour: The question is, do we need to change the `tabSize` attribute in the document itself? So that for example it works correctly when importing/exporting.

VladaHarbour: I was thinking about this one and decided to leave it for now like this. This attribute may be helpful for tab stops implementation which I'm working on right now. But in general, we have 2 types of tab definitions in docx: general tab tag which doesn't have size attribute and tab stops defined for the whole paragraph. Tab stops values have to be defined when exporting and general tab will look like this _<w:tab/>_, in contrast.

VladaHarbour:
@artem-harbour I pushed fixes for your comments and changed tab height calculation in order to get rid of getComputedStyle. I'm working on the different tab case right now(tab stops) and will create another PR for this
cc @harbournick

packages/super-editor/src/extensions/tab/tab.js
```diff
@@ -1,16 +1,32 @@
...
+        class: 'tab',
+        // this works together with content prop: 
+        // since tab can't have content inside but content prop is defined I have to manually add attribute
+        contentEditable: false
+      },
     };
   },
 
+  parseDOM() {
+    return [{tag: 'span.tab'}];
```
artem-harbour, line 27: You will probably need to add a priority here so that these spans are not parsed as the `textStyle` marks on copy/paste (but it's better to check this first).

Example from `field-annotations.js`
```javascript
parseDOM() {
    return [{ 
      tag: `span.${this.options.annotationClass}`,
      priority: 60,
    }];
  },
  ```

PR #147, opened 2024-10-30
SuperDoc: Collaboration updates

harbournick:
- Adds all collaboration infra to superdoc
- Tweaks to editor.js for collaboration service use (in node)


packages/super-editor/src/style.css
```diff
@@ -63,7 +63,33 @@
   border: 1px solid #1355FF !important;
 }
 
-/* li,ol,ul {
-  margin: 0;
-  padding: 0;
-} */
+/* Collaboration cursors */
```
artem-harbour: I would probably prefer to keep these styles here.
`packages/super-editor/src/core/config/style.js`

But our styles are not very standardized and require some revision in any case.

packages/super-editor/src/extensions/document-attributes/document-attributes.js
```diff
@@ -0,0 +1,43 @@
+import { Node, Attribute } from '@core/index.js';
+
+export const DocumentAttributes = Node.create({
```
artem-harbour: What is this node used for?

packages/super-editor/src/components/SuperEditor.vue
```diff
@@ -36,34 +37,73 @@ const props = defineProps({
...
+  const metaMap = ydoc.getMap('meta');
 
-  const [content, media] = await Editor.loadXmlData(props.fileSource);
+  const checkData = () => {
+    const docx = metaMap.get('docx');
+    if (docx) {
+      initEditor(docx);
+    } else if (retries > 0) {
+      console.debug(`Waiting for 'docx' data... retries left: ${retries}`);
+      setTimeout(checkData, interval); // Retry after the interval
```
artem-harbour: Should we call `clearTimeout` on component unmount?

harbournick: Ah yes, good call. Adding that now.

packages/super-editor/src/extensions/document-attributes/document-attributes.js
harbournick: Actually I'll remove it - ended up using default values for this instead. 
The reason I added initially was that the 'document' node is a special case and can't have attrs associated with it (or at least, they are not retained in collaboration), so we were losing the <w:document> tags' attrs. I figured I'd sync it with its own hidden node, but using default values here seems to work fine instead and its simpler.

PR #153, opened 2024-11-04
HAR-7987 - track changes

artem-harbour:
**BEFORE MERGE:**
- Let me know what to do with the test buttons in the toolbar.
- I think we should discuss document normalization first.

List of changes:
- Track changes has been completely redesigned from scratch.
`packages/super-editor/src/extensions/track-changes/`
- The previous implementation has been moved to the `track-changes-deprecated` path for reference only and will be removed over time.
`packages/super-editor/src/extensions/track-changes-deprecated/`
- Fixed implementation of track changes marks.
- Fixed commands for accepting and rejecting tracked changes by selection (Also can be used ID).
- Fixed ID assignment when adding or deleting text. Using `findTrackedMarkBetween` function.
`packages/super-editor/src/extensions/track-changes/trackChangesHelpers/findTrackedMarkBetween.js`
- Changed approach with `amendTransaction` function and removal of tracked changes if plugin is disabled. This will lead to various complications and makes no sense. Instead we use the `trackedTransaction` function if the plugin is enabled or simply return `tr` (transaction) without any changes.
- Implemented track changes for formatting.
- Other issues have been fixed, such as when we remove tracked text inside an added.

Notes:
- Track changes need to be connected to comments.
- There is a need to discuss the issue of document normalization so that it works well for users' documents.
- Currently Chrome has some issue in the latest update that track changes does not always work as expected, in other browsers everything is fine (**looks like I found a workaround for this**).

### DIFF
--- a/packages/super-editor/src/core/config/style.js
+++ b/packages/super-editor/src/core/config/style.js
@@ -123,6 +123,8 @@ img.ProseMirror-separator {
   border-top: 1px dashed #CB0E47;
   border-bottom: 1px dashed #CB0E47;
   background-color: #CB0E4722;
+  text-decoration: line-through;
+  text-decoration-thickness: 2px;
 }
 
 .ProseMirror .track-format-dec.highlighted {


packages/super-editor/src/core/helpers/createDocument.js
```diff
@@ -6,8 +8,18 @@
  */
 export function createDocument(converter, schema) {
   const documentData = converter.getSchema();
-  if (documentData) {
-    return schema.nodeFromJSON(documentData);
+
+  // Experimental.
+  let normalizedData;
+  try {
+    normalizedData = normalizeDocument(documentData);
```
artem-harbour, line 15: This is an item for discussion before the merge.

packages/super-editor/src/core/Editor.js
```diff
@@ -519,13 +520,23 @@ export class Editor extends EventEmitter {
 
     let state;
     try {
-      const trackedTr = amendTransaction(transaction, this.view, this.options.user);
-      const { state: newState } = this.view.state.applyTransaction(trackedTr);
+      const trackChangesState = TrackChangesBasePluginKey.getState(this.view.state);
+      const isTrackChangesActive = trackChangesState?.isTrackChangesActive ?? false;
+
+      const tr = isTrackChangesActive
```
harbournick, line 526: much better!

packages/super-editor/src/core/helpers/createDocument.js
harbournick: leaving it in for now, until import refactor

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/documentHelpers.js
```diff
@@ -0,0 +1,91 @@
...
+
+  if (toArr) {
+    return marksFound;
+  }
+
+  return markFound;
+};
+
+// https://discuss.prosemirror.net/t/expanding-the-selection-to-the-active-mark/
+export const findMarkPosition = (doc, pos, markName) => {
```
harbournick, line 34: ah, fascinating

PR #164, opened 2024-11-17
Pagination

harbournick:
Adds (optional) pagination v1 to SuperEditor

### DIFF
--- a/packages/superdoc/package.json
+++ b/packages/superdoc/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@harbour-enterprises/superdoc",
   "type": "module",
-  "version": "0.1.17",
+  "version": "0.2.0",
   "license": "AGPL-3.0",
   "readme": "../../README.md",
   "files": [


packages/super-editor/src/core/config/style.js
```diff
@@ -101,6 +101,16 @@ img.ProseMirror-separator {
   margin-bottom: 1.5px;
 }
 
+.ProseMirror {
+  width: 100%;
+  outline: none;
+  bordeR: none;
```
artem-harbour, line 107: `border`

packages/super-editor/src/extensions/pagination/pagination.js
```diff
@@ -0,0 +1,363 @@
...
+    let shouldUpdate = false;
+
+    // Track wether the first load has occured or not
+    let hasInitialized = false;
+
+    const paginationPlugin = new Plugin({
+      key: paginationPluginKey,
+      state: {
+        init() {
+          return DecorationSet.empty;
```
artem-harbour, line 37: Probably here it is necessary to have prepared decorations right away.

packages/super-editor/src/extensions/pagination/pagination.js
```diff
@@ -0,0 +1,363 @@
...
+    footerData,
+    location,
+    sectionHeight,
+    suppressFooter,
+    suppressHeader,
+    editor,
+  };
+
+  const header = createApp({
+    render: () => h(PageBreak, { data, coords, sectionHeight, styles, isDebugging })
```
artem-harbour, line 325: I would not recommend using this method for many reasons.

- For each such case, Vue app will be created, which can be very expensive (including decorations, all recalculations and so on).
- Vue render functions are less optimized than regular components (and are more suitable for small UI).
- Dependency on Vue in PM editor.

In the case of PM it is always preferable to use NodeViews, decorations, native work with DOM.

harbournick:
Closing this one as making major changes to the approach to remove createApp etc

PR #170, opened 2024-11-27
HAR-8229 - normalize doc text nodes

artem-harbour:
- Moved text node normalization to import level.
- Processed text nodes in paragraphs and lists.


packages/super-editor/src/core/super-converter/v2/importer/mergeTextNodes.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  return true;
+};
+
+const areAttrsEqual = (attrsA = {}, attrsB = {}) => {
+  return objectIncludes(attrsA, attrsB);
+};
+
+// const compare = (obj1, obj2) => {
```
harbournick: Should we remove the commented code?

artem-harbour: Done.

PR #171, opened 2024-11-29
HAR-8253 - handle image uploading

artem-harbour:
- Handled uploading of new images.
- New images are uploaded as base64 by default (I think this makes sense instead of storing them in memory), but handler can be replaced through the config.
- Image resize if exceeding the max-width.
- Export for new added images.

Screenshot 2024-12-02 at 14 31 32

Screenshot 2024-12-02 at 14 31 57

### DIFF
--- a/packages/super-editor/src/extensions/image/imageHelpers/getFileOpener.js
+++ b/packages/super-editor/src/extensions/image/imageHelpers/getFileOpener.js
@@ -1,0 +1,32 @@
+const ACCEPT_IMAGE_TYPES = [
+  '.jpg',
+  '.jpeg',
+  '.png',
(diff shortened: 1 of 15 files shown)


packages/super-editor/src/extensions/image/imageHelpers/getFileOpener.js
```diff
@@ -0,0 +1,32 @@
+const ACCEPT_IMAGE_TYPES = [
+  '.jpg',
+  '.jpeg',
+  '.png',
+  'image/jpeg',
+  'image/png',
+];
+
+export const getFileOpener = () => {
```
harbournick, line 9: nice!

PR #182, opened 2024-12-04
add meta attr to field-annotation

artem-harbour:
A meta attribute that allows storing additional information regarding a field/annotation.


artem-harbour:
Paused on this for now.

PR #183, opened 2024-12-04
HAR-8248 Superdoc multiple image upload support

VladaHarbour:


packages/super-editor/src/extensions/field-annotation/FieldAnnotationPlugin.js
```diff
@@ -50,6 +51,33 @@ export const FieldAnnotationPlugin = (options = {}) => {
...
+            content,
+            editor,
+          });
+        }
+        return false;
+      },
+
+      transformPasted(slice) {
+        const addMultipleAttributeForImageField = (node) => {
+          if (node.attrs.fieldType === 'IMAGEINPUT') {
```
artem-harbour: `node.attrs.type === 'image'` - use this instead. `IMAGEINPUT ` is an internal name in the webapp that has no place here.

packages/super-editor/src/extensions/field-annotation/FieldAnnotationPlugin.js
```diff
@@ -50,6 +51,33 @@ export const FieldAnnotationPlugin = (options = {}) => {
...
+                {
+                  ...node.attrs,
+                  multiple: true,
+                },
+                node.content
+            );
+          }
+          return node.copy(node.content);
+        }
+        return mapSlice(slice, addMultipleAttributeForImageField);
```
artem-harbour: I'm not sure what exactly is going on here without looking into it, but please make sure you've handled/tested cases where not only the field-annotation node is copied and pasted, but for example text containing one or more annotations.

packages/super-editor/src/extensions/field-annotation/FieldAnnotationPlugin.js
```diff
@@ -101,3 +129,29 @@ function handleDropOutside({
...
 }
+
+/**
+ * Helpers to transform pasted node
+ * Used to modify multiple attribute for image annotations
+ * 
+ * https://discuss.prosemirror.net/t/modify-specific-node-on-copy-and-paste-in-clipboard/4901
+ */
+
+function mapFragment(fragment, callback) {
```
artem-harbour: Move to helpers folder.

packages/super-editor/src/extensions/field-annotation/FieldAnnotationPlugin.js
```diff
@@ -101,3 +129,29 @@ function handleDropOutside({
...
+        mapFragment(node.content, callback)
+      );
+    }
+
+    return callback(node);
+  })
+);
+}
+
+function mapSlice(slice, callback) {
```
artem-harbour: Move to helpers folder.

packages/super-editor/src/extensions/field-annotation/field-annotation.js
```diff
@@ -178,6 +178,11 @@ export const FieldAnnotation = Node.create({
         default: true,
         rendered: false,
       },
+      
+      multiple: {
```
artem-harbour: Maybe `multipleImage` or something more clear? Are you sure we don't need to render this attr otherwise it won't be accepted when we copy/paste?

@harbournick - Do we need to handle new attributes for import/export to work in the webapp agreements"?

packages/super-editor/src/extensions/field-annotation/field-annotation.js
```diff
@@ -402,6 +429,45 @@ export const FieldAnnotation = Node.create({
...
       },
 
+      /**
+       * Delete a portion of annotations associated with a field.
+       * @param fieldIdOrArray The field ID or array of field IDs.
+       * @param end index at which to end extraction
+       * @example
+       * editor.commands.deleteFieldAnnotations('123')
+       * @example
+       * editor.commands.deleteFieldAnnotations(['123', '456'])
```
artem-harbour: Update examples.

packages/super-editor/src/extensions/field-annotation/FieldAnnotationPlugin.js
VladaHarbour: I decided to get rid of this overcomplicated logic since it doesn't really create a lot of sense for clients

packages/super-editor/src/extensions/field-annotation/field-annotation.js
VladaHarbour: @artem-harbour @harbournick added attribute handling in import/export

packages/super-editor/src/core/super-converter/importer.js
```diff
@@ -1204,13 +1204,15 @@ export class DocxImporter {
     const fieldType = sdtPr?.elements.find((el) => el.name === 'w:fieldType')?.attributes['w:val'];
     const type = sdtPr?.elements.find((el) => el.name === 'w:fieldTypeShort')?.attributes['w:val'];
     const fieldColor = sdtPr?.elements.find((el) => el.name === 'w:fieldColor')?.attributes['w:val'];
+    const isMultipleImage = sdtPr?.elements.find((el) => el.name === 'w:fieldMultipleImage')?.attributes['w:val'];
```
artem-harbour: Is there no need to parse the value from a string here?

PR #187, opened 2024-12-06
HAR-8201 - exposing core editor for paragraph editing

mattConnHarbour:


packages/superdoc/src/core/Superdoc.js
```diff
@@ -300,6 +300,18 @@ export class Superdoc extends EventEmitter {
...
+  getHTML() {
+    const editors = []
+    this.superdocStore.documents.forEach((doc) => {
+      const editor = doc.getEditor();
+      if (editor) {
+        editors.push(editor);
+      }
+    });
+
+    return editors.map((editor) => editor.getHTML()).join('\n');
```
harbournick: Let's return an array of HTML data here rather than a long string. Reason being we could have a mix of document types (docx, pdf, docx) in which case the user needs to know which document which HTML belongs to.

mattConnHarbour: done

PR #190, opened 2024-12-10
for paragraph editing: allowing filtering of toolbar items

mattConnHarbour:


packages/superdoc/src/index.js
```diff
@@ -15,8 +15,6 @@ export {
   BlankDOCX,
   getFileObject,
   compareVersions,
-  Editor,
```
harbournick: Let's keep these in here for now, they might be used somewhere

PR #191, opened 2024-12-11
fix editor initialization

artem-harbour:
These callbacks override those that are in the `props.options` and break a lot of functionality (for example, drag&drop in AE).

Here need to either subscribe to the events or combine callbacks.


artem-harbour:
FYI @harbournick

PR #192, opened 2024-12-12
Fix collaboration images, properly detroy ydoc on superdoc destroy

harbournick:
- Place document images into ydoc for collaboration sharing
- Export document images along with added media
- Destroy ydoc and provider on superdoc destroy
- Images will now work going from full screen modal to AE, or between multiple clients

### DIFF
--- a/packages/super-editor/src/core/Editor.js
+++ b/packages/super-editor/src/core/Editor.js
@@ -15,7 +15,7 @@ import { initComments } from '@features/index.js';
 import { style } from './config/style.js';
 import { trackedTransaction } from "@extensions/track-changes/trackChangesHelpers/trackedTransaction.js";
 import { TrackChangesBasePluginKey } from '@extensions/track-changes/plugins/index.js';
-import { initBlobs } from '@core/utilities/imageBlobs.js';
+import { getMediaObjectUrls } from '@core/utilities/imageBlobs.js';
 import DocxZipper from '@core/DocxZipper.js';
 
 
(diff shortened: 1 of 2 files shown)


harbournick:
@andrii-harbour fyi this will fix the missing images issue

packages/super-editor/src/core/Editor.js
```diff
@@ -733,11 +747,16 @@ export class Editor extends EventEmitter {
...
+      json,
+      this.schema,
+      this.storage.image.media,
+      isFinalDoc
+    );
     const relsData = this.converter.convertedXml['word/_rels/document.xml.rels'];
     const rels = this.converter.schemaToXml(relsData.elements[0]);
     const media = this.converter.addedMedia;
-
+g
```
artem-harbour: What is this mystical symbol?) Please remove.

packages/super-editor/src/core/utilities/imageBlobs.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Convert object of Uint8Array to blob URL
+ * @param {Object} media Object where keys are docx file names and values are Uint8Arrays
+ * @returns {Object} Object where keys are docx file names and values are blob URLs
+ */
+export const initBlobs = (media) => {
```
artem-harbour: The name of the function is quite general, but the function is very specific to the media.

Maybe change the name to something like - `getMediaObjectUrls` or `getMediaBlobUrls`?

PR #200, opened 2024-12-13
Polimorphic lists issue

VladaHarbour:


harbournick:
@VladaHarbour after testing some of our more complex custom lists, I found that our tab approach unfortunately breaks down. However, I took the beginning of your approach and made a different PR which appears to have solved it a different way. Closing this one. Thank you for all the research and work on this one!

PR #204, opened 2024-12-16
HAR-7992 Table fixes + highlight mark

VladaHarbour:


harbournick:
Nice to have 'highlight' now - we'll be able to add text background color tool now too!

PR #212, opened 2024-12-23
feat(ai): add phase 1 of local browser model writer and rewriter

palmer-cl:
# Summary
- Adds the ability for a user to use an AI button to generate text or update text

# Developer Summary
- Adds AI button to the toolbar
- Adds custom popout styled menu when button is hit
- Implements [writer and rewriter](https://github.com/webmachinelearning/writing-assistance-apis/blob/main/README.md) apis -- to test make sure you have required experimental flags turned on

image

- Adds AI layer component to handle highlighting (similar to comments layer)
- When suggesting, generated text are inserted as tracked changes

# Concerns
- The models are very unreliable. Ive experienced about 25% failure rate. While we handle the errors, its not the best UX. The same exact prompt will fail, and on retry, succeed. We should weigh the bonuses of a lightweight model being able to work offline, with the expectation users have of simple AI actions like generating and re-writing text. This is cost of admission now for all major models and it doesn't seem like this functionality can do it reliably in its current form. Ill attach a demo video below.

# Demo

https://github.com/user-attachments/assets/89565f1c-b2cd-438f-b6ed-d39abc370916


palmer-cl:
I added additional functionality to call out. This feature now optionally accepts an OpenAI key, and uses `gpt-4o-mini` when the key is present. Im attaching another demo here. You can see the performance is better in both speed and accuracy. 

Demo:

https://github.com/user-attachments/assets/0e27726f-457f-49d4-bc28-fd3903a95d69

palmer-cl:
Closing this in favor of: https://github.com/Harbour-Enterprises/SuperDoc/pull/444

PR #215, opened 2025-01-02
HAR-8546 Telemetry service example

VladaHarbour:
Hi @harbournick! Could you please take a look on this example implementation for unknown marks? Just wonder if you had different ideas


caio-pizzol:
@harbournick This is an initial proposal, but we still need to identify what data we need to collect to properly fix any issues around XML Elements/Tags later.

I've also proposed using DSN (similar to Sentry), but we might have to have two separate authorizations/identifications since this will be an Open-Source solution. One only to authenticate in our WebService (that will collect data and store it in GCS) and another to monitor usage more explicitly. Worth having a chat with @edoversb

harbournick:
hey! yeah this makes sense to me. The only thing I would say is that since
we also want to know the frequency of *all* tags, is it better to collect
all tags throughout the code and then calculate the 'missing' tags after
the fact. Same with attributes. But either way, as long we know tag and
attribute frequency and what our importer is missing, we can prioritize
adding new tags strategically.

PR #219, opened 2025-01-09
HAR-8702: superdoc toolbar - auto focus link input

mattConnHarbour:


packages/super-editor/src/components/toolbar/LinkInput.vue
```diff
@@ -65,12 +65,24 @@ const isAnchor = computed(() => props.href.startsWith('#'));
...
 );
+
+const focusInput = () => {
+  const input = document.querySelector('.link-input-ctn input');
+  if (!input) return;
+  input.focus();
+};
+
+onMounted(() => {
+  if (!props.showInput) return;
```
harbournick: Should this check happen inside focusInput() instead? Since onMounted might have other unrelated code

harbournick: or, if (props.showInput) focusInput() perhaps if you prefer it outside the function

mattConnHarbour:
Done!

PR #222, opened 2025-01-09
Import marks for annotations when re-importing a doc

harbournick:


packages/super-editor/src/core/super-converter/v2/importer/annotationImporter.js
```diff
@@ -32,10 +33,20 @@ export const handleAnnotationNode = (nodes, docx, nodeListHandler, insideTrackCh
     multipleImage: isMultipleImage === 'true',
   };
 
+  const annotationMarks = [];
+  const innerContent = node.elements.find((el) => el.name === 'w:sdtContent');
```
artem-harbour, line 37: I looked at this a bit earlier and it seems to me that it needs recursive processing (for all internal .elements properties) to collect all marks. You can export a doc that contains annotation with some marks and check the import step if it works as expected.

Note:
Also, we may not need annotation marks at this point to add y-prosemirror support. So I'll block this for now.

harbournick:
sounds good!

harbournick:
converting back to draft as we will likely close this one soon

harbournick:
No longer relevant

PR #223, opened 2025-01-09
HAR-8754 - fieldAnnotation formatting marks as attrs

artem-harbour:
Handled formatting marks as attrs for the `fieldAnnotation` node.

List of marks:
- bold
- italic
- underline
- textStyle (font-family, font-size)

Handled the conversion of marks to attributes when adding a new annotation so that the new annotation has styles of the context.

Using `appendTransaction` I remove all the marks from fieldAnnotations as they cause issues when there are both marks and attributes.

Note:
- At this point I think this solution should be temporary and when `y-prosemirror` fixes the issue we should go back to using marks on `fieldAnnotation`.


packages/super-editor/src/extensions/field-annotation/field-annotation-marks-impl.js
```diff
@@ -0,0 +1,639 @@
+import { Node, Attribute } from '@core/index.js';
```
artem-harbour, line 1: This is the original file that I want to keep for reference, as I'm pretty sure we'll be getting back to working with marks later.

PR #237, opened 2025-01-14
feat: add code examples

caio-pizzol:
**Problem**:  There were no examples showing how to integrate Superdoc with popular frontend frameworks like React, Vue, and Vanilla JS. This made it difficult for developers to quickly get started using the library in their projects.

**Solution**: This PR adds new example projects for React, Vue, and Vanilla JS, demonstrating how to initialize Superdoc, load documents, and handle basic events.  These examples provide clear and concise starting points for developers using these frameworks.

**Changes**:
- Added a React example application (`examples/react-example`).
- Added a Vue example application (`examples/vue-example`).
- Added a Vanilla JS example application (`examples/vanilla-example`).
- Included basic styling for each example.
- Added file input functionality to load documents in each example.

**Testing**:  Each example application was manually tested by running the development server (`npm run dev`) and verifying the Superdoc editor initialized correctly and loaded documents as expected.

**Notes**: These examples provide a basic integration and can be further extended to showcase more advanced Superdoc features. Consider adding more complex examples demonstrating collaborative editing, custom toolbar configurations, and other advanced use cases in future PRs.

### DIFF
--- a/examples/react-example/src/components/DocumentEditor.jsx
+++ b/examples/react-example/src/components/DocumentEditor.jsx
@@ -1,0 +1,71 @@
+import { useEffect, useRef } from 'react';
+import { Superdoc } from '@harbour-enterprises/superdoc';
+import '@harbour-enterprises/superdoc/style.css';
+
+const DocumentEditor = ({ 
+  documentId, 
+  documentType = 'docx', 
+  initialData = null,
+  readOnly = false,
+  onEditorReady 
+}) => {
+  const editorRef = useRef(null);
+
+  useEffect(() => {
+    const editor = new Superdoc({
+      selector: '#superdoc',
+      toolbar: 'superdoc-toolbar',
+      documentMode: readOnly ? 'viewing' : 'editing',
+      documents: [{
+        id: documentId,
+        type: documentType,
+        data: initialData
+      }],
+      onReady: () => {
+        if (onEditorReady) {
+          onEditorReady(editor);
+        }
+      },
+      onEditorCreate: () => {
+        console.log('Editor created');
+      },
+      onEditorDestroy: () => {
+        console.log('Editor destroyed');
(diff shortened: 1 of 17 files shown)


examples/react-example/src/components/DocumentEditor.jsx
```diff
@@ -0,0 +1,71 @@
...
+  documentId, 
+  documentType = 'docx', 
+  initialData = null,
+  readOnly = false,
+  onEditorReady 
+}) => {
+  const editorRef = useRef(null);
+
+  useEffect(() => {
+    const editor = new Superdoc({
```
chittolinag, line 15: The `useEffect` hook will be executed any time `documentId`, `documentType`, `initialData`, `readOnly` or `onEditorReady` value changes. I'm wondering if that's the expected behavior here? We will be instantiating a new super doc every time they change. I'm not super familiar with what happens when you create a new instance of super doc, does it create a new document, or it updates if it already exists? Just to confirm we're implementing the right behavior here.

examples/react-example/src/components/DocumentEditor.jsx
```diff
@@ -0,0 +1,71 @@
...
+      onEditorDestroy: () => {
+        console.log('Editor destroyed');
+      }
+    });
+
+    editorRef.current = editor;
+
+    // Cleanup on unmount
+    return () => {
+      if (editorRef.current) {
```
chittolinag, line 41: Why exactly do you need this? Did you face any issues without this logic? The reason I'm asking is because the component will be unmounted anyways (and as far as I know React doesn't preserve state in this case), so I'm not sure if we really need to set the value to `null`

harbournick:
Yeah, this makes sense at least for now. If any of those change we should
instantiate a new super doc.

PR #245, opened 2025-01-15
HAR-8702: link input in toolbar - auto close and move document cursor

mattConnHarbour:


packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -125,6 +126,27 @@ export class SuperToolbar extends EventEmitter {
...
+        // move cursor to end
+        const endPos = this.activeEditor.view.state.selection.$to.pos;
+        const selection = new TextSelection(this.activeEditor.view.state.doc.resolve(endPos));
+        const tr = this.activeEditor.view.state.tr.setSelection(selection)
+        const state = this.activeEditor.view.state.apply(tr)
+        this.activeEditor.view.updateState(state)
+
+        setTimeout(() => {
+          this.activeEditor.view.focus();
+        }, 100)
```
artem-harbour: Why do we need a timeout here?

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -125,6 +126,27 @@ export class SuperToolbar extends EventEmitter {
...
       this.updateToolbarState();
     },
+
+    toggleLink: ({ item, argument }) => {
+      let command = item.command;
+
+      if (command in this.activeEditor.commands) {
+        this.activeEditor.commands[command](argument);
+        
+        // move cursor to end
```
artem-harbour, line 136: I would probably abstract this piece of code into a separate command (more general), but let's leave it like this for now. Let's just improve readability a little bit (and add semicolons or use prettier).

```javascript
const { view } = this.activeEditor;
const endPos = view.state.selection.$to.pos;
const selection = new TextSelection(view.state.doc.resolve(endPos));
const tr = view.state.tr.setSelection(selection);
const state = view.state.apply(tr);
view.updateState(state);
```

mattConnHarbour: Done!

artem-harbour (approved):
LGTM (fyi @harbournick )

PR #250, opened 2025-01-16
Pagination upgrades

harbournick:
- Fix/improve page break calculation by adding more detail to the search
- Improve debug
- Fixes to page threshold calculation
- Fixes to hard break node calculation
- Fixes to final page padding calculation


edoversb (approved):
LGTM

PR #258, opened 2025-01-22
fix pdf preview scale

artem-harbour:
**Please test this before merging.**

I noticed that the fields in PDF preview are not in their proper positions on the mobile, so it seems to me that the calculations are not entirely correct.

The issue can be reproduced if you open PDF preview with fields on a small screen with the zoom already changed (different from 1).

Since zoom only changes the target element (and not the page size), it seems to me that need to take the original page size and multiply it by pdf.js default scale factor (1.333), and not calculate the scale based on the current dimensions of the page element.

Screenshot 2025-01-22 at 15 57 23

Before:
Screenshot 2025-01-22 at 15 34 36

After:
Screenshot 2025-01-22 at 15 41 33


edoversb (approved):
LGTM

PR #272, opened 2025-01-28
HAR-8777 Clicking in margin should place cursor at start or end of nearest line

VladaHarbour:
Issue: clicking on page in left or right margin does move cursor to start/end of line - it should.
Added a handler which tracks click event on editor margins and calculates position for the cursor.


harbournick (approved):
LGTM - seems to work great!

PR #273, opened 2025-01-28
use core editor class instead of vue wrapper

artem-harbour:
I think we should use core Editor class (as before) instead of Vue wrapper and avoid unnecessary overhead.

It also seems to me that this wrapper doesn't affect anything now, since we update the toolbar state in a different way and don't rely on the reactive properties from here. The toolbar with core Editor class works the same way.
https://github.com/Harbour-Enterprises/SuperDoc/blob/main/packages/superdoc/src/SuperDoc.vue#L279

@harbournick, please let me know if you think differently.


harbournick (approved):
LGTM - I like the simplification!

PR #274, opened 2025-01-29
HAR-8946 - Standardize how params are passed around importer functions

harbournick:
- Since the importer has grown so much, the params that handlers require were getting very confusing. This cleans up how params are passed around the recursive importer handlers.
- Fixes issue with images not importing into headers/footers


edoversb (approved):
Lgtm

PR #275, opened 2025-01-29
HAR-8865 Completed Agreement adding an extra line space between paragraphs

VladaHarbour:
We don't really consider paragraph properties for list items. Here I added spacing and alignment styles for list items.

### DIFF
--- a/packages/super-editor/src/core/super-converter/exporter.js
+++ b/packages/super-editor/src/core/super-converter/exporter.js
@@ -528,11 +528,13 @@ function getListParagraphProperties(node, level, type) {
 
   // numbering.xml reference
   if (node.attrs.numId) listType = node.attrs.numId;
+  const importAttributes = node.attrs.attributes?.parentAttributes?.paragraphProperties?.elements;
 
   return {
     name: 'w:pPr',
     type: 'element',
     elements: [
+      ...importAttributes.filter(attrs => attrs.name !== 'w:numPr'),
       {
         name: 'w:numPr',
         type: 'element',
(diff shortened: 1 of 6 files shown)


packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -528,11 +528,13 @@ function getListParagraphProperties(node, level, type) {
 
   // numbering.xml reference
   if (node.attrs.numId) listType = node.attrs.numId;
+  const importAttributes = node.attrs.attributes?.parentAttributes?.paragraphProperties?.elements;
 
   return {
     name: 'w:pPr',
     type: 'element',
     elements: [
+      ...importAttributes.filter(attrs => attrs.name !== 'w:numPr'),
```
VladaHarbour, line 537: @harbournick filtering numPr elements here since User could edit the list

packages/super-editor/src/core/super-converter/v2/importer/listImporter.js
```diff
@@ -177,14 +177,20 @@ function handleListNodes(
       nodeAttributes['textStyle'] = textStyle;
       nodeAttributes['order'] = start;
       nodeAttributes['lvlText'] = lvlText;
-      nodeAttributes['lvlJc'] = lvlJc;
+      nodeAttributes['lvlJc'] = textStyle?.attrs.textAlign;
```
VladaHarbour, line 180: @harbournick noticed that text align value taken from the numbering.xml doesn't reflect the real alignment. Taking it from paragraph props instead

PR #276, opened 2025-01-29
HAR-8767 Spacing in Document Preview Differs from Final Document

VladaHarbour:
We don't really consider paragraph properties for list items. Here I added spacing and alignment styles for list items.


packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -528,11 +528,13 @@ function getListParagraphProperties(node, level, type) {
 
   // numbering.xml reference
   if (node.attrs.numId) listType = node.attrs.numId;
+  const importAttributes = node.attrs.attributes?.parentAttributes?.paragraphProperties?.elements;
 
   return {
     name: 'w:pPr',
     type: 'element',
     elements: [
+      ...importAttributes.filter(attrs => attrs.name !== 'w:numPr'),
```
VladaHarbour, line 551: @harbournick filtering numPr elements here since User could edit the list

packages/super-editor/src/core/super-converter/v2/importer/listImporter.js
```diff
@@ -177,14 +177,20 @@ function handleListNodes(
       nodeAttributes['textStyle'] = textStyle;
       nodeAttributes['order'] = start;
       nodeAttributes['lvlText'] = lvlText;
-      nodeAttributes['lvlJc'] = lvlJc;
+      nodeAttributes['lvlJc'] = textStyle?.attrs.textAlign;
```
VladaHarbour, line 177: @harbournick noticed that text align value taken from the numbering.xml doesn't reflect the real alignment. Taking it from paragraph props instead

harbournick: it may be a case of hierarchy, too. Sometimes if its missing in one place its inherited from another, or one will override another (Case in point: w:spacing, which seems to 'combine' nodes by hierarcy). Just noting this here in case we need to check anything else later.

harbournick, line 177: @VladaHarbour Can we add a test case for this?

packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -503,7 +503,19 @@ function translateList(params) {
       const pPr = getListParagraphProperties(listNode, level, type);
       const content = outputNode.elements.filter((e) => e.name !== 'w:pPr');
       if (!content.length) {
-        const spacer = { name: 'w:p', elements: [] };
+        // Some empty nodes could have spacing defined
+        const spacingProp = pPr?.elements.find((e) => e.name === 'w:spacing');
```
harbournick, line 509: Let's add a test case here.

harbournick (changes requested):
LGTM, and it looked good on manual testing here. Please add test cases to the main changes here (we will try to do this on all import/export changes)

harbournick (approved):
LGTM

PR #277, opened 2025-01-30
HAR-8952 Line Height issue in SuperDoc

VladaHarbour:
@harbournick My proposal is to use line-height values without units like is's done in Word

P.S. https://github.com/Harbour-Enterprises/SuperDoc/pull/276 is dependent on this one and requires some changes in case this PR is merged


harbournick (approved):
LGTM

PR #281, opened 2025-01-31
find field annotations by predicate

artem-harbour:
Add `findFieldAnnotations` helper for `fieldAnnotation`.


harbournick (approved):
LGTM

PR #283, opened 2025-01-31
HAR-8956 - Add collaboration migration system, add first migration, store editor version in docx

harbournick:
- Adds new collaboration doc migration system
- Adds first migration for post 0.4.14
- Store editor version inside the docx xml


edoversb (approved):
LGTM

PR #284, opened 2025-02-03
HAR-8934 Text justify toolbar icon issue

VladaHarbour:
Issue: Weird behavior in the toolbar because when I highlight the left justified text, the tool bar updates to the justified icon. But to actually justify it you have to click the icon.

### DIFF
--- a/packages/super-editor/src/core/super-converter/v2/importer/listImporter.js
+++ b/packages/super-editor/src/core/super-converter/v2/importer/listImporter.js
@@ -144,16 +144,19 @@ function handleListNodes(
       item.seen = true;
 
       const schemaElements = [];
-
+      
       let parNode = {
         type: 'paragraph',
-        content: nodeListHandler.handler({ ...params, nodes: elements })?.filter((n) => n),
+        content: nodeListHandler.handler({ ...params, nodes: [ attributes.paragraphProperties, ...elements ] })?.filter((n) => n),
       };
 
       // Normalize text nodes.
       if (parNode.content) {
         parNode = {
           ...parNode,
+          attrs: {
+            textAlign: textStyle?.attrs.textAlign
+          },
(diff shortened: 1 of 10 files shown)


packages/super-editor/src/core/super-converter/v2/importer/listImporter.js
```diff
@@ -174,15 +174,15 @@ function handleListNodes(
       nodeAttributes['textStyle'] = textStyle;
       nodeAttributes['order'] = start;
       nodeAttributes['lvlText'] = lvlText;
-      nodeAttributes['lvlJc'] = textStyle?.attrs.textAlign;
+      nodeAttributes['textAlign'] = textStyle?.attrs.textAlign;
```
harbournick: Something in the importer is creating an issue - not exactly sure where, such that when a list is imported, and you add a new list item in the editor, we are getting an error and the item is inserted in the wrong place. Here is a video: https://cdn.zappy.app/9589da12497997e679fd2a5973d3574d.mp4

I reverted this file only and the issue was fixed, so I'm guessing something about passing marks through. I believe block-level objects aren't supposed to have marks, so we may need to pass these marks down to the paragraph element that is inserted into every list item (all list items are created with a paragraph inside them) - perhaps this might be the cause?

harbournick (changes requested):
See comment. LMK if you need any more info!

packages/super-editor/src/core/super-converter/v2/importer/listImporter.js
VladaHarbour: Hi @harbournick! I wasn't able to reproduce this issue with available documents but I did some fixes anyway. Could you please give it another try?

harbournick (approved):
LGTM

PR #285, opened 2025-02-04
List item spacing issue

VladaHarbour:
Hi, @harbournick! I believe it has to fix the issue with import exported document with list item spacing. I tested it with the documents I have but please let me know if something doesn't work for you


harbournick (approved):
LGTM

PR #291, opened 2025-02-05
HAR-8846 table import/export fixes

VladaHarbour:
Tables styles fixes and nil borders export fixes


harbournick:
@VladaHarbour Setting a meeting for Monday to chat through the unit change. More info in the calendar invite.

harbournick (changes requested):
We need to add a 'unit' attribute to nodes where we are changing the unit moving forward, or let's brainstorm any other ways to keep this sort of change (which I am sure we will come across again) from becoming a breaking change for existing docs.

I requested a meeting for Monday fyi.

harbournick (approved):
LGTM

PR #295, opened 2025-02-11
HAR-9023 Superdoc mobile improvements

VladaHarbour:


packages/super-editor/src/extensions/pagination/pagination.js
```diff
@@ -279,24 +279,23 @@ function generateInternalPageBreaks(doc, view, editor, sectionData) {
   doc.descendants((node, pos) => {
     coords = view?.coordsAtPos(pos);
     if (!coords) return;
-
-    let shouldAddPageBreak = coords.bottom > pageHeightThreshold * scale;
```
VladaHarbour, line 283: @harbournick noticed that when we calculate page size based on scale (which is always less than 1 on mobile) it creates an enormous empty space before the page break

harbournick (approved):
LGTM

PR #300, opened 2025-02-15
Remove DOM observer in pagination, use transaction only, add header m…

harbournick:
- removes to DOM observer that recalculated pagination
- uses transaction to trigger recalculation of left margins
- Adds header/footer margin change detection


edoversb (approved):
LGTM

PR #302, opened 2025-02-17
HAR-8820 - Initial TS support

chittolinag:
**Problem**: TypeScript is widely used in the web ecosystem. According to Stack Overflow's survey, it was the 5th most used language in 2024 and the 4th most admired too. According to GitHub's octoverse insights from 2024, continues strong growth as the 3rd most popular language. TypeScript developers usually expect that types are defined in a way that it can be easily integrate with IDEs - providing autocomplete suggestions, static type checking and much more.

**Solution**:  I ended up following the following approach:

1. We define JSDoc types.
  1.1 We can create them on the same file, or in separate files.
2. We use TS to output declaration types from it

**Changes**:

- Added TS as a dev dependency on the `superdoc` package and `super-editor` packages. This runs on build time and outputs `.d.ts` files alongside the build files, on `dist`.
- Added JSDoc definitions for most of the config object that `SuperDoc` expects and its class attributes.
- Created an example folder with TS, like we have for react and vue.

**Next steps**: 

If we agree this is the right way to go, we just need to ensure every developer creates JSDoc definitions for new code, and export them so TS developers can make use of it. Besides that, we should figure out what are the next most important things we expose to our customers (for existing code), create types for them and expose them.


packages/superdoc/src/core/types.js
```diff
@@ -0,0 +1,60 @@
...
+ * @property {string} type The type of the document
+ * @property {File | null} [data] The initial data of the document
+ */
+
+/**
+ * @typedef {Object} SuperDocTypes.Config
+ * @property {string} [superdocId] The ID of the SuperDoc
+ * @property {string} selector The selector to mount the SuperDoc into
+ * @property {'editing' | 'viewing' | 'suggesting'} documentMode The mode of the document
+ * @property {'editor' | 'viewer' | 'suggester'} [role] The role of the user in this SuperDoc
```
harbournick: Just curious, why are some of the properties in square brackes and some not? ie: [role] vs documentMode

packages/superdoc/src/core/types.js
```diff
@@ -0,0 +1,60 @@
...
+ * @property {Array<SuperDocTypes.Document>} documents The documents to load
+ * @property {SuperDocTypes.User} [user] The current user of this SuperDoc
+ * @property {Array<SuperDocTypes.User>} [users] All users of this SuperDoc (can be used for "@"-mentions)
+ * @property {Array<string>} [colors] Colors to use for user awareness
```
harbournick: We should mark optional params as such. This one is optional. If the client doesn't provide one, we will use a default set.

packages/superdoc/src/core/types.js
```diff
@@ -0,0 +1,60 @@
...
+ * @property {string} id The ID of the document
+ * @property {string} type The type of the document
+ * @property {File | null} [data] The initial data of the document
+ */
+
+/**
+ * @typedef {Object} SuperDocTypes.Config
+ * @property {string} [superdocId] The ID of the SuperDoc
+ * @property {string} selector The selector to mount the SuperDoc into
+ * @property {'editing' | 'viewing' | 'suggesting'} documentMode The mode of the document
```
harbournick: SuperDoc will set some defaults on certain params, if they are not provided. For instance, this one defaults to 'editing' if not provided. Can we indicate _both_ that it is optional and has a default value somehow?

packages/superdoc/src/core/types.js
```diff
@@ -0,0 +1,60 @@
...
+ * @property {SuperDocTypes.User} [user] The current user of this SuperDoc
+ * @property {Array<SuperDocTypes.User>} [users] All users of this SuperDoc (can be used for "@"-mentions)
+ * @property {Array<string>} [colors] Colors to use for user awareness
+ * @property {Object} [modules] Modules to load
+ * @property {boolean} [pagination] Whether to show pagination in SuperEditors
+ * @property {string} [toolbar] Optional DOM element to render the toolbar in
+ * @property {Array<string>} [toolbarGroups] Toolbar groups to show
+ * @property {Object} [toolbarIcons] Icons to show in the toolbar
+ * @property {boolean} [isDev] Whether the SuperDoc is in development mode
+ * @property {SuperDocTypes.Telemetry} [telemetry] Telemetry configuration
```
harbournick: Not for this PR necessarily, just noting that a logical next step for types will likely be the configuration objects for telemetry and any config modules (mainly comments and collaboration)

packages/superdoc/src/core/types.js
```diff
@@ -0,0 +1,60 @@
...
+ * @property {(editor: Editor) => void} [onEditorBeforeCreate] Callback before an editor is created
+ * @property {(editor: Editor) => void} [onEditorCreate] Callback after an editor is created
+ * @property {() => void} [onEditorDestroy] Callback after an editor is destroyed
+ * @property {(params: { error: object, editor: Editor, documentId: string, file: File })} [onContentError] Callback when there is an error in the content
+ * @property {(editor: { superdoc: SuperDoc }) => void} [onReady] Callback when the SuperDoc is ready
+ * @property {(params: { type: string, data: object}) => void} [onCommentsUpdate] Callback when comments are updated
+ * @property {(params: { context: SuperDoc, states: Array }) => void} [onAwarenessUpdate] Callback when awareness is updated
```
harbournick: This is a collaboration-specific callback, and actually I think what you have here is great.

harbournick (changes requested):
Just a couple of minor comments here but looking great - a great start to adding JSDoc and exporting types here

packages/superdoc/src/core/types.js
chittolinag: the ones in square brackets are optional

packages/superdoc/src/core/types.js
chittolinag: the square brackets marks them as optional :)

packages/superdoc/src/core/types.js
chittolinag: yes, we can! I'll work on it. good catch

PR #305, opened 2025-02-19
Telemetry v1 updates

VladaHarbour:


packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -271,6 +271,10 @@ class SuperConverter {
   getSchema(editor) {
     this.getDocumentInternalId();
     const result = createDocumentJson({...this.convertedXml, media: this.media }, this, editor);
+    
+    setTimeout(() => {
```
caio-pizzol: We're sending events after 6s?

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -175,140 +182,77 @@ const createNodeListHandler = (nodeHandlers) => {
...
-                  internalId: converter?.documentInternalId, 
-                }
-              );
-              index = 0; // Reset to safe value
+          } else {
+            converter?.telemetry?.trackStatistic('node', context);
+            
+            // Use Telemetry to track list item attributes
+            if (context.type === 'orderedList' || context.type === 'bulletList') {
+              context.content.forEach((item) => {
```
caio-pizzol, line 208: Nice

shared/common/Telemetry.js
```diff
@@ -84,129 +67,162 @@ class Telemetry {
...
-      source: this.getSourceData(),
-      category,
-      document: processedDoc,
-      name,
-      path,
-      ...(metadata && { metadata }),
-    };
+    // Style attributes
+    if (data.attributes && Object.keys(data.attributes).length) {
+      const styleAttributes = [
```
caio-pizzol, line 155: Can't we get those automatically instead of hardcoding?

shared/common/Telemetry.js
```diff
@@ -220,84 +236,104 @@ class Telemetry {
...
 
   /**
-   * Flush queued events to server
+   * Sends current report
    * @returns {Promise<void>}
    */
-  async flush() {
-    if (!this.enabled || !this.events.length) return;
+  async sendReport() {
+    if (!this.enabled || !this.isTelemetryDataChanged()) return;
```
caio-pizzol, line 260: Nice.

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -175,140 +182,77 @@ const createNodeListHandler = (nodeHandlers) => {
...
-              converter?.documentId,
-              'node', 
-              'unhandled', 
-              `/word/${filename || 'document.xml'}`, 
-              {
-              context,
-              internalId: converter?.documentInternalId,
-            });
             
+            const ignoreElements = ['w:pPr', 'w:tcPr'];
```
caio-pizzol: Same here - or at least make them more explicitly defined? env or global vars?

caio-pizzol (commented):
@VladaHarbour overall LGTM - just a few comments.
@harbournick @artem-harbour want to hear your thoughts as well.

packages/super-editor/src/core/super-converter/SuperConverter.js
VladaHarbour: I did it in purpose just to be sure that all the data was retrieved. But I'm thinking that maybe there is some event listener for the case when content was loaded
will check it

shared/common/Telemetry.js
VladaHarbour: We can just monitor all of them. Note: there are not only attributes from xml but also things which was added by us during parsing

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
VladaHarbour: Will check it one more time
Not sure why those are treated as unhandled

caio-pizzol (approved):
LGTM

PR #308, opened 2025-02-21
HAR-9081 Import existing annotations as {{ }}

VladaHarbour:
Annotations presented in the imported doc are now transformed into text nodes {{ ${displayLabel} }}


harbournick (approved):
LGTM

PR #309, opened 2025-02-25
Add docxtemplater demo

mattConnHarbour:


mattConnHarbour:
@edoversb PR to add docxtemplater example! Also, can we host this example ourselves so that we can link to it in the readme?

edoversb (approved):
LGTM

PR #311, opened 2025-02-25
HAR-8995 Blurred signature fields(typed signature)

VladaHarbour:
Getting a proper signature scale in order to make it more clear in the final doc

### DIFF
--- a/packages/super-editor/src/core/super-converter/exporter.js
+++ b/packages/super-editor/src/core/super-converter/exporter.js
@@ -3,7 +3,14 @@ import { DOMParser as PMDOMParser } from 'prosemirror-model';
 import { EditorState } from 'prosemirror-state';
 import { SuperConverter } from './SuperConverter.js';
 import { toKebabCase } from '@harbour-enterprises/common';
-import { inchesToTwips, linesToTwips, pixelsToEightPoints, pixelsToEmu, pixelsToTwips } from './helpers.js';
+import {
+  emuToPixels,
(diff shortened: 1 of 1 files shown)


packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -1095,18 +1102,64 @@ function translateMark(mark) {
...
+
 function translateImageNode(params, imageSize) {
   const {
     node: { attrs = {}, marks = [] },
   } = params;
 
   let imageId = attrs.rId;
-  const size = attrs.size
+  
+  const { originalWidth, originalHeight } = getPngDimensions(attrs.imageSrc);
```
harbournick, line 1145: How will this affect non-signature images (from annotations, or directly in the document)? This function might encounter jpg as well (possibly others but less likely)

harbournick (changes requested):
See my question - thanks!

packages/super-editor/src/core/super-converter/exporter.js
VladaHarbour: This approach won’t work for jpeg. That’s why I check base64 type in this function. For jpeg will stick with current approx
For png images(image upload input specifically) it also will calculate the ratio according to the default image container size. Tested it for both signatures and image upload input

harbournick (approved):
LGTM

harbournick:
Actually let's hold this one for a moment, want to put it in 0.6

PR #314, opened 2025-02-25
feat: added tests on ci/cd

chittolinag:
- Run unit tests before publish (needs to succeed, or the publish won't be executed)
- Run unit tests on PR creations to `main`


harbournick (approved):
LGTM

PR #318, opened 2025-02-26
feat: Add LocalStorageManager class with persistent ruler

palmer-cl:
- Added a LocalStorageManager class to interact with localStorage
- Complete with get, set, merge, getSetting functions 

Example usage:
```LocalStorageManager.getSetting('rulers')``` -> ```true```

Stored as:
```superdoc::settings | { rulers: true } ```

To update rulers: 
```LocalStorageManager.merge('settings', { rulers: this.config.rulers });```


harbournick:
@palmer-cl Thanks for the PR here. I'll shoot you a note on the discord, but I think we would prefer (at least for now) to keep this functionality outside of SuperDoc itself, and leave it up to the end user to choose if-and-how they want to make the ruler sticky. We can implement this as an example in superdoc.dev directly for now.

PR #321, opened 2025-02-26
Nick/docs updates

harbournick:
- Removing any invalid docs
- Changed events to 'hooks' - our preferred way for end users to work with SuperDoc events. Hooks are possible, too, but they require additional lower level management.


edoversb (approved):
Lgtm

PR #325, opened 2025-02-26
added file polyfill for node js < v22

andrii-harbour:


harbournick (approved):
LGTM
Noted - polyfill is needed for Node <= 22

PR #329, opened 2025-02-27
HAR-8998 Highlight toolbar icon

VladaHarbour:


packages/super-editor/src/core/super-converter/helpers.js
```diff
@@ -85,6 +85,39 @@ const getContentTypesFromXml = (contentTypesXml) => {
   return Array.from(defaults).map((item) => item.getAttribute('Extension'));
 };
 
+const getHexColorFromDocxSystem = (docxColor) => {
```
VladaHarbour, line 88: @harbournick Noticed that when you add highlight in word it has limited colour options and colour is defined as text in the xml. When we add highlight in the google docx, for example, it will be defined as hex

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -50,6 +50,7 @@ class SuperConverter {
     { name: 'w:spacing', type: 'lineHeight', mark: 'textStyle', property: 'lineHeight' },
     { name: 'link', type: 'link', mark: 'link', property: 'href' },
     { name: 'w:highlight', type: 'highlight', mark: 'highlight', property: 'color' },
+    { name: 'w:shd', type: 'highlight', mark: 'highlight', property: 'color'}
```
VladaHarbour, line 53: @harbournick It's possible to have w:highlight and w:shd tags with different attributes. Word uses w:highlight, google docs w:shd

harbournick: ahh I wondered where the w:shd came from, have come across it before. Nice!

harbournick (approved):
LGTM

PR #333, opened 2025-02-28
HAR-9082 - Create default styles for styles list when they dont exist

chittolinag:
Adds default styles to the document if they are not present.

- Normal
- Title
- Subtitle
- Heading1
- Heading2
- Heading3

Changes are also applied to the exported document.

**Known bug**:

1 - I found out that DOCX viewers can't handle dealing with some special characters (I've tried with `&` for now). We're still not sure which others are unsupported (if there are any more). This will be fixed in a later version, in case there are other cases.


VladaHarbour:
Hi @chittolinag. Commits history in the PR looks a bit odd. Do you mind rebasing with the latest main?

VladaHarbour (approved):

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -29,6 +29,661 @@ import { getDefaultStyleDefinition } from './paragraphNodeImporter.js';
  * @typedef {{handlerName: string, handler: NodeHandler}} NodeHandlerEntry
  */
 
+export const DEFAULT_LINKED_STYLES = {
```
harbournick: let's move this definition out of the main docxImporter file
we could use: packages/super-editor/src/core/super-converter/exporter-docx-defs.js instead

harbournick (changes requested):
Just a quick change here otherwise looks good!

chittolinag:
> Hi @chittolinag. Commits history in the PR looks a bit odd. Do you mind rebasing with the latest main?

yeah you're right... the reason we're seeing those commits is because I created this branch off of another PR created by @harbournick , but that's already merged to `main`. So even though they are showing up here, they are not producing any diffs as they are already on `main`. I tried to rebase but the conflicts were weird and I didn't want to possibly break something

harbournick (approved):
LGTM

PR #336, opened 2025-03-03
HAR-8936 - tables

artem-harbour:
Screenshot 1
Screenshot 2025-03-03 at 12 58 09

Screenshot 2
Screenshot 2025-03-03 at 12 58 26

Note:
- Import/export has been changed to work with the new table functionality, but overall it needs improvement.


harbournick (approved):
LGTM - nice work!

PR #337, opened 2025-03-03
HAR-9135 Set default color for highlight icon

VladaHarbour:
Hi @harbournick! I did it like it's done in Word. Set default color to light grey  so when User select text and click the highlight icon it will be immediately highlighted and color picker will be offered to select a different color


harbournick (approved):
LGTM

PR #339, opened 2025-03-03
HAR-9150 - Re-add consumed check in telemetry logic

harbournick:


packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -223,6 +223,26 @@ const createNodeListHandler = (nodeHandlers) => {
             }
           }
 
+          if (consumed > 0) {
+            index += consumed - 1;
+            if (index < 0) {
```
VladaHarbour: Hi @harbournick! This check with telemetry call is not required anymore. Looks like I grabbed one more extra if
Just 
`if (consumed > 0) {
            index += consumed - 1;
}
`should work as well

VladaHarbour (approved):

PR #348, opened 2025-03-04
HAR-9165 Fix comments export for docx files that don't have previous comments infra

harbournick:
- fix comments export for docx files that don't have previous comments infra (ie: comments.xml etc)
- fix pdf export


edoversb (approved):
Lgtm

PR #350, opened 2025-03-05
HAR-9165 - Refactor comments to use marks instead of nodes

harbournick:
- Refactors comments to use marks instead of nodes (fixes 'extra hidden nodes' issue)
- Adds document post-import pre-processing and pre-export processing for comments


edoversb (approved):
LGTM

PR #351, opened 2025-03-05
HAR-8873 - fix: prevent duplicate editor initialization on collaboration sync

chittolinag:
### Issue

Collaboration cursor starts to flick from time to time. 

unnamed

### How to reproduce:

1. Open a SuperDoc and share it with someone
2. Leave the page open for ~5 minutes.
3. Come back to the page and try to make any edits. You will see the cursor will start to flick.

The reason this happens is because we use Cloud Run for the collaboration service, and by default it times out after 5 minutes. When it times out, the connection will be lost, but our provider (hocus pocus)  will reconnect automatically.

When that happens: 

1. The `synced` event will be triggered again
2. It will call the `pollForMetaMapData` function. 
3. The `pollForMetaMapData` function will call `initEditor` again.
4. When that happens, internally we are reinitializing  `Editor.js`, and hence the error starts to happen.

### Solution

Remove the listener to the `synced` event once it's triggered for the first time.

### DIFF
--- a/packages/super-editor/src/components/SuperEditor.vue
+++ b/packages/super-editor/src/components/SuperEditor.vue
@@ -60,6 +60,10 @@ const pollForMetaMapData = (ydoc, retries = 10, interval = 500) => {
     if (docx) {
       stopPolling();
       initEditor({ content: docx });
+      // Remove the synced event listener.
+      // Avoids re-initializing the editor in case the connection is lost and reconnected
+      const provider = props.options.collaborationProvider;
+      provider.off('synced');
     } else if (retries > 0) {
       console.debug(`Waiting for 'docx' data... retries left: ${retries}`);
       dataPollTimeout = setTimeout(checkData, interval); // Retry after the interval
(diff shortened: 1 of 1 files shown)


packages/super-editor/src/components/SuperEditor.vue
```diff
@@ -60,6 +60,10 @@ const pollForMetaMapData = (ydoc, retries = 10, interval = 500) => {
     if (docx) {
       stopPolling();
       initEditor({ content: docx });
+      // Remove the synced event listener.
+      // Avoids re-initializing the editor in case the connection is lost and reconnected
+      const provider = props.options.collaborationProvider;
+      provider.off('synced');
```
harbournick: I think we should try provider.off right inside the provider.on call
likely have to do something like:

function handleSynced () {
  pollForMetaMapData(ydoc);
  provider.off('synced', handleSynced);
}

provider.on('synced', handleSynced);

The reason being that we shouldn't delay turning this event off (+ we should pass the specific function we are removing here). The editor uses 'synced' for other things later on so we likely don't want to turn it off altogether

harbournick (changes requested):
Just a quick change on where to place the .off here

packages/super-editor/src/components/SuperEditor.vue
chittolinag: oh yeah, much better.

chittolinag: done

harbournick (approved):
LGTM

PR #354, opened 2025-03-06
fix: remove only handleSynced listener after it's triggered

chittolinag:
After the `synced` event is triggered, we need to ensure only the same event listener is unbind as we have other listeners for the `synced` event.


harbournick (approved):
LGTM

PR #355, opened 2025-03-06
HAR-9170 - fix toolbar reactivity and responsiveness

artem-harbour:
Fixed toolbar reactivity and responsiveness following the current approach.

Note:
I think the approach should be reconsidered and changed later.


harbournick (approved):
LGTM - works so much better!

PR #358, opened 2025-03-09
HAR-7608 Create document version when collaboration is closed

VladaHarbour:
Hi @harbournick. When I was working on this "When the editor is closed (or there are no more people) + 5min <- Should be done inside Collab" inside collaboration I noticed that when the User closes the Editor we don't really have provider disconnected. It results in having clientsCount = 1 in onDisconnect hook in collaboration. Here is the issue that seems similar - https://github.com/ueberdosis/hocuspocus/issues/594
I also added 'preserveConnection: false' to the hocuspocus provider config which guarantees that provider will be disconnected. Let me know if it has some sense?


harbournick (approved):
LGTM
I had run into a previous issue with preserve connection but its unrelated, and I never came back around to add it again. Good thinking here and let's test it thoroughly.

PR #359, opened 2025-03-10
Fix header height min calculation, fix hidden editor styles missing

harbournick:
- Fixes issue with hidden editor (for pagination) not getting the right styles, causing various calculations to be off
- Fixes header height calculation


edoversb (approved):
LGTM

PR #362, opened 2025-03-11
HAR-9198 - Fix resolving dual add/delete tracked change

harbournick:
- When tracked change has both add/delete, we need to resolve or reject both at the same time.


edoversb (approved):
Lgtm

PR #366, opened 2025-03-13
HAR-9157 - Trigger floating comment init on sync or locations emit 

harbournick:
- Fixes floating comments init
- Fixes race case between comments collaboration sync and editor locations emit
- Fixes issue with floating comments having no selectionBounds on init


edoversb (approved):
LGTM

PR #367, opened 2025-03-13
HAR-9224 - fix export images inside table

artem-harbour:
Fixed images and field annotations (images/signatures) in import/export.

Note:
- This does not fix the image export issue that was introduced with one of the previous PRs.


harbournick (approved):
LGTM

PR #375, opened 2025-03-14
HAR-9245 - Fix export issue with some paragraphs missing elements key when coming from paragraph fields

harbournick:


edoversb (approved):
LGTM

PR #378, opened 2025-03-17
HAR-9254 - At-mention issues

harbournick:
- Ensure viewers can't be at-mentioned in a comment
- Ensure only comment owners can edit a comment
- use user email when there is no user name in at-mentions


edoversb (approved):
LGTM

PR #380, opened 2025-03-17
HAR-9280 - Fix pending comment being overriden by setActiveComment when change and comment overlap

harbournick:
- Fix pending comment being overriden by setActiveComment when change and comment overlap
- Allow setting user role, email from url params in SuperDoc dev


edoversb (approved):
LGTM

PR #385, opened 2025-03-18
HAR-9295 - Fix tracked changes not using the same emit system as comments

harbournick:


harbournick:
Closing - this has a bug

PR #388, opened 2025-03-18
HAR-9299 - Fix: improve selection bounds calculation for floating comments

chittolinag:
**Problem**: Floating comments are not correctly aligned with the selected text. There was some code in place to determine  `Math.max(96, (top - 100))`. This ensures that the floating comment won't be too close to the header, but the `-100` part was causing the misalignment. 

**Before:**

Screenshot 2025-03-18 at 18 59 53

**After:** 

Screenshot 2025-03-18 at 19 01 14


harbournick (approved):
LGTM

PR #391, opened 2025-03-19
HAR-9297 - pagination: break table row instead of cell if found

artem-harbour:
Added changes that break tables more accurately (for pagination). If a node exceeds the page size and it is located in a table cell, then other cells are not taken into account. I think in this case we should break the tableRow instead.

Before.
Screenshot 2025-03-19 at 15 47 14

After.
Screenshot 2025-03-19 at 15 49 32


harbournick (approved):
VERY nice and simple fix! LGTM

PR #397, opened 2025-03-19
HAR-9300 - fix: improve floating comments offset calculation and scrolling behavior

chittolinag:
**Problem**: When there are multiple comments, the selected comment might not be visible in the screen. 

**Solution**: To solve that, we are now scrolling into the element so it's on the viewport.

Note that I also changed the transition duration to `.1s` as we need to wait for the transition to complete and only then scroll to the element. Otherwise it would take `.3s` to scroll to the component.


harbournick (approved):
LGTM

PR #400, opened 2025-03-20
HAR-9211 - fix content types and hex

harbournick:
- Fix issue with content types becoming corrupted if any comments type is double (ie: if imported file already had it)
- Fix paraId generator to use correct 32-bit hex


edoversb (approved):
Lgtm

PR #402, opened 2025-03-20
HAR-9250 - fix images export

artem-harbour:
For new media we always have to create a new relationship because a new image id is generated every time.


harbournick (approved):
LGTM

PR #403, opened 2025-03-20
HAR-8992 Get run properties from styles.xml by the styleID

VladaHarbour:


harbournick (changes requested):
Looks good, but this PR is breaking some tests. Can you double check what is breaking?

harbournick (approved):
LGTM

PR #407, opened 2025-03-20
HAR-9326 - Handle imported author from docx imports, add it to comments composble, show it if it exists

harbournick:


edoversb (approved):
Lgtm

PR #412, opened 2025-03-24
HAR-9336 - fix toolbar state updates

artem-harbour:
Fixed updating the toolbar state when initialized and when updating the document mode.

@harbournick Please double check before merging as I'm not aware of the details of  `SuperToolbar` and `SuperDoc` classes.

### DIFF
--- a/packages/super-editor/src/components/toolbar/super-toolbar.js
+++ b/packages/super-editor/src/components/toolbar/super-toolbar.js
@@ -198,7 +198,7 @@ export class SuperToolbar extends EventEmitter {
     this.config = { ...this.config, ...config };
     this.toolbarItems = [];
     this.overflowItems = [];
-    this.documentMode = 'editing';
+    this.documentMode = config.documentMode || 'editing';
     this.isDev = config.isDev || false;
     this.superdoc = config.superdoc;
     this.role = config.role || 'editor';
--- a/packages/superdoc/src/core/SuperDoc.js
+++ b/packages/superdoc/src/core/SuperDoc.js
@@ -187,7 +187,6 @@ export class SuperDoc extends EventEmitter {
 
     // If a toolbar element is provided, render a toolbar
     this.addToolbar(this);
-  
   }
 
   get requiredNumberOfEditors() {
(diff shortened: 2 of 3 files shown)


packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -198,7 +198,7 @@ export class SuperToolbar extends EventEmitter {
     this.config = { ...this.config, ...config };
     this.toolbarItems = [];
     this.overflowItems = [];
-    this.documentMode = 'editing';
```
artem-harbour, line 201: The document mode was set to 'editing' and never changed.

artem-harbour: But since the document mode is checked on toolbar state update, I can assume that it should change (on doc mode change).

packages/superdoc/src/core/SuperDoc.js
```diff
@@ -402,26 +403,38 @@ export class SuperDoc extends EventEmitter {
...
+      if (firstEditor) this.setActiveEditor(firstEditor);
     }
 
     this.superdocStore.documents.forEach((doc) => {
       doc.restoreComments();
       const editor = doc.getEditor();
       if (editor) editor.setDocumentMode('editing');
     });
+
+    if (this.toolbar) {
```
artem-harbour, line 415: Set document mode for toolbar and update the state.

packages/superdoc/src/core/SuperDoc.js
```diff
@@ -354,11 +353,13 @@ export class SuperDoc extends EventEmitter {
       role: this.config.role,
       pagination: this.config.pagination,
       icons: this.config.toolbarIcons,
+      documentMode: this.config.documentMode,
       superdoc: this,
     };
 
     this.toolbar = new SuperToolbar(config);
     this.toolbar.on('superdoc-command', this.onToolbarCommand.bind(this));
+    this.once('editorCreate', () => this.toolbar.updateToolbarState());
```
artem-harbour, line 362: We need to wait for the editor to be created to update the toolbar state because there are a lot of asynchronous things behind.

harbournick (approved):
LGTM

PR #414, opened 2025-03-24
HAR-9211 Comments export fixes

VladaHarbour:


packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -390,8 +390,10 @@ class SuperConverter {
     const relsData = this.convertedXml['word/_rels/document.xml.rels'];
     const relationships = relsData.elements.find((x) => x.name === 'Relationships');
     const newRels = [];
-
-    let largestId = Math.max(...relationships.elements.map((el) => Number(el.attributes.Id.replace('rId', ''))));
+    
+    const regex = /rId|mi/g;
```
VladaHarbour, line 394: @harbournick the document I tested had a media relation with id started with 'mi' and it results in largestId being NaN

packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -1604,11 +1604,14 @@ function translateFieldAnnotation(params) {
 
 export function translateHardBreak() {
   return {
-    name: 'w:br',
-    type: 'element',
-    attributes: { 'w:type': 'page' }
+    name: 'w:r',
```
VladaHarbour, line 1607: @harbournick When Word recovers our documents it wraps hard breaks into run for some reason

packages/super-editor/src/core/super-converter/v2/exporter/commentsExporter.js
```diff
@@ -38,11 +38,14 @@ export function translateCommentNode(params, type) {
 
   let commentSchema = getCommentSchema(type, commentIndex);
   if (type === 'End') {
-    const commentReference = { name: 'w:commentReference', attributes: { 'w:id': String(commentIndex) } };
+    const commentReference = {
```
VladaHarbour, line 41: @harbournick  Same with wrapping here

packages/super-editor/src/core/super-converter/v2/exporter/commentsExporter.js
```diff
@@ -185,15 +188,14 @@ export const updateCommentsExtendedXml = (comments = [], commentsExtendedXml) =>
     const attributes = {
       'w15:paraId': comment.commentParaId,
       'w15:done': comment.resolvedTime ? '1' : '0',
-      'w:rsid': comment.commentId || comment.importedId,
```
VladaHarbour, line 188: @harbournick removed it because Word doesn't add this attribute for own comments

packages/super-editor/src/core/super-converter/v2/exporter/commentsExporter.js
```diff
@@ -240,7 +242,7 @@ export const updateCommentsIdsAndExtensible = (comments = [], commentsIds, exten
       "name": "w16cex:commentExtensible",
       "attributes": {
           "w16cex:durableId": newDurableId,
-          "w16cex:dateUtc": "2025-03-06T23:32:00Z"
+          "w16cex:dateUtc": toIsoNoFractional()
```
VladaHarbour, line 245: not sure it has to be hardcoded

harbournick: oh yeah, that was my mistake. thanks for fixing!

packages/super-editor/src/core/super-converter/SuperConverter.js
harbournick: whoa, interesting! TIL

packages/super-editor/src/core/super-converter/v2/exporter/commentsExporter.js
harbournick: SG. I started adding this to try and find a place where we could put our own comment ID into the export, so that if we re-import, we can re-match comments. but we will circle back to this, this may not be the right place.

VladaHarbour: I can bring it back because I don't this it affects the export

PR #420, opened 2025-03-25
Prevent trying to connect the global superdoc ydoc connection if there is no superdocId

harbournick:
Prevent trying to connect the global superdoc ydoc connection if there is no superdocId


edoversb (approved):
LGTM

PR #423, opened 2025-03-25
HAR-9366 - Fix issue with floating comments ignoring tracked changes when no comments

harbournick:
- Fixes issue with floating comments sidebar flickering when there are no comments (only tracked change comments)
- Fixes transaction dispatch on selection change in comments plugin


edoversb (approved):
LGTM

PR #431, opened 2025-03-28
HAR-9246 - handle import issues

artem-harbour:
Fixed import issues that prevented import of a specific document.

### DIFF
--- a/packages/super-editor/src/core/super-converter/SuperConverter.js
+++ b/packages/super-editor/src/core/super-converter/SuperConverter.js
@@ -53,7 +53,7 @@ class SuperConverter {
     { name: 'w:spacing', type: 'lineHeight', mark: 'textStyle', property: 'lineHeight' },
     { name: 'link', type: 'link', mark: 'link', property: 'href' },
     { name: 'w:highlight', type: 'highlight', mark: 'highlight', property: 'color' },
-    { name: 'w:shd', type: 'highlight', mark: 'highlight', property: 'color'}
+    // { name: 'w:shd', type: 'highlight', mark: 'highlight', property: 'color'}
   ];
 
   static propertyTypes = Object.freeze({
--- a/packages/super-editor/src/core/super-converter/v2/importer/annotationImporter.js
+++ b/packages/super-editor/src/core/super-converter/v2/importer/annotationImporter.js
@@ -35,6 +35,10 @@ export const handleAnnotationNode = (params) => {
     fieldColor,
     multipleImage: isMultipleImage === 'true',
   };
+
+  if (!attrs.fieldId || !attrs.displayLabel) {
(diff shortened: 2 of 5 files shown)


packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -53,7 +53,7 @@ class SuperConverter {
     { name: 'w:spacing', type: 'lineHeight', mark: 'textStyle', property: 'lineHeight' },
     { name: 'link', type: 'link', mark: 'link', property: 'href' },
     { name: 'w:highlight', type: 'highlight', mark: 'highlight', property: 'color' },
-    { name: 'w:shd', type: 'highlight', mark: 'highlight', property: 'color'}
+    // { name: 'w:shd', type: 'highlight', mark: 'highlight', property: 'color'}
```
artem-harbour, line 56: This added highlighting in places where there was no highlighting in the original document.

packages/super-editor/src/core/super-converter/v2/importer/listImporter.js
```diff
@@ -110,7 +110,7 @@ function handleListNodes(
     // Spacers in the XML and need to be appended to the last item.
     if (item.elements && !hasTextNode(item.elements)) {
       const n = handleStandardNode({ ...params, nodes: [item] }).nodes[0];
-      parsedListItems[parsedListItems.length - 1]?.content.push(n);
+      if (n) parsedListItems[parsedListItems.length - 1]?.content.push(n);
```
artem-harbour, line 113: At this point we could push a `null` element that could break the doc schema.

packages/super-editor/src/core/super-converter/v2/importer/annotationImporter.js
```diff
@@ -35,6 +35,10 @@ export const handleAnnotationNode = (params) => {
     fieldColor,
     multipleImage: isMultipleImage === 'true',
   };
+
+  if (!attrs.fieldId || !attrs.displayLabel) {
```
artem-harbour, line 39: We need to check some attributes that this is really an annotation and not something else.

packages/super-editor/src/core/super-converter/v2/importer/standardNodeImporter.js
```diff
@@ -13,6 +13,10 @@ export const handleStandardNode = (params) => {
   const node = nodes[0];
   const { name } = node;
   const { attributes, elements, marks = [] } = parseProperties(node, docx);
+
+  if (name === 'w:sdt') {
```
artem-harbour, line 17: If 'w:sdt' element has not been processed before (for example as an annotation) then we need to skip it here, otherwise it will create elements that do not match the schema.

harbournick (approved):
LGTM

PR #432, opened 2025-03-28
Lists input rules

VladaHarbour:
1. Creates bullet list by typing '- '
2. Creates ordered list by typing '1. + Enter'

### DIFF
--- a/packages/super-editor/src/core/commands/undoInputRule.js
+++ b/packages/super-editor/src/core/commands/undoInputRule.js
@@ -1,0 +1,32 @@
+export const undoInputRule = () => ({ state, dispatch }) => {
+  const plugins = state.plugins;
+
+  for (let i = 0; i < plugins.length; i += 1) {
+    const plugin = plugins[i];
+    let undoable;
+    
+    // eslint-disable-next-line
+    if (plugin.spec.isInputRules && (undoable = plugin.getState(state))) {
+      if (dispatch) {
+        const tr = state.tr;
+        const toUndo = undoable.transform;
+
+        for (let j = toUndo.steps.length - 1; j >= 0; j -= 1) {
+          tr.step(toUndo.steps[j].invert(toUndo.docs[j]));
+        }
+
+        if (undoable.text) {
+          const marks = tr.doc.resolve(undoable.from).marks();
+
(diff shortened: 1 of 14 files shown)


harbournick:
@artem-harbour mind reviewing this one as well pls?

packages/super-editor/src/core/extensions/keymap.js
```diff
@@ -20,6 +20,7 @@ export const Keymap = Extension.create({
 
     const handleBackspace = () =>
       this.editor.commands.first(({ commands, tr }) => [
+        () => commands.undoInputRule(),
```
harbournick, line 23: nice !

packages/super-editor/src/core/inputRules/wrappingInputRule.js
```diff
@@ -0,0 +1,59 @@
...
+
+        if (marks) {
+          const filteredMarks = marks.filter(mark => splittableMarks.includes(mark.type.name));
+
+          tr.ensureMarks(filteredMarks);
+        }
+      }
+      if (config.keepAttributes) {
+        /** If the nodeType is `bulletList` or `orderedList` set the `nodeType` as `listItem` */
+        const nodeType = config.type.name === 'bulletList' || config.type.name === 'orderedList' ? 'listItem' : 'taskList';
```
artem-harbour: We do not have `taskList` node type, please fix this.

packages/super-editor/src/core/commands/undoInputRule.js
```diff
@@ -0,0 +1,32 @@
+export const undoInputRule = () => ({ state, dispatch }) => {
```
artem-harbour, line 1: I would prefer not do `undoInputRule` and keep it as simple as possible.

@harbournick, what do you think?

artem-harbour (changes requested):
Added my comments.

FYI @harbournick

packages/super-editor/src/core/commands/undoInputRule.js
VladaHarbour: When used with ordered list it’s more essential (from my point of view) to undo the rule on delete to leave the initial typing
If not the initial typing will be removed

harbournick (approved):
LGTM

PR #433, opened 2025-03-28
HAR-9057 - getting super editor keydown event for enabling agreement link updates

mattConnHarbour:
Companion to Webapp PR: https://github.com/Harbour-Enterprises/Harbour-WebApp/pull/5890

### DIFF
--- a/packages/super-editor/src/components/SuperEditor.vue
+++ b/packages/super-editor/src/components/SuperEditor.vue
@@ -138,7 +138,7 @@ const initEditor = async ({ content, media = {}, mediaFiles = {}, fonts = {} } =
 };
 
 const handleSuperEditorKeydown = (event) => {
-  emit('editor-keydown', { editor: editor.value });
+  props.options.onKeydown({editor: editor.value })
 };
 
 const handleSuperEditorClick = (event) => {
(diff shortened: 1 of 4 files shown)


packages/super-editor/src/components/SuperEditor.vue
```diff
@@ -138,7 +138,7 @@ const initEditor = async ({ content, media = {}, mediaFiles = {}, fonts = {} } =
 };
 
 const handleSuperEditorKeydown = (event) => {
-  emit('editor-keydown', { editor: editor.value });
+  props.options.onKeydown({editor: editor.value })
```
mattConnHarbour, line 141: @harbournick Am I using this as intended? I was having issues just using an "emit" before.

harbournick: I would prefer to not add _more_ hooks here that will complicate the API. Couldn't we just use the existing onUpdate hook?

harbournick (changes requested):
See note

harbournick:
Closing this one after discussion. Based on the requirements, it seems that an existing hook like onUpdate will do the trick.

harbournick:
Closing this one after discussion. Based on the requirements, it seems that an existing hook like onUpdate will do the trick.

PR #440, opened 2025-03-31
HAR-9057 - superdoc event - bubbling up update event

mattConnHarbour:


packages/superdoc/src/core/SuperDoc.js
```diff
@@ -229,6 +231,7 @@ export class SuperDoc extends EventEmitter {
     this.on('pdf-document-ready', this.config.onPdfDocumentReady);
     this.on('sidebar-toggle', this.config.onSidebarToggle);
     this.on('collaboration-ready', this.config.onCollaborationReady);
+    this.on('update', this.config.onEditorUpdate);
```
harbournick: I don't think the event here should just be 'update' because this _specifically_ relates to editor updates. SuperDoc also handles multiple documents, comments... PDF documents, all of which can have 'update' events but that won't be relevant here. we should keep something like 'editor-update'

harbournick (changes requested):
Just a small change

harbournick (approved):
LGTM

PR #442, opened 2025-04-01
HAR-9405 - update rich text extensions

artem-harbour:


harbournick:
Actually its failing CICD - we'll need to fix that before merging this one. I had tried to do this previously and noticed CICD fails on Git but wasn't failing locally

packages/super-editor/src/tests/export/export-helpers/export-helpers.js
```diff
@@ -1,7 +1,7 @@
 import { join } from 'path';
 import { readFile } from 'fs/promises';
 import { Editor } from '@core/Editor.js';
-import { getRichTextExtensions } from '@harbour-enterprises/super-editor';
```
artem-harbour, line 4: This is the issue.

artem-harbour:
> Actually its failing CICD - we'll need to fix that before merging this one. I had tried to do this previously and noticed CICD fails on Git but wasn't failing locally

Fixed.

packages/super-editor/src/tests/export/export-helpers/export-helpers.js
harbournick: ahhh makes sense - thank you for finding!

harbournick (approved):
LGTM

PR #444, opened 2025-04-03
feat(ai): add ai text generation and rewriting using insights endpoint

palmer-cl:
## Adds the v1 functionality for AI Writing and Re-writing

- Adds AI button to the toolbar
- Adds custom popout styled menu when button is hit
- Adds AI layer component to handle highlighting (similar to comments layer)
- When suggesting, generated text are inserted as tracked changes
- Adds AI button to side tools menu

### Requesting eyes on 
- Emitting and Listening for events (specifically emitting from toolbar vs superdoc or if we could optimize this)
- AIWriter component is built in SuperEditor and then exported to SuperDoc. It is consumed by both. Is there a better pattern for this? Do we have a place we store shared components? 

### Caveats
- The `processStream` function will change when the insights endpoint update the data structure returned. This is in progress by @falsaadehh 

Screenshot 2025-04-02 at 7 27 30 PM


packages/superdoc/src/core/SuperDoc.js
```diff
@@ -101,7 +100,8 @@ export class SuperDoc extends EventEmitter {
     user: { name: null, email: null },
     users: [],
 
-    modules: {}, // Optional: Modules to load
+    modules: {}, // Optional: Modules to load. Use modules.ai.{your_key} to pass in your key
```
artem-harbour, line 113: `modules.ai.{your_key}` - I think the list of available modules should be listed in jsdoc.

packages/superdoc/src/core/SuperDoc.js
```diff
@@ -364,11 +364,16 @@ export class SuperDoc extends EventEmitter {
...
+      aiApiKey: this.config.modules?.ai?.apiKey,
       ...moduleConfig,
     };
 
     this.toolbar = new SuperToolbar(config);
+
     this.toolbar.on('superdoc-command', this.onToolbarCommand.bind(this));
+    // AI highlight is not related to document editing, should be separate events
+    this.toolbar.on('ai-highlight-add', (data) => this.emit('ai-highlight-add', data));
+    this.toolbar.on('ai-highlight-remove', () => this.emit('ai-highlight-remove'));
```
artem-harbour: I would prefer to have a more general event so as not to pollute the event list.

```javascript
// type - add, remove, other event type
this.toolbar.on('ai-highlight', ({ type, data }) => {
  this.emit('ai-highlight', { type, data });
});
```

packages/superdoc/src/components/AiLayer/AiLayer.vue
```diff
@@ -0,0 +1,94 @@
...
+
+<template>
+  <div class="ai-highlight-container" id="aiHighlightContainer">
+    <div class="ai-highlight-layer"></div>
+  </div>
+</template>
+
+<style>
+/* Global style for the prosemirror AI highlight */
+.ai-highlight {
```
artem-harbour: To avoid conflicts in the future, need to have more unique classes for global styles or use a prefix.

Like
`.sd-ai-highlight {}`

packages/superdoc/src/components/AiLayer/AiLayer.vue
```diff
@@ -0,0 +1,94 @@
...
+  };
+};
+
+const addAiHighlight = () => {
+  // Add the AI mark using the editor if available
+  if (props.editor && !props.editor.isDestroyed) {
+    props.editor.commands.insertAiMark();
+  } else {
+    // Fallback to DOM method if editor is not available
+    const layer = document.querySelector('.ai-highlight-layer');
```
artem-harbour: Need to use `ref` instead.

packages/superdoc/src/components/AiLayer/AiLayer.vue
```diff
@@ -0,0 +1,94 @@
...
+};
+
+const removeAiHighlight = () => {
+  // Remove the AI mark using the editor if available
+  if (props.editor && !props.editor.isDestroyed) {
+    props.editor.commands.removeAiMark();
+  } 
+  
+  // Always clear the DOM layer as a safety measure
+  const layer = document.querySelector('.ai-highlight-layer');
```
artem-harbour: Need to use `ref` instead.

packages/superdoc/src/SuperDoc.vue
```diff
@@ -338,10 +347,14 @@ onMounted(() => {
   if (isCommentsEnabled.value && !modules.comments.readOnly) {
     document.addEventListener('mousedown', handleDocumentMouseDown);
   }
+  proxy.$superdoc.on('ai-highlight-add', handleAiHighlightAdd);
+  proxy.$superdoc.on('ai-highlight-remove', handleAiHighlightRemove);
```
artem-harbour: A more general event is preferred.

packages/superdoc/src/SuperDoc.vue
```diff
@@ -604,6 +705,21 @@ watch(getFloatingComments, () => {
         />
       </div>
     </div>
+
+    <!-- AI Writer at cursor position -->
+    <div class="ai-writer-container" v-if="showAiWriter" :style="aiWriterPosition">
+      <AIWriter 
+        :selected-text="proxy.$superdoc.activeEditor && proxy.$superdoc.activeEditor.state ?
```
artem-harbour: Probably better to use composable here.

packages/superdoc/src/SuperDoc.vue
```diff
@@ -73,24 +74,30 @@ const layers = ref(null);
 const commentsLayer = ref(null);
 const toolsMenuPosition = reactive({ top: null, right: '-25px', zIndex: 101 });
 
+// Ai Layer
+const showAiLayer = ref(false);
```
artem-harbour: It is better to move properties of methods that relate to a separate feature to pinia store or composable.

packages/superdoc/src/SuperDoc.vue
```diff
@@ -505,6 +518,87 @@ watch(getFloatingComments, () => {
...
+    // If we have selected text, add AI highlighting
+    if (!selection.empty) {
+      // Emit the highlight event to trigger the AI highlighting
+      proxy.$superdoc.emit('ai-highlight-add');
+    }
+    
+    let coords;
+    try {
+      // Try to get coordinates from the selection head
+      coords = view.coordsAtPos(selection.$head.pos);
```
artem-harbour: It seems to me that working directly with PM should be inside the editor, but here we use only helpers and commands. @harbournick, what do you think?

packages/super-editor/vite.config.js
```diff
@@ -61,6 +62,7 @@ export default defineConfig(({ mode }) => {
             'toolbar': ['@components/toolbar/Toolbar.vue'],
             'super-input': ['@components/SuperInput.vue'],
             'file-zipper': ['@core/super-converter/zipper.js'],
+            'ai-writer': ['@components/toolbar/AIWriter.vue'],
```
artem-harbour, line 65: Why is this placed in a separate chunk?

packages/super-editor/src/extensions/ai/ai-marks.js
```diff
@@ -0,0 +1,74 @@
...
+    return {
+      id: {
+        default: null,
+        rendered: false,
+      },
+      class: {
+        default: null,
+        rendered: true,
+      },
+      'data-mark-id': {
```
artem-harbour: For attributes it is better to use camel case. And if we want to render it as a date attribute, then it should not be reflected in the PM attribute.

Example
```javascript
myId: {
  renderDOM: (attrs) => {
    if (!attrs.myId) return {};
    return {
      'data-my-id': attrs.myId,
    };
  },
},
```

packages/super-editor/src/extensions/ai/ai-plugin.js
```diff
@@ -0,0 +1,127 @@
...
+
+        tr.addMark(
+          $from.pos,
+          $to.pos,
+          this.editor.schema.marks[AiMarkName].create({
+            id: 'ai-highlight'
+          })
+        );
+
+        dispatch(tr);
```
artem-harbour: `if (dispatch) dispatch(tr)`

packages/super-editor/src/extensions/ai/ai-plugin.js
```diff
@@ -0,0 +1,127 @@
...
+          const aiMark = marks.find((mark) => mark.type.name === markName);
+
+          if (aiMark) {
+            markFound = true;
+            tr.removeMark(pos, pos + node.nodeSize, state.schema.marks[markName]);
+          }
+        });
+
+        if (markFound) {
+          dispatch(tr);
```
artem-harbour: `if (dispatch) dispatch(tr)`

packages/super-editor/src/extensions/ai/ai-plugin.js
```diff
@@ -0,0 +1,127 @@
...
+        const { doc } = state;
+        let markFound = false;
+
+        doc.descendants((node, pos) => {
+          const { marks = [] } = node;
+          const aiMark = marks.find((mark) => mark.type.name === markName);
+
+          if (aiMark) {
+            markFound = true;
+            tr.removeMark(pos, pos + node.nodeSize, state.schema.marks[markName]);
```
artem-harbour, line 49: I'm not sure if this command will work for many marks in a document and the positions won't be shifted when deleting, although it probably works differently for marks than for nodes. But I would suggest double checking this.

packages/super-editor/src/components/toolbar/AIWriter.vue
```diff
@@ -0,0 +1,454 @@
...
+<style scoped>
+/* Add isolation styles */
+.prosemirror-isolated {
+  /* Make sure the component is above ProseMirror in z-index */
+  z-index: 100;
+  position: relative;
+}
+
+.paper-plane {
+  --webkit-mask-image: url('@harbour-enterprises/common/icons/paper-plane-regular.svg');
```
artem-harbour: I think it's not a very good idea to use svg icons in css like that. It would be better to use inline svg instead.

packages/super-editor/src/components/toolbar/AIWriter.vue
```diff
@@ -0,0 +1,454 @@
...
+  if (aiWriterRef.value && !aiWriterRef.value.contains(event.target)) {
+    props.handleClose();
+  }
+};
+
+// Add ref for the textarea
+const editableRef = ref(null);
+
+// Save selection when component is mounted
+onMounted(() => {
```
artem-harbour: It is better to move the logic inside hooks into separate functions.

packages/super-editor/src/components/toolbar/AIWriter.vue
```diff
@@ -0,0 +1,454 @@
...
+const handleTextChunk = (text) => {
+  try {
+    // If this is the first chunk and we're rewriting, remove the selected text
+    if (props.selectedText && !textProcessingStarted.value) {
+      props.editor.commands.deleteSelection();
+      textProcessingStarted.value = true;
+    }
+
+    // If the text is null, undefined or empty, don't process it
+    if (text === null || text === undefined || text === '') {
```
artem-harbour, line 133:
```javascript
if (!text) return;
```

packages/super-editor/src/components/toolbar/ai-helpers.js
```diff
@@ -0,0 +1,301 @@
...
+ *       apiKey: 'your-harbour-api-key'
+ *     }
+ *   }
+ * };
+ * ```
+ */
+
+// @todo: Figure out logic for self hosted vs Harbour hosted and which endpoint
+// should be used based on that
+const API_ENDPOINT = 'https://api.myharbourshare.com/v2/insights';
```
artem-harbour: I'm pretty sure this shouldn't be hardcoded and be inside the repo.

packages/super-editor/src/components/toolbar/ai-helpers.js
```diff
@@ -0,0 +1,301 @@
...
+const GATEWAY_ENDPOINT = 'https://sd-dev-express-gateway-i6xtm.ondigitalocean.app/insights';
```
artem-harbour: This too.

packages/super-editor/src/components/toolbar/ai-helpers.js
```diff
@@ -0,0 +1,301 @@
...
+}
+
+/**
+ * UTILITY - Extracts content from a non-streaming response
+ * @param {Response} response - The API response
+ * @returns {Promise<string>} - The extracted content
+ */
+async function returnNonStreamingJson(response) {
+  const jsonResponse = await response.json();
+  if (jsonResponse.custom_prompt) return jsonResponse.custom_prompt[0].value;
```
artem-harbour:
```javascript
if (jsonResponse.custom_prompt) {
  return jsonResponse.custom_prompt[0].value;
} else {
  throw new Error('No custom prompt found in response');
}
```
OR
```javascript
if (jsonResponse.custom_prompt) return jsonResponse.custom_prompt[0].value;
else throw new Error('No custom prompt found in response');
```

artem-harbour (changes requested):
Reviewed the PR and added my comments. @harbournick, please review too.

packages/super-editor/src/components/toolbar/ai-helpers.js
palmer-cl: I may have misunderstood the requirements, but I thought we actually did want all instances to hit this endpoint so that we could track usage and metrics. 

Happy to make this configurable if thats what the team wants.

artem-harbour: Hey @palmer-cl, I am not familiar with the requirements of this feature, just reviewed what I see. Let's wait for @harbournick's review too.

However, it seems to me that such URLs should always be configurable and provided via props, config, etc (in my opinion).

packages/superdoc/src/core/SuperDoc.js
palmer-cl: great idea. updated in 0c27840ae55bf89d69584aaba297d4b7c30a5b79

packages/superdoc/src/core/SuperDoc.js
palmer-cl: sounds good! added in a2a94fc3a1e7a42a600c37cd317fe66869f6a14c

packages/superdoc/src/components/AiLayer/AiLayer.vue
palmer-cl: Yes! Was able to prefix and was also able to remove one class and just do the styles inline. 112c087a1587df7e7ff9517556524adc07a842e9

packages/superdoc/src/SuperDoc.vue
palmer-cl: This also has been updated here: a2a94fc3a1e7a42a600c37cd317fe66869f6a14c

packages/superdoc/src/components/AiLayer/AiLayer.vue
palmer-cl: Added ref in e40adf269f9e2333f9b994368a223c53cb51a844

packages/superdoc/src/components/AiLayer/AiLayer.vue
palmer-cl: Added ref in e40adf269f9e2333f9b994368a223c53cb51a844

packages/super-editor/vite.config.js
palmer-cl: AiWriter is built as a component in SuperEditor, but rendered directly in both SuperDoc (side tool menu) and SuperEditor (toolbar). In order to import it in SD, im exporting from SE. 

If theres a better way to do this let me know.

packages/super-editor/src/extensions/ai/ai-plugin.js
palmer-cl: Updated in 3550f8033776bab508d6c83d81d8c28d85c128ac

packages/super-editor/src/extensions/ai/ai-plugin.js
palmer-cl: Updated in 3550f8033776bab508d6c83d81d8c28d85c128ac

packages/super-editor/src/components/toolbar/ai-helpers.js
palmer-cl: Fixed conditional in 28d5adebfcd787710c23f2cc059d5d221cc35097

packages/super-editor/vite.config.js
harbournick: @palmer-cl what if we placed the AI writer component inside the super-editor package and handle everything there? Since the AI writer has to call prosemirror commands (+ look for an active editor every time), it seems better composed entirely within that package, and then superdoc just needs to turn it on/off via config flag. 

This also relates to @artem-harbour's comment below in **showAiWriterAtCursor**. 

Aside from a bit more of a learning curve, I think it should be a fairly simple refactor to move the logic there but LMK if you see major issues there. I would suggest to `cd packages/super-editor` and run the `npm run dev` from there which will run the DeveloperPlayground.dev component, which in turn is just importing the SuperEditor.vue component and Editor.js directly.

If everything is handled inside Editor.js, we likely don't need this export either?

I'll also ping you in case you want to chat through this.

harbournick: The other option here, in the interest of time, could be to move forward with the architecture as-is, and then I could refactor it to put everything inside Editor, which I'm happy to do if it helps move this along @palmer-cl

packages/superdoc/src/SuperDoc.vue
harbournick: Yeah, I agree. See my comment above for more. I think we would benefit from an additional Editor extension to centralize AI related things, and moving everything to the Editor package.

harbournick (changes requested):
See my comments below. We currently try to keep a separation between what is Editor.js and what is SuperDoc which is not necessarily obvious.

harbournick:
@palmer-cl and I discussed my comments, and we will move forward after @artem-harbour's review without architectural changes, and I will make the necessary changes in an additional PR.

packages/superdoc/src/SuperDoc.vue
palmer-cl: Update to composable pattern here: 27ba4eb247e42ecf621ced0a91af7088805f063f

packages/superdoc/src/SuperDoc.vue
palmer-cl: Refactored all of the AI logic into a single composable: c6db2d3beca4a7e7e2844115b119a254c047a177

packages/superdoc/src/SuperDoc.vue
palmer-cl: Synced with @harbournick - he is going to move this after merging.

packages/super-editor/vite.config.js
palmer-cl: @harbournick feel free to edit or move this when you do the SD / SE changes.

packages/super-editor/src/extensions/ai/ai-marks.js
palmer-cl: 274fd419 updates to use camel case. 

Were not using it as a date attribute per se, the date function is just generating a unique ID so that PM doesnt merge the marks together. This allows the "streaming" affect of text to animate in rather than animate all in one large block of text.

packages/super-editor/src/extensions/ai/ai-plugin.js
palmer-cl: Looking at my element tree here i see 2 <aiMark> wrappers when split by a <p>. 

image

Clicking off the text and calling removeAiMark does remove both marks:

image

Is that what you mean?

packages/super-editor/src/components/toolbar/AIWriter.vue
palmer-cl: I think the original solution is ok, seems cleaner to me, but just in case others feel inline svg is better I updated styles in 15f29604

packages/super-editor/src/components/toolbar/AIWriter.vue
palmer-cl: Added helper functions to remove from inside the hooks here 53c261d6

packages/super-editor/src/components/toolbar/ai-helpers.js
palmer-cl: Made this configurable in 5f6c8dfedbb0de22df014f3936d85f5d4a3d351d with default to the gateway

@harbournick @edoversb do we want to default to our intermediary service? or the harbour endpoint directly? or no default?

packages/super-editor/src/components/toolbar/ai-helpers.js
palmer-cl: Conversation above.

palmer-cl:
@harbournick it looks like https://github.com/Harbour-Enterprises/SuperDoc/pull/444/commits/c6db2d3beca4a7e7e2844115b119a254c047a177 caused the e2e tests to fail. The reactivity chain may have been altered slightly when moving the state variables and refs into a composable pattern. Would you mind confirming if we should update the test, or change the code.

packages/super-editor/src/extensions/ai/ai-plugin.js
artem-harbour: If that works, then everything is ok. It seems that deleting several marks does not shift positions and deleting the following marks is successful.

In the case of nodes, I do something like this.
- Collect the needed nodes with positions into an array.
- In a loop, I go through the nodes and map positions between transactions.
```javascript
// in a loop
let newPosFrom = tr.mapping.map(pos); // map the position between transaction steps
let newPosTo = tr.mapping.map(pos + node.nodeSize);

let currentNode = tr.doc.nodeAt(newPosFrom);
if (node.eq(currentNode)) {
  tr.delete(newPosFrom, newPosTo);
}
```

packages/super-editor/src/components/toolbar/AIWriter.vue
```diff
@@ -0,0 +1,413 @@
...
+<style scoped>
+/* Add isolation styles */
+.prosemirror-isolated {
+  /* Make sure the component is above ProseMirror in z-index */
+  z-index: 100;
+  position: relative;
+}
+
+/* Base icon styles */
+[class^="fa-"] {
```
artem-harbour, line 299: Minor, but as far as I know font icons are no longer used. This is most likely unnecessary.

harbournick (approved):
LGTM

harbournick:
Thanks for all the hard work on this one @palmer-cl !

palmer-cl:
Thanks @harbournick and @artem-harbour ! Credit to the complete team effort!

PR #450, opened 2025-04-09
HAR-9472 - show pages when pdf document is rendered

artem-harbour:
This causes issues with document resizing and fields positions in mobile. Before resizing the document for mobile width, we need to show the pages at their original size so that the fields have the correct positions. 

We will show the loader first until all pages are rendered and only then append.


harbournick (approved):
LGTM

PR #451, opened 2025-04-09
HAR-9488 - Allow external toolbar customization, add customization api

harbournick:
- allow customizing (including/excluding) which toolbar buttons to render in a given toolbar
- move config to 'modules' for superdoc


edoversb (approved):
LGTM

PR #452, opened 2025-04-10
HAR-9432  Formatting - Line spacing

VladaHarbour:


packages/super-editor/src/core/super-converter/v2/importer/paragraphNodeImporter.js
```diff
@@ -49,33 +49,50 @@ export const handleParagraphNode = (params) => {
 
   const pPr = node.elements?.find((el) => el.name === 'w:pPr');
   const styleTag = pPr?.elements?.find((el) => el.name === 'w:pStyle');
+  const nestedRPr = pPr?.elements?.find((el) => el.name === 'w:rPr');
```
VladaHarbour, line 52: @harbournick sometimes empty paragraphs could have run properties with font details defined. We don't parse it and that is why SuperToolbar shows default font for such paragraphs. Since there is no text we can't apply marks. I decided to implement it through additional attributes but will be happy to hear your thoughts
@artem-harbour if you have any ideas or suggestions I will be happy to hear as well

packages/super-editor/src/core/super-converter/v2/importer/paragraphNodeImporter.js
```diff
@@ -117,17 +137,27 @@ export const getParagraphSpacing = (node, docx) => {
 
   const beforeSpacing = inLineSpacing?.['w:before'] || lineSpaceBefore || pDefaultSpacing?.['w:before'];
   if (beforeSpacing) spacing.lineSpaceBefore = twipsToPixels(beforeSpacing);
+
```
VladaHarbour, line 142: @harbournick I tried to parse spaces and indents as accurate as possible but here is a document with w:beforeAutospacing and w:afterAutospacing defined. As far as I understood Word has its own way to handle it and I didn't found any specifics about it. I parse spaces based on font size in this case (according chatgpt suggestion).

harbournick (approved):
LGTM

PR #455, opened 2025-04-14
HAR-9430 - fix paragraph text indentation

artem-harbour:
The first line indent should not be mixed and used for the overall indentation of the entire paragraph.


harbournick (approved):
LGTM

PR #456, opened 2025-04-15
HAR-9430 - fix tab styles update

artem-harbour:


packages/super-editor/src/extensions/tab/tab.js
```diff
@@ -82,7 +81,7 @@ const getTabDecorations = (state, view) => {
       let textWidth = 0;
 
       try {
-        view.state.doc.nodesBetween(pos - prevNodeSize - 1, pos - 1, (node, nodePos) => {
```
artem-harbour, line 85: Looks like this was the cause of the possible errors.

It seems to me that view here still contains the old state, and we take positions from the new state (from args) for iteration.

packages/super-editor/src/extensions/tab/tab.js
```diff
@@ -55,7 +55,6 @@ export const TabNode = Node.create({
         },
 
         apply(tr, oldDecorationSet, oldState, newState) {
-          if (!tr.docChanged) return oldDecorationSet;
```
artem-harbour, line 58: I think this check should be removed otherwise the tab styles are not updated when the document is initialized.

artem-harbour:
@VladaHarbour, please check the changes for tab extension only.

VladaHarbour (approved):

artem-harbour:
fyi @harbournick rebased the branch. Ready to review/merge.

harbournick (approved):
LGTM

PR #458, opened 2025-04-15
HAR-8863 - Lists v1

harbournick:
Next iteration of lists import/export/rendering
- Fixes many issues around lists that are broken by other block nodes (ie: paragraph, table)
- Fixes sublists that are broken by other block nodes (ie: paragraph, table)
- adds import list tracking for list continuity
- adds list definitions from styles.xml 
- adds 24 more tests (27% increase) all around import/export of lists]
- adds framework for testing list editor commands (to test generated lists, plus export)


packages/super-editor/src/core/extensions/keymap.js
```diff
@@ -2,6 +2,36 @@ import { Extension } from '../Extension.js';
 import { isIOS } from '../utilities/isIOS.js';
 import { isMacOS } from '../utilities/isMacOS.js';
 
+export const handleEnter = (editor) => {
```
harbournick, line 5: note: separating these into exported functions allows us to call these from inside our unit tests, therefore allowing us to easily edit test docs and test at each stage (ie: add list, add another node, hit enter, hit enter, add another list item, export the whole thing, test the export)

edoversb (approved):
LGTM

PR #463, opened 2025-04-17
HAR-9454 - take into account borders for tableWrapper width

artem-harbour:


packages/super-editor/src/extensions/table/TableView.js
```diff
@@ -121,3 +135,31 @@ function updateTable(editor, node, table) {
...
+
+  let borderLeftMax = parseFloat(table.style.borderLeftWidth || 0);
+  let borderRightMax = parseFloat(table.style.borderRightWidth) || 0;
+
+  let firstColumnCells = [...table.querySelectorAll(':scope > tbody > tr > td:first-child')];
+  let lastColumnCells = [...table.querySelectorAll(':scope > tbody > tr > td:last-child')];
+
+  for (let cell of firstColumnCells) {
+    let borderLeft = parseFloat(cell.style.borderLeftWidth) || 0;
+    borderLeftMax = Math.max(borderLeftMax, borderLeft);
```
artem-harbour, line 155: Note: We also need to take into account the left and right cell borders, because if their width is greater than the table border width, they are added to the table width.

PR #464, opened 2025-04-18
HAR-9454 - handle table colspan and colgroup

artem-harbour:


packages/super-editor/src/core/super-converter/v2/importer/tableImporter.js
```diff
@@ -159,10 +160,33 @@ export function handleTableCellNode(
...
+
+        if (typeof colwidthValue !== 'undefined') {
+          colwidth.push(colwidthValue);
+        } else {
+          colwidth.push(defaultColwidth);
+        }
+      }
+
+      if (colwidth.length) {
+        attributes['colwidth'] = [...colwidth];
```
artem-harbour, line 183: If the column has `colspan` greater than 1, then it is important that `colwidth ` attribute has this form `[200, 400]` (not `[600]` for colspan 2).

artem-harbour: Or `[200,300,300]` for colspan 3 (not `[800]`).

packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -992,17 +994,32 @@ function generateTableBorders(node) {
...
+    colgroup = colgroupValues;
+  } catch(err) {
+    colgroup = [];
+  }
+
   const elements = [];
-  gridColumnWidths?.forEach((width) => {
+  colgroup?.forEach((width) => {
     elements.push({
       name: 'w:gridCol',
```
artem-harbour, line 1018: It is important to save `w:gridCol` element for the table, it was not saved before. Now we generate the same values ​​that were for the table node in the editor.

artem-harbour:
@harbournick this is ready for review.

harbournick (approved):
LGTM

PR #465, opened 2025-04-18
Refactor header/footer import logic

harbournick:
Refactor header/footer import logic to account for documents that have multiple <w:sectPr> tags across document.xml


edoversb (approved):
LGTM

PR #466, opened 2025-04-18
HAR-8863 - Fix missing custom list numbering

harbournick:
- Fix missing custom list numbering (ie: SECTION 1, (a), etc)
- Fix stylePr overriding numId
- Add test for custom numbering
- fix styles


edoversb (approved):
LGTM

PR #467, opened 2025-04-21
Fixes various recursive list definitions, outline numbering

harbournick:
- Fixes various issues around recursive list definitions
- adds support for outlineLvl instead of ilvl tags


edoversb (approved):
Lgtm

PR #471, opened 2025-04-23
HAR-9454 - fix text centering/wrapping in cells

artem-harbour:


packages/super-editor/src/assets/styles/elements/prosemirror.css
```diff
@@ -157,22 +157,18 @@ https://github.com/ProseMirror/prosemirror-tables/blob/master/demo/index.html
...
   box-sizing: border-box;
+  overflow-wrap: anywhere;
 }
 
 .ProseMirror th {
   font-weight: bold;
   text-align: left
 }
 
-.ProseMirror td p {
```
artem-harbour, line 171: Unnecessary styles, width 100% for p breaks styles inside cells.

packages/super-editor/src/assets/styles/elements/prosemirror.css
```diff
@@ -157,22 +157,18 @@ https://github.com/ProseMirror/prosemirror-tables/blob/master/demo/index.html
 .ProseMirror td,
 .ProseMirror th {
   min-width: 1em;
-  padding: 3px 5px;
+  padding: 3px 2px;
   position: relative;
   vertical-align: top;
   box-sizing: border-box;
+  overflow-wrap: anywhere;
```
artem-harbour, line 164: Approximately the same as 'break-word'.
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#anywhere

packages/super-editor/src/assets/styles/elements/prosemirror.css
```diff
@@ -157,22 +157,18 @@ https://github.com/ProseMirror/prosemirror-tables/blob/master/demo/index.html
 .ProseMirror td,
 .ProseMirror th {
   min-width: 1em;
-  padding: 3px 5px;
+  padding: 3px 2px;
```
artem-harbour, line 160: The side padding has been reduced, but I would prefer to leave it for now, otherwise tables created from scratch and new cells will not have it. It's probably a good idea to remove this when it becomes possible to edit table properties.

Note:
- Usually the cells have inline padding, but if it is not, then sometimes the style raises issues.

harbournick (approved):
LGTM

PR #472, opened 2025-04-23
Adds an Editor.js HTML example

harbournick:
- HTML example of using Editor.js directly
- Minor API tweaks to allow more selector types in SuperToolbar and SuperEditor


edoversb (approved):
LGTM

PR #473, opened 2025-04-23
HAR-9556 - Updates to SuperDoc AP

harbournick:
- Adds 'document' key to pass in single file to superdoc (does not remove documents key)
- Adds even simpler default blank docx init
- Adds HTML-to-docx init in SuperDoc


edoversb (approved):
lgtm

PR #474, opened 2025-04-23
HAR-9556 - Simpler docx API, docx template with html content

harbournick:
- Simplifies API even further by allowing 'document: /my-docx.docx' to be passed in to default to docx doc
- Adds support to load docx document just from html
- Adds support to load an existing docx document but replace content with html (ie: a template, with html content)
- Adds example of using SuperDoc with a .docx template and HTML content


edoversb (approved):
LGTM

PR #480, opened 2025-04-25
Fix parsing error in Ideals doc

VladaHarbour:
@harbournick This document from Ideals has font inconsistencies but document itself has no font definitions. SD renders Arial but Word has Times new roman. As far as I understood it uses 'Normal' styles for paragraphs and 'Heading 1'. Probably this document was created in older version of Word which had Times new roman as default font but I didn't find any signs of it in xml


harbournick (approved):
LGTM

PR #481, opened 2025-04-25
[WIP] HAR-9460 - handle shape textbox

artem-harbour:


artem-harbour:
@harbournick - ready for review.

PR #482, opened 2025-04-29
HAR-9577 - Serialize/deserialize external/custom marks

harbournick:
- Add a mechanism to serialize/deserialize custom external marks


edoversb (approved):
Lgtm

PR #486, opened 2025-04-29
HAR-9460 - handle alternate content node

artem-harbour:
Changes from this PR are used here as a basis.
https://github.com/Harbour-Enterprises/SuperDoc/pull/478

### DIFF
--- a/packages/super-editor/src/core/Editor.js
+++ b/packages/super-editor/src/core/Editor.js
@@ -690,6 +690,7 @@ export class Editor extends EventEmitter {
     element.style.paddingLeft = pageMargins.left + 'in';
     element.style.paddingRight = pageMargins.right + 'in';
     element.style.boxSizing = 'border-box';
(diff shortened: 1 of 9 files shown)


packages/super-editor/src/core/Editor.js
```diff
@@ -690,6 +690,7 @@ export class Editor extends EventEmitter {
     element.style.paddingLeft = pageMargins.left + 'in';
     element.style.paddingRight = pageMargins.right + 'in';
     element.style.boxSizing = 'border-box';
+    element.style.isolation = 'isolate'; // to create new stacking context.
```
artem-harbour, line 693: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Stacking_context

To make `z-index: -1` work for the element. This is probably a safer way than assigning z-index to the editor.

artem-harbour:
FYI @harbournick - this is ready for review.

harbournick (approved):
LGTM

PR #492, opened 2025-05-07
HAR-9637 - SD header/footer fixes

harbournick:
- Fixes around haders/footers when there's a sectPr page break
- Fixes to first page header/footer id
- Adds page breaks on sectPr


edoversb (approved):
lgtm

PR #502, opened 2025-05-14
HAR-9653 - improve import and export of html annotation

artem-harbour:
Added comments in the code explaining the changes.

### DIFF
--- a/packages/super-editor/src/core/super-converter/exporter.js
+++ b/packages/super-editor/src/core/super-converter/exporter.js
@@ -1731,17 +1731,25 @@ function prepareCheckboxAnnotation(params) {
  */
 function prepareHtmlAnnotation(params) {
   const {
-    node: { attrs = {} },
+    node: { attrs = {}, marks = [] },
+    editorSchema,
   } = params;
 
   const parser = new window.DOMParser();
   const paragraphHtml = parser.parseFromString(attrs.rawHtml || attrs.displayLabel, 'text/html');
+  const marksFromAttrs = translateFieldAttrsToMarks(attrs);
+  const allMarks = [...marks, ...marksFromAttrs]
 
-  const state = EditorState.create({
-    doc: PMDOMParser.fromSchema(params.editorSchema).parse(paragraphHtml),
+  let state = EditorState.create({
+    doc: PMDOMParser.fromSchema(editorSchema).parse(paragraphHtml),
   });
 
+  if (allMarks.length) {
+    state = applyMarksToHtmlAnnotation(state, allMarks);
+  }
+
   const htmlAnnotationNode = state.doc.toJSON();
+  
   return {
     name: 'htmlAnnotation',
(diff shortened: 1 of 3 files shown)


packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -1896,6 +1904,20 @@ function translateFieldAnnotation(params) {
               'w:val': attrs.multipleImage,
             },
           },
+          {
```
artem-harbour, line 1907: Save fontFamily and fontSize as attributes to re-import annotations correctly.

packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -2103,3 +2125,46 @@ function resizeKeepAspectRatio(width, height, maxWidth) {
   }
   return { width, height };
 }
+
+function applyMarksToHtmlAnnotation(state, marks) {
```
artem-harbour, line 2129: Apply font marks (*attributes as marks) of html annotation to inner content of annotation, but do not override if text node already has font styles. This is necessary for the final doc export, so the paragraph text look the same (as annotation).

packages/super-editor/src/core/super-converter/v2/importer/annotationImporter.js
```diff
@@ -34,23 +36,27 @@ export const handleAnnotationNode = (params) => {
     fieldType,
     fieldColor,
     multipleImage: isMultipleImage === 'true',
+    fontFamily: fontFamily !== 'null' ? fontFamily : null,
+    fontSize: fontSize !== 'null' ? fontSize : null,
   };
 
+  const allAttrs = { ...attrs, ...marksAsAttrs };
```
artem-harbour, line 43: For regular text fields, fontFamily and fontSize will be in `attrs` and `marksAsAttrs`. It will be overridden from the last object.

For html fields, fontFamily and fontSize will be only in `attrs` object.

packages/super-editor/src/core/super-converter/v2/importer/annotationImporter.js
```diff
@@ -65,8 +71,20 @@ export const handleAnnotationNode = (params) => {
...
  */
-export const parseAnnotationMarks = (content = {}) => {
-  const run = content.elements?.find((el) => el.name === 'w:r');
+export const parseAnnotationMarks = (content = {}, type) => {
+  let mainContent = content;
+
+  if (type === 'html') {
+    /// Note: html annotation has a different structure and can include 
+    /// several paragraphs with different styles. We could find the first paragraph 
+    /// and take the marks from there, but we take fontFamily and fontSize from the annotation attributes.
```
artem-harbour, line 80: There is no processing for html field, since getting fontFamily and fontSize from attributes is a more correct solution in my opinion.

harbournick (approved):
LGTM

PR #506, opened 2025-05-15
Add jsdoc annotations

johanneswilm:
First attempt to add jsdoc annotations to Editor.js and super-toolbar.js.
In some cases, I was not entirely sure, for example when arguments are not in use anywhere in a given method like here: 
https://github.com/Harbour-Enterprises/SuperDoc/compare/main...johanneswilm:SuperDoc:add-jsdoc-annotations?expand=1#diff-cf578a6f4b5c61edc4820b992b2ecf85ac935fd7c6019e26b6b59076c7f0067fL563-L566


packages/super-editor/src/core/Editor.js
```diff
@@ -988,9 +1243,9 @@ export class Editor extends EventEmitter {
   /**
    * Perform any post conversion pre prosemirror import processing.
    * Comments are processed here.
-   * 
-   * @param {import('prosemirror-model').Node} doc The prosemirror document
-   * @returns {import('prosemirror-model').Node} The updated prosemirror document
+   * @private
+   * @param {Object} doc The prosemirror document
+   * @returns {Object} The updated prosemirror document
```
chittolinag, line 1264: @johanneswilm Hey! Gabriel here from Harbour 👋  Thanks for the help here!

Just to understand better - why exactly did this change?

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -13,7 +13,61 @@ import { getQuickFormatList } from '@extensions/linked-styles/linked-styles.js';
 import { getAvailableColorOptions, makeColorOption, renderColorOptions } from './color-dropdown-helpers.js';
 import { isInTable } from '@helpers/isInTable.js';
 
+/**
+ * @typedef {Object} ToolbarConfig
+ * @property {string} [selector] - CSS selector for the toolbar container
```
chittolinag, line 18: I have a feeling that the `selector` parameter is not optional here. Can you please double check @johanneswilm ? I did a quick search and on every instance I found the `selector` is given as a parameter.

chittolinag: Please double check the rest of them too.

chittolinag: @harbournick I believe we also need an `editor` parameter here, can you please help clarifying? I found 3 instances where we are passing an `editor` parameter when creating a new instance of `SuperToolbar` - but I'm not quite sure if that's an optional param or not.

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -13,7 +13,61 @@ import { getQuickFormatList } from '@extensions/linked-styles/linked-styles.js';
...
+ * @property {string[]} [excludeItems=[]] - Items to exclude from the toolbar
+ * @property {Object} [groups=null] - Custom groups configuration
+ * @property {Object} [editor=null] - The editor instance
+ * @property {string} [aiApiKey=null] - API key for AI integration
+ * @property {string} [aiEndpoint=null] - Endpoint for AI integration
+ */
+
+/**
+ * @typedef {Object} ToolbarItem
+ * @property {Object} name - The name of the toolbar item
```
chittolinag, line 35: I think we might be missing some attributes here (see `type` for example) - can you please cross check with the `use-toolbar-item.js` composable?

packages/super-editor/src/core/Editor.js
johanneswilm: Hey @chittolinag ,
I ran it through JSDoc and it marked this as invalid syntax. I then looked through the JSDoc documentation and didn't find an option to do this either. Based on the remainder of the sources, it seems like every time a more complex object related to ProseMirror is described, `{Object}` is used.

However - I am not sure how you use the annotations. If you only need a syntax in the style of JSDoc without actually running JSDoc, or maybe you run JSDoc in some customized way, then this syntax may be fine.

johanneswilm: I see this syntax is mentioned in the TypeScript documentation on JSDoc [1]. So it seems this is valid in the TypeScript variant of JSDoc, but not according to the generic JSDoc syntax [2]. So yes, if this is to make it available to TypeScript, then it will have to look differently. In that case there will be a number of other objects that have to be written differently.

[1] https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#import-types

[2] https://jsdoc.app/

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -105,36 +181,86 @@ export class SuperToolbar extends EventEmitter {
...
+     */
     setHighlight: ({ item, argument }) => {
       this.#runCommandWithArgumentOnly({ item, argument });
     },
 
+    /**
+     * Toggles the ruler visibility
+     * @param {Object} params - Command parameters
+     * @param {CommandItem} params.item - The command item
+     * @param {*} params.argument - Not used
```
chittolinag: It turns out we don't use any of the parameters. Instead of documenting them, can we remove them completely?

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -105,36 +181,86 @@ export class SuperToolbar extends EventEmitter {
...
+    /**
+     * Initiates the image upload process
+     * @async
+     * @param {Object} params - Command parameters
+     * @param {CommandItem} params.item - The command item
+     * @param {*} params.argument - Not used
+     * @returns {Promise<void>}
+     */
     startImageUpload: async ({ item, argument }) => {
       let open = getFileOpener();
```
chittolinag, line 323: Same as above, let's remove unused parameters.

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -232,13 +400,28 @@ export class SuperToolbar extends EventEmitter {
...
+     * @returns {void}
+     */
     insertTable: ({ item, argument }) => {
       this.#runCommandWithArgumentOnly({ item, argument });
     },
 
+    /**
+     * Executes a table-related command
+     * @param {Object} params - Command parameters
+     * @param {CommandItem} params.item - The command item
```
chittolinag, line 475: This doesn't seem to be used. Can we remove?

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -261,21 +454,40 @@ export class SuperToolbar extends EventEmitter {
 
   /**
    * The toolbar expects an active Super Editor instance.
-   * @param {*} editor
+   * @param {Object} editor - The editor instance to attach to the toolbar
```
chittolinag, line 515: We can use the `Editor` type here

packages/super-editor/src/core/Editor.js
```diff
@@ -1030,6 +1285,12 @@ export class Editor extends EventEmitter {
 
   /**
    * Export the editor document to DOCX.
+   * @async
+   * @param {Object} options - The export options
+   * @param {boolean} [options.isFinalDoc=false] - Whether this is the final document version
+   * @param {string} [options.commentsType] - The type of comments to include
+   * @param {Array} [options.comments=[]] - Array of comments to include in the document
```
chittolinag, line 1308: Not sure if this within the scope of your task here, but it would be useful to add typing for the comments. @harbournick cc ^

packages/super-editor/src/core/Editor.js
```diff
@@ -549,6 +728,15 @@ export class Editor extends EventEmitter {
   /**
    * Load the data from DOCX to be used in the schema.
    * Expects a DOCX file.
+   * @static
+   * @async
+   * @param {File|Blob|Buffer} fileSource - The DOCX file to load (File/Blob in browser, Buffer in Node.js)
+   * @param {boolean} [isNode=false] - Whether the method is being called in a Node.js environment
+   * @returns {Promise<Array>} - A promise that resolves to an array containing:
+   *   - [0] xmlFiles - Array of XML files extracted from the DOCX
```
chittolinag, line 752: Can we rewrite the `returns` part? Example: 
```js
* @returns {Promise<Array<{xmlFiles: Array<string>, mediaFiles: Object, mediaFilesBase64: Object, fonts: Object}>>} - Array of objects containing xmlFiles, mediaFiles, mediaFilesBase64, and fonts

```

Before:

Screenshot 2025-05-20 at 09 03 45

After:
Screenshot 2025-05-20 at 09 05 29

packages/super-editor/src/core/Editor.js
```diff
@@ -250,74 +382,88 @@ export class Editor extends EventEmitter {
...
    * Get the editor element
+   * @returns {HTMLElement} The editor element
    */
   get element() {
     return this.options.element;
   }
 
   /**
    * Get possible users of the editor.
+   * @returns {Array.<Object>} List of users
```
chittolinag: We have a `User` type that could be used here. See `packages/superdoc/src/core/SuperDoc.js`.

packages/super-editor/src/core/Editor.js
```diff
@@ -26,19 +26,109 @@ import DocxZipper from '@core/DocxZipper.js';
...
 
+  /**
+   * ProseMirror schema for the editor
+   * @type {Object}
+   */
   schema;
 
+  /**
+   * ProseMirror view instance
+   * @type {Object}
```
chittolinag, line 134: I am not super familiar with `prosemirror` tbh - but I wonder if we can reference their types here? Same with `schema` above.

packages/super-editor/src/core/Editor.js
```diff
@@ -26,19 +26,109 @@ import DocxZipper from '@core/DocxZipper.js';
 /**
  * Editor main class.
  */
+/**
+ * @typedef {Object} EditorOptions
+ * @property {HTMLElement} [element] - The container element for the editor
```
chittolinag, line 47: I have a feeling that at least `element` or `selector` must be provided. Even if they are initialized with `null` values, when initializing (constructor method) we need either one or the other. Please ensure we are not adding other important values marked as optional when they are required - please let us know if you have any questions here.

chittolinag (changes requested):
@johanneswilm thanks a lot for the help here! I know it's not a trivial task to add types for a somewhat unfamiliar codebase... I added a few comments and requested some changes - please let us know if you have any questions.

@harbournick cc ^

packages/super-editor/src/core/Editor.js
chittolinag: @johanneswilm got it, and yeah, that's true. Our main goal here is to enable rich support for TS users, so anything we can do to help them infer types without the need of reverse-engineering, we'll try to do so. 

However, I see this is a private function, so you can leave as is as it won't be exposed to customers anyways.

johanneswilm: Hey @chittolinag,
ok, yeah so given that there is a different purpose for this than what I understood (running JSDoc), a lot of this will have to be redone from scratch. I can do thatnext.
I just wonder - if this is for TS rather than JSDoc, would it not maybe make more sense to create TS bindings directly? I'm maintaining two libraries that I've convertes to TS. Even that is not that hard. 

Or maybe at least we should document somewhere what the purpose is? (maybe it is and I overlooked that.)

chittolinag: By TS bindings you mean creating the TS definition files manually, right? I personally like that idea as we would be able to create a richer experiences for customers, especially given we can't achieve everything TS supports with JS Doc.  At the beginning I gave up on that idea because it would require more effort and not everyone in the team is familiar with TypeScript, but that might be a good solution for the long term. I'll run this with our team and will ping you back.

chittolinag: Just to give you some more context: 

We probably won't use TypeScript in SuperDoc, but we still want to create a great experience for users that use TypeScript. We ended up using JS Docs for that purpose because it not only allow us to generate TS types automatically, but also because it helps us, maintainers of the library, to have a better coding experience. For example, VS Code picks up types from JS Doc, which helps a lot with intellisense and more. 

Back then I gave up on the idea of defining the types manually because of what I said above, not everyone in the team is familiar with TypeScript, and this would require more manual effort. JS Doc is something that is already part of our development process in general, so it would just feel smoother.  However, I'm running this again with our team to figure out what makes more sense for us.

Have you ever done something similar? Ideas are super welcome here.

johanneswilm: Hey, sure. In the end any of these solutions will work. I think I was given this task to get to know the source code a little bit. I think I am here due to my experience with ProseMirror and pagination, not because I need to change your policy on TS. I am just curious asking question more than anything. I hope I can be of help when it comes to ProseMirror, pagination, input events, etc. I can work with any of the possibly ways to connect to TS, but I think it's probably a good idea to document the fact that this project is using one particular variant of JSDoc annotations and why that is.
As for code editor - I am switching between Pulsar and Zed. I don't like that VSCode is doimng too much stuff in the background that I feel is difficult to control. :) But I get that the most devs will benefit from things that integrate with VSCode.

chittolinag: Sure, no problem!  Let's tackle that separately. 

If you can please take a look at the other comments I made and submit the fixes, I'll review again and merge the PR. And then we can take a look at these other optimizations for TypeScript later.

harbournick: hey @johanneswilm great to connect on here! And thank you for the great work. This is a lot of great feedback for thought and we're always looking for ways to improve this. Will discuss with the team. Resolving this thread.

packages/super-editor/src/components/toolbar/super-toolbar.js
harbournick: Let's make selector required here.
But technically editor is optional. I'm cool with it as-is for that one

packages/super-editor/src/components/toolbar/super-toolbar.js
harbournick: +1 let's remove it then

packages/super-editor/src/components/toolbar/super-toolbar.js
harbournick: I'm ok with this one, we may decide to use it (and its available)

packages/super-editor/src/core/Editor.js
harbournick: In theory, yes, but its actually _technically_ fine to not pass anything here so I'm ok with optional. We generate the element dynamically if this is not passed in. Not really used anywhere right now but could be useful in the future.

packages/super-editor/src/core/Editor.js
harbournick: nice! yes let's do that

packages/super-editor/src/core/Editor.js
harbournick: _very_ useful, but let's leave it for a future PR - thanks for flagging!

packages/super-editor/src/components/toolbar/super-toolbar.js
harbournick: Deleted my previous comment. I'm actually fine with this as-is. We will generate the element dynamically if no element is passed in shortly so this will, in fact, be optional.

packages/super-editor/src/components/toolbar/super-toolbar.js
harbournick: Good call - @johanneswilm can we create a typedef for ToolbarItem so we don't need to use Object here? I think that will be very useful!

harbournick:
@johanneswilm thanks for the great work here! I went through and resolved any comments no longer needed. Could you please resolve the merge conflict and resolve the final 5 comments? Thank you!

johanneswilm:
will do

packages/super-editor/src/components/toolbar/super-toolbar.js
johanneswilm: Hey @harbournick - sorry, but I am a little confused. In line 32 above I created a typedef for ToolbarItem. Is that not the same thing you are asking for?

packages/super-editor/src/components/toolbar/super-toolbar.js
johanneswilm: Hey,
I don't know your entire source code and the source code of all commercial extensions that have been made, but I can see that a lot of these editing commands are all called with a single argument that contains (at least) the parameters `item` and `argument`. For some of these commands these are not actually needed - like in this case. However, for consistency, all the commands are likely still all called with the same argument as al other commands. 

Of course it's possible to define the method to just not receive any argument at all and it won't fail if it is still called with an argument. However, in a TS or otherwise linted or typechecked environment, this is likely going to cause to throw an error as now this function is called with an argument while it is defined without one.

So instead I would propose something like this:

```js
  /**
     * Toggles the ruler visibility
     * @param {Object} _params - Command parameters (not used)
     * @returns {void}
     */
    toggleRuler: (_params) => {
      this.superdoc.toggleRuler();
    },
```

The underscore at the beginning of `_params` marks that this argument is not actually used in the function. 

If we want to allow calling this function without any arguments, we could even make it:

```js
  /**
     * Toggles the ruler visibility
     * @param {Object} _params - Command parameters (not used)
     * @returns {void}
     */
    toggleRuler: (_params=null) => {
      this.superdoc.toggleRuler();
    },
```

packages/super-editor/src/core/Editor.js
johanneswilm: @chittolinag I may be incorrect on this, but as far as I can tell, the syntax `Array<{}>` means "array of objects". You can find some examples of that here on Github [1].

However, that is not what we have here. There are no objects. The return is a promise of a fixed-length array with the members being as described.

Let me know if I am confused on this one.

[1] https://github.com/search?q=%22%40returns+Array%3C%7B%22&type=code

packages/super-editor/src/core/Editor.js
johanneswilm: Yes, so given that we use the TS syntax (and not JSDoc compliant syntax) we should be able to import here. It's probably a good idea to lint this so that we catch cases where imported types have moved. Would this be something to add to the `pre-commit` configuration? Are either of you aware of a good linter that works with the TS-JSDoc syntax? Or maybe one can use TS for that itself?

packages/super-editor/src/core/Editor.js
johanneswilm: @chittolinag This would be the logical way to write it: 
```@returns {Promise<[xmlFiles: Object, mediaFiles: Object, mediaFilesAlias: Object, fonts: Object]>}```

And some write it like that [1]. But it's not valid JSDoc syntax (JSDoc throws an error.).

[1] https://github.com/search?q=%22%40returns+%7B%5B%22&type=code

packages/super-editor/src/components/toolbar/super-toolbar.js
johanneswilm: @chittolinag Done. I've gone through a few files, as far as I can tell, the **output object** from the `useToolbarItem` function ends up as being instances of `ToolbarItem` in `super-toolbar.js` (after filtering).

packages/super-editor/src/core/Editor.js
johanneswilm: I noticed that that definition is in the other package. So in order to keep jsdoc compatible and avoid any issues with typedef not being available if only one package is used, I simply copy-pasted the definition.

packages/super-editor/src/core/Editor.js
johanneswilm: Tl;DR: No, unfortunately it's not possible and there seems to be no good way to do this with JSDoc. If we want better/tighter control, then the solution is probably to use TypeScript directly.

packages/super-editor/src/components/toolbar/super-toolbar.js
johanneswilm: Done as case above.

johanneswilm:
@harbournick @chittolinag I have resolved the remaining issues - have a look though before merging as some of the solutions I used were different from what was initially suggested.

chittolinag (approved):

PR #509, opened 2025-05-17
feat(ai): add chunk formatting capabilities

palmer-cl:
Adds real-time formatting to editor coming from streaming API

- adds formatRegistry that can be extended to all format types, currently starts with bold text as that is the only current contract set by insights endpoint API as `**_____**`
- Updates handleChunk to format document
- Format document function seeks to optimistically find match in one node, and if that fails, formats across nodes
- Updates HandleDone function to wait for any pending formats to be completed, and calls format one single time to account for the last node being rendered


palmer-cl:
@harbournick this can be combined with https://github.com/Harbour-Enterprises/SuperDoc/pull/501 or kept separate

harbournick (approved):
LGTM

PR #510, opened 2025-05-19
Ideals: add search match id

VladaHarbour:


harbournick (approved):
LGTM - nice solution!

PR #511, opened 2025-05-19
HAR-9578 - return result from keymap handlers

artem-harbour:
Commands (in this case handlers) must return the result.


harbournick (approved):
Lgtm

PR #512, opened 2025-05-19
HAR-9669 - fix header and footer for first page

artem-harbour:
Show default header/footer for the first page.

Previous changes probably addressed some specific document issue, related to compatibility for example. But these changes raise header/footer issue for many other docs.

For reference.
https://github.com/Harbour-Enterprises/SuperDoc/pull/492/files.


harbournick (approved):
LGTM

PR #513, opened 2025-05-19
HAR-9663 - add zero width text when create a table

artem-harbour:
Add zero width text to empty cell when create a table in order to apply marks to the text.

Note: please check my comment in the ticket before merging.


harbournick (approved):
LGTM

PR #515, opened 2025-05-20
HAR-8849 make header/footer editable

VladaHarbour:


packages/super-editor/src/core/Editor.js
```diff
@@ -531,6 +532,22 @@ export class Editor extends EventEmitter {
     // If we are not sync'd yet, wait for the event then insert the data
     else provider.on('synced', postSyncInit);
   }
+  
+  replaceContent(content) {
```
harbournick, line 564: Let's add JSDOC if this is going to be a public method

packages/super-editor/src/core/Editor.js
```diff
@@ -867,8 +890,34 @@ export class Editor extends EventEmitter {
       ...this.options.editorProps,
       dispatchTransaction: this.#dispatchTransaction.bind(this),
       state: EditorState.create(state),
+      handleDoubleClick: async (view, pos, event) => {
+        const isHeader = hasSomeParentWithClass(event.target, 'pagination-section-header');
```
harbournick, line 926: I added an 'isHeaderFooter' flag to options. Is that useful here?

packages/super-editor/src/extensions/pagination/pagination.js
```diff
@@ -487,63 +490,113 @@ function createHeader(pageMargins, pageSize, sectionData, headerId) {
...
-  if (!sectionContainer) sectionContainer = document.createElement('div');
-
-  const autoPageNumber = sectionContainer?.querySelector('span[data-id="auto-page-number"]');
-  if (autoPageNumber) {
-    const fontSize = autoPageNumber.previousElementSibling?.style?.fontSize ||
-      autoPageNumber.nextElementSibling?.style?.fontSize;
-    if (fontSize) autoPageNumber.style.fontSize = fontSize;
-    autoPageNumber.innerText = currentPageNumber;
+  
+  // const sectionContainer = footerDef?.sectionContainer?.cloneNode(true);
```
harbournick, line 559: is this not working?

packages/super-editor/src/core/Editor.js
VladaHarbour: Unfortunately, no. Double click event fires on main editor instance always and not being propagated down to the child (probably it is prohibited somewhere internal in prosemirror). Since our header and footer editors are placed inside main editor I have to pass this event down manually

packages/super-editor/src/extensions/pagination/pagination.js
VladaHarbour: Forgot to uncomment it, my fault

packages/super-editor/src/extensions/pagination/pagination-helpers.js
```diff
@@ -56,64 +57,125 @@ export const initPaginationData = async (editor) => {
...
+    content: data,
+    extensions: getStarterExtensions(),
+    documentId: sectionId || 'sectionId',
+    media: editor.options.media,
+    mediaFiles: editor.options.mediaFiles,
+    fonts: editor.options.fonts,
+    isHeaderOrFooter: true,
+    onCreate: (evt) => setEditorToolbar(evt, editor),
+    onBlur: (evt) => onHeaderFooterDataUpdate(evt, editor, sectionId, type),
+  });
```
harbournick, line 132: @artem-harbour we could access the hook from editor.options.onFieldAnnotationDropped

harbournick (approved):
LGTM

PR #518, opened 2025-05-21
HAR-9668 Fix issue with inability to open the document after export

VladaHarbour:
The issue appeared due to exporting sdt annotations together with text nodes in one paragraph within table cell.
`Word cannot open a .docx file when a <w:sdt> (Structured Document Tag) and a regular text run (<w:r>) are siblings directly inside a single paragraph (<w:p>) inside a table cell (<w:tc>) due to a strict XML schema rule in WordprocessingML. This structure violates the expected content model for Word, even though it's technically allowed by the Open XML standard.`


VladaHarbour:
@artem-harbour @harbournick tested it within Webapp and add a couple of more fixes. Was able to download and open agreement with html annotations

harbournick (approved):
LGTM

PR #519, opened 2025-05-21
HAR-9709, HAR-9712 - highlight for field annotation and table cell

artem-harbour:
- Text highlight and color for field annotations (including html/paragraph fields).
- Background for table cells.


artem-harbour:
@harbournick - ready for review.

PR #521, opened 2025-05-21
Fix editor type exports when rolling into SuperDoc, add annotate() types

harbournick:
- Fixes types export in super-editor package
- Adds initial types for new annotate() function in super-editor
- Fixes super-editor type defs roll up in SuperDoc


edoversb (approved):
LGTM

PR #529, opened 2025-05-27
HAR-9550 - [experimental] floating page breaks

artem-harbour:
**DO NOT MERGE!**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -675,6 +675,31 @@
         "node": ">=12"
       }
     },
+    "node_modules/@floating-ui/core": {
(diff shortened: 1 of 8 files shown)


harbournick (commented):
@artem-harbour thanks for working on this, sorry for the delay here. I think the approach works well - as well as we can with the current implementation. The flickering is very minimal and there could be optimizations made there later if needed. I would say the result here is miles better than the current result so we should move forward with this until a larger v2 of pagination

artem-harbour:
Closing this one as I found a better solution.

PR #532, opened 2025-05-28
HAR-9736 - fix link export (and url field)

artem-harbour:
Added comments in the code (more details in the ticket).

### DIFF
--- a/packages/super-editor/src/core/super-converter/SuperConverter.js
+++ b/packages/super-editor/src/core/super-converter/SuperConverter.js
@@ -11,6 +11,7 @@ import {
   prepareCommentParaIds,
   prepareCommentsXmlFilesForExport,
 } from './v2/exporter/commentsExporter.js';
+import { HYPERLINK_RELATIONSHIP_TYPE } from './constants.js';
 
 class SuperConverter {
   static allowedElements = Object.freeze({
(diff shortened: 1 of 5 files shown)


packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -504,15 +505,16 @@ class SuperConverter {
...
+      if (existingTarget && !isNewMedia && !isNewHyperlink) {
         return;
       }
       
       // Update the target to escape ampersands
       rel.attributes.Target = rel.attributes?.Target?.replace(/&/g, '&amp;');
 
-      // Update the ID. If we've assigned a long ID (ie: images) we leave it alone
+      // Update the ID. If we've assigned a long ID (ie: images, links) we leave it alone
       rel.attributes.Id = existingId.length > 6 ? existingId : `rId${++largestId}`;
```
artem-harbour, line 518: This is the main issue why the links did not work when exporting. We create a new id for the link node every time, but we don't create a new relationship for this id.

Note:
- Previous relationships are not deleted, but this does not seem to be an issue. It can also be quite risky to delete prev relationships by checking the value. In any case, we can consider this later.

packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -601,25 +601,69 @@ function translateLinkNode(params) {
...
       'r:id': rId,
     },
-    elements: [outputNode],
+    elements: [contentNode],
   };
 
   return newNode;
 }
 
+function processLinkContentNode(node) {
```
artem-harbour, line 627: Add default properties and color to the link node so it looks like a link in a document.

packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -1817,6 +1865,7 @@ function prepareUrlAnnotation(params) {
   const newId = addNewLinkRelationship(params, attrs.linkUrl);
 
   const linkTextNode = getTextNodeForExport(attrs.linkUrl, marks, params);
+  const contentNode = processLinkContentNode(linkTextNode);
```
artem-harbour, line 1868: Same for url annotation.

PR #534, opened 2025-05-28
HAR-9755 - Handle non-field w:sdt tags in import/render/export

harbournick:
- Adds import/export for non field w:sdt fields (common in docs coming from Google Docs)
- Adds import and export tests for new field
- Adds new schema node for rendering these types


edoversb (approved):
LGTM

PR #535, opened 2025-05-29
update docs

artem-harbour:
I made some small updates to the SD configs in the docs, although I think the documentation needs some revision.


artem-harbour:
> Are we removing completely the "documents" `arg`? Or just in the examples?

Just in the examples. It's still supported under the hood, but I guess we want to use the new property now. fyi @harbournick

artem-harbour:
@harbournick @caio-pizzol - can we merge this?

caio-pizzol (approved):
LGTM

PR #540, opened 2025-06-01
Node resizer

johanneswilm:
I've tried a number of different things and finally found that this general solution should be the most versatile. By using the bounding box and placing the resizer controls somewhere else in the document, it should work with arbitrary element types. However, it does require for there to be size attributes on the node that is to change in size. 

I  have enabled it to be used on images in this PR. Field annotations seem to lack such size attributes for now. Should they be added @harbournick ?

### DIFF
--- a/packages/superdoc/CHANGELOG.md
+++ b/packages/superdoc/CHANGELOG.md
@@ -1,6 +1,18 @@
 # Changelog
 
-## [0.12.35](https://github.com/Harbour-Enterprises/SuperDoc/compare/v0.12.34-next.2...v0.12.35) (2025-06-06)
+## [0.12.36-next.1](https://github.com/Harbour-Enterprises/SuperDoc/compare/v0.12.35-next.1...v0.12.36-next.1) (2025-06-06)
+
+## [0.12.35-next.1](https://github.com/Harbour-Enterprises/SuperDoc/compare/v0.12.34-next.2...v0.12.35-next.1) (2025-06-06)
+
+
+### Features
+
+* (accessibility) simulate click with enter ([f9cd0b4](https://github.com/Harbour-Enterprises/SuperDoc/commit/f9cd0b4f5f8148ad0d0afad091288b32333f4cd3))
+* added extra accessibility features ([6a0f9d8](https://github.com/Harbour-Enterprises/SuperDoc/commit/6a0f9d80ae281f21ebc39beb8a14be027ef24619))
+* added initial toolbar keyboard navigation ([e37d79b](https://github.com/Harbour-Enterprises/SuperDoc/commit/e37d79bd284265f9b22f9a1ec23d05e58c21263a))
+* added roving tabindex ([6c77015](https://github.com/Harbour-Enterprises/SuperDoc/commit/6c770155532b805888d37bc91cb4d120129d541f))
+
+## [0.12.34-next.2](https://github.com/Harbour-Enterprises/SuperDoc/compare/v0.12.34-next.1...v0.12.34-next.2) (2025-06-06)
 
 
 ### Features
--- a/packages/superdoc/package.json
+++ b/packages/superdoc/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@harbour-enterprises/superdoc",
   "type": "module",
-  "version": "0.12.35",
+  "version": "0.12.36-next.1",
   "license": "AGPL-3.0",
   "readme": "../../README.md",
   "files": [


johanneswilm:
I am not sure why these tests are failing. They don't seem to be directly related to this code.

If the overall approach of this is OK, I can try to add tests. I am not sure it will be possible to do the dragging with automatic tests. If it is not, I can make those tests be manual.

packages/super-editor/src/extensions/image/image.js
```diff
@@ -1,145 +1,153 @@
...
+
+	addStorage() {
+		return {
+			media: {},
+		};
+	},
+
+	addAttributes() {
+		return {
+			src: {
```
artem-harbour: Please use two spaces to format the code. Tuned Prettier should do the trick.

packages/super-editor/src/extensions/image/image.js
```diff
@@ -1,145 +1,153 @@
-import { Node, Attribute } from '@core/index.js';
-import { ImagePlaceholderPlugin } from './imageHelpers/imagePlaceholderPlugin.js';
-import { ImagePositionPlugin } from './imageHelpers/imagePositionPlugin.js';
+import { Attribute, Node } from "@core/index.js";
+import { ImagePlaceholderPlugin } from "./imageHelpers/imagePlaceholderPlugin.js";
+import { ImagePositionPlugin } from "./imageHelpers/imagePositionPlugin.js";
```
artem-harbour: Same for this, I would avoid such changes in PRs.

packages/super-editor/src/extensions/resize-plugin/resize.css
```diff
@@ -0,0 +1,87 @@
+/* Resize handles container */
+.resize-container {
```
artem-harbour: Move the styles here as all the styles are already stored there.
`/superdoc/packages/superdoc/src/assets/styles`

Use prefixes for class names so that it does not affect other environments.

For `packages/super-editor` use `sd-editor-some-class` or `sde-some-class`
For `packages/superdoc` use `sd-some-class`

packages/super-editor/src/extensions/resize-plugin/index.js
```diff
@@ -0,0 +1,2 @@
+export * from "./resize-plugin.js";
+import "./resize.css";
```
artem-harbour: Move to `packages/super-editor/src/assets`.

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
```diff
@@ -0,0 +1,331 @@
...
+			case "ne":
+			case "sw":
+				return "nesw-resize";
+			default:
+				return "default";
+		}
+	}
+};
+
+export const ResizePlugin = Extension.create({
```
artem-harbour: I would recommend using a more general name for the extension because it can contain more than just a plugin (or multiple plugins). Like NodeResizer.

packages/super-editor/src/extensions/resize-plugin/resize.css
```diff
@@ -0,0 +1,87 @@
...
+	position: absolute;
+	pointer-events: none;
+	z-index: 1000;
```
artem-harbour: Such a large z-index won't create any ui issues?

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
```diff
@@ -0,0 +1,331 @@
...
+					// Show/hide resize handles based on selection changes
+					const { selection } = view.state;
+					const prevSelection = prevState.selection;
+
+					if (
+						selection.from !== prevSelection.from ||
+						selection.to !== prevSelection.to
+					) {
+						setTimeout(() => {
+							const selectedResizableWrapper = document.querySelector(
```
artem-harbour: Won't this way of searching for an element create issues when we have multiple `resizable-wrapper` elements/decorations?

packages/super-editor/src/extensions/image/image.js
johanneswilm: Hey @artem-harbour Yes, please have a look at https://github.com/Harbour-Enterprises/SuperDoc/pull/541 . That should avoid such problems in the future. I also think there are some linters that are still missing in this repo.

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
```diff
@@ -0,0 +1,331 @@
...
+			startY: event.clientY,
+			startWidth: rect.width,
+			startHeight: rect.height,
+			handle,
+			pos,
+			resizableElement,
+			aspectRatio: rect.width / rect.height,
+		};
+
+		document.addEventListener("mousemove", handleMouseMove);
```
artem-harbour: I think we can try using pointer events instead - in theory it should work exactly the same and enable support for other devices.

`pointermove, pointerup, pointerdown, lostpointercapture, etc`
https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
johanneswilm: OK, sure. I'll rename "ResizePlugin" to "NodeResizer".

I don't know your roadmap, so if you want to turn this into multiple plugins a different name makes sense. Also, I tried looking at the other plugins and I didn't quite get the naming convention. Some plugins are called something with "Plugin" and others are not.

packages/super-editor/src/extensions/resize-plugin/resize.css
johanneswilm: Not in my testing. But I don't know all your code. Are you using higher numbers for something specific?

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
johanneswilm: That shouldn't be able to happen as it only sets the decoration on the currently selected node, if any. So there should always just be either 1 or 0 `resizable-wrapper` decorations.

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
```diff
@@ -0,0 +1,331 @@
...
+		if (resizeState.handle.includes("w")) deltaX = -deltaX;
+		if (resizeState.handle.includes("n")) deltaY = -deltaY;
+
+		// Calculate new dimensions maintaining aspect ratio
+		const newWidth = Math.max(20, resizeState.startWidth + deltaX);
+
+		// Apply the new size immediately for visual feedback
+		if (resizeState.resizableElement) {
+			resizeState.resizableElement.style.width = `${newWidth}px`;
+			resizeState.resizableElement.style.height = "auto";
```
artem-harbour: I think this will work for images, but other node types might require height resizing?

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
```diff
@@ -0,0 +1,331 @@
...
+			startWidth: rect.width,
+			startHeight: rect.height,
+			handle,
+			pos,
+			resizableElement,
+			aspectRatio: rect.width / rect.height,
+		};
+
+		document.addEventListener("mousemove", handleMouseMove);
+		document.addEventListener("mouseup", handleMouseUp);
```
artem-harbour: Is there any difference/advantage to adding listeners to a document instead of handles (`["nw", "ne", "sw", "se"]`)?

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
johanneswilm: For now, only image nodes have size attributes, AFAICT. So therefore this plugin can only work for images. Once we do give other nodes size attributes, then it depends on what kind of size attributes they have. I have written this plugin in a way that is generic to be usable for any kind of node once those other nodes have size attributes. I think it's probably best to wait and see which kinds of size attributes those nodes will have before adding extra logic that will only to apply to nodes other than images. Maybe they don't have height at all, in which case such logic wouldn't be useful.

As for aspect ratio: For now this plugin only allows resizing without changing the aspect ratio. So that effectively means the height is changed automatically when the with is changed.

Allow the change of aspect ratio has pros and cons. Pro is that it allows the user more freedom. Con is that if the user accidentally changes only height or only width, it can be very difficult to get back to the correct aspect ratio (unless one provides extra controls for that).

But of course I can add height controls if that is what the users want.

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
johanneswilm: In theory, that should work as the mousedown should capture the events even if the mouse moves away from the resize handlers. Not sure this will work across all browsers, but probably.

packages/super-editor/src/extensions/resize-plugin/resize.css
johanneswilm: OK,

I'll move it to `/superdoc/packages/super-editor/src/assets/styles/extensions/node-resizer.css`

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
johanneswilm: I'll try.

packages/super-editor/src/extensions/resize-plugin/resize.css
johanneswilm: I've moved it and imported it the same way the other extension styles are imported. Only downside to that in comparison to my previous setup is that now the style will be loaded even if the plugin is not loaded.

packages/super-editor/src/extensions/resize-plugin/index.js
johanneswilm: Done

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
johanneswilm: Done

packages/super-editor/src/extensions/image/image.js
artem-harbour: Hey @johanneswilm, thanks for creating PR for that! Will ask @harbournick to review it.

packages/super-editor/src/extensions/resize-plugin/resize.css
artem-harbour: Thanks! And I think that's ok.

packages/super-editor/src/extensions/resize-plugin/index.js
artem-harbour: There is still this import with styles inside the extension folder. Can we delete this?

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
artem-harbour: We didn't really have a naming convention when we were working on these extensions. But it seems to me that it makes sense to give more general names for extensions (rather than tying them to one plugin).

packages/super-editor/src/extensions/resize-plugin/resize.css
artem-harbour: I mean, won't these elements be highlighted on top when we scroll? Or won't they be visible at that moment (only when resizing)?

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
artem-harbour: I've worked on similar functionality before, I don't think I had any issues. Just need to take `lostpointercapture` event into account as well. But that's optional (pointer events), I think.

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
artem-harbour: Makes sense to me. I will also let @harbournick check this out.

packages/super-editor/src/extensions/resize-plugin/index.js
johanneswilm: It should not be there in the new folder now.

packages/super-editor/src/extensions/resize-plugin/resize.css
johanneswilm: Only when the resizer is show - that is when a node is selected. They should scroll with the page content. If you have some kind of animation with high z-index values that kick in when you scroll, then this could interfere when you are scrolling past a selected image (for example). But there should always just be max one such item. Are you doing something like that? If yes, point me to where and I'll try to make this compatible.

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
johanneswilm: I used mousedown as I saw it being used a bunch of places in this code while there is no mentioning of pointerdown. But given that it is now supported across browsers [1], maybe it's time to switch to pointer events.

[1] https://caniuse.com/?search=pointerdown

packages/super-editor/src/extensions/resize-plugin/resize.css
artem-harbour: It doesn't seem to work exactly like that in dev env. But as an example - a fixed toolbar, which is always on top when we scroll the document.

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
artem-harbour: I think we can keep it as is, but it should be very easy to switch to pointer events if needed.

packages/super-editor/src/extensions/resize-plugin/resize.css
johanneswilm: Ok, gotcha. Do you have a convention for which z-index values that are safe to use within the editor itself?

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
johanneswilm: @artem-harbour I tried out doing this, but unfortunately it doesn't seem to work in Chromium (Brave). So I'll leave the event listeners on the document. Or do you have a policy against doing that?

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
johanneswilm: @artem-harbour I was actually trying out pointer events yesterday and it didn't immediately work, so I thought I'll try to modify it until it works. But if you think it's not needed, I'll not go any further with it for now.

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
artem-harbour: @johanneswilm No policy for that. Feel free to keep the current events.

packages/super-editor/src/extensions/resize-plugin/resize.css
artem-harbour: There is no convention on this, but I always use the minimum required z-index in these cases. No more than 10-15, unless there is a special need for it.

For example, in Harbour webapp the toolbar has `z-index: 10` and it is fixed in relation to the editor.

johanneswilm: I've looked through the other CSS files for z-indexes and I've adjusted it to 11/12 which should be fine. There is though at least one case of 1000 nd another of 100.

artem-harbour (approved):
LGTM

packages/super-editor/src/extensions/resize-plugin/resize-plugin.js
harbournick: SG

harbournick (approved):
LGTM

PR #542, opened 2025-06-02
HAR-9741 - add select class to dropdowns

artem-harbour:
Added for:
- zoom.
- font-family.
- font-size.
- line-height.
- linked styles.


harbournick (approved):
LGTM

PR #547, opened 2025-06-03
HAR-9739 - add toolbar text customization

artem-harbour:
And PR with docs updates.
https://github.com/Harbour-Enterprises/SuperDoc/pull/546


harbournick (approved):
LGTM

PR #548, opened 2025-06-03
feat: allow to export the raw blob without triggering a download

ybrodsky:
Adds a new param to the export function to allow exporting the document(s) blob without triggering a download.

The use case is you wanting to do something else with the document: save in some backend, upload to S3, etc.


harbournick (approved):
LGTM - Thanks @ybrodsky I think this is a nice addition!

PR #552, opened 2025-06-04
HAR-9797 - feat: added initial support for high contrast mode

chittolinag:
Added support for high contrast mode.

- When enabled, the toolbar and the editor borders should have maximum contrast.
- Users can enable/disable it by calling `setHighContrastMode` on the main SuperDoc instance.

**How to enable:**
```js
const config = {
  // ...
  onReady: (event) => {
    event.superdoc.setHighContrastMode(true);
  }
  // ...
}

const editor = new SuperDoc(config);
// Or at any point (after the editor is ready)
editor.setHighContrastMode(true);
```

This function can be called programatically at any point.

**UI changes:**

- Toolbar items (hover state, dropdowns). 
- Editor borders.

Screenshot 2025-06-05 at 11 14 13


examples/react-example/package-lock.json
harbournick, line 1: Let's remove any changes in this lock file unless strictly necessary to prevent any potential issues with install (we've had issues with unintentional changes in lock files)

harbournick (changes requested):
one quick comment on the package-lock that looks to be unintended

chittolinag:
@artem-harbour @harbournick The idea I came up with for now is to keep two composables, one for superdoc, one for supereditor. We do have to change colors on both packages, but given the timeline I couldn't really figure out a way to keep a same shared state - but I'm super open to suggestions in case you know a better approach to do so!

harbournick (approved):
LGTM

PR #553, opened 2025-06-05
feat(ai): Add support for italic and underline text formatting in AI …

palmer-cl:
- Adds support for italic and underline formatting
- Underline is handled by using both <ins> and <u> tags as thats what it appears like Gemini is sending back from the insights endpoint


harbournick (approved):
LGTM

PR #560, opened 2025-06-05
HAR-9796 -  Accessibility keyboard navigation and focus management

chittolinag:
- Added basic keyboard navigation support on the toolbar and its items
- Add shortcut to go back to toolbar when you're focused on the editor

https://github.com/user-attachments/assets/e4784d31-ead5-435e-9c48-aeccea491776


harbournick (approved):
LGTM

PR #564, opened 2025-06-06
HAR-9550 - floating page breaks V2 (absolute strategy)

artem-harbour:
Fixes page break inside a table (apply only to page breaks within tables).

Screenshot 2025-06-11 at 13 05 06


packages/super-editor/src/extensions/pagination/pagination.js
```diff
@@ -229,7 +241,18 @@ const performUpdate = (editor, view, previousDecorations) => {
     );
 
     view.dispatch(updateTransaction);
-  }
+
+    requestAnimationFrame(() => {
```
harbournick, line 245: Just curious - why is the double requestAnimationFrame needed?

harbournick (commented):
Just one comment. Code looks good. Will just do some testing in the UI before merge.

packages/super-editor/src/extensions/pagination/pagination.js
artem-harbour: To make sure that DOM is updated and the necessary elements are found there (for clearing). Single requestAnimationFrame does not work reliably in this case. And I don't want to use setTimeout.

PR #568, opened 2025-06-10
HAR-9813 - fix line-height and extra spacing

artem-harbour:


packages/super-editor/src/core/super-converter/helpers.js
```diff
@@ -171,11 +171,7 @@ const getLineHeightValueString = (lineHeight, defaultUnit, lineRule = '', isObje
   let [value, unit] = parseSizeUnit(lineHeight);
   if (Number.isNaN(value) || value === 0) return {};
   if (lineRule === 'atLeast' && value < 1) return {};
-  
   unit = unit ? unit : defaultUnit;
-
-  // MS Word has a slightly bigger gap with line spacing equal to Superdoc's
-  value += unit ? 4 : 0.2;
```
artem-harbour, line 178: This is what increases the line-height in the editor/preview and adds extra spacing. Which also creates a difference with the final pdf.

packages/super-editor/src/extensions/paragraph/paragraph.js
```diff
@@ -28,8 +28,9 @@ export const Paragraph = Node.create({
         renderDOM: (attrs) => {
           const { spacing } = attrs;
           if (!spacing) return {};
-
-          const style = getSpacingStyleString(spacing);
+          const spacingCopy = { ...spacing };
+          if (attrs.lineHeight) delete spacingCopy.line; // we'll get line-height from lineHeight
```
artem-harbour, line 32: This is necessary to avoid a case like this, otherwise the change in line-height in the editor is not reflected.
Screenshot 2025-06-10 at 15 06 17

PR #570, opened 2025-06-10
HAR-9859 - Add new handling for page numbers and total page count after editable headers/footers

harbournick:
- Fixes auto page number nodes after editable headers/footers changes
- Adds new node view to page-number and total-page-numer
- imports styles and checks styles of neighboring nodes to fix various style issues
- adds export to both numbering nodes
- adds keyboard shortcuts for both numbering nodes


edoversb (approved):
Lgtm

PR #574, opened 2025-06-11
Feature: superdoc-yjs-collaboraton package [WIP]

harbournick:


packages/collaboration-yjs/README.md
```diff
@@ -0,0 +1,119 @@
...
+## Quick Start (Example)
+
+Below is an example using Fastify, but you can adapt it to any server framework.
+
+```js
+import Fastify from 'fastify';
+import websocket from '@fastify/websocket';
+import { v4 as uuidv4 } from 'uuid';
+import SuperDocCollaboration from '@harbour-enterprises/superdoc-yjs-collaboration';
+import hooks from './hooks'; // your hook implementations
```
chittolinag: Minor, but since this is supposed to be a quick example, I think it would be better to have the hooks inside the file, just so the user can see it all at once and doesn't have to reason about how the hooks should look like (ie, the parameters it receives, etc).

packages/collaboration-yjs/src/collaboration/collaboration.js
```diff
@@ -0,0 +1,49 @@
...
+import { createLogger } from '../internal-logger/logger.js';
+
+import { DocumentManager } from '../document-manager/manager.js';
+import { ConnectionHandler } from '../connection-handler/handler.js';
+
+/**
+ * Create and configure a SuperDoc collaboration service.
+ * Simply call service.welcome(socket, request) in your WebSocket endpoint.
+ *
+ * @property {config} config - Configuration object
```
chittolinag: If we don't have time that's ok, but it would be good to create a type definition for the config object, this would be super helpful for TS users - and even for JS users, because it would be simple to know what are the possible parameters and their types.

packages/collaboration-yjs/src/connection-handler/handler.js
```diff
@@ -0,0 +1,131 @@
...
+   * Authenticate the WebSocket connection.
+   * This method checks if the `authenticate` hook is defined and calls it.
+   * If the authentication fails, it closes the socket with an error code.
+   * If an error occurs during authentication, it also closes the socket with an error code.
+   * @param {import('../types.js').WebSocket} socket The WebSocket connection to authenticate.
+   * @param {import('../types.js').CollaborationParams} params - The parameters for this connection.
+   * @returns {Promise<import('../types.js').UserContext | void>} Context object if authentication is successful, else closes the socket.
+   */
+  async #authenticate(socket, params) {
+    if (!this.#hooks.authenticate) {
```
chittolinag, line 108: It seems like authentication is not required. If that's the case, I suggest we make it very explicit in the documentation to avoid any inconveniences.

packages/collaboration-yjs/src/document-manager/manager.js
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * @param {SharedSuperDoc} doc - The SharedSuperDoc instance.
+   */
+  #scheduleSave(doc, userParams) {
+    const documentId = doc.name;
+    if (this.debounceMs > 0) {
+      clearTimeout(this.#timers.get(documentId));
+
+      this.#timers.set(
+        documentId,
```
chittolinag, line 90: Since the `documentId` is the `doc.name` - could it be an issue in case there's more than a document with the same name?

packages/collaboration-yjs/src/shared-doc/callback.js
```diff
@@ -0,0 +1,83 @@
...
+  ? JSON.parse(process.env.CALLBACK_OBJECTS)
+  : {};
+
+export const isCallbackSet = !!CALLBACK_URL;
+
+/**
+ * @param {import('../types.js').SharedSuperDoc} doc - The SharedSuperDoc instance.
+ */
+export const callbackHandler = (doc) => {
+  const room = doc.name;
```
chittolinag, line 16: Same as I mentioned above - could this be an issue in case there's more than one document with the same name?

packages/collaboration-yjs/README.md
harbournick: I like it. Will place in-line

packages/collaboration-yjs/src/collaboration/collaboration.js
harbournick: There is a 'config' type defined, but the name is too generic. I renamed it to ServiceConfig so its easier to identify. But the type def exists in service-types

packages/collaboration-yjs/src/document-manager/manager.js
harbournick: These should be unique. Worth noting in the docs as well.

packages/collaboration-yjs/src/shared-doc/callback.js
harbournick: doc.name ends up being an internal reference. All docs should be created with unique IDs (which are the names). Its a good call, will make sure its clear in docs.

PR #579, opened 2025-06-12
HAR-9842 - Accessibility keyboard navigation focus management part 2

chittolinag:
- Better keyboard navigation on the toolbar
- All items on the toolbar are accessible and usable with the keyboard only (except the "overflow" button)

Known issues:

1. When there's a overflow button (grouping items that otherwise wouldn't be visible on the screen), the keyboard navigation doesn't work on its items.
2. When pressing "Esc" on a currently open dropdown, it should close the dropdown. However, the following items have a problem with that (meaning it won't close the dropdown): color dropdowns, link, table, overflow, document mode. This doesn't really block the user from navigating, the only thing is that they have to select something and hence the dropdown will close.


artem-harbour (approved):
LGTM

PR #580, opened 2025-06-12
Feature: Add y-websocket provider as default provider

harbournick:
- Adds y-websocket (Yjs default) provider as the default collaboration provider
- Requires new providerType flag in collaboration config to use any other provider
- Minor change to awareness


edoversb (approved):
LGTM

PR #586, opened 2025-06-14
[Bugfix] Remove white background from `.pagination-inner`

Chee7ah:
The header / footer of the pages have a white background set in the `.pagination-inner` class. The "body" of the pages however have no background color set and are transparent.

This results in the document looking "broken" if SuperDoc is used on anything but a white background. Hence this commit removing the explicit white background to leave the whole page transparent.

*Illustration of the issue:*
Screenshot 2025-06-14 at 12 21 10 PM

### DIFF
--- a/packages/super-editor/src/assets/styles/extensions/pagination.css
+++ b/packages/super-editor/src/assets/styles/extensions/pagination.css
@@ -54,7 +54,6 @@
   left: 0;
   display: flex;
   flex-direction: column;
-  background-color: white;
 }
 
 /**


harbournick:
@Chee7ah Thank you for submitting this! We are actually currently underway on a huge Pagination V2 project that fixes and improves all sorts of things around pagination, and will include a consistent background color between the page and the break there. It is due to be completed and released by the end of July. In the mean time, for most docs we tend to assume the editor itself has a white background for now so the color makes sense. Are you able to adjust the styling on your end in the meantime?

Chee7ah:
> @Chee7ah Thank you for submitting this! We are actually currently underway on a huge Pagination V2 project that fixes and improves all sorts of things around pagination, and will include a consistent background color between the page and the break there. It is due to be completed and released by the end of July. In the mean time, for most docs we tend to assume the editor itself has a white background for now so the color makes sense. Are you able to adjust the styling on your end in the meantime?

Yes, already doing that. You can close this PR if it's superfluous.

PR #588, opened 2025-06-16
HAR-9857, HAR-9832 - field annotation commands

artem-harbour:
Extended schema, added `defaultDisplayLabel` attribute, which will be filled with `displayLabel` attr value when the annotation is added.

Commands added:
- `replaceFieldAnnotationsWithLabel` - Convert annotations back to text node.
- `replaceFieldAnnotationsWithLabelInSelection` - Convert annotations back to text node (In selection).
- `resetFieldAnnotations` - Reset all annotations to default value.

Helpers added:
- `resetHeaderFooterFieldAnnotations` - Reset all annotations to default value (in header/footer).

### DIFF
--- a/packages/super-editor/src/extensions/field-annotation/field-annotation.js
+++ b/packages/super-editor/src/extensions/field-annotation/field-annotation.js
@@ -55,8 +55,19 @@ export const FieldAnnotation = Node.create({
         },
       },
 
+      defaultDisplayLabel: {
+        default: '',
+        parseDOM: (elem) => elem.getAttribute('data-default-display-label'),
+        renderDOM: (attrs) => {
+          if (!attrs.defaultDisplayLabel) return {};
+          return {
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/extensions/field-annotation/field-annotation.js
```diff
@@ -463,30 +485,156 @@ export const FieldAnnotation = Node.create({
...
+
+          return true;
+        },
+      
+      /**
+       * Resets all annotations to default values.
+       * @example
+       * editor.commands.resetFieldAnnotations()
+       */
+      resetFieldAnnotations:
```
harbournick, line 600: @artem-harbour Just a note: Let's add some documentation about these commands on Friday when we do a round up

harbournick (approved):
LGTM

PR #595, opened 2025-06-18
feat(menus): implement slash and right click menus

palmer-cl:
- Adds logic for a slash command menu (5 second cooldown)
- Adds logic for right click menu
- Adds GenericPopover component to handle positioning
- Simplify the AIWriter component around emitters in favor of using commands
- Removes links from the Toolbar, and adds inlinne editor link control

`The flow is right click to open menu or slash, if menu item has a component, use the generic popover to position it, render the component inside the popover`

Areas to review:
- menuItems file
- cursorHelpers 
- Component files (SlashMenu, GenericPopover)
- Add link by selecting text and right click, remove link by just clicking on it

Screenshot: 

image

image


artem-harbour (approved):
LGTM

packages/super-editor/src/components/slash-menu/SlashMenu.vue
```diff
@@ -0,0 +1,369 @@
+<script>
+import { ref, onMounted, onBeforeUnmount, watch, nextTick, computed, h, markRaw } from 'vue';
+import { SlashMenuPluginKey } from '@/extensions/slash-menu';
+import { getPropsByItemId } from './utils.js';
+import { moveCursorToMouseEvent } from '../cursor-helpers.js';
+import { getItems } from './menuItems.js';
+import { getEditorContext } from './utils.js';
+import GenericPopover from '../popovers/GenericPopover.vue';
+
+export default {
```
harbournick: @palmer-cl just curious - how come this component is using the options API? I think we should refactor it for consistency, unless there's a specific reason we should stick to it?

packages/super-editor/src/components/slash-menu/menuItems.js
```diff
@@ -0,0 +1,139 @@
+import { toolbarIcons } from '../toolbar/toolbarIcons.js';
+import TableGrid from '../toolbar/TableGrid.vue';
+import AIWriter from '../toolbar/AIWriter.vue';
+import TableActions from '../toolbar/TableActions.vue';
+import LinkInput from '../toolbar/LinkInput.vue';
+import { selectionHasNodeOrMark } from '../cursor-helpers.js';
+import { serializeSelectionToClipboard, writeToClipboard } from '../../core/utilities/clipboardUtils.js';
```
harbournick: should we use '@core' alias?

packages/super-editor/src/components/slash-menu/utils.js
```diff
@@ -0,0 +1,106 @@
...
+ * @param {Object} props
+ * @returns {Object}
+ */
+export const getPropsByItemId = (itemId, props) => {
+    // Common props that are needed regardless of trigger type
+    const editor = props.editor;
+
+    const baseProps = {
+        editor: markRaw(props.editor),
+
```
harbournick: clean up blank line

packages/super-editor/src/components/toolbar/defaultItems.js
```diff
@@ -1065,7 +1004,6 @@ export const makeDefaultItems = ({
     colorButton,
     highlight,
     separator,
```
harbournick, line 867: let's remove table actions buttons from the toolbar as well

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -766,8 +724,6 @@ export class SuperToolbar extends EventEmitter {
   onToolbarResize = () => {
     this.#makeToolbarItems({
       superToolbar: this,
-      icons: this.config.icons,
```
harbournick: I think we need to support passing this into defaultItems rather than just importing it. There are cases where the end user will provide custom icons and texts in the config here.

packages/super-editor/src/components/toolbar/defaultItems.js
```diff
@@ -17,14 +17,120 @@ import checkIconSvg from '@harbour-enterprises/common/icons/check.svg?raw';
 import SearchInput from './SearchInput.vue';
 import { TOOLBAR_FONTS } from './constants.js';
 
+//Do we need to pass this in to "makeDefaultItems" or is it better to just import it?
```
harbournick: I think we do need to pass it in (see comment below too). Users can provide their own custom icons and texts so this needs to be dynamic

packages/super-editor/src/extensions/slash-menu/index.js
```diff
@@ -0,0 +1,195 @@
+import { Plugin, PluginKey } from 'prosemirror-state';
+import { Extension } from '@core/Extension.js';
```
harbournick: For consistency, we should probably leave index.js as just a list of exports, and move this code to slash-menu.js

packages/super-editor/src/extensions/slash-menu/index.js
```diff
@@ -0,0 +1,195 @@
...
+            }
+            return true;
+          }
+
+          return false;
+        },
+      },
+    });
+
+    return [slashMenuPlugin];
```
harbournick: Let's not add the slashMenuPlugin if this.editor.options.isHeadless (ie: its a headless / node js editor only). Can just do `return this.editor.options.isHeadless ? [] : [slashMenuPlugin]` probably

harbournick (changes requested):
Really nice work @palmer-cl ! It works super smoothly too, nice experience.
Just added a few minor comments and tweaks. Thank you!

packages/super-editor/src/components/toolbar/defaultItems.js
palmer-cl: Can you confirm if you want to leave insert table in the toolbar, and just remove the table actions? Or remove both?

packages/super-editor/src/components/slash-menu/utils.js
palmer-cl: Fixed in bd5a79bd9b7f29f2884e2a7bd1a9bcb20a25fe45

packages/super-editor/src/components/slash-menu/menuItems.js
palmer-cl: Updated in 6f42b4c5745a460c7330e8d0ca748f822cfd270b

packages/super-editor/src/components/slash-menu/SlashMenu.vue
palmer-cl: Updated in 8c2bc4038914fdadb2f35c72445dceffe8a42f08

packages/super-editor/src/components/toolbar/defaultItems.js
palmer-cl: This makes sense. I think we are going to be removing these items from the toolbar (table actions and link etc.) So do we want to separate out passing in the toolbar icons and text (no longer in toolbar) and slash menu config? Do we want to create a new config to allow the user to pass text and icons to slash menu?

Said another way, do we want to be using `toolbarIcons._` at all here?

packages/super-editor/src/extensions/slash-menu/index.js
palmer-cl: Good catch. Fixed in ec2975fffa7645c0c5addd7cac590c2435fe3657

packages/super-editor/src/extensions/slash-menu/index.js
palmer-cl: Also updated in ec2975fffa7645c0c5addd7cac590c2435fe3657

packages/super-editor/src/components/toolbar/defaultItems.js
harbournick: That's true, we are removing those from the toolbar. We can use all-default icons for the context menu (since we have no specific way for the user to configure _those_ icons yet) if that makes sense (and they are ultimately all different buttons from the toolbar)

packages/super-editor/src/components/toolbar/defaultItems.js
harbournick: Let's keep the insert table button. Remove only table actions

palmer-cl: Removed 👍

packages/super-editor/src/components/toolbar/super-toolbar.js
palmer-cl: Fixed in latest version 👍

PR #598, opened 2025-06-19
HAR-9896 - split single table cell command

artem-harbour:
Split table cell into smaller ones if colspan equal to one.


harbournick (approved):
LGTM - works great!

PR #599, opened 2025-06-19
HAR-9781 - Highlight field annotation on export

falsaadehh:


packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -2488,3 +2489,19 @@ const getAutoPageJson = (type, outputMarks = []) => {
...
 };
+
+const getFieldHighlightJson = () => {
+  return {
+    "name": "w:rPr",
+    "elements": [
+      {
+        "name": "w:shd",
+        "attributes": {
+          "w:fill": '7AA6FF',
```
harbournick, line 2500: Could we use attrs.fieldColor for the exported highlight, if it exists?
And, fall back to 7AA6FF if needed.

It might provide a bit more continuity between the view in the editor and the docx. And, in the future, we could also probably re-import field colors to complete the cycle.

harbournick (changes requested):
Just one quick change to try though I do like the 7AA6FF as default

packages/super-editor/src/core/super-converter/exporter.js
falsaadehh: @harbournick  because we can't add transparency to fill color in MS word like in html this is how it will look like using the attrs.fieldColor do you want that?
Screenshot 2025-06-20 at 6 58 24 PM

falsaadehh: [Test document (2).docx](https://github.com/user-attachments/files/20841002/Test.document.2.docx)

harbournick: ah! right, you did mention that. Nah, let's leave it as is for now. Will likely revisit when we get to various fields-related UX/UI tickets that exist in the backlog (ie: using non-transparent colors everywhere, with hex values that are "closer to white" to achieve the same effect.

harbournick (approved):
LGTM

PR #603, opened 2025-06-20
Pagination memory issue

harbournick:
- re-use already existing DOM element for headers/footers 
- fixes issue with losing font sizes/family
- please double check this doesn't re-introduce the memory issue


artem-harbour (approved):
LGTM

@VladaHarbour, please check this too.

PR #604, opened 2025-06-21
Feature: Lists v2 

harbournick:
- Refactors lists to work more closely how MS Word lists work
- All lists are root-level nodes and a list is a single node
- There is no concept of 'sub lists' in the document schema. It is provided by numId, ilvl etc from numbering.xml
- Adds several files of tests
- Adds support for 'custom' list numbering and ordinal lists
- Adds support for user-defined bullets
- Introduces the node view for list items


edoversb (approved):
LGTM

PR #605, opened 2025-06-23
HAR-9901 fix selection issues for image nodes in header/footer

VladaHarbour:


packages/super-editor/src/extensions/pagination/pagination.js
```diff
@@ -82,10 +82,7 @@ export const Pagination = Extension.create({
           const imagePluginTransaction = tr.getMeta(ImagePlaceholderPluginKey);
           if (imagePluginTransaction) {
             if (imagePluginTransaction.type === 'remove') {
-              const imagePos = imagePluginTransaction.pos;
```
VladaHarbour, line 85: Hi @harbournick! Just wanted to explain this. We don't really have actual image inserted at this point and querySelector will give us either first image node or prosemirror separator image node which both have complete set as true. But header/footer editor can have undefined here

PR #609, opened 2025-06-25
HAR-9833 Implement docChanged getter

VladaHarbour:


packages/super-editor/src/core/Editor.js
```diff
@@ -260,6 +263,14 @@ export class Editor extends EventEmitter {
     initMode();
   }
 
+  /**
+   * Getter which indicates if any changes happen in Editor
+   * @returns {boolean}
+   */
+  get docChanged() {
+    return this.options.isHeaderFooterChanged || this.options.isCustomXmlChanged || !this.options.initialState.doc.eq(this.state.doc);
```
harbournick, line 271: Can you double check if this works in collaboration as well? state is a bit weird there since its managed the by plugin, though I do think it _should_ work - just want to be sure.

packages/super-editor/src/core/Editor.js
```diff
@@ -954,11 +965,13 @@ export class Editor extends EventEmitter {
     // Only initialize the doc if we are not using Yjs/collaboration.
     const state = { schema: this.schema };
     if (!this.options.ydoc) state.doc = doc;
+    
+    this.options.initialState = EditorState.create(state);
```
harbournick, line 969: oh - actually now that I see this here. We likely need to do this inside the collaboration init stuff, after sync

harbournick (changes requested):
Just a quick check on collaboration version of this

packages/super-editor/src/core/Editor.js
```diff
@@ -1199,6 +1212,7 @@ export class Editor extends EventEmitter {
 
     this.options.onCollaborationReady({ editor, ydoc });
     this.options.collaborationIsReady = true;
+    this.options.initialState = this.state;
```
VladaHarbour, line 1215: @harbournick this should work for collaboration

harbournick (approved):
LGTM

PR #612, opened 2025-06-26
CSP compatibility

johanneswilm:
This adds basic compatibility for Content Security Policies with a `style-src-attr` set to "self".

Issues to resolve: 

* Tests have only been done with some basic documents. Tables needed extra adjustments. There may be other more advanced features that also require this that have not yet been found.

* ~A reference to the document node (by default `window.document`) is needed. I am not sure if Superdoc is ever run with a different document node than `window.document`.~

* ~Code license. The function "renderSpec" was imported and slightly modified from prosemirror. It likely still falls under the ProseMirror license terms. A decision needs to be made how to handle this - for now I have included the function in a special file with a reference to the origin and the MIT license of ProseMirror. But there are other options - such as releasing as a separate package or merging upstream.~


harbournick:
It looks good to me and seemed to be really solve the core CSP issue in editor.js. @artem-harbour can you give it a review too please? 

Also @johanneswilm mind resolving the conflicts? Now that 0.14.1 is out I'd love to get this one merged into develop for the next release.

johanneswilm:
Merge conflicts have been resolved.

johanneswilm:
A large part of this has now been moved upstream. If you want dependencies with version numbers, wait until `prosemirror-model` version 1.25.2 or 1.26.0 before merging this.

artem-harbour (commented):
LGTM

And a few questions

1. @harbournick - should we wait for the next version of `prosemirror-model` or are we ok with such a dependency in package.json?
`"prosemirror-model": "git://github.com/ProseMirror/prosemirror-model.git#e73b74ca17cc2e1ca8d7a6a608310278502e9370"`

2. As far as I understand, this workaround solves the issue with inline styles, won't there be issues with other dynamically added styles?
`initDefaultStyles`, `initMobileStyles`, `initFonts`, `updateEditorStyles`, etc
https://github.com/Harbour-Enterprises/SuperDoc/blob/develop/packages/super-editor/src/core/Editor.js

johanneswilm:
2. With other CSPs, there could be other problems that will require a separate PR. Let me know which ones we need to handle and I'll take a look if it's possible to do so. This PR removes all error messages thrown with a style-src-attr set to "self" when loading my test files.

johanneswilm:
@artem-harbour: Note that we now link to a released version of prosemirror-model.

artem-harbour (approved):
LGTM

fyi @harbournick

harbournick (approved):
LGTM

PR #614, opened 2025-06-26
feat: prevent custom menu from opening when user holds control

palmer-cl:
- prevents our custom menu from opening on right click if control key is also held


harbournick (approved):
LGTM

PR #618, opened 2025-06-27
HAR-9854 - fix: text mark font family precedes linked style

chittolinag:
- Ensure linked styles doesn't override the active mark
- Add initial tests for the toolbar


harbournick (approved):
LGTM - nice

PR #620, opened 2025-06-30
fix: allow toolbar when typing font size

palmer-cl:
- update the control of the font size input
- allow typing in the toolbar
- focuses the input on click 

image


harbournick (approved):
LGTM

PR #625, opened 2025-07-02
feat: click context menu enhancements 

palmer-cl:
- Prevent slash menu in view only modes
- Share the logic for pasting from core/InputRule for the slash menu


harbournick (approved):
@palmer-cl actually since this PR doesn't break anything, and does get us one step closer to the goal, plus includes a critical fix for 'viewing' mode, I decided to dismiss previous review and approve this one. Let's make additional changes as a separate PR.

PR #627, opened 2025-07-02
feat: add LinkInput popover for cmd/ctrl + K shortcut in SuperEditor

palmer-cl:
- adds keyboard listener for command / control K to generate the link popover

### DIFF
--- a/packages/super-editor/src/components/SuperEditor.vue
+++ b/packages/super-editor/src/components/SuperEditor.vue
@@ -1,7 +1,7 @@
 <script setup>
 import 'tippy.js/dist/tippy.css';
 import { NSkeleton } from 'naive-ui';
-import { ref, onMounted, onBeforeUnmount, computed, shallowRef, reactive, nextTick } from 'vue';
+import { ref, onMounted, onBeforeUnmount, computed, shallowRef, reactive, nextTick, markRaw } from 'vue';
 import { Editor } from '@/index.js';
 import { getStarterExtensions } from '@extensions/index.js';
 import SlashMenu from './slash-menu/SlashMenu.vue';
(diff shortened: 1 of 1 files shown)


harbournick (changes requested):
Thanks Cole! 

I think there's two issues we should consider in this PR, even though they weren't caused by the PR. But while we're in here: https://cdn.zappy.app/885162e5bc3d52c2058df388e9abb418.mp4

1. In gdocs if you hit CMD+K it offers you input boxes for 'text' and 'link'. In our case, if you hit CMD+K with nothing selected, and enter a link, nothing will happen. We could just handle the CMD+K with no selection case the way MS word does: just insert the link URL as the text of the link itself
2. If there is text, and I apply a link, I am now unable to edit the link text itself - the popover grabs the selection

palmer-cl:
@harbournick can you check the latest updates? 


- Improved selection handling for link text, accommodating empty and non-empty selections.
- Updated URL validation to consider anchors as valid.
- Refactored link submission logic to handle text insertion and link updates more effectively.
- Add text input field. Enhanced styling for link input fields and improved user experience in the toolbar.

harbournick (changes requested):
@palmer-cl looks so good! The text box is so handy.
Just one corner case: starting with SOME text, but adding more on link creation, does not actually insert the added text: https://cdn.zappy.app/f4797dab1450e48768ffe91825e07e06.mp4

palmer-cl:
> @palmer-cl looks so good! The text box is so handy. Just one corner case: starting with SOME text, but adding more on link creation, does not actually insert the added text: https://cdn.zappy.app/f4797dab1450e48768ffe91825e07e06.mp4

Ack! Will fix now.

harbournick (approved):
LGTM

PR #630, opened 2025-07-03
feat: add disableContextMenu option to SuperDoc

palmer-cl:
- Do not install plugin if headless or disabled
- Do not render slash menu if disable flag exists in SuperDoc config


harbournick (approved):
LGTM

PR #634, opened 2025-07-04
Paragraph lists v2 migrations, copy paste handling of lists from html…

harbournick:
- Adds migration tool for lists in html fields
- HTML must be migrated before being previewed in fields, if necessary
- Adds "child editor" concept of linked editors
- Migration tweaks
- List item node view tweaks, field annotation view change for html view


edoversb (approved):
Lgtm

PR #635, opened 2025-07-04
HAR-10010 Cleanup empty paragraphs when annotation deleted

VladaHarbour:
Hi @harbournick could you please check if it's something we wanted to achieve here
WebApp PR: https://github.com/Harbour-Enterprises/Harbour-WebApp/pull/6319

P.S. What about unit tests is that we don't have tests for this part of SuperDoc. Should I do some?


packages/super-editor/src/core/helpers/annotator.js
```diff
@@ -483,6 +483,35 @@ export const cleanUpListsWithAnnotations = (fieldsToDelete = [], editor) => {
   editor.view.dispatch(tr);
 };
 
+export const cleanUpParagraphWithAnnotations = (fieldsToDelete = [], editor) => {
```
artem-harbour: This is not a helper, but a command. And all commands in PM must return a boolean.

For proper integration with the editor, all commands should be in `packages/super-editor/src/core/commands` or organized in an extension if they are for some specific task (as in this case). Helpers should be organized in a separate folder and exported separately too.

I understand that it is already organized this way and works, but it is not entirely correct. I will let @harbournick  look into it when he gets back.

artem-harbour (commented):
Added my comments, not ready for merge now.

@falsaadehh @VladaHarbour

packages/super-editor/src/core/helpers/annotator.js
VladaHarbour: Hi @artem-harbour! Moved cleanup functions to commands. Could you please check if one more time
cc @harbournick

caio-pizzol:
@nick can you review this one?

harbournick:
Holding this one as it needs some extensive testing in other services.

packages/super-editor/src/core/helpers/annotator.js
harbournick: Thanks for moving this to commands @VladaHarbour. There were some issues created by this but fixed now.
@artem-harbour noted on commands. Good distinction to make moving forward:
 - commands: change the document, trigger a transaction
 - helpers: are read-only, don't change doc
 LMK if that is the correct thinking there. There's probably a few other functions I will migrate later.

harbournick:
Apologies for the delay on this one. The issue was that moving the 'helper' to commands creates issues for anyone using this API and needs to be tested carefully. I've added two commits fixing a few issues, and tested thoroughly now and we can move this forward.

harbournick (approved):
LGTM

PR #636, opened 2025-07-05
fix: update link selection accuracy

palmer-cl:
- Update links to trim document whitespace when selecting and replacing text
- Handle edge case where user can start selection off a link, but end on a link
- Simplify button logic


newfadel (changes requested):
I've reviewed the PR, and the changes in `packages/super-editor/src/extensions/link/link.js` are a significant improvement!

Specifically, the updates to trim leading and trailing whitespace when selecting and replacing text address a common usability issue. This will prevent accidental inclusion of whitespace in the link, leading to cleaner URLs.

Furthermore, handling the edge case where a user can start selection *off* a link but end *on* one is a very robust addition. This makes the link selection more forgiving and intuitive, reducing user frustration.

Could you provide a small example or snippet of how the whitespace trimming is implemented? Also, for the edge case, does this involve adjusting the selection boundaries or a more complex parsing of the text? Understanding these details would help in a more in-depth review.

packages/super-editor/src/components/toolbar/LinkInput.vue
```diff
@@ -239,7 +236,7 @@ const handleRemove = () => {
   </div>
 </template>
 
-<style scoped>
+<style scoped lang="scss">
```
artem-harbour: We use `postcss` (when needed) instead scss. Can you please delete this?

palmer-cl: thanks @artem-harbour updated in latest commit!

artem-harbour: Thanks!

PR #637, opened 2025-07-07
Fix page break consider safe position to break

falsaadehh:


packages/super-editor/src/extensions/pagination/pagination.js
```diff
@@ -531,6 +541,23 @@ function generateInternalPageBreaks(doc, view, editor, sectionData) {
   return decorations;
 }
 
+/**
+ * Helper function to not split inside creation blocks
+ * @param {Number} pos The position of the outermost node that exceeds threshold
+ * @returns {Number} safe pos to insert page break
+ */
+function getSafeBreakPos(pos) {
+  const UNSPLITTABLE = new Set(['listItem', 'tableRow', 'table']);
```
artem-harbour: In this case, the tables will probably not be split at all and will not create a page break. Make sure this is the expected behavior.

falsaadehh:
> LGTM
> 
> Note: The code looks good to me, but make sure it is tested enough and the behavior here is expected (tables case).

added for listItems only

PR #639, opened 2025-07-07
fix: wrong anchor image position in header (HAR-9893)

VladaHarbour:
Hi @harbournick!. Vivienne experienced issues with header anchor images which are placed outside of the table structure. 
The main issue is that Word treats those images like ones which are attached to the document not header. I did some fix but it's still not ideal since image is cut and plugin applied only when header is activated. Let's talk about this one once you're back. Attaching video of how it works

https://github.com/user-attachments/assets/5565610b-c05b-4e06-9617-cd5c5054dd98

### DIFF
--- a/packages/super-editor/src/core/Editor.js
+++ b/packages/super-editor/src/core/Editor.js
@@ -250,6 +250,8 @@ export class Editor extends EventEmitter {
     isCustomXmlChanged: false,
 
     focusTarget: null,
+
+    parentEditor: null,
   };
 
   /**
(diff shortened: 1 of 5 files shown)


harbournick:
@VladaHarbour this one is quite old. Do you recall if we still need this? feel free to ping me on chat

VladaHarbour:
Hi @nick! There is an anchor image issue in this doc https://linear.app/harbour/issue/IDE-7/docx-editing-images-are-absent-in-file-content-fps-4785
I will check if this PR is relevant for that doc and in general

PR #648, opened 2025-07-09
HAR-10039 Can't open docx with image annotations

VladaHarbour:


packages/super-editor/src/extensions/field-annotation/field-annotation.js
```diff
@@ -317,6 +318,11 @@ export const FieldAnnotation = Node.create({
       generatorIndex: {
         rendered: false,
         default: null,
+      },
+      
+      hash: {
```
artem-harbour, line 323: Can we use fieldId instead? Or is it docx specific?

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -670,8 +670,6 @@ class SuperConverter {
     });
 
     relationships.elements = [...relationships.elements, ...newRels];
-
-    this.convertedXml['word/_rels/document.xml.rels'] = relsData;
```
artem-harbour, line 674: Will this affect anything or is it no longer needed?

packages/super-editor/src/extensions/field-annotation/field-annotation.js
VladaHarbour: Since it's possible to have multiple image input we need something more unique that fieldId to store those images. We used to generate this hash on export before but we also do export inside updateYdocDocxData which led to inconsistencies in image annotations export

packages/super-editor/src/core/super-converter/SuperConverter.js
VladaHarbour: relsData is mutated in previous line anyway. There is no need to assign it again

artem-harbour (approved):
LGTM

PR #653, opened 2025-07-10
HAR-10050 Strikethrough from Google Docs is missing

VladaHarbour:


artem-harbour (approved):
LGTM

fyi @falsaadehh

PR #654, opened 2025-07-10
HAR-10078 Set custom text selection when default can't be applied

VladaHarbour:


packages/super-editor/src/assets/styles/layout/global.css
```diff
@@ -14,3 +14,7 @@
 a {
   text-decoration: auto;
 }
+
+.custom-selection {
```
artem-harbour: Please always add prefix to the class name.

Like
`sd-my-class` - for SD
`sd-editor-my-class` - for SD editor

packages/super-editor/src/components/toolbar/ButtonGroup.vue
```diff
@@ -75,6 +75,11 @@ const closeDropdowns = () => {
   currentItem.value = null;
 };
 
+const clearSelection = () => {
```
artem-harbour: It seems to me that this functionality related to switching the selection should not be part of UI, buttons and toolbar.

Is it possible to make this selection work separately from UI? For example, hide the native selection and always apply decoration for the current selection, or do a smarter switch when needed (and if possible).

packages/super-editor/src/components/toolbar/ToolbarButton.vue
```diff
@@ -58,9 +59,14 @@ const { isHighContrastMode } = useHighContrastMode();
...
 const handleClick = () => {
   if (hasInlineTextInput) {
+    
+    const editor = getActiveEditor();
+    const selection = editor ? editor.view.state.selection : null;
     nextTick(() => {
       inlineInput.value?.focus();
       inlineInput.value?.select();
+
+      if (editor) editor.toggleCustomSelectionDecorations(selection);
```
artem-harbour: Same comment about UI and selection logic.

packages/super-editor/src/core/Editor.js
```diff
@@ -422,6 +423,25 @@ export class Editor extends EventEmitter {
     this.options.onFocus({ editor, event });
   }
 
+  /**
+   * Set custom selection decorations for editor when no custom selection could be applied
+   * @param {Object} selection - Current selection
+   * @returns {void}
+   */
+  toggleCustomSelectionDecorations(selection = null) {
```
artem-harbour: It seems to me that public editor methods should be something more general that a developer can use from the outside.

In this case, this method is very specific and it seems to me that this should not be a public editor method, or even an editor method at all.

artem-harbour (changes requested):
@VladaHarbour please check my comments and let me know if that makes sense.

fyi @falsaadehh @harbournick

harbournick:
@VladaHarbour mind resolving the conflict?

harbournick (approved):
LGTM

artem-harbour:
@VladaHarbour - can you double check the changes? This probably broke the font selection in the toolbar.

fyi @harbournick

PR #656, opened 2025-07-11
HAR-10096 Strikethrough toolbar item

VladaHarbour:


packages/super-editor/src/components/toolbar/defaultItems.js
```diff
@@ -919,6 +931,7 @@ export const makeDefaultItems = ({
     bold,
     italic,
     underline,
+    strikethrough,
```
artem-harbour, line 934: Please test the toolbar in webapp with this new item, does it fit on a small screen (mac) or not. As far as I remember, even without this new item there were already issues related to the width of the toolbar.

VladaHarbour: Looks good to me on those breakpoints
Screenshot 2025-07-11 at 15 34 51
Screenshot 2025-07-11 at 15 35 40
Screenshot 2025-07-11 at 15 36 46

artem-harbour: And desktop (1440px)?

VladaHarbour: Pushed an adjustment and it looks like this
Screenshot 2025-07-11 at 16 42 57
But there is no space left for another icon without hiding something in overflow

PR #665, opened 2025-07-16
update table cellMinWidth

artem-harbour:
I noticed that some tables may have smaller cell widths than specified in the options.


harbournick (approved):
LGTM

PR #666, opened 2025-07-16
to solve some complex docx loading errors

rex-0085:
sometimes encountered an error about undefined of localeCompare. especially complex docx documents

### DIFF
--- a/packages/super-editor/src/extensions/linked-styles/linked-styles.js
+++ b/packages/super-editor/src/extensions/linked-styles/linked-styles.js
@@ -242,6 +242,6 @@ export const getQuickFormatList = (editor) => {
       return style.type === 'paragraph' && style.definition.attrs;
     })
     .sort((a, b) => {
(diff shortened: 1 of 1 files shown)


harbournick:
Hi @rex-0085 ! Thanks for this. I'm curious if you would be able to include a stripped down version of a doc that was causing the issue? I'd love to investigate further to get to the bottom of this beyond preventing the error if possible (+ add to unit tests once we find the root cause). Thank you!

rex-0085:
image
It will occur when I use editorRef.current.goToSearchResult(searchResults[0])
the attachment you can have a try. for now, I have only found this error sample.
[errorsample.docx](https://github.com/user-attachments/files/21415071/errorsample.docx)

harbournick:
Actually, will fix this along with some other linked styles things. 
@rex-0085 would you be interested in joining our Discord? it may be a better place to ask quick questions: https://discord.gg/WjkQEwZHBK

PR #673, opened 2025-07-17
fix(har-10122): toolbar styles are missing

VladaHarbour:


packages/superdoc/src/SuperDoc.vue
```diff
@@ -1,5 +1,6 @@
 <script setup>
 import '@harbour-enterprises/common/styles/common-styles.css';
+import '@harbour-enterprises/super-editor/style.css';
```
VladaHarbour, line 3: @harbournick also not sure why we didn't import those styles before and it used to work without issues

packages/super-editor/tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "emitDeclarationOnly": true,
     "outDir": "dist",
     "declarationMap": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "typeRoots": ["./node_modules/@types/*"]
```
VladaHarbour, line 10: @harbournick faced TS2688: Cannot find type definition file for 'minimatch' error during the package build process and this line is supposed to fix it
https://github.com/strapi/strapi/issues/23859

harbournick (approved):
Lgtm

PR #676, opened 2025-07-18
chore: improve tables keyboard UX

VladaHarbour:


packages/super-editor/src/extensions/custom-selection/custom-selection.js
```diff
@@ -34,7 +34,7 @@ export const CustomSelection = Extension.create({
       props: {
         handleDOMEvents: {
           focusout: (view, event) => {
-            if (document.activeElement && !event.relatedTarget) {
+            if (document.activeElement && !event.relatedTarget && !view.state.selection.empty) {
```
VladaHarbour, line 37: @harbournick this fix is essential for tables functionality (didn't notice that custom selection can affect tables)

harbournick:
@VladaHarbour thank you for the work on this one. I think we're going to hold off on adding this one, I think selection in general needs some more attention at a wider level before we focus in on tables, and table functionality also needs some upgrades.

PR #682, opened 2025-07-21
fix: page number styling + table import fixes

VladaHarbour:


chittolinag:
Hey @VladaHarbour ! We now have automated screenshot-based testing running on Pull Requests here. 

I noticed that there's a couple tests that failed - and I was taking a look it seems to be related to tables. There should be a message posted on your Linear ticket with more details.

PR #700, opened 2025-07-24
refactor(selection): new approach

VladaHarbour:
Hi @harbournick! I decided to move to different selection approach and try to use as much native selection as possible to avoid bugs and issues. Custom selection will jump in when toolbar item is clicked (to make it unified for the whole toolbar) and will have a slightly different colour from the native one. Could you please check this one and let me know what do you think
cc @artem-harbour

### DIFF
--- a/packages/super-editor/src/extensions/custom-selection/custom-selection.js
+++ b/packages/super-editor/src/extensions/custom-selection/custom-selection.js
@@ -1,41 +1,67 @@
 import { Extension } from '@core/Extension.js';
-import { AllSelection, Plugin, PluginKey, TextSelection } from 'prosemirror-state';
+import { Plugin, PluginKey, TextSelection } from 'prosemirror-state';
 import { Decoration, DecorationSet } from 'prosemirror-view';
 
 export const CustomSelectionPluginKey = new PluginKey('CustomSelection');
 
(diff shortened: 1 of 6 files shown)


packages/super-editor/src/extensions/custom-selection/custom-selection.js
```diff
@@ -1,53 +1,108 @@
 import { Extension } from '@core/Extension.js';
-import { AllSelection, Plugin, PluginKey, TextSelection } from 'prosemirror-state';
+import { Plugin, PluginKey, TextSelection } from 'prosemirror-state';
 import { Decoration, DecorationSet } from 'prosemirror-view';
 
 export const CustomSelectionPluginKey = new PluginKey('CustomSelection');
 
+const handleClickOutside = (event, editor) => {
+  const editorElem = document.querySelector('.editor-element');
```
artem-harbour: Need to fix this and get editor element from `editor.options.element`.

artem-harbour (dismissed):
Added a comment to fix.

packages/super-editor/src/extensions/custom-selection/custom-selection.js
harbournick: Great call!

PR #704, opened 2025-07-24
fix: paragaph fields inside lists for export

harbournick:
- fix exporting paragraph fields (final doc and non) inside lists


edoversb (approved):
LGTM

PR #713, opened 2025-07-29
HAR-10180 - formatting, import/export fixes

artem-harbour:


packages/super-editor/src/core/super-converter/v2/importer/markImporter.js
```diff
@@ -44,6 +47,11 @@ export function parseMarks(property, unknownMarks = [], docx = null) {
         return;
       }
 
+      // this probably requires a more thorough check.
+      if (['w:bCs'].includes(m.name) && langAttrs['w:eastAsia']) {
```
artem-harbour, line 51: It's a bit questionable, but fixes one case with the issue.

harbournick (approved):
LGTM

PR #716, opened 2025-07-30
 HAR-10195 highlight tracked changes comment

falsaadehh:


packages/superdoc/src/components/CommentsLayer/CommentDialog.vue
```diff
@@ -125,6 +125,13 @@ const setFocus = () => {
 const handleClickOutside = (e) => {
   if (e.target.classList.contains('n-dropdown-option-body__label')) return;
   if (e.target.classList.contains('sd-editor-comment-highlight')) return;
+  if (e.target.classList.contains('sd-editor-tracked-change-highlight')) return;
```
artem-harbour: Can you add all class names to the array and use `.some()` for checking?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

harbournick (approved):
LGTM

PR #728, opened 2025-08-04
refactor: add fallback logging for init routine

mattConnHarbour:


packages/superdoc/src/core/SuperDoc.js
```diff
@@ -479,7 +482,7 @@ export class SuperDoc extends EventEmitter {
   }
 
   log(...args) {
-    console.debug('🦋 🦸‍♀️ [superdoc]', ...args);
+    this.logger('🦋 🦸‍♀️ [superdoc]', ...args);
```
harbournick: just curious, why not just do the console.debug ? console.debug : console.log; check in here rather than adding this.logger? Seems we don't need both this.log and this.logger

harbournick (changes requested):
looks good just a quick q

packages/superdoc/src/core/SuperDoc.js
mattConnHarbour: Just didn't want to add an if to every log call! But if you think this.logger is more clutter than it's worth, I can take that out.

harbournick: or we should make logger private. I've been meaning to make log() private too (#log()) but haven't gotten around to it. Basically if its not something to expose to the public API of the class then let's not. But i'm fine with a private logger or a check in #log

mattConnHarbour: Done! Checking for console.debug in #log.

harbournick (approved):
LGTM

PR #730, opened 2025-08-05
feat: add tab stop support

johanneswilm:


chittolinag:
Hey @johanneswilm - thanks for the help here. Can you please create a PR using a branch from the `SuperDoc` repo itself, instead of a fork? The reason is that we have a set of tests that are executed internally, but I'm guessing it won't be triggered because we need to use a token from the repository, and I don't think that a fork has access to it.

johanneswilm:
@chittolinag I know. But at least yesterday I didn't have access to this repo.

johanneswilm:
@chittolinag I still don't have write access.

johanneswilm:
@chittolinag Given that you have write access, can you just create a branch on this repo and pull into that and then create the PR? Alternatively, can you give me write access?

harbournick (approved):
LGTM - great to have this!

PR #735, opened 2025-08-05
feat: extend editor to allow addHelpers in extensions

harbournick:
- Extends the extensions service and editor to allow addHelpers in extensions
- adds editor.helpers with extension namespace


edoversb (approved):
lGTM

PR #736, opened 2025-08-06
fix: handle click outside if the target class is focused

falsaadehh:


falsaadehh:
@artem-harbour if you remember i fixed the comments plugin where the track change activated but the floating comment get active and inactive very quickly 
https://linear.app/harbour/issue/HAR-10195/selecting-text-in-tracked-change-does-not-highlight-the-comment#comment-010eb897
I found that after clicking on track change another target class being activated which is ProseMirror-focused so i added it to the exclude list. let me know if you think this is a good idea

packages/superdoc/src/components/CommentsLayer/CommentDialog.vue
```diff
@@ -133,6 +133,8 @@ const handleClickOutside = (e) => {
     'track-delete-dec',
     'track-format',
     'track-format-dec',
+    'ProseMirror',
+    'ProseMirror-focused',
```
artem-harbour, line 137: I'm not entirely sure what exactly this change fixes, I tested without it and it seems to work as without the change.

Also I don't think it's a good idea to add a main editor class here. @harbournick what do you think?

harbournick: @artem-harbour I tend to agree with you. From what I can see in this issue there's more going on with this sync and double event triggering. I think we need to fix it at the event trigger source, maybe there's transactions we need to suppress in the plugin or something to do with wether or not selection has changed in the tr

let's hold off on adding this for now as I did not see any change either. The bug continues even sporadically.

harbournick:
Closing this one per discussion. More investigation needed.

PR #737, opened 2025-08-06
HAR-10132 - apply font to list item from annotation

artem-harbour:
Apply font/font-size to list item from annotation if no text nodes.


harbournick:
@artem-harbour merging this one but it looks like you might not have the pre-commit hooks installed locally. Mind running `npm run prepare` locally to make sure commit titles don't break CICD commitlint checks?

PR #738, opened 2025-08-06
HAR-10241 - fix table import

artem-harbour:
Fixed incorrect handling of cell width if inline width is missing and other issues.

See comments in the code.

### DIFF
--- a/packages/super-editor/src/core/super-converter/v2/importer/tableImporter.js
+++ b/packages/super-editor/src/core/super-converter/v2/importer/tableImporter.js
@@ -419,18 +419,19 @@ export function handleTableRowNode(node, table, rowBorders, styleTag, params) {
   const cellNodes = node.elements.filter((el) => el.name === 'w:tc');
 
   let currentColumnIndex = 0;
-  const content = cellNodes?.map((n, index) => {
-    let colWidth = gridColumnWidths?.[currentColumnIndex] || null;
+  const content =
+    cellNodes?.map((n, index) => {
+      let colWidth = gridColumnWidths?.[currentColumnIndex] || null;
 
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/core/super-converter/v2/importer/tableImporter.js
```diff
@@ -416,14 +416,21 @@ export function handleTableRowNode(node, table, rowBorders, styleTag, params) {
   }
 
   const gridColumnWidths = getGridColumnWidths(table);
-
   const cellNodes = node.elements.filter((el) => el.name === 'w:tc');
-  const content =
-    cellNodes?.map((n, index) => {
-      const colWidth = cellNodes.length > 1 ? gridColumnWidths[index] : null;
```
artem-harbour, line 423: It is not clear what the significance of this check (`cellNodes.length > 1`) was, but we need to get the width value regardless of the number of cells in the row.

packages/super-editor/src/core/super-converter/v2/importer/tableImporter.js
```diff
@@ -416,14 +416,21 @@ export function handleTableRowNode(node, table, rowBorders, styleTag, params) {
...
-      return handleTableCellNode(n, node, table, borders, colWidth, styleTag, params, index);
-    }) || [];
 
+  let currentColumnIndex = 0;
+  const content = cellNodes?.map((n, index) => {
+    let colWidth = gridColumnWidths?.[currentColumnIndex] || null;
```
artem-harbour: We also need to take the width value at the correct index (taking into account the previous merged cells), and not at the index of the current iteration.

PR #753, opened 2025-08-07
HAR-10241 - fix: list items

artem-harbour:
See comment in linear.


harbournick:
Left a comment in the internal ticket. Removing 'ready for review' label pending changes but please re-add when ready. Thank you @artem-harbour !

PR #756, opened 2025-08-08
docs: ref new super doc api for conversion

caio-pizzol:


docs/src/guide/modules.md
```diff
@@ -291,117 +291,51 @@ const mySuperDocConfig = {
...
-   > - Convert documents to PDF format
-   > - Maintains formatting and layout of the original document
-   > - Default format if none specified
-
-**Notes**
-
-- Conversion operations are asynchronous and may take a few seconds to complete
-- If the base64 is invalid, the conversion won't be possible
-- This approach is secure as data is not persisted on our servers and uploading a document is not required
+Your new API is much better than the old one:
```
harbournick: Should change this "Your" to "The"  or remove/change entire line

docs/src/guide/modules.md
```diff
@@ -291,117 +291,51 @@ const mySuperDocConfig = {
...
 
+## Example
 
+```javascript
+// Backend endpoint example
+app.post('/convert-pdf', async (req, res) => {
+  const formData = new FormData();
+  formData.append('file', req.file.buffer, req.file.originalname);
+  
+  const response = await fetch('https://api.superdoc.dev/v1/convert?from=docx', {
```
harbournick, line 331: Are we deprecating the old API endpoint or will it continue to work for existing users for now?

harbournick (changes requested):
One small change, and one q

docs/src/guide/modules.md
caio-pizzol: It will continue to exist (for now) - but we should deprecate shortly.<br><br>Since OpenGov is the only one using - I think it is the right timing?

docs/src/guide/modules.md
caio-pizzol: Nice thanks!

harbournick (approved):
Lgtm

PR #758, opened 2025-08-09
fix: add support for vrect nodes

palmer-cl:
- Adds support for vrect nodes in the super editor
- Updates contentBlock to allow for percentage widths along with px values


palmer-cl:
@harbournick - added import and export tests.

One thing I did find in my investigation. Sometimes Word saves the rects at extremely small height values (.05pt) and even when we parse/convert it, it rounds to 0px and does not show up. 

I made a fix here to put a min height of 1px on our element. Let me know if you think this is ok or want to do it a different way. My goal was to preserve the raw height data but just change the appearance in the doc so at least something does render.

harbournick:
> @harbournick - added import and export tests.
> 
> One thing I did find in my investigation. Sometimes Word saves the rects at extremely small height values (.05pt) and even when we parse/convert it, it rounds to 0px and does not show up.
> 
> I made a fix here to put a min height of 1px on our element. Let me know if you think this is ok or want to do it a different way. My goal was to preserve the raw height data but just change the appearance in the doc so at least something does render.

This makes sense to me!

harbournick (approved):
LGTM - nice work! vrect has come up so many times, so glad to finally have this!

PR #778, opened 2025-08-15
feat: added duration to transaction callback

chittolinag:
Added transaction duration to allow us to measure the performance of SuperDoc interactions.


packages/super-editor/src/core/Editor.js
```diff
@@ -1313,6 +1313,7 @@ export class Editor extends EventEmitter {
    */
   #dispatchTransaction(transaction) {
     if (this.isDestroyed) return;
+    const start = Date.now();
```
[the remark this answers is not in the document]
harbournick: sticking with Date.now for this

packages/superdoc/src/core/SuperDoc.js
```diff
@@ -182,7 +182,7 @@ export class SuperDoc extends EventEmitter {
     onCommentsListChange: () => null,
     onException: () => null,
     onListDefinitionsChange: () => null,
-
+    onTransaction: () => null,
```
harbournick, line 186: can you update the SuperDoc.js Config type if we're adding this so it matches?

chittolinag: great catch. just added!

harbournick (approved):
LGTM

PR #783, opened 2025-08-18
chore: add jsdocs + extension docs auto-gen

caio-pizzol:


eslint.config.docs.mjs
```diff
@@ -0,0 +1,93 @@
+// .eslintrc.jsdoc.js
```
harbournick: remove this line (not the name of the file anyway)

eslint.config.docs.mjs
```diff
@@ -0,0 +1,93 @@
...
+          object: 'Object',      // Use Object not object
+          array: 'Array',        // Use Array not array
+          'Array.<>': 'Array<>', // Use Array<Type> not Array.<Type>
+        }
+      }
+    }
+  },
+
+  // Optional: Stricter rules for core types file
+  {
```
harbournick: remove for now if we don't need to specify any additional rules here?

package.json
```diff
@@ -37,6 +39,11 @@
       "eslint --fix",
       "prettier --write"
     ],
+    "packages/super-editor/src/extensions/field-annotation/field-annotation.js": [
+      "eslint -c eslint.config.docs.mjs",
```
harbournick: what do we use this for? is it temporary/testing or needed? We should make sure we won't have to do this for each extension fwiw

packages/super-editor/tsconfig.json
```diff
@@ -1,14 +1,15 @@
 {
-  "include": ["./src/**/*"],
   "compilerOptions": {
+    "checkJs": true,
```
harbournick, line 3: is this needed now? It might break the build or trigger a lot of issues since we don't have full coverage yet. For now we've been using // @ts-check to explicitly allow certain files to be checked when they are ready with 100% coverage

packages/super-editor/tsconfig.json
```diff
@@ -1,14 +1,15 @@
...
-    "outDir": "dist",
-    "declarationMap": true,
-    "skipLibCheck": true,
-    "module": "NodeNext",
-    "moduleResolution": "NodeNext",
-    "typeRoots": ["./node_modules/@types/*"]
-  }
+    "outDir": "dist"
+  },
+  "include": ["src/extensions/field-annotation/field-annotation.js", "src/core/types/index.d.ts"],
```
harbournick: why do we need to explicitly set this?

packages/superdoc/src/core/SuperDoc.js
```diff
@@ -50,7 +50,17 @@ import { initSuperdocYdoc, initCollaborationComments, makeDocumentsCollaborative
  * @property {Object} [toolbar] Toolbar module configuration
  */
 
-/** @typedef {import('@harbour-enterprises/super-editor').Editor} Editor */
+/**
+ * @typedef {Object} Editor
```
harbournick: Importing the Editor type from the original (import('@harbour-enterprises/super-editor').Editor) _should_ work - does it not? We shouldn't define the type here. The Super Editor build needs to be creating a usable type for the class and this should just consume it on the SuperDoc level

packages/superdoc/src/core/SuperDoc.js
```diff
@@ -50,7 +50,17 @@ import { initSuperdocYdoc, initCollaborationComments, makeDocumentsCollaborative
...
+ * @property {Object} options - Editor options
```
harbournick: continuing from the last comment: for instance, Editor class options have type EditorOptions. This should be automatically imported from the original definition.

if its not working here LMK we can investigate that separately

packages/super-editor/src/extensions/field-annotation/types.d.ts
```diff
@@ -0,0 +1,18 @@
+export type FieldType = 'text' | 'checkbox' | 'image' | 'html' | 'link' | 'signature';
```
harbournick: these types should be auto-generated when we build. Any reason we have to manually create it here?

harbournick (changes requested):
Great that we're adding this! Generally looking good - just have a lot of questions.
I tend to think we should actually stop using field-annotation for this initially and use a much simpler node (maybe document sections?) to validate the whole process since there so much happening here between ensuring types are generated correctly, JSDoc validations etc.

package.json
caio-pizzol: It will be needed for every new extension that we add JSDocs.

This will then run lint check (with pre-commit hook and ci workflow), to make sure it matches the defined standard

packages/super-editor/tsconfig.json
caio-pizzol: good question! I was adding @ts-ignore for the things that were breaking. but this approach this better

packages/super-editor/tsconfig.json
caio-pizzol: goes back to allowing checkJs above

packages/superdoc/src/core/SuperDoc.js
caio-pizzol: I was trying to fix type checks (after I allowed `checkJs`) will roll it back

packages/super-editor/src/extensions/field-annotation/types.d.ts
caio-pizzol: We need to have types file during development not only when building it, otherwise we lose the real-time type checking.

But we could use the auto-generated ones indeed.

caio-pizzol:
> Great that we're adding this! Generally looking good - just have a lot of questions. I tend to think we should actually stop using field-annotation for this initially and use a much simpler node (maybe document sections?) to validate the whole process since there so much happening here between ensuring types are generated correctly, JSDoc validations etc.

I liked your approach! Let me switch to Document Sections then (new docs are not official as well, so we can test it end to end first)

packages/super-editor/tsconfig.json
caio-pizzol: in order to make it work with Intellisense we have the make `checkJs=true`

e.g. not complaining when declare a `string` but passing a `bool`
image

with `checkJs`
image

added a `"types:watch": "tsc --emitDeclarationOnly --watch"` for realtime type checking but it is not the same thing

@chittolinag thoughts?

packages/super-editor/tsconfig.json
```diff
@@ -1,14 +1,14 @@
 {
-  "include": ["./src/**/*"],
   "compilerOptions": {
+    "checkJs": true,
```
caio-pizzol, line 3: `checkJs` is enabled for localdev but disabled for build (check file above)

packages/super-editor/tsconfig.json
```diff
@@ -1,14 +1,14 @@
...
     "declaration": true,
     "emitDeclarationOnly": true,
-    "outDir": "dist",
-    "declarationMap": true,
-    "skipLibCheck": true,
-    "module": "NodeNext",
-    "moduleResolution": "NodeNext",
-    "typeRoots": ["./node_modules/@types/*"]
-  }
+    "outDir": "types"
```
caio-pizzol: Types are saved in `types/` folder for IntelliSense type checking

packages/super-editor/tsconfig.build.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "checkJs": false,
+    "outDir": "dist"
```
caio-pizzol: Types are saved in `dist` for build

PR #788, opened 2025-08-19
WIP - Trackable id - HAR-10208

falsaadehh:


VladaHarbour:
@falsaadehh @harbournick looks good to me!
The only thing I would like to mention is that I noticed that all block nodes of the same type have the same id. I'm not aware of the ticket details but just wondering whether this is desired behavior

packages/super-editor/src/extensions/heading/heading.js
```diff
@@ -25,6 +26,16 @@ export const Heading = Node.create({
         rendered: false,
       },
       tabStops: { rendered: false },
+      sdBlockId: {
```
VladaHarbour, line 28: @falsaadehh let's make renderHTML the same as for other nodes to make it consistent

falsaadehh:
@VladaHarbour can you give this another look now? 
@harbournick I tried to implement this without a plugin but it wasn't a good solution, if default return the id then it's executed once per node type, hence multiple nodes of same type will have the same id even if i have keepOnSplit: false, also it wasn't an elegant solution overall. if i'm using it then i need to use parseHTML instead of parseDOM to make sure the id is in the node not only in the dom which also is not a good practice here
based on this conversation the best way to do it via appendTransaction 
https://discuss.prosemirror.net/t/how-i-can-attach-attribute-with-dynamic-value-when-new-paragraph-is-inserted/751/3
let me know what you think

VladaHarbour (changes requested):
Hi @falsaadehh! I think that onCreate hook of BlockNode extension is extra and we can proceed with appendTransaction which is executed right after onCreate every time we face block node. We can reduce the number of calculations. @harbournick can you please confirm my point
But what is more, I found one lists related issue. Every paragraph added after newly inserted list has the same id which is weird. Also @harbournick do you think every list item(which is a separate list) should have the same or different id? For now it's the same

https://github.com/user-attachments/assets/75e6451a-d1a6-4a43-9efd-8dd1cc114c80

falsaadehh:
> Hi @falsaadehh! I think that onCreate hook of BlockNode extension is extra and we can proceed with appendTransaction which is executed right after onCreate every time we face block node. We can reduce the number of calculations. @harbournick can you please confirm my point But what is more, I found one lists related issue. Every paragraph added after newly inserted list has the same id which is weird. Also @harbournick do you think every list item(which is a separate list) should have the same or different id? For now it's the same if we create new list but different if we upload a document
> 
>  Screen.Recording.2025-08-20.at.15.51.25.mov

if onCreate is removed then uploaded documents won't have unique ids wrapped around their nodes. for lists having the same unique id I think it's better if i move unique id to list item or not add unique ids to lists at all since it has many attributes that identify it and I think that will fix the issue for the following paragraph having same id as will @harbournick should we remove lists from our block node types?

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+  name: 'BlockNode',
+
+  addCommands() {
+    return {
+      replaceBlockNodeById:
+        (id, contentNode) =>
+        ({ dispatch, tr }) => {
+          let blockNode = this.editor.helpers.BlockNode.getBlockNodeById(id);
+          if (!blockNode || blockNode.length > 1) {
+            return true;
```
harbournick: we should return false here to indicate the command cannot be executed

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+
+          return true;
+        },
+
+      deleteBlockNodeById:
+        (id) =>
+        ({ dispatch, tr }) => {
+          let blockNode = this.editor.helpers.BlockNode.getBlockNodeById(id);
+          if (!blockNode || blockNode.length > 1) {
+            return true;
```
harbournick: same here, return false if we cannot execute

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+            }
+          }
+
+          return true;
+        },
+
+      updateBlockNodeAttributes:
+        (id, attrs = {}) =>
+        ({ dispatch, tr }) => {
+          if (!dispatch) return true;
```
harbournick: let's remove this line and replace with if (dispatch) check below the next check (same as previous commands here)

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+
+          let blockNode = this.editor.helpers.BlockNode.getBlockNodeById(id);
+          if (!blockNode || blockNode.length > 1) {
+            return true;
```
harbournick: same as above (return false)

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+      },
+
+      getBlockNodesInRange: (from, to) => {
+        let blockNodes = [];
+
+        this.editor.state.doc.nodesBetween(from, to, (node, pos) => {
+          if (!node || node?.nodeSize === undefined) {
+            return;
+          }
+          if (node.type.groups.includes('block')) {
```
harbournick: use node.isBlock

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+  addPmPlugins() {
+    return [
+      new Plugin({
+        appendTransaction: (transactions, _oldState, newState) => {
+          if (!transactions.some((tr) => tr.docChanged)) return null;
+
+          let tr = null;
+          let changed = false;
+          newState.doc.descendants((node, pos) => {
+            if (!node.type.spec.group?.split(' ')?.includes('block')) return;
```
harbournick: same note re: block checks. would suggest we do something like https://github.com/Harbour-Enterprises/SuperDoc/pull/793

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+      },
+    };
+  },
+  onCreate() {
+    const { state, view } = this.editor;
+    const tr = state.tr;
+    let changed = false;
+
+    state.doc.descendants((node, pos) => {
+      if (!node.type.spec.group?.split(' ')?.includes('block')) return;
```
harbournick: use node.isBlock

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+
+export const BlockNode = Extension.create({
+  name: 'BlockNode',
+
+  addCommands() {
+    return {
+      replaceBlockNodeById:
+        (id, contentNode) =>
+        ({ dispatch, tr }) => {
+          let blockNode = this.editor.helpers.BlockNode.getBlockNodeById(id);
```
harbournick: use const (clearer intent)

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+            }
+          }
+
+          return true;
+        },
+
+      deleteBlockNodeById:
+        (id) =>
+        ({ dispatch, tr }) => {
+          let blockNode = this.editor.helpers.BlockNode.getBlockNodeById(id);
```
harbournick: use const

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+
+          return true;
+        },
+
+      updateBlockNodeAttributes:
+        (id, attrs = {}) =>
+        ({ dispatch, tr }) => {
+          if (!dispatch) return true;
+
+          let blockNode = this.editor.helpers.BlockNode.getBlockNodeById(id);
```
harbournick: use const

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+
+      getBlockNodesByType: (type) => {
+        return findChildren(this.editor.state.doc, (node) => node.type.name === type);
+      },
+
+      getBlockNodesInRange: (from, to) => {
+        let blockNodes = [];
+
+        this.editor.state.doc.nodesBetween(from, to, (node, pos) => {
+          if (!node || node?.nodeSize === undefined) {
```
harbournick: can remove this check since we only perform an action inside the next if statement

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+        return blockNodes;
+      },
+    };
+  },
+  onCreate() {
+    const { state, view } = this.editor;
+    const tr = state.tr;
+    let changed = false;
+
+    state.doc.descendants((node, pos) => {
```
harbournick: Let's avoid duplicated code here and in the appendTransaction hook below. We can probably extract the logic to a single helper here like const changed = assignMissingBlockIds(tr, state.doc) that returns wether something changed or not

packages/super-editor/src/core/utilities/sdBlockUniqueId.js
```diff
@@ -0,0 +1,9 @@
+let counter = 0;
+
+export function generateBlockUniqueId(type) {
+  counter++;
```
harbournick: I don't love the use of counter here tbh. why not just generate UUIDs rather than the Math.random stuff? just curious. We don't really need to have the node type in the prefix IMO. I'm fine with this format but just curious why not just go UUID

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+        (id, attrs = {}) =>
+        ({ dispatch, tr }) => {
+          if (!dispatch) return true;
+
+          let blockNode = this.editor.helpers.BlockNode.getBlockNodeById(id);
+          if (!blockNode || blockNode.length > 1) {
+            return true;
+          }
+
+          let { pos, node } = blockNode[0];
```
harbournick: if we remove if (!dispatch) above, then we would wrap this next block in if (dispatch)

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+
+          return true;
+        },
+    };
+  },
+
+  addHelpers() {
+    return {
+      getBlockNodes: () => {
+        return findChildren(this.editor.state.doc, (node) => node.type.groups.includes('block'));
```
harbournick: I think for the checks of wether or not this block is a valid type for sdBlockId we should explore other options. I know the initial specification called for "block" type nodes but it might be better to only return blocks that have the sdBlockId. I created a PR for you to consider on top of your PR that adds that functionality (with some utilities and a ton of tests for those utils)

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+      );
+      changed = true;
+      return false;
+    });
+
+    if (changed) view.dispatch(tr);
+  },
+  addPmPlugins() {
+    return [
+      new Plugin({
```
harbournick, line 117: let's create a specific key for this plugin. example: https://github.com/Harbour-Enterprises/SuperDoc/blob/5e944d2bf9a1d45ebabf511fcdeb4d1f6e000b2a/packages/super-editor/src/extensions/linked-styles/plugin.js#L5

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+        appendTransaction: (transactions, _oldState, newState) => {
+          if (!transactions.some((tr) => tr.docChanged)) return null;
```
harbournick: in https://github.com/Harbour-Enterprises/SuperDoc/pull/793 I suggest we remove onCreate and just use a hasInitialized flag in order to always do a first-pass

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,167 @@
...
+
+          let tr = null;
+          let changed = false;
+          newState.doc.descendants((node, pos) => {
```
harbournick, line 127: since this will run on every keystroke, this will become an issue on large docs. I suggest we only process IF a transaction has created a new valid block node ie: https://github.com/Harbour-Enterprises/SuperDoc/pull/793

harbournick: but even that can be optimized further in the future (maybe we need to get the changed ranges from the transaction and _only_ check nodes in those ranges. but we can return to that later

harbournick (changes requested):
Great call using appendTransaction here! Please see my comments and https://github.com/Harbour-Enterprises/SuperDoc/pull/793

harbournick:
One more comment: let's please add some unit tests for each of the helpers and commands

packages/super-editor/src/core/utilities/sdBlockUniqueId.js
falsaadehh: hmm felt that uid is long but I think more robust, I'll put it back

falsaadehh:
done @harbournick @VladaHarbour considered all the review comments here

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,184 @@
+import { Extension } from '@core/Extension.js';
+import { helpers } from '@core/index.js';
+import { Plugin, PluginKey } from 'prosemirror-state';
+import { ReplaceStep } from 'prosemirror-transform';
+import { v4 as uuidv4 } from 'uuid';
+
+const { findChildren } = helpers;
+const SD_BLOCK_ID_ATTRIBUTE_NAME = 'sdBlockId';
+export const BlockNodePluginKey = new PluginKey('blockNdePlugin');
```
harbournick: Let's fix typo in **blockNdePlugin**

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -0,0 +1,184 @@
...
+export const BlockNode = Extension.create({
+  name: 'BlockNode',
```
harbournick: let's go with blockNode for consistency (lower case b) in the actual name key

harbournick (changes requested):
Just two super minor fixes but otherwise looks great!

falsaadehh:
> Just two super minor fixes but otherwise looks great!

done

harbournick (approved):
LGTM

chittolinag:
I see the unit tests are failing. Is this expected @harbournick @falsaadehh ?

harbournick:
> I see the unit tests are failing. Is this expected @harbournick @falsaadehh ?

@chittolinag great catch! you're right - by changing the name of the block we introduced a namespace issue. And that's why we have unit tests! 
pushed a commit to fix

PR #794, opened 2025-08-21
fix: guarding replaceSpecialCharacters against non-strings

mattConnHarbour:
Making sure element text is string before passing to `replaceSpecialCharacters`. This was failing with images. Element text was equal to something similar to:
```
[
  {
    imagegcspath: 'path/to/storage',
    filename: 'picture.jpeg',
    ismultiple: false
  }
]
```

### DIFF
--- a/packages/super-editor/src/core/super-converter/exporter.js
+++ b/packages/super-editor/src/core/super-converter/exporter.js
@@ -1966,7 +1966,12 @@ function prepareUrlAnnotation(params) {
  * @param {String} annotationType
  * @returns {Function} handler for provided annotation type
  */
-function getTranslationByAnnotationType(annotationType) {
+function getTranslationByAnnotationType(annotationType, annotationFieldType) {
+  // invalid annotation
+  if (annotationType === 'text' && annotationFieldType === 'FILEUPLOADER') {
+    return null;
+  }
+
   const imageEmuSize = {
     w: 4286250,
     h: 4286250,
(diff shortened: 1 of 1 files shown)


packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -2283,7 +2283,10 @@ export class DocxExporter {
     if (elements) {
       if (name === 'w:instrText') {
         tags.push(elements[0].text);
-      } else if (name === 'w:t' || name === 'w:delText' || name === 'wp:posOffset') {
+      } else if (
+        (name === 'w:t' || name === 'w:delText' || name === 'wp:posOffset') &&
+        typeof elements[0].text === 'string'
```
harbournick: I do think we can add an additional safety here to prevent crashed but can we instead of checking if its of type string:
- if its not of type string (here in w:t we expected string), do a try/catch to attempt to convert to a String. if it fails, skip handling and ignore it. Log an error to the console.

This way we don't _force_ check in the if statement (we actually want to know if this happens again rather than skipping it) and attempt to parse it.

harbournick:
Discussed with @mattConnHarbour via chat here. The fix in the initial PR here is useful (+ see comment) but it only addresses the _symptom_ not the root cause. Matt will add a commit for the root cause.

packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -2284,8 +2289,15 @@ export class DocxExporter {
...
-        const text = this.#replaceSpecialCharacters(elements[0].text);
-        tags.push(text);
+        try {
+          // test for valid string
+          let text = String(elements[0].text);
+          text = this.#replaceSpecialCharacters(text);
+          tags.push(text);
+        } catch (error) {
+          console.error('Text element does not contain valid string:', error);
+          throw new Error('Invalid text element in XML generation');
```
harbournick: let's not throw an error here - we actually want to continue processing if we do encounter an error. the console.error is good

harbournick (changes requested):
tiny change! looks great otherwise

packages/super-editor/src/core/super-converter/exporter.js
mattConnHarbour: Done!

PR #796, opened 2025-08-21
feat: integrate markdown parsing into SuperDoc

caio-pizzol:


packages/super-editor/src/core/Editor.js
```diff
@@ -989,6 +1001,59 @@ export class Editor extends EventEmitter {
...
+
+  /**
+   * Convert Markdown to HTML with proper structure
+   * @private
+   * @param {string} markdown - Markdown content
+   * @returns {string} HTML content
+   */
+  #convertMarkdownToHTML(markdown) {
+    // Configure marked for better compatibility
+    marked.setOptions({
```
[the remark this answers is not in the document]
harbournick: worth double checking if the copilot is right

packages/super-editor/src/core/Editor.js
```diff
@@ -989,6 +1001,59 @@ export class Editor extends EventEmitter {
...
+    return html;
+  }
+
+  /**
+   * Prepare markdown-generated HTML for SuperDoc/DOCX compatibility
+   * @private
+   * @param {string} html - HTML from markdown parser
+   * @returns {string} HTML prepared for SuperDoc
+   */
+  #prepareHTMLForSuperDoc(html) {
```
harbournick: I think we should extract this out of the class and put it in a markdown helpers area. example: import { createLinkedChildEditor } from '@core/child-editor/index.js';

There may be more additions to make here and this will likely grow over time so we should design it in a way that it can grow easily. And add a quick set of unit tests for it!

harbournick: actually go one level higher and extract convertMarkdownToHTML too

harbournick (changes requested):
Looking good! just a minor comment to move convertMarkdownToHTML and prepareHTMLForSuperDoc to some helpers folder rather than making this Editor private methods. likely means the use of 'marked' will move to the helper and the import can be removed from Editor.js

Just an effort to keep this already huge class smaller!

packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -169,6 +170,30 @@ const generateDefaultHeaderFooter = (type, id) => {
...
+    content: node.content,
+    attrs: {
+      ...otherAttrs,
+      styleId: `Heading${level}`, // Maps to Heading1, Heading2, etc. in Word
+    },
+  };
+
+  // Use existing paragraph translator with the modified node
+  return translateParagraphNode({ ...params, node: paragraphNode });
+}
```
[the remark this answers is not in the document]
caio-pizzol: @harbournick is this relevant?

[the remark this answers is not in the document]
harbournick: nah, it actually uses the same attr so all good

packages/super-editor/src/core/Editor.js
```diff
@@ -37,6 +37,8 @@ import {
 import { createLinkedChildEditor } from '@core/child-editor/index.js';
 import { unflattenListsInHtml } from './inputRules/html/html-helpers.js';
 import { SuperValidator } from '@core/super-validator/index.js';
+import { createDocFromMarkdown } from './helpers/importMarkdown.js';
+import { createDocFromHTML } from './helpers/importHtml.js';
```
harbournick: in ./helpers we should have an index.js file that exports from all helpers
can you add these to that file and just import from both the index? (for consistency)

packages/super-editor/src/components/SuperEditor.vue
```diff
@@ -95,8 +100,14 @@ const pollForMetaMapData = (ydoc, retries = 10, interval = 500) => {
 };
 
 const loadNewFileData = async () => {
+  fileSource.value = props.fileSource;
+  if (!fileSource.value || fileSource.value.type !== DOCX) {
+    fileSource.value = await getFileObject(BlankDOCX, 'blank.docx', DOCX);
+    console.debug('file source', fileSource.value);
```
harbournick: I know this was me but mind removing this console log?

harbournick (changes requested):
All looks good but could you please add some unit tests (in the same folder, next to the file) for:
packages/super-editor/src/core/helpers/importHtml.js
packages/super-editor/src/core/helpers/importMarkdown.js

Also add // @ts-check at the top of these and verify any typing issues are working. Since these are small self-contained files we should be able to pass ts validation

packages/super-editor/src/core/Editor.js
caio-pizzol: Yes!

packages/super-editor/src/components/SuperEditor.vue
caio-pizzol: Yes!

PR #797, opened 2025-08-21
fix: automatically insert blank docx if SuperEditor file is not docx type

harbournick:
allow .md files in developer playground for SuperEditor


caio-pizzol (approved):
LGTM! Nice

PR #798, opened 2025-08-22
Doc gen: Add aliases to tsconfig, remove ts-ignore in document sections 

harbournick:
- Fixes issues in document-sections preventing build if ts-ignore is removed
- Adds vite aliases to tsconfig


caio-pizzol:
Nice! Thanks Nick

PR #799, opened 2025-08-22
Draft: fix(underline): underline handling that allows for run underline to o…

palmer-cl:
- Allows for override of paragraph underlines with run underlines that contain no w:val.

Additional context based on my findings:
- Word may have <w:u> tags but does not express them as underlines if they have no `val` property
- The document sent via email has mostly all of the text underlined, but not expressed because at the run level the underlines have no `val=single` or any other type of underline
- So if on the paragraph you have <w:u val=single> but on a run you then have <w:u color='fff'> this underline should not be expressed

Pay close attention to:
- we swap the order of mark processing to process paragraph first, then run marks so that the run marks take precendence. I didn't see any ramifications but good to double check this. 

Example p tag:

image

### DIFF
--- a/packages/super-editor/src/assets/styles/layout/global.css
+++ b/packages/super-editor/src/assets/styles/layout/global.css
@@ -14,3 +14,8 @@
 a {
   text-decoration: auto;
 }
+
+u:has(u.underline-hidden),
+u.underline-hidden {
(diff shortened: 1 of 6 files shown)


palmer-cl:
Closing this going to take a different approach

PR #800, opened 2025-08-22
HAR-10247: add validator rules to clean up numberingxml

falsaadehh:


packages/super-editor/src/core/super-validator/super-validator.js
```diff
@@ -42,35 +46,42 @@ export class SuperValidator {
...
   #initializeValidators() {
     const requiredNodes = new Set();
     const requiredMarks = new Set();
 
-    const validators = Object.fromEntries(
-      Object.entries(StateValidators).map(([key, factory]) => {
-        const validatorLogger = this.logger.withPrefix(key);
-        /** @type {ValidatorFunction} */
-        const validator = factory({ editor: this.#editor, logger: validatorLogger });
+    const initializeValidatorSet = (validatorFactories) => {
```
harbournick, line 64: nice!

packages/super-editor/src/core/super-validator/super-validator.js
```diff
@@ -174,4 +185,35 @@ export class SuperValidator {
...
+
+  /**
+   * Validate the exported document in the editor. Triggered automatically on editor export.
+   * @returns {{ modified: boolean, results: Array<{ key: string, results: string[] }> }}
+   */
+  validateDocumentExport() {
+    const { tr } = this.#editor.state;
+    const { dispatch } = this.#editor.view;
+
+    const documentAnalysis = this.#analyzeDocument();
```
harbournick: This is where we'll want to diverge between validateActiveDocument and validateDocumentExport. There _may_ be cases where we want to validate something about editor state in the export validator, not sure, but primarily we should use it to validate actual XML files. 

The current numberingValidator (while it looks good) is actually validating state (ie: takes listItems from the current state). We _could_ validate these - and we may want to explore wether or not it makes sense to add these validators to the active document validation (not sure - needs a bit of investigation). But let's remove them from the export validator. In this case we want to take editor.converter.convertedXml and ensure things are correct there.

For the current task, we should take editor.converter.convertedXml['word/numbering.xml'] - you should be able to confirm in it that there is a null numId, and we will want to remove that definition from the list. 

Later on (not necessary for this initial case), we would also want to look at document.xml (and possibly styles.xml) to check for any numPr tags that have a "null" numId. But that would be a separate (bigger) check we'll do later.

harbournick: One note: the null ID is easy to reproduce if you update your branch from develop (but don't include https://github.com/Harbour-Enterprises/SuperDoc/pull/804) and use a markdown file that has a list in SuperDoc dev. The markdown conversion is generating numbering.xml files with this null numId so its an easy way to test this one out

packages/super-editor/src/core/super-validator/super-validator.js
```diff
@@ -174,4 +185,35 @@ export class SuperValidator {
...
+      this.logger.debug(`🕵 Validating export with ${key}...`);
+
+      const { results, modified } = validator(tr, documentAnalysis);
+      validationResults.push({ key, results });
+
+      hasModifiedDocument = hasModifiedDocument || modified;
+    });
+
+    if (!this.dryRun) dispatch(tr);
+    else this.logger.debug('DRY RUN: No export changes applied to the document.');
```
harbournick, line 211: in this validator, we will need to actually update editor.convert.convertedXml files if we make any updates to them here.

packages/super-editor/src/core/super-validator/validators/xml/numbering/rules/list-item-numid-level.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Ensure each listItem has numId and level. If missing, create parent list type,
+ * create/generate a numbering definition and set attrs via tr.setNodeMarkup.
+ * @param {import('../../../../types.js').ElementInfo[]} listItems
+ * @param {import('../../../../types.js').Editor} editor
+ * @param {import('prosemirror-state').Transaction} tr
+ * @param {import('../../../../types.js').ValidatorLogger} logger
+ * @returns {{ modified: boolean, results: string[] }}
+ */
+export function ensureListItemHasNumIdAndLevel(listItems, editor, tr, logger) {
```
harbournick: See note above. Let's remove this validation rule for now (its state validation). I'm not sure if we should or should not add it to active document validation. It would be necessary to find an example file where this validation actually takes effect and verify that the behavior is needed. It is very possible that it is needed (and would be most obvious in collaborative files but maybe the markdown test above might work).

packages/super-editor/src/core/super-validator/validators/xml/numbering/rules/numid-has-definition.js
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Ensure each listItem.numId has a corresponding numbering definition.
+ * If missing, generate it based on the parent list type.
+ * @param {import('../../../../types.js').ElementInfo[]} listItems
+ * @param {import('../../../../types.js').Editor} editor
+ * @param {import('prosemirror-state').Transaction} tr
+ * @param {import('../../../../types.js').ValidatorLogger} logger
+ * @returns {{ modified: boolean, results: string[] }}
+ */
+export function ensureNumIdHasDefinition(listItems, editor, tr, logger) {
```
harbournick: same here as the other rule

packages/super-editor/src/core/super-validator/super-validator.js
harbournick: Now that createNumberingValidator only needs the XML we likely won't need to analzeDocument here fyi

falsaadehh:
> The new approach looks great! Let's clean up validateDocumentExport() and add tests (+ types where poss) and that should do it.

@harbournick ready for a final look :D

harbournick (approved):
LGTM - nice work!

PR #803, opened 2025-08-22
HAR-10370 - fix: strikethrough not active on toolbar

chittolinag:
Strikethrough is not active on the toolbar even if the selected text has it. This happens because the mark is added with the `strike` name, but the toolbar is added as `strikethrough`. So, when the [updateToolbarState](https://github.com/Harbour-Enterprises/SuperDoc/blob/develop/packages/super-editor/src/components/toolbar/super-toolbar.js#L713) function is called, it doesn't set it as active. 

Screenshot 2025-08-22 at 14 57 53


harbournick (approved):
LGTM

PR #814, opened 2025-08-26
refactor: update document version handling and add SuperDoc ID [SD-428]

caio-pizzol:
- Renamed `updateDocumentVersion` to `setDocumentVersion` for clarity.
- Introduced methods to get and set SuperDoc ID and version in `SuperConverter`.
- Improved error handling and structure for custom properties in docx.
- Added unit tests for new functionality in `SuperConverter`.

Notes:
- MD5 hash (when GUID present)
- Insert GUID on key stroke
- Do not count on blank document

### DIFF
--- a/packages/super-editor/src/core/Editor.js
+++ b/packages/super-editor/src/core/Editor.js
@@ -937,7 +937,7 @@ export class Editor extends EventEmitter {
     return SuperConverter.getDocumentGuid(doc);
   }
 
-  // Deprecated method for backward compatibility
+  // Deprecated
   /**
    * @deprecated use setDocumentVersion instead
    */
@@ -1386,8 +1386,14 @@ export class Editor extends EventEmitter {
       return;
     }
 
-    // Document modified - promote to GUID if needed
-    this.#ensureDocumentGuid();
+    // Track document modifications and promote to GUID if needed
+    if (transaction.docChanged && this.converter) {
+      if (!this.converter.documentGuid) {
+        this.converter.promoteToGuid();
+        console.debug('Document modified - assigned GUID:', this.converter.documentGuid);
+      }
+      this.converter.documentModified = true;
+    }
 
     this.emit('update', {
       editor: this,
@@ -1396,34 +1402,19 @@ export class Editor extends EventEmitter {
   }
 
   /**
-   * Ensure document has a permanent GUID if it's been modified
-   * @private
-   * @returns {void}
+   * Get document identifier for telemetry (async - may generate hash)
+   * @returns {Promise<string>} GUID for modified docs, hash for unmodified
    */
-  #ensureDocumentGuid() {
-    if (!this.converter) return;
-
-    // If we only have a hash, promote to GUID
-    if (this.converter.hasTemporaryId()) {
-      this.converter.promoteToGuid();
-      console.debug('Document modified - assigned permanent GUID');
-    }
+  async getDocumentIdentifier() {
+    return (await this.converter?.getDocumentIdentifier()) || null;
   }
 
   /**
-   * Get the document GUID
-   * @returns {string|null} The unique document GUID
+   * Get permanent document GUID (sync - only for modified documents)
+   * @returns {string|null} GUID or null if document hasn't been modified
    */
   getDocumentGuid() {
-    return this.converter?.getDocumentGuid() || null;
-  }
-
-  /**
-   * Get document identifier for telemetry (works for both modified and unmodified)
-   * @returns {string|null} GUID for modified docs, hash for unmodified
-   */
-  getDocumentIdentifier() {
-    return this.converter?.getDocumentIdentifier() || null;
+    return this.converter?.documentGuid || null;
   }
 
   /**
@@ -1435,16 +1426,21 @@ export class Editor extends EventEmitter {
   }
 
   /**
-   * Get telemetry data including document identifier
-   * @returns {Object} Telemetry data object
+   * Get telemetry data (async because of lazy hash generation)
    */
-  getTelemetryData() {
+  async getTelemetryData() {
     return {
-      documentId: this.getDocumentIdentifier(), // Hash or GUID
+      documentId: await this.getDocumentIdentifier(),
       isModified: this.isDocumentModified(),
-      isPermanentId: !this.converter?.hasTemporaryId(),
+      isPermanentId: !!this.converter?.documentGuid,
       version: this.converter?.getSuperdocVersion(),
     };
+  }
+
+  // Deprecated for backward compatibility
+  getDocumentId() {
+    console.warn('getDocumentId is deprecated, use getDocumentGuid instead');
+    return this.getDocumentGuid();
   }
 
   /**
@@ -1482,8 +1478,7 @@ export class Editor extends EventEmitter {
     return {
       ...json,
       metadata: {
-        documentId: this.getDocumentIdentifier(),
-        documentGuid: this.converter?.documentGuid || null, // Only if modified
+        documentGuid: this.converter?.documentGuid || null,
         isModified: this.isDocumentModified(),
         version: this.converter?.getSuperdocVersion() || null,
       },
@@ -1646,6 +1641,7 @@ export class Editor extends EventEmitter {
       const json = this.#prepareDocumentForExport(comments);
 
       // Export the document to DOCX
+      // GUID will be handled automatically in converter.exportToDocx if document was modified
       const documentXml = await this.converter.exportToDocx(
         json,
         this.schema,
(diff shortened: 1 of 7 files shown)


packages/super-editor/src/core/Editor.js
```diff
@@ -904,8 +904,8 @@ export class Editor extends EventEmitter {
    * @param {string} version - New version
    * @returns {Object}
    */
-  static updateDocumentVersion(doc, version) {
-    const updatedContent = SuperConverter.updateDocumentVersion(doc, version);
+  static setDocumentVersion(doc, version) {
```
harbournick, line 926: Since this changes a potentially in-use API I think we should keep the old name and redirect it to the new name.

packages/super-editor/src/core/Editor.js
```diff
@@ -1400,6 +1400,22 @@ export class Editor extends EventEmitter {
     return html;
   }
 
+  /**
+   * Get the document ID from the converter
+   * @returns {string|null} The unique SuperDoc ID for this document
+   */
+  getDocumentId() {
```
harbournick: I think we should make this getDocumentUID or getDocumentXMLUID because the documentId seems more like it might point to the options.documentId config key (which is different from the XML UID)

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -132,6 +132,9 @@ class SuperConverter {
     this.fileSource = params?.fileSource || null;
     this.documentId = params?.documentId || null;
 
+    // SuperDoc ID - unique identifier for this document
+    this.superdocId = null;
```
harbournick: let's make this a bit more clear that its the UID for xml

harbournick (changes requested):
Looks good and I like the ideas. Just a few comments for clarity

packages/super-editor/src/core/Editor.js
caio-pizzol: Good point!

packages/super-editor/src/core/Editor.js
caio-pizzol: Spot on!

caio-pizzol:
@edoversb @harbournick I think we need a more deep analysis here before merging this

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -168,53 +176,254 @@ class SuperConverter {
...
 
-  static getStoredSuperdocVersion(docx) {
+  /**
+   * Generic method to get a stored custom property from docx
+   * @static
+   * @param {Array} docx - Array of docx file objects
+   * @param {string} propertyName - Name of the property to retrieve
+   * @returns {string|null} The property value or null if not found
+   */
+  static getStoredCustomProperty(docx, propertyName) {
```
harbournick, line 275: Let's write some unit tests for this one. I know its not a new function, but since there are some bigger changes its probably a good time to cover it.

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -168,53 +176,254 @@ class SuperConverter {
...
+  /**
+   * Generic method to set a stored custom property in docx
+   * @static
+   * @param {Object} docx - The docx object to store the property in
+   * @param {string} propertyName - Name of the property
+   * @param {string|Function} value - Value or function that returns the value
+   * @param {boolean} preserveExisting - If true, won't overwrite existing values
+   * @returns {string} The stored value
+   */
+  static setStoredCustomProperty(docx, propertyName, value, preserveExisting = false) {
```
harbournick, line 305: let's add unit tests here too

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -168,53 +176,254 @@ class SuperConverter {
...
+      this.documentHash = await this.generateDocumentHash();
+    }
+
+    return this.documentHash;
+  }
+
+  /**
+   * Check if using temporary ID
+   */
+  hasTemporaryId() {
```
harbournick: can be private?

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -168,53 +176,254 @@ class SuperConverter {
...
+
+    // Then check custom property
+    return SuperConverter.getStoredCustomProperty(docx, 'DocumentGuid');
+  }
+
+  /**
+   * Get the permanent document GUID
+   * @returns {string|null} The document GUID (only for modified documents)
+   */
+  getDocumentGuid() {
```
harbournick, line 407: Curious: why do we need this in addition to the static of the same name?

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -168,53 +176,254 @@ class SuperConverter {
...
+      return this.documentInternalId.replace(/[{}]/g, '');
+    }
+    return null;
+  }
+
+  /**
+   * Generate document hash for telemetry (async, lazy)
+   */
+  async generateDocumentHash() {
+    if (!this.fileSource) return `HASH-${Date.now()}`;
```
harbournick, line 456: This won't work in the collaboration case where we don't have a fileSource available. In collaboration we load 'converter.docx' from the yjs binary. We might need to handle this? That said, all files opened collaboratively will already have run through _at least one import_ process with the original file.

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -168,53 +176,254 @@ class SuperConverter {
...
+    if (this.documentInternalId) {
+      return this.documentInternalId.replace(/[{}]/g, '');
+    }
+    return null;
+  }
+
+  /**
+   * Generate document hash for telemetry (async, lazy)
+   */
+  async generateDocumentHash() {
```
harbournick: can be made private

packages/super-editor/src/core/Editor.js
```diff
@@ -1381,7 +1449,15 @@ export class Editor extends EventEmitter {
    * @returns {Object} Editor content as JSON
    */
   getJSON() {
-    return this.state.doc.toJSON();
+    const json = this.state.doc.toJSON();
+    return {
```
harbournick, line 1489: I don't think think we should change the intent of this function (editor content as JSON). No metadata here. We can add a separate getMetadata() function if needed.

harbournick (changes requested):
Just a few comments. Overall it looks good to me. However main points:
- Need to check if this works in collaboration
- Need to test what happens if converter.docx = [] (ie: we are not in docx mode)
- Please add unit tests to all new functions
- We need to create a few test files to test the various cases (missing GUID etc). How can we add this to automated testing (either within this repo, or in the separate tests repo)

packages/super-editor/src/core/super-converter/SuperConverter.js
luccas-harbour: Added tests to `packages/super-editor/src/core/super-converter/SuperConverter.test.js`

packages/super-editor/src/core/super-converter/SuperConverter.js
luccas-harbour: Also added tests for this in `packages/super-editor/src/core/super-converter/SuperConverter.test.js`

packages/super-editor/src/core/super-converter/SuperConverter.js
luccas-harbour: I made it private and then noticed it was not being referenced anywhere but tests so I just created a helper function in the test itself and removed this method. Let me know if that works.

packages/super-editor/src/core/super-converter/SuperConverter.js
luccas-harbour: I noticed that the static method was used to read the document GUID from a docx file, while the non-static one was simply a getter for the `documentGuid` attribute. I renamed the static one to `extractDocumentGuid` to avoid confusion.

packages/super-editor/src/core/super-converter/SuperConverter.js
luccas-harbour: Done 👍

packages/super-editor/src/core/Editor.js
luccas-harbour: I changed this method back to what it was and added a new `getMetadata` method.

packages/super-editor/src/core/super-converter/SuperConverter.js
luccas-harbour: I suppose in that case it would simply use a hash containing the current date which would be ok? Also, if the document being opened collaboratively has already been exported once, it will contain a permanent GUID so a hash would not need to be generated. What do you think @harbournick?

harbournick (approved):
LGTM

PR #816, opened 2025-08-26
fix(track changes): fix permissions issue and group replace changes

VladaHarbour:
1. Show correct comment dialog for track changes
2. Group replace changes into one comment

### DIFF
--- a/packages/superdoc/src/stores/comments-store.js
+++ b/packages/superdoc/src/stores/comments-store.js
@@ -122,6 +122,7 @@ export const useCommentsStore = defineStore('comments', () => {
       authorEmail,
       date,
       author: authorName,
+      importedAuthor,
       documentId,
       coords,
     } = params;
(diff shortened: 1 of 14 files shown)


packages/superdoc/src/stores/comments-store.js
```diff
@@ -489,22 +498,59 @@ export const useCommentsStore = defineStore('comments', () => {
...
       }
     });
   };
 
+  /**
+   * Combines replace transaction which is represented by insertion + deletion
+   *
+   * @returns {Array} grouped track changes array
+   */
+  const groupChanges = (changes) => {
```
harbournick: Since we are trying to add unit tests for all new fns, can you add a test for this one?

Since it isn't so straight forward to unit test a pinia store, feel free to move this to a helper, and have the store call the (tested) helper.

PR #818, opened 2025-08-27
feat: add table module documentation + keyboard shortcuts

caio-pizzol:
- Added keyboard shortcuts for bold (Mod-b, Mod-B), italic (Mod-i, Mod-I), underline (Mod-u, Mod-U), highlight (Mod-Shift-h), and strikethrough (Mod-Shift-s).
- Enhanced table module with detailed configuration options, commands, and helper functions for better usability and documentation clarity.

### DIFF
--- a/packages/super-editor/src/extensions/table-cell/table-cell.js
+++ b/packages/super-editor/src/extensions/table-cell/table-cell.js
@@ -71,14 +71,26 @@ export const TableCell = Node.create({
 
   addAttributes() {
     return {
+      /**
+       * @category Attribute
+       * @param {number} [colspan=1] - Number of columns this cell spans
+       */
       colspan: {
         default: 1,
       },
 
+      /**
+       * @category Attribute
+       * @param {number} [rowspan=1] - Number of rows this cell spans
+       */
       rowspan: {
         default: 1,
       },
 
+      /**
+       * @category Attribute
+       * @param {number[]} [colwidth=[100]] - Column widths array in pixels
+       */
       colwidth: {
         default: [100],
         parseDOM: (elem) => {
@@ -94,16 +106,10 @@ export const TableCell = Node.create({
         },
       },
 
-      /* width: {
-        renderDOM: ({ width, widthType, widthUnit }) => {
-          if (!width) return {};
-          let unit = widthUnit === 'px' ? widthUnit : 'in';
-          if (widthType === 'pct') unit = '%';
-          const style = `width: ${width}${unit}`;
-          return { style };
-        },
-      }, */
-
+      /**
+       * @category Attribute
+       * @param {CellBackground} [background] - Cell background color configuration
+       */
       background: {
         renderDOM({ background }) {
           if (!background) return {};
@@ -113,6 +119,10 @@ export const TableCell = Node.create({
         },
       },
 
+      /**
+       * @category Attribute
+       * @param {string} [verticalAlign] - Vertical content alignment (top, middle, bottom)
+       */
       verticalAlign: {
         renderDOM({ verticalAlign }) {
           if (!verticalAlign) return {};
@@ -121,6 +131,10 @@ export const TableCell = Node.create({
         },
       },
 
+      /**
+       * @category Attribute
+       * @param {CellMargins} [cellMargins] - Internal cell padding
+       */
       cellMargins: {
         renderDOM({ cellMargins }) {
           if (!cellMargins) return {};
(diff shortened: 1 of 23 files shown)


packages/super-editor/src/extensions/table-cell/table-cell.js
```diff
@@ -121,4 +187,29 @@ export const TableCell = Node.create({
...
+       *
+       * // Using default values
+       * const borders = createCellBorders()
+       *
+       * // Using custom values
+       * const borders = createCellBorders({ size: 1, color: '#cccccc' })
+       * @note Creates uniform borders for all four sides of a cell
+       * @note Default size matches Word's default cell border width
+       */
+      createCellBorders: ({ size = 0.66665, color = '#000000' } = {}) => {
```
harbournick: I think this helper is intended to be internal (not to expose via API) since its just for setting up the node. I think we should remove from here

packages/super-editor/src/extensions/table/table.js
```diff
@@ -636,20 +955,108 @@ export const Table = Node.create({
...
+       * @param {number} colsCount - Number of columns
+       * @param {boolean} withHeaderRow - Create first row as header
+       * @param {Object} [cellContent=null] - Initial cell content
+       * @returns {Object} Complete table node with borders
+       * @example
+       * const table = createTable(schema, 3, 3, true)
+       * @example
+       * const table = createTable(schema, 2, 4, false, paragraphNode)
+       */
+      createTable: (schema, rowsCount, colsCount, withHeaderRow, cellContent = null) => {
```
harbournick: Let's remove this one from the external API. Internal helper for the command.

packages/super-editor/src/extensions/table/table.js
```diff
@@ -636,20 +955,108 @@ export const Table = Node.create({
...
+       * @returns {TableBorders} Complete borders object for all sides
+       * @example
+       * // Using default values
+       * const borders = createTableBorders()
+       *
+       * // Using custom values
+       * const borders = createTableBorders({ size: 1, color: '#cccccc' })
+       * @note Creates uniform borders for all sides including inside borders
+       */
+      createTableBorders: ({ size = 0.66665, color = '#000000' } = {}) => {
```
harbournick: This might be intended as an internal helper too. @artem-harbour can you confirm?

packages/super-editor/src/extensions/table/table.js
```diff
@@ -636,20 +955,108 @@ export const Table = Node.create({
...
+       * Check if selection is a cell selection
+       * @category Helper
+       * @param {*} value - Selection to check
+       * @returns {boolean} True if cell selection
+       * @example
+       * if (isCellSelection(editor.state.selection)) {
+       *   // Handle cell selection
+       * }
+       */
+      isCellSelection: (value) => {
```
harbournick: let's leave internal only

packages/super-editor/src/extensions/table/table.js
```diff
@@ -636,20 +955,108 @@ export const Table = Node.create({
...
+       * Check if cursor is inside a table
+       * @category Helper
+       * @param {Object} state - Editor state
+       * @returns {boolean} True if cursor is in table
+       * @example
+       * if (isInTable(state)) {
+       *   // Enable table-specific commands
+       * }
+       */
+      isInTable: (state) => {
```
harbournick: let's leave internal only for now.

I think it will be helpful to have an API for determining selection context, but I don't think it should be delegated to each individual extension. We will figure out a way to make a simpler API for this

harbournick (changes requested):
Mainly comments to remove internal helpers. It does highlight the point that we need to more clearly decide what is an internal helper (ie: to be used inside a command) vs what should be external 'helpers' API

packages/super-editor/src/extensions/table-cell/table-cell.js
caio-pizzol: Spot on! I made @private arg to filter everything that we shouldn't expose - let's use that!

packages/super-editor/src/extensions/table/table.js
caio-pizzol: Removed for now!

packages/super-editor/src/extensions/table/table.js
caio-pizzol: Done

harbournick (approved):
LGTM

PR #821, opened 2025-08-29
fix: jsonOverride documentation

VladaHarbour:
We expect jsonOverride to be an actual json to override content with in generatePmData function of Editor.js but documentation says that it's boolean. Eric faced this issue while trying to create an example of Superdoc usage


harbournick (approved):
LGTM

PR #823, opened 2025-08-29
HAR-10176 - SuperDoc automated testing demo gallery apps

chittolinag:
### Description

This PR introduces a base infrastructure with Playwright to run tests on *pretty much* every examples on the `examples/` folder. I noticed that not everything can be tested right away because they need different setups, and some of them depend on external services (ie, word-addin for example). Same for the node example, which we wouldn't be able to use playwright.

In short, I'm testing pretty much all examples that share the same setup and that are working properly.

### Examples that were ignored here for now

1. agentic-slack-redlining-example  -> require its own testing
2. chrome-extension  -> require its own testing
3. docxtemplater-example  -> screenshot not being taken reliably
4. grading-papers-example -> doesn't actually test SD directly unless we write a specific test because the SD is not rendered on the main page
5. nodejs-example  -> requires its own testing
6. vue-custom-node-example -> something fails when we try to open it
7. vue-html-editor -> needs a different setup


.github/workflows/run-tests-example-apps.yml
```diff
@@ -0,0 +1,30 @@
+name: Run tests for example apps
+
+on:
+  workflow_call:
+  # TODO: Enable this when we test this workflow manually
+  # pull_request:
```
chittolinag, line 6: Once we merge this PR, I will trigger this workflow manually. If it works, I will open a subsequent PR simply to enable this on every PR creation to `develop`.

harbournick (approved):
LGTM

PR #834, opened 2025-09-01
feat: register images from paste + HTML insertion, fixes #790

### LINKED ISSUE #790: how to export image(it's invalid when I imported by editorRef.current.activeEditor.commands.insertContent(`<img src=${base64}>`))
image export not working when I use editorRef.current.activeEditor.commands.insertContent(``)

when I open the docx file, I found this and I clicked yes

ps: 
1. no error occur when I insertContent or export
2. I tried to upload manual by toolbar Image item, which is exported normally.
3.superdoc 0.14.15

johanneswilm:


packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -307,7 +281,7 @@ function generateParagraphProperties(node) {
     if (hanging || hanging === 0) attributes['w:hanging'] = pixelsToTwips(hanging);
 
     if (textIndent && !attributes['w:left']) {
-      attributes['w:left'] = getTextIndentExportValue(textIndent);
+      attributes['w:left'] = inchesToTwips(textIndent);
```
VladaHarbour: I suggest not to remove this code because we faced some documents with pasted content that contain text indent value with points not inches
https://linear.app/harbour/issue/HAR-10339/whalar-copy-and-pasted-terms-and-conditions-pushed-to-one-side-in

johanneswilm: Hey @VladaHarbour That was not the intention. This is not a file I touched. This must have occurred during auto merging. I did not see this earlier today. I'll try to clean up.

johanneswilm: Done. It seems to have been a merge error.

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
```diff
@@ -0,0 +1,210 @@
...
+          set = meta.set;
+          return { set };
+        }
+        // Adjust decoration positions to changes made by the transaction
+        set = set.map(tr.mapping, tr.doc);
+
+        return { set };
+      },
+    },
+    appendTransaction: (trs, _oldState, state) => {
```
artem-harbour, line 41: I would suggest adding something like this here.

```javascript
let docChanges = trs.some((tr) => tr.docChanged) && !_oldState.doc.eq(state.doc);

if (!docChanges) {
  return;
}
```

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
```diff
@@ -0,0 +1,210 @@
...
+        .forEach(({ node, pos }) => {
+          tr.delete(pos, pos + node.nodeSize);
+        });
+      // Map the decoration set through the transaction to adjust positions
+      set = set.map(tr.mapping, tr.doc);
+
+      // Set the updated decoration set in the transaction metadata
+      tr.setMeta(key, { set });
+
+      return tr;
```
artem-harbour, line 110: I think here we only need to return the transaction if changes were actually made, otherwise - `null` or `undefined`.

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -5,7 +5,12 @@ import { makeDefaultItems } from './defaultItems';
 import { getActiveFormatting } from '@core/helpers/getActiveFormatting.js';
 import { vClickOutside } from '@harbour-enterprises/common';
 import Toolbar from './Toolbar.vue';
-import { startImageUpload, getFileOpener } from '../../extensions/image/imageHelpers/index.js';
+import {
+  checkAndProcessImage,
+  replaceSelectionWithImagePlaceholder,
+  uploadAndInsertImage,
+  getFileOpener,
+} from '../../extensions/image/imageHelpers/index.js';
```
artem-harbour, line 13: I'm sure we have a lot of imports like this, but I think we need to minimize it because it increases the coupling of extensions with the rest of the codebase.

fyi @harbournick

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
johanneswilm: @artem-harbour That would mean you'd compare the full input and output document of every transaction (`trs` is usually just one `tr`). In my experience, that is not exactly performant.

johanneswilm: The second check of tr.docChanged is already included below and we would then do this check twice.

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
johanneswilm: @artem-harbour Yes, that is already happening. See above line 72-74.

artem-harbour (commented):
Added a couple of comments for consideration.

Note: also need to pay extra attention to adding new `appendTransaction` so that it doesn't lead to infinite loops and affect performance.

fyi @harbournick @johanneswilm

packages/super-editor/src/components/toolbar/super-toolbar.js
artem-harbour: This is just for info, no actions needed here now.

artem-harbour (approved):
LGTM

cc @harbournick

packages/super-editor/src/components/toolbar/super-toolbar.js
johanneswilm: @artem-harbour I am trying to reuse parts of what the image uploader is using. The issue here is that of 3 parts the image uploader is using, we need parts 1 and 3 and skip part 2. That's why it needed to be split into three parts.

The other option would have been to make the image uploader work together with this new code directly. That would also have been weird, as it would make the image uploader first insert the image in the document, only for this for this new code to remove it again, then to clean it and potentially insert it a second time.

For that reason, there isn't really a much better way of doing this.

Btw - I noticed you guys were exporting the plugin keys in other files. That's not how it's intended in PM. One is supposed to keep these private within one file to ensure that all changes to the meta of a given plugin are only done within functions within that same file.

johanneswilm:
> Added a couple of comments for consideration.

I believe I answered all of those.
 
> Note: also need to pay extra attention to adding new `appendTransaction` so that it doesn't lead to infinite loops and affect performance.

True, that is always important to consider. But as long as one has it limited to only acting in very specific cases and filtering out all other cases and not performing any dom lookups/operations before determining that another `tr` is needed, it should not have much of an impact on performance.

If one would want to limit the number of appendTransactions in the future, one could turn this into a single function that is executed to find all relevant content that enters through `paste` and the `insertContent` command. It could then make sense to move it to a clipboard handling plugin or similar.

harbournick:
I moved everything to an internal PR and merged from there FYI: https://github.com/Harbour-Enterprises/SuperDoc/pull/1010
Closing this one. Just seemed quicker as I was testing and resolving merge conflicts. Feature looks great @johanneswilm !

PR #843, opened 2025-09-02
fix(internal): remove pdfjs from build

artem-harbour:
This PR removes pdfjs from the final SD build and significantly optimize the final SD build size - `30,5 MB -> 22,4 MB`.

See linear ticket for more info.


packages/superdoc/src/components/PdfViewer/pdf/pdf-viewer.css
```diff
@@ -0,0 +1 @@
+@import 'pdfjs-dist/web/pdf_viewer.css';
```
artem-harbour, line 1: pdf.js viewer styles must be imported in the component itself in order to be processed by postcss to be scoped to SD class. Otherwise they would be global and affect other app components, styles.

packages/superdoc/package.json
```diff
@@ -64,7 +64,7 @@
   },
   "peerDependencies": {
     "@hocuspocus/provider": "^2.13.6",
-    "pdfjs-dist": "4.3.136",
+    "pdfjs-dist": ">=4.3.136 <=4.6.82",
```
artem-harbour, line 67: Need to keep specific pdfjs versions here as breaking changes for pdf viewer styles come next (and it breaks the layout).

I think to be more independent from pdfjs versions and their styles, SD pdf viewer should be redesigned later and use pdfjs api instead of `pdfjs-dist/web/pdf_viewer` functionality. Or maybe another library for pdf rendering.

artem-harbour:
Hey @caio-pizzol, which commit prefix would work better here (refactor or fix)?

I think it's not as important here as updating the version, since at the moment this is internal functionality. But we need to update the version here anyway for webapp related updates.

fyi @harbournick

caio-pizzol:
> Hey @caio-pizzol, which commit prefix would work better here (refactor or fix)?
> 
> I think it's not as important here as updating the version, since at the moment this is internal functionality. But we need to update the version here anyway for webapp related updates.
> 
> fyi @harbournick

There're 2 ways to do it:

1) Use specific scopes: `fix(internal): remove pdfjs from build pkg` (what I recomend for now)

2) Create custom prefix types
```json
// .releaserc.json
{
  "plugins": [
    ["@semantic-release/commit-analyzer", {
      "releaseRules": [
        {"type": "refactor", "release": "patch"},  // Make refactor trigger patch
        {"type": "internal", "release": "patch"}   // Custom type
      ]
    }]
  ]
}
```

PR #859, opened 2025-09-03
fix: add safety check for clipboard usage

palmer-cl:
- Adds checks for clipboard read access 
- Fails safely to no content in clipboard if permissions are not granted or navigator does not exist


harbournick (changes requested):
Could you add a test file next to clipboardUtils.js (clipboardUtils.test.js) with some tests to cover this function? also let's include one test case for the current bug

palmer-cl:
Tests added, also removed some unused utils

harbournick (approved):
LGTM

PR #863, opened 2025-09-03
fix: restore stored marks if they exist

artem-harbour:


packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -122,13 +123,12 @@ export const BlockNode = Extension.create({
...
           if (hasInitialized && !checkForNewBlockNodesInTrs(transactions)) return null;
 
-          let tr = null;
+          const { tr } = newState;
           let changed = false;
           newState.doc.descendants((node, pos) => {
             // Only allow block nodes with a valid sdBlockId attribute
             if (!nodeAllowsSdBlockIdAttr(node) || !nodeNeedsSdBlockId(node)) return null;
 
-            tr = tr ?? newState.tr;
```
artem-harbour, line 131: Not sure why it was like that.

PR #868, opened 2025-09-03
feat: add custom toolbar button example (HAR-10436)

luccas-harbour:
Adds an example on how to add a custom button to Superdoc's toolbar

[Linear ticket](https://linear.app/harbour/issue/HAR-10436/luccas-starter-project)


examples/customized-toolbar-example/package.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "vanilla-superdoc-example",
```
harbournick: let's change the package name to the example name

examples/customized-toolbar-example/package.json
```diff
@@ -0,0 +1,15 @@
...
+    "private": true,
+    "version": "0.0.1",
+    "type": "module",
+    "scripts": {
+        "dev": "vite"
+    },
+    "dependencies": {
+        "@harbour-enterprises/superdoc": "^0.11.45"
```
harbournick: I know most of our examples are using this version, but mind updating to the latest version (0.16.0)? We'll eventually update all the other ones (not for this task).

examples/customized-toolbar-example/src/main.js
```diff
@@ -0,0 +1,152 @@
...
+            });
+        },
+      };
+    },
+    addPmPlugins() {
+        return [
+            new Plugin({
+                key: new PluginKey('gifNodePlugin'),
+                props: {
+                    handleClickOn: (view, pos, node, nodePos, event, direct) => {
```
harbournick, line 78: Since adding functionality for 'on click' to the image is a pretty specific use case (and unrelated to the custom toolbar button) let's add a comment here about what this code does so that anyone looking at this example in the future will know this is _optional_ and an example of _something else_

harbournick (changes requested):
The example works great! Nice work. Aside from the comments below, let's also:
- Add a super brief README.md file (some examples have this, trying to make them all have it) so that if someone is just browsing the folder they'll know what to expect from this example. I would add a small blurb about both the custom toolbar button and handling the click event on the image in this case. 
- There is a bug introduced by this plugin, having to do with normal image nodes. If you insert a normal image, and click on it, _it_ will also change to a cat gif. Not really a big deal, but can you think of a way to scope out the behavior only to cat gifs?

examples/customized-toolbar-example/package.json
luccas-harbour: sounds good, done

examples/customized-toolbar-example/package.json
luccas-harbour: gotcha, updated!

examples/customized-toolbar-example/src/main.js
luccas-harbour: That makes sense. I added a comment, let me know if that's good

examples/customized-toolbar-example/package.json
luccas-harbour: by the way, I fixed the bug where clicking on any image would turn it into a cat 😅

harbournick:
@mattConnHarbour Can you show @luccas-harbour how to add this example to our superdoc.dev examples gallery please?

PR #874, opened 2025-09-04
fix: inserting html with heading tags does not render as expected (HAR-10430)

luccas-harbour:


packages/super-editor/src/core/commands/insertContentAt.js
```diff
@@ -60,7 +60,7 @@ export const insertContentAt =
 
       // If the original input is plainly textual, prefer insertText regardless of how parsing represents it.
       const forceTextInsert =
```
harbournick, line 62: while we are here, mind adding some unit tests for this function? Not directly related to the task, but worth it since we are making a change here (and it should be a fairly easy one to test). You can crate insertContentAt.test.js next to this file

luccas-harbour: done! that test file already existed actually but I added a new test simulating this

package.json
```diff
@@ -16,7 +16,8 @@
     "watch:super-editor": "npm run build:watch --workspace=packages/super-editor",
     "build:superdoc": "npm run build --workspace=packages/superdoc",
     "build:super-editor": "npm run build --workspace=packages/super-editor",
-    "build": "npm run build:super-editor && npm run build:superdoc",
+    "build:ooxml-inspector": "npm run build --workspace=packages/ooxml-inspector",
+    "build": "npm run build:ooxml-inspector && npm run build:super-editor && npm run build:superdoc",
```
harbournick, line 20: nice catch

harbournick (approved):
LGTM

PR #880, opened 2025-09-05
fix: add processing for line-height defined in px

VladaHarbour:
Hi @harbournick. Not sure if I can add some tests for it because most probably this content was either copied and pasted from somewhere or added via API. It's not possible to have px for line height when importing fro docx


harbournick (approved):
LGTM

PR #881, opened 2025-09-05
feat: add document.xml.rels validator

falsaadehh:


chittolinag (changes requested):
Hi @falsaadehh ! Our automated test suite caught an issue with a document. It seems like MS Word is having trouble to load a document after it is exported in SuperDoc. 

I will send you the document via DM. Can you please double check? 

Screenshot 2025-09-05 at 16 43 44

harbournick:
Adding don't merge flag pending automated testing fixes

falsaadehh:
@harbournick should be good now, but needs your review whenver you have a moment please

packages/super-editor/src/core/super-validator/validators/xml/relationships/relationships-validator.js
```diff
@@ -0,0 +1,330 @@
...
+ * valid relationship entries that reference existing files and follow Word's conventions.
+ *
+ * @typedef {import('../../../types.js').Editor} Editor
+ * @typedef {import('../../../types.js').ValidatorLogger} ValidatorLogger
+ * @param {Object} params - Validator parameters
+ * @param {Editor} params.editor - The editor instance containing the document
+ * @param {ValidatorLogger} params.logger - Logger for validation messages
+ * @returns {Function} Validator function that returns {results: string[], modified: boolean}
+ */
+export function createRelationshipsValidator({ editor, logger }) {
```
artem-harbour, line 14: Why is a factory function used here to create a validator function? Not sure I see why this was done.

packages/super-editor/src/core/super-validator/validators/xml/relationships/relationships-validator.js
```diff
@@ -0,0 +1,330 @@
...
+    }
+    const candidateKeys = [
+      'word/_rels/document.xml.rels',
+      'word/document.xml.rels',
+      '_rels/document.xml.rels',
+      'document.xml.rels',
+    ];
+
+    // Find the first existing relationships file
+    let relsKey = candidateKeys.find((k) => convertedXml?.[k]?.elements);
```
artem-harbour: Other rels in the array will not be processed? Is this expected?

packages/super-editor/src/core/super-validator/validators/xml/relationships/relationships-validator.js
```diff
@@ -0,0 +1,330 @@
...
+    root.attributes = root.attributes || {};
+    if (root.attributes.xmlns !== RELS_NS) {
+      root.attributes.xmlns = RELS_NS;
+      results.push(`Set relationships xmlns to ${RELS_NS}`);
+      modified = true;
+    }
+
+    // Clean up the root element's children to ensure only valid Relationship elements
+    const nextChildren = [];
+    for (const child of Array.isArray(root.elements) ? root.elements : []) {
```
artem-harbour: `Array.isArray(root.elements) ? root.elements : []` - for this it is better to create a new variable and take it out of the loop.

packages/super-editor/src/core/super-validator/validators/xml/relationships/relationships-validator.js
artem-harbour: I also think that this validation function is quite large and responsible for many things. It would be better to have many small functions, and this one composes them.

I also see that these validators do not just validate data, but also change and mutate it. Is this how it is expected to work? Because in general these are two completely different responsibilities (validation and processing). 

сс @harbournick

packages/super-editor/src/core/super-validator/validators/xml/relationships/relationships-validator.js
```diff
@@ -0,0 +1,330 @@
...
+}
+
+/**
+ * Recursively processes document.xml to find and fix missing relationship references.
+ *
+ * @param {any} node - XML node to process (can be element, text, etc.)
+ * @param {Set<string>} usedIds - Set of valid relationship IDs from the relationships file
+ * @param {string[]} fixed - Array to collect descriptions of fixes made
+ */
+function processDocumentForMissingRefs(node, usedIds, fixed) {
```
artem-harbour, line 416: I think this function needs to be reworked.

artem-harbour (changes requested):
Added my comments for consideration. @harbournick - could you check that too?

packages/super-editor/src/core/super-validator/validators/xml/relationships/relationships-validator.js
falsaadehh: Working on it thanks

falsaadehh:
> Added my comments for consideration. @harbournick - could you check that too?

Ready for a second look

packages/super-editor/src/core/super-validator/validators/xml/relationships/relationships-validator.js
harbournick: Hey! you're right. I debated not calling this a "validator" because it is a validator-fixer, but I just didn't find another punchy name to call the whole thing. Open to suggestions! but the main goal of "SuperValidator" overall is to actually make changes where it finds invalid XML (or document schema depending on the validator). The good news is that its an internal tool so we can rename it to something more appropriate whenever. Open to suggestions! could call it super-editor-normalizer or super-editor-schema-cleaner

harbournick: re: factory function: I originally designed it this way for dependency injection and trying to make testing easier, but its not strictly necessary.

harbournick (approved):
Code LGTM. @falsaadehh this is nice! I know its a complex validator-fixer logic here, and there are others we'll want to do like this that have to do with ensuring "orphan" rels get removed (if not used anywhere in any xml)

one note: I am generally hoping to keep the super-validator sub project 100% ts checked so can add '// @ts-check at the top of the file (and then double check it still builds with npm run build). I also notice there's a lint warning. Neither is a blocker so merging this now but just a note for the future. 

Ideally all functions should have JSDoc signatures, too

PR #890, opened 2025-09-06
fix(table): add support for table row w:cantSplit

palmer-cl:
- If a row would overlap a page cutoff, render the row on the next line


harbournick (approved):
LGTM

PR #901, opened 2025-09-08
feat: remove the heading node from superdoc

luccas-harbour:
- Modifies how headings are parsed so that they are represented as styled paragraphs
- Turns the heading node into an extension
- Adds tests to commands in the linked-styles extension
- Adds tests to commands in the heading extension

### DIFF
--- a/packages/super-editor/src/extensions/heading/heading.js
+++ b/packages/super-editor/src/extensions/heading/heading.js
@@ -1,5 +1,5 @@
 // @ts-check
-import { Node, Attribute } from '@core/index.js';
+import { Extension } from '@core/index.js';
 
 /**
  * Heading attributes
@@ -18,14 +18,8 @@ import { Node, Attribute } from '@core/index.js';
  * @shortcut Mod-Alt-5 | toggleHeading | Toggle heading level 5
  * @shortcut Mod-Alt-6 | toggleHeading | Toggle heading level 6
  */
-export const Heading = Node.create({
+export const Heading = Extension.create({
   name: 'heading',
-
-  group: 'block',
-
-  content: 'inline*',
-
-  defining: true,
 
   addOptions() {
     return {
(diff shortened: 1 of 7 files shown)


packages/super-editor/src/extensions/heading/heading.js
```diff
@@ -1,5 +1,5 @@
 // @ts-check
-import { Node, Attribute } from '@core/index.js';
+import { Extension, Attribute } from '@core/index.js';
```
harbournick: Just to keep our lint happy, let's remove Attribute if its unused now

packages/super-editor/src/extensions/linked-styles/linked-styles.js
```diff
@@ -33,13 +88,22 @@ export const LinkedStyles = Extension.create({
...
       },
 
+      /**
+       * Get a linked style by its ID
+       * @param {string} styleId The style ID (e.g., 'Heading1')
+       * @returns {object|null} The linked style object or null if not found
+       */
       getStyleById: (styleId) => {
-        const styles = this.getStyles();
+        const styles = this.editor.helpers[this.name].getStyles();
```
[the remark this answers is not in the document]
harbournick: This doesn't really seem necessary, but mind taking a quick look to double check if you agree/disagree with the bot's assessment here?

harbournick:
@artem-harbour please review this one as well - thank you!

packages/super-editor/src/extensions/heading/heading.js
```diff
@@ -18,15 +18,9 @@ import { Node, Attribute } from '@core/index.js';
  * @shortcut Mod-Alt-5 | toggleHeading | Toggle heading level 5
  * @shortcut Mod-Alt-6 | toggleHeading | Toggle heading level 6
  */
-export const Heading = Node.create({
+export const Heading = Extension.create({
```
artem-harbour, line 21: I am not sure I understand the decision behind this change, so I will block this PR for now. At the very least it will be a breaking change.

packages/super-editor/src/extensions/heading/heading.js
```diff
@@ -18,15 +18,9 @@ import { Node, Attribute } from '@core/index.js';
...
   name: 'heading',
 
-  group: 'block',
-
-  content: 'inline*',
-
-  defining: true,
-
   addOptions() {
```
artem-harbour, line 24: `htmlAttributes` in options can be deleted with this change.

packages/super-editor/src/extensions/linked-styles/linked-styles.js
```diff
@@ -13,11 +14,65 @@ export const LinkedStyles = Extension.create({
...
+        (style, nodeType = null) =>
+        (params) => {
+          const { tr } = params;
+          if (tr.selection.empty) {
+            console.debug('[toggleLinkedStyle] empty selection, nothing to do...');
+            return false;
+          }
+          let node = tr.doc.nodeAt(tr.selection.$from.pos);
+
+          if (nodeType && node.type.name !== nodeType) {
```
artem-harbour: `node` - can be `null` here

packages/super-editor/src/extensions/linked-styles/linked-styles.js
```diff
@@ -13,11 +14,65 @@ export const LinkedStyles = Extension.create({
...
+       * @param {string|null} nodeType The node type to restrict the toggle to (e.g., 'paragraph'). If null,
+       * the style can be toggled on any node type.
+       * @returns {boolean} Whether the style was correctly applied/removed
+       */
+      toggleLinkedStyle:
+        (style, nodeType = null) =>
+        (params) => {
+          const { tr } = params;
+          if (tr.selection.empty) {
+            console.debug('[toggleLinkedStyle] empty selection, nothing to do...');
```
artem-harbour: Let's remove all logs from the command.

packages/super-editor/src/extensions/linked-styles/linked-styles.js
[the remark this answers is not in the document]
luccas-harbour: Hmm I think it's actually ok to reference other helper methods so that we don't need to repeat ourselves

packages/super-editor/src/extensions/heading/heading.js
luccas-harbour: Hey Artem! The idea behind this was to unify the behavior around headings.

Previously if you wrote some text into the editor and then via the UI you clicked to style it as a heading, the document would contain a paragraph node that had a `styleId` attribute equal to `Heading1`. At the same time, if you programmatically inserted a title using `editor.commands.insertContent("<h1>My Title</h1>")` it would instead create a `heading` node without a style. After exporting both documents to MS Word, they had different spacing as the nodes were not behaving the exact same way.

We could have either modified the existing `heading` node to behave like a styled paragraph (with the correct spacing, etc.) or just unify them so we don't have to maintain them separately and still get the same behavior whether you add a heading via the UI or programmatically.

I hope this clarifies it!

packages/super-editor/src/extensions/heading/heading.js
luccas-harbour: Done!

packages/super-editor/src/extensions/heading/heading.js
luccas-harbour: Sounds good, deleted

packages/super-editor/src/extensions/linked-styles/linked-styles.js
luccas-harbour: Nice catch! Fixed

packages/super-editor/src/extensions/linked-styles/linked-styles.js
luccas-harbour: Ok, removed

artem-harbour (approved):
LGTM

PR #913, opened 2025-09-09
feat: transform lists in copied content to match GoogleDocs/Word standard

VladaHarbour:


harbournick (changes requested):
Looks good! Seems to work very nicely.
the one note is that we're now putting test files next to the files they test. Maybe mode transformCopiedLists.test.js to transform-copied-lists.test.js (in the same html folder)

Also double check that all functions in here have at least some tests!

PR #942, opened 2025-09-11
feat: upgrade w:hyperlink import/export to v3 (HAR-10497)

luccas-harbour:
- Implements a NodeTranslator for the w:hyperlink element
- Adds support for all w:hyperlink attributes as defined in ECMA-376
- Fixes issue where a hyperlink using a different linked style would have its style overwritten with `Hyperlink` during import

### DIFF
--- a/packages/super-editor/src/core/super-converter/v3/handlers/w/hyperlink/hyperlink-translator.test.js
+++ b/packages/super-editor/src/core/super-converter/v3/handlers/w/hyperlink/hyperlink-translator.test.js
@@ -1,0 +1,244 @@
+// @ts-check
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { config, translator } from './hyperlink-translator.js';
+import { generateDocxRandomId } from '@helpers/generateDocxRandomId.js';
+import { exportSchemaToJson } from '@core/super-converter/exporter';
+
+vi.mock('@helpers/generateDocxRandomId.js', () => ({
+  generateDocxRandomId: vi.fn(),
+}));
+
+vi.mock('@core/super-converter/exporter', () => ({
+  exportSchemaToJson: vi.fn(),
+}));
+
+describe('w:hyperlink translator', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    vi.mocked(exportSchemaToJson).mockImplementation((p) => ({
+      name: 'w:r',
+      elements: [{ name: 'w:t', elements: [{ type: 'text', text: p.node.text || 'link text' }] }],
+    }));
+  });
+
+  describe('attribute handlers', () => {
+    const findAttr = (sdName) => config.attributes.find((a) => a.sdName === sdName);
+
+    it('handles w:anchor -> anchor', () => {
+      const handler = findAttr('anchor');
+      expect(handler.encode({ 'w:anchor': 'foo' })).toBe('foo');
+      expect(handler.decode({ anchor: 'foo' })).toBe('foo');
+    });
+
+    it('handles w:docLocation -> docLocation', () => {
+      const handler = findAttr('docLocation');
+      expect(handler.encode({ 'w:docLocation': 'bar' })).toBe('bar');
+      expect(handler.decode({ docLocation: 'bar' })).toBe('bar');
+    });
+
+    it('handles w:history -> history', () => {
+      const handler = findAttr('history');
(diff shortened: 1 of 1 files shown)


packages/super-editor/src/core/super-converter/v3/handlers/w/hyperlink/hyperlink-translator.js
```diff
@@ -0,0 +1,180 @@
...
+});
+
+/**
+ * The attributes that can be mapped between OOXML and SuperDoc.
+ * Note: These are specifically OOXML valid attributes for a given node.
+ * @type {import('@translator').AttrConfig[]}
+ * @see {@link https://ecma-international.org/publications-and-standards/standards/ecma-376/} "Fundamentals And Markup Language Reference", page 1276
+ */
+const validXmlAttributes = [
+  _createAttributeHandler('w:anchor', 'anchor'),
```
artem-harbour, line 32: It seems the approach to handling attributes has changed a bit in the latest examples, please check it out.

packages/super-editor/src/core/super-converter/v3/node-translator/node-translator.js
```diff
@@ -170,7 +170,7 @@ export class NodeTranslator {
    */
   decode(params) {
     const decodedAttrs = this.decodeAttributes(params);
-    return this.decodeFn ? this.decodeFn(params, decodedAttrs) : undefined;
+    return this.decodeFn ? this.decodeFn.call(this, params, decodedAttrs) : undefined;
```
artem-harbour, line 173: Good point.

packages/super-editor/src/core/super-converter/v3/handlers/w/hyperlink/hyperlink-translator.js
luccas-harbour: Hmm I see that for other elements each attribute handler is defined in its own separate file under an `attributes` directory. 
Most of these attribute encode/decode functions are simply returning a specific key from the attributes object, without any other logic, so it seemed simpler to just keep everything in the same file (so that a glance at that file gives you all the info about how the import/export is handled) but I am fine with changing this if the goal is to have all of them following the same standard.
What do you think @harbournick?

artem-harbour: +1

I agree that multiple files for attrs complicate things.

packages/super-editor/src/core/super-converter/v3/node-translator/node-translator.js
artem-harbour: Just a note: this will work only for regular functions/methods (not arrow functions).

luccas-harbour: Yes, good point. I had to switch to a regular function for it to work

packages/super-editor/package.json
```diff
@@ -43,6 +43,7 @@
     "types:check": "tsc --noEmit",
     "types:build": "tsc -p tsconfig.build.json",
     "test": "vitest",
+    "test:debug": "vitest --inspect-brk --pool threads --poolOptions.threads.singleThread",
```
harbournick, line 46: nice! TIL

packages/super-editor/src/core/super-converter/v3/node-translator/node-translator.js
harbournick, line 173: good call!

packages/super-editor/src/core/super-converter/v3/handlers/utils.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Generates a handler entity for a given node translator.
+ * @param {string} handlerName - The name of the handler.
+ * @param {import('../node-translator/').NodeTranslator} translator - The node translator object.
+ * @returns { import("../../v2/importer/docxImporter").NodeHandlerEntry } The handler entity with the specified name.
+ */
+export const generateV2HandlerEntity = (handlerName, translator) => ({
```
harbournick, line 7: YES! Great catch here.  I started pondering this this weekend. Love this!

harbournick (approved):
LGTM - really great contributions to the import/export v3 upgrades here - above and beyond! Much appreciated 🎉

PR #944, opened 2025-09-11
feat: implement unified content processing for insertContent command

caio-pizzol:
- Added a new `processContent` function to handle various content types (HTML, Markdown, Text, Schema) in a unified manner.
- Updated `insertContent` command to utilize the new processor, ensuring backward compatibility.
- Introduced content injection functionality in Developer Playground for testing different content types.
- Added helper functions for processing lists and stripping HTML styles to maintain document integrity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,4 +40,4 @@ test-results
 *.tgz
 
 !examples/chrome-extension/dist
-dev/
+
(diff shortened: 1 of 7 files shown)


caio-pizzol:
> Hey @caio-pizzol ! you might have noticed already, but the automated test suite found some regressions. Can you please double check? The failing document was recently added by @luccas-harbour. Maybe confirm with him/Missy what's the expected output there. It is possible that your changes are actually an improvement instead of a bug, and then we need to update the reference screenshots.
> 
> Thanks!

this is exactly what we're changing in the PR. HTML import behavior - we will need to update the base screenshot @gabriel.
image

PR #953, opened 2025-09-15
feat: add v3 translator for w:tr and w:trPr elements (HAR-10480)

luccas-harbour:
- Implements translators for `w:tr` and `w:trPr` elements covering all properties from ECMA 376


packages/super-editor/src/extensions/table-row/table-row.js
```diff
@@ -25,31 +68,60 @@ export const TableRow = Node.create({
     return {
       /**
        * @category Attribute
-       * @param {number} [rowHeight] - Fixed row height in pixels
+       * @param {TableRowProperties} [tableRowProperties] - Properties for the table row.
+       * @see {@link https://ecma-international.org/publications-and-standards/standards/ecma-376/} "Fundamentals And Markup Language Reference", page 377-482
        */
-      rowHeight: {
```
artem-harbour: I would be very careful with such changes. And while this can be handled in the same way in another attribute `tableRowProperties`, it changes the schema. And if you already have a saved document in json, then most likely the subsequent loading and display of the table row will be incorrect.

packages/super-editor/src/extensions/pagination/pagination.js
```diff
@@ -441,7 +441,7 @@ function generateInternalPageBreaks(doc, view, editor, sectionData) {
         : coords.bottom > pageHeightThreshold;
 
     // handle the w:cantSplit attribute for table rows
-    if (currentNode.type.name === 'tableRow' && currentNode.attrs?.cantSplit) {
```
artem-harbour: Same. This looks like a breaking change.

artem-harbour (commented):
Added a couple of comments. @harbournick please review this as well.

packages/super-editor/src/extensions/table-row/table-row.js
luccas-harbour: I see. Yes, this makes sense.

It would be nice if I could implement a getter/setter that would reference the new attribute but I am not sure if that's possible

luccas-harbour: I ended up reverting this change @artem-harbour. I modified the translator so that the `rowHeight` and `cantSplit` properties are pulled one level up so that we don't cause any breaking changes.

Also, you'll notice that with these latest changes, the number of files in this PR increased quite a bit. I didn't really change anything, I just moved all the translators that were being defined inside the `trPr-translator.js` file to separate files. 

This was done after talking with Nick as we noticed that some of the elements can be children of multiple elements (for example `w:tblCellSpacing` can be a child of `w:trPr` but also `w:tblPr`) and so defining them globally makes sharing the logic easier.

packages/super-editor/src/extensions/pagination/pagination.js
luccas-harbour: This was also reverted so as not to cause breaking changes.

packages/super-editor/src/extensions/table-row/table-row.js
artem-harbour: Thanks! I think the "schema changes" is something we need to discuss and how to deal with it.

packages/super-editor/src/core/super-converter/v3/handlers/utils.js
```diff
@@ -19,3 +19,154 @@ export const generateV2HandlerEntity = (handlerName, translator) => ({
...
+ * @param {object|Array} [attributes] The attributes object or array to which encoded properties will be added.
+ */
+export function encodeProperties(node, translatorsByXmlName, attributes) {
+  node.elements.forEach((el) => {
+    const translator = translatorsByXmlName[el.name];
+    if (translator) {
+      const encodedAttr = translator.encode({ ...el.attributes, nodes: [el] });
+      if (encodedAttr !== undefined && encodedAttr !== null) {
+        if (Array.isArray(attributes)) {
+          attributes.push({ [translator.sdNodeOrKeyName]: encodedAttr });
```
artem-harbour, line 113: Can we avoid side effects and convert to pure function? For example, return new attributes or a node with updated attributes?

packages/super-editor/src/core/super-converter/v3/handlers/utils.js
```diff
@@ -19,3 +19,154 @@ export const generateV2HandlerEntity = (handlerName, translator) => ({
...
+ * @param {object} [properties] The properties object containing attributes to be decoded.
+ * @param {Array} [elements] The array to which decoded elements will be added.
+ */
+export function decodeProperties(translatorsBySdName, properties, elements) {
+  Object.entries(properties).forEach(([key, value]) => {
+    const translator = translatorsBySdName[key];
+    if (translator) {
+      const attributes = translator.decode({ [key]: value, node: { attrs: properties[key] } });
+      if (attributes !== undefined && attributes !== null) {
+        elements.push({ name: translator.xmlName, attributes });
```
artem-harbour, line 136: I would recommend to avoid side effects and not mutate args, but return the result.

packages/super-editor/src/core/super-converter/v3/handlers/utils.js
```diff
@@ -19,3 +19,154 @@ export const generateV2HandlerEntity = (handlerName, translator) => ({
...
+export const booleanToString = (value) => (value ? '1' : '0');
+
+/**
+ * Parses a value to an integer.
+ * Returns undefined if the value is undefined, null, or cannot be parsed to a valid integer.
+ * @param {any} value The value to parse.
+ * @returns {number|undefined} The parsed integer or undefined.
+ */
+export const parseInteger = (value) => {
+  if (value === undefined || value === null) return undefined;
```
artem-harbour: Such checks can be simplified:
`value != undefined` or `value != null` or `value == null` - it works great in this particular case.

packages/super-editor/src/extensions/table-row/table-row.js
```diff
@@ -30,7 +71,7 @@ export const TableRow = Node.create({
       rowHeight: {
         renderDOM({ rowHeight }) {
           if (!rowHeight) return {};
-          const style = `height: ${rowHeight}px`;
```
artem-harbour: Most likely this is also a breaking change. I can assume that previously the data in the attribute was stored in px.

packages/super-editor/src/core/super-converter/v3/handlers/w/tr/tr-translator.js
```diff
@@ -0,0 +1,138 @@
...
+
+  let tableRowProperties = {};
+  const tPr = row.elements.find((el) => el.name === 'w:trPr');
+  if (tPr) {
+    ({ attributes: tableRowProperties } = trPrTranslator.encode({
+      ...params,
+      nodes: [tPr],
+    }));
+  }
+  encodedAttrs['tableRowProperties'] = tableRowProperties;
```
artem-harbour: Shouldn't the attributes here be frozen to avoid mutation? fyi @harbournick

artem-harbour:
Hey @luccas-harbour, I have added a few more comments for consideration, but I will let @harbournick have the final say.

Note: I also see that this PR has a lot of translators at the top level, which on the one hand makes it more granular, but on the other hand it seems to me that all related functionality should be in one place.

packages/super-editor/src/core/super-converter/v3/handlers/utils.js
luccas-harbour: You're right, that's better. Fixed!

packages/super-editor/src/core/super-converter/v3/handlers/utils.js
luccas-harbour: Fixed!

packages/super-editor/src/core/super-converter/v3/handlers/utils.js
luccas-harbour: Nice! Fixed

packages/super-editor/src/extensions/table-row/table-row.js
luccas-harbour: Yes, you are correct.

I did this because when we convert to pixels and then back, we lose precision and when the document is exported, the value is not the same as it was originally. If you import/export the same document a couple of times, the height keeps changing each time.

I guess I can store the raw value inside the `tableProperties` key and keep the pixel value in the `rowHeight` attribute. I'll do that.

packages/super-editor/src/core/super-converter/v3/handlers/w/tr/tr-translator.js
luccas-harbour: Fixed!

luccas-harbour:
> Hey @luccas-harbour, I have added a few more comments for consideration, but I will let @harbournick have the final say.
> 
> Note: I also see that this PR has a lot of translators at the top level, which on the one hand makes it more granular, but on the other hand it seems to me that all related functionality should be in one place.

Hey @artem-harbour!
Yes, I agree that this adds a lot of translators to the top level. The main thing is that these elements may show up as children of other elements as well, so having them at the top level could make code reuse easier. But I can see that as we work on more nodes, the number of directories in the `packages/super-editor/src/core/super-converter/v3/handlers/w/` directory will grow quite a bit.

luccas-harbour:
I had to make a few more changes as I noticed that I wasn't passing the parameters to the decode/encode functions in the right format from my utility functions.

packages/super-editor/src/core/super-converter/v3/handlers/utils.js
```diff
@@ -19,3 +19,162 @@ export const generateV2HandlerEntity = (handlerName, translator) => ({
...
+/**
+ * Helper to create simple property handlers with one-to-one mapping for properties with a single attribute (eg: 'w:val')
+ * @param {string} xmlName The XML attribute name (with namespace).
+ * @param {string|null} sdName The SuperDoc attribute name (without namespace). If null, it will be derived from xmlName.
+ * @param {string} [attrName='w:val'] The specific attribute name to map to/from. Default is 'w:val'.
+ * @param {function} [transformEncode=(v) => v] Optional transformation function to apply during encoding.
+ * @param {function} [transformDecode=(v) => v] Optional transformation function to apply during decoding.
+ * @returns {import('@translator').NodeTranslatorConfig} The attribute handler config with xmlName, sdName, encode, and decode functions.
+ */
+export function createSingleAttrPropertyHandler(
```
harbournick, line 32: this is cool! nice idea!

PR #956, opened 2025-09-15
fix: move undo/redo from collaboration to history

VladaHarbour:


packages/super-editor/package.json
```diff
@@ -76,8 +76,8 @@
     "naive-ui": "^2.38.2",
     "tippy.js": "^6.3.7",
     "vite-plugin-node-polyfills": "^0.23.0",
-    "y-prosemirror": "^1.2.12",
-    "yjs": "13.6.19"
+    "y-prosemirror": "latest",
```
harbournick: I think we should just pin to ^1.3.7 in case of breaking changes

harbournick (changes requested):
just a comment on "latest" but otherwise looks good.
maybe a good opportunity to add unit tests to history commands?

VladaHarbour:
@harbournick just wondering what exactly has to be tested here since we call a library command. Maybe it's better to test this with end-to-end tests?

PR #961, opened 2025-09-15
import/export/display of transformed images

johanneswilm:
Implemented features:
* import of transformed images (flip h, flip v, rotated) with `wp:effectExtent` 
* export of transformed images with `wp:effectExtent`
* display of transformed images using custom formula to calculate needed extra space
* tests

**TODO:**
1. Custom formula and values from `wp:effectExtent` should be the same - but this is only the case sometimes. Custom formula gives values as they are needed for display in browser, while values from `wp:effectExtent` are only used for export. Once formulas fully align, values for export should be calculated based on formula.
2. Custom command to change the rotation and flip state of individual images (will need to recalculate values for `wp:effectExtent` based on custom formula - issue 1).
3. Code works with provided test file, but other test files produced with LibreOffice currently don't work in import.

### DIFF
--- a/packages/super-editor/src/core/super-converter/helpers.js
+++ b/packages/super-editor/src/core/super-converter/helpers.js
@@ -84,6 +84,16 @@ function ptToTwips(pt) {
   return pt * 20;
 }
 
+function rotToDegrees(rot) {
+  if (rot == null) return;
+  return rot / 60000;
+}
+
+function degreesToRot(degrees) {
+  if (degrees == null) return;
+  return degrees * 60000;
+}
+
 /**
  * Get the export value for text indent
  * @param {string|number} indent - The text indent value to export
@@ -236,6 +246,8 @@ export {
   halfPointToPoints,
   eigthPointsToPixels,
   pixelsToEightPoints,
+  rotToDegrees,
+  degreesToRot,
   getArrayBufferFromUrl,
   getContentTypesFromXml,
   getHexColorFromDocxSystem,
--- a/packages/super-editor/src/extensions/image/image.js
+++ b/packages/super-editor/src/extensions/image/image.js
@@ -1,6 +1,7 @@
 import { Attribute, Node } from '@core/index.js';
 import { ImagePlaceholderPlugin } from './imageHelpers/imagePlaceholderPlugin.js';
 import { ImagePositionPlugin } from './imageHelpers/imagePositionPlugin.js';
+import { getRotationMargins } from './imageHelpers/rotation.js';
 
 /**
  * @module Image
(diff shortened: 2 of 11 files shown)


johanneswilm:
I am currently looking at other open source software and how they deal with calculating the values of `wp:effectExtent` in order to better understand how the fomula may need to be modified.

packages/super-editor/src/extensions/image/image.js
```diff
@@ -172,8 +208,26 @@ export const Image = Node.create({
        */
       padding: {
         default: {},
-        renderDOM: ({ padding, marginOffset }) => {
-          const { left = 0, top = 0, bottom = 0, right = 0 } = padding ?? {};
+        renderDOM: ({ size, padding, marginOffset, transformData }) => {
+          let { left = 0, top = 0, bottom = 0, right = 0 } = padding ?? {};
+          // TODO: The wp:effectExtent (transformData.sizeExtension) sometimes
```
VladaHarbour, line 213: @johanneswilm what I found is that wp:effectExtent only stores margins needed for effect but doesn't store values set by User. wp:effectExtent is also not rotation padding. It only accounts for visual effects (shadow, glow, outline, etc.)

VladaHarbour (approved):

packages/super-editor/src/extensions/image/image.js
johanneswilm: @VladaHarbour The wp:effectExtent in the example files extends the images so that they don't overlap even though they are rotated. Depending on the wordprocessor, they seem to do this as well - change the effectExtent value when the image is rotated so that they don't overlap. 

We also need this when working in HTML - the images need to be extended when rotated so that they don't overlap with other content - unless that is the effect one wants. My formula calculates how much that is and the formula works for Superdoc. 

The strange thing is that the values stored in the DOCX files are the same for some in images as stored in the wp:effectExtent - but totally different for others. After reading some LibreOffice source files, I understand that H and W are switched on some rotated images if the file was created with a Word version after 2007. I continue to investigate this. Hopefully, I can make my formula and the wp:effectExtent values converge so that we can have a function to rotate images and update the values needed for export.

packages/super-editor/src/core/super-converter/helpers.js
```diff
@@ -84,6 +84,16 @@ function ptToTwips(pt) {
   return pt * 20;
 }
 
+function rotToDegrees(rot) {
+  if (rot == null) return;
```
artem-harbour, line 88: I think that in this case, need to either return `0` or explicitly `null/undefined` (in this case, check from outside the function too).

packages/super-editor/src/core/super-converter/helpers.js
```diff
@@ -84,6 +84,16 @@ function ptToTwips(pt) {
...
+  return rot / 60000;
+}
+
+function degreesToRot(degrees) {
+  if (degrees == null) return;
```
artem-harbour, line 93: I think that in this case, need to either return 0 or explicitly null/undefined (in this case, check from outside the function too).

packages/super-editor/src/core/super-converter/helpers.js
johanneswilm: @artem-harbour That's also how I would have written it, but if you look at the other conversion helper functions in that file, they all seem to handle it this way. In order to stay consistent, I applied the same logic here.

packages/super-editor/src/core/super-converter/helpers.js
johanneswilm: @artem-harbour See comment above.

harbournick:
Nice work! 
Resolved merged conflicts, visual tests etc in https://github.com/Harbour-Enterprises/SuperDoc/pull/1011
Merging via that PR, closing this one.

PR #963, opened 2025-09-16
fix: emit editor exception on import/export errors

VladaHarbour:
Hi @harbournick! Tested Superdoc's onException hook and it catches unhandled import exceptions. Added another event on document export error. 
We have some try/catch blocks in the import/export process. How do you think should I also emit exception event there if there is a handler for error?

### DIFF
--- a/packages/super-editor/src/components/SuperEditor.vue
+++ b/packages/super-editor/src/components/SuperEditor.vue
@@ -117,6 +117,9 @@ const loadNewFileData = async () => {
     return { content: docx, media, mediaFiles, fonts };
   } catch (err) {
     console.debug('Error loading new file data:', err);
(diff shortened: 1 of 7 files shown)


harbournick (changes requested):
Can we also handle any errors here: https://github.com/Harbour-Enterprises/SuperDoc/blob/01809a4001588fac09ef15c271b8a92432b3d08b/packages/super-editor/src/components/SuperEditor.vue#L116

In this case we should still be able to pass it up to SuperDoc level somehow, and make sure all editor errors end up here: https://github.com/Harbour-Enterprises/SuperDoc/blob/01809a4001588fac09ef15c271b8a92432b3d08b/packages/superdoc/src/core/SuperDoc.js#L184

We might consider toolbar errors too. Anywhere else?

PR #967, opened 2025-09-16
feat: support Blob type in fileSource and document properties

caio-pizzol:
Updated the fileSource and initialData properties to accept both File and Blob types across components. Enhanced documentation to reflect these changes. Added a utility function to convert Blob to File and implemented tests to ensure proper handling of Blob objects in the SuperDoc store.


harbournick (approved):
LGTM

PR #974, opened 2025-09-17
fix: html import parse styleId attr for round-trip fidelity

caio-pizzol:
- Added `exportHTML` method in SuperdocDev.vue to allow users to export editor content as HTML.
- Updated `getHTML` method in Editor.js to include options for unflattening lists.
- Refactored attribute handling in htmlSanitizer.js and paragraph.js to support new `styleid` attribute and improve attribute preservation logic.

### DIFF
--- a/packages/super-editor/src/extensions/paragraph/paragraph.js
+++ b/packages/super-editor/src/extensions/paragraph/paragraph.js
@@ -189,12 +189,35 @@ export const Paragraph = OxmlNode.create({
       {
         tag: 'p',
         getAttrs: (node) => {
-          let extra = {};
+          const styleId = node.getAttribute('styleid') || null;
+
+          const extra = {};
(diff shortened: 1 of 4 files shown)


packages/super-editor/src/extensions/paragraph/paragraph.js
```diff
@@ -189,13 +189,36 @@ export const Paragraph = OxmlNode.create({
       {
         tag: 'p',
         getAttrs: (node) => {
-          let extra = {};
+          const styleId = node.getAttribute('styleid') || null;
+
+          const extra = {};
+          Array.from(node.attributes).forEach((attr) => {
```
harbournick: You can probably just do something like: 
const { styleid: styleId, ...extra } = node.attrs;

more succint / clear

harbournick (approved):
LGTM

PR #977, opened 2025-09-17
fix: enhance list migration and HTML sanitizer logic

caio-pizzol:
- Updated the Editor class to conditionally migrate lists based on additional options (markdown, html).
- Improved the htmlSanitizer to retain data-* attributes while stripping unsupported attributes.
- Modified bullet and ordered list extensions to handle specific list attributes during parsing.

### DIFF
--- a/packages/super-editor/src/extensions/bullet-list/bullet-list.js
+++ b/packages/super-editor/src/extensions/bullet-list/bullet-list.js
@@ -48,20 +48,7 @@ export const BulletList = Node.create({
   },
 
   parseDOM() {
-    return [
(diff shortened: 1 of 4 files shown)


packages/super-editor/src/extensions/bullet-list/bullet-list.js
```diff
@@ -48,7 +48,20 @@ export const BulletList = Node.create({
   },
 
   parseDOM() {
-    return [{ tag: 'ul' }];
+    return [
```
caio-pizzol: Not the correct approach, just a temp work around for now @harbournick

PR #987, opened 2025-09-19
feat: structured content node views

artem-harbour:


packages/super-editor/src/extensions/structured-content/StructuredContentViewBase.js
```diff
@@ -0,0 +1,235 @@
...
+
+  update(node, decorations, innerDecorations) {
+    if (node.type !== this.node.type) {
+      return false;
+    }
+
+    this.node = node;
+    this.decorations = decorations;
+    this.innerDecorations = innerDecorations;
+    this.updateHTMLAttributes();
```
harbournick, line 58: should we return true here?

packages/super-editor/src/extensions/structured-content/structured-content-block.js
```diff
@@ -1,40 +1,65 @@
...
+
       sdtPr: {
         rendered: false,
       },
     };
   },
 
   parseDOM() {
-    return [{ tag: `div.${this.options.structuredContentClass}` }];
+    return [{ tag: `span[data-structured-content-block]` }];
```
harbournick: should this remain 'div' not 'span'?

harbournick (changes requested):
Two minor changes - but looks great and seems to work well.
The prosemirror concern is reproducible but it doesn't seem like a blocker to me.

packages/super-editor/src/extensions/structured-content/StructuredContentViewBase.js
artem-harbour: Fixed.

packages/super-editor/src/extensions/structured-content/structured-content-block.js
artem-harbour: Fixed.

harbournick (approved):
LGTM

PR #989, opened 2025-09-19
refactor: enhance documentation and type definitions across extensions

caio-pizzol:


packages/super-editor/src/extensions/line-break/line-break.js
```diff
@@ -61,6 +67,23 @@ export const LineBreak = Node.create({
   },
 });
 
+/**
+ * Configuration options for HardBreak
```
caio-pizzol, line 71: I think right now we only document 1 extension per file.

Does it make sense? Or we should also allow to have multiple in 1 file? @harbournick

harbournick: there are places where multiple extensions may exist in a single file but we should fix that so they all have their own file. documenting 1 per file makes sense

PR #1020, opened 2025-09-24
refactor: v3 handler for mc:AlternateContent

VladaHarbour:


packages/super-editor/src/core/super-converter/v2/importer/bookmarkEndImporter.js
```diff
@@ -1,5 +1,5 @@
 // @ts-check
-import { translator as wBookmarkEndTranslator } from '../../v3/handlers/w/bookmark-end/index.js';
+import { translator as wBookmarkEndTranslator } from '@converter/v3/handlers/mc/altermateContent';
```
harbournick: This should be reverted I think?  should use the bookmark importer in this file. I think this is breaking tests too

harbournick (changes requested):
One quick comment

packages/super-editor/src/core/super-converter/v2/importer/bookmarkEndImporter.js
VladaHarbour: Yes, this was changed by mistake
already fixed

PR #1028, opened 2025-09-25
Image anchor wrapping options [SD-173]

johanneswilm:
Additions:
* Reading of image anchor wrapping settings
* Writing of image anchor wrapping settings
* Setting of wrapping options via command
* Switching between inline and anchor mode
* Tests
* Rendering of images


VladaHarbour:
Hi @johanneswilm! I'm not sure whether your PR is meant to fix anchor images position issue. But there is getImagePositionDecorations function in imagePositionPlugin.js which may overlap with your code. I set some margins there and could you please check if something about margins in getImagePositionDecorations is still relevant. Anchor images also require absolute positioning which I try to calculate in getImagePositionDecorations but it doesn't work well enough

johanneswilm:
@VladaHarbour This PR is related to [SD-173] - making all the various wrapping options available in SuperDoc. I put some screenshots into Linear on that issue there.  

I'll check the method you mention next.

VladaHarbour:
@johanneswilm getImagePositionDecorations to be honest might be outdated already especially if pay attention to the pagination changes that will be introduced by @harbournick soon

johanneswilm:
@VladaHarbour I am reading up on it and trying it out. I think in some cases we are addressing the same nodes. However, a lot of it is gone if I simply add a condition to the plugin:

```js
if (style.length || className.length) {
        decorations.push(Decoration.inline(pos, pos + node.nodeSize, { style, class: className }));
}
```

I get that this plugin has to do this as a decoration as the styling depends on whether or not there is a page break. If this is handled differently in the future, then maybe there is no need for making the two work well together? Is there a branch for that? Or has it been merged into `main`?

johanneswilm:
@VladaHarbour No, there is no break there as Tight and Through will be treated the same
way - in rendering as well as import and export. So the two cases have just one code block. The reason for this is that
they are quite similar, but that due to the limitations we have in support
of rendering around polygons in the DOM which don't permit us to let the
text flow continue inside of islands, they will look the exact same (until
we decide to provide to do the rendering of this in JavaScript). They also have the same attributes, which is why they can be treated the same when importing and exporting.

VladaHarbour:
@johanneswilm I don't think that new pagination is ready and probably this plugin can be removed for now. The reason it was made is that sometimes anchor images have defined vertical and horizontal positions with prop relativeFrom(margin, page, paragraph). If relativeFrom is set to page, for example, images have to appear out of the context and top/left properties have to be calculated based on the current page. Since we don't have page containers I used to look for the nearest pagination block.
This comment is just fyi since you've been working on anchor images too

VladaHarbour (approved):

VladaHarbour:
@johanneswilm could you please also check failing tests here. I suppose it's because of changed measurement helpers

johanneswilm:
@VladaHarbour Sure. Although actually I thought these tests were unrelated to this code as they seem to deal with list indentations, etc. . That's why I left them aside. But if they are impacted by this, I will of coruse figure out what is going on.

PR #1061, opened 2025-10-02
fix: exported annotations are suppressed in Google Docs

VladaHarbour:


packages/super-editor/src/core/super-converter/v3/handlers/w/hyperlink/hyperlink-translator.js
```diff
@@ -107,12 +107,6 @@ function decode(params) {
 
   const linkMark = node.marks.find((m) => m.type === 'link');
   const linkAttrs = this.decodeAttributes({ ...params, node: linkMark });
-  let { anchor, href: link } = linkMark.attrs;
```
VladaHarbour, line 110: Hi @luccas-harbour! Made some changes here since we used to add a new relationship on every export for existing hyperlinks. When we import document with hyperlinks we have all the relationship and there is no need to add it. Also params.relationships won't contain any items for imported links and is used to be used only for newly created ones.  When we create new hyperlink in SD relationship is added here:
https://github.com/Harbour-Enterprises/SuperDoc/blob/47471457a54193f6eea02e5bce20ebd03e5108b0/packages/super-editor/src/extensions/link/link.js#L218
Please let me know if it makes sense

luccas-harbour: Hi @VladaHarbour! Got it. Thank you for the explanation! This sounds good to me.

luccas-harbour: I guess the whole `_addNewLinkRelationship` function can be removed since it's not used anywhere else in the translator.

VladaHarbour: Make an update there to consider anchor links and add a relationship just in case node miss rId attribute for some reason

PR #1062, opened 2025-10-02
SD-284 - Handler for commentRangeStart/commentRangeEnd

chittolinag:


packages/super-editor/src/core/super-converter/v3/handlers/w/commentRange/comment-range-translator.js
```diff
@@ -0,0 +1,89 @@
...
+const SD_NODE_NAME = 'commentRange';
+
+/**
+ * Decode the commentRange(Start|End) node back into OOXML <w:commentRange(Start|End)>.
+ * @param {import('@translator').SCDecoderConfig} params
+ * @param {import('@translator').DecodedAttributes} [decodedAttrs] - The already decoded attributes
+ * @returns {import('@translator').SCDecoderResult}
+ */
+const decode = (params, decodedAttrs) => {
+  // TODO: Is exportedCommentDefs needed?
```
chittolinag: I'm not sure if we really need this - does anyone have more context about why this is needed? It seems like this variable is only used to return in case it's empty, but we don't do anything with it.

packages/super-editor/src/core/super-converter/v3/handlers/w/commentRange/comment-range-translator.js
```diff
@@ -0,0 +1,89 @@
...
+    commentSchema = [commentSchema, commentReference];
+  }
+
+  return commentSchema;
+};
+
+/**
+ * Generate a w:commentRangeStart or w:commentRangeEnd node
+ *
+ * @param {string} type Must be 'commandRangeStart' or 'commandRangeEnd'
```
luccas-harbour: I think there was a typo here `commandRangeStart` => `commentRangeStart`.

Just a tip, you can type this as:
```javascript
/**
* @param { 'commentRangeStart' | 'commentRangeEnd' } type
*/
```

chittolinag: ah yeah! great catch! thanks

PR #1075, opened 2025-10-03
fix: use SUPERDOC_PAT to trigger e2e tests

chittolinag:
It seems like we are not able to trigger E2E tests due to a recent change to the token here. I'm adding the same token we used before to fix it for now.


edoversb (approved):
LGTM

PR #1090, opened 2025-10-07
feat: add handler for w:tcpr (SD-76)

luccas-harbour:
- Adds a translator for `w:tcPR` and all its children
- Modifies handling of `w:tc` to make use of `w:tcPR` translator
- Adds a new `tableCellProperties` attribute to the `tableCell` node

### DIFF
--- a/packages/super-editor/src/extensions/table-cell/table-cell.js
+++ b/packages/super-editor/src/extensions/table-cell/table-cell.js
@@ -135,7 +135,8 @@ export const TableCell = Node.create({
 
       verticalAlign: {
         renderDOM({ verticalAlign }) {
-          const style = `vertical-align: ${verticalAlign ?? 'top'}`;
+          if (!verticalAlign) return {};
+          const style = `vertical-align: ${verticalAlign}`;
           return { style };
         },
       },
--- a/packages/super-editor/src/core/super-converter/helpers.js
+++ b/packages/super-editor/src/core/super-converter/helpers.js
@@ -74,9 +74,9 @@ function pixelsToHalfPoints(pixels) {
   return Math.round((pixels * 72) / 96);
 }
 
-function eigthPointsToPixels(eigthPoints) {
-  if (eigthPoints == null) return;
-  const points = parseFloat(eigthPoints) / 8;
+function eighthPointsToPixels(eighthPoints) {
+  if (eighthPoints == null) return;
+  const points = parseFloat(eighthPoints) / 8;
   const pixels = points * 1.3333;
   return pixels;
 }
@@ -368,7 +368,7 @@ export {
   pixelsToEmu,
   pixelsToHalfPoints,
   halfPointToPoints,
-  eigthPointsToPixels,
+  eighthPointsToPixels,
   pixelsToEightPoints,
   rotToDegrees,
   degreesToRot,
--- a/packages/super-editor/src/core/super-converter/v2/importer/tableImporter.js
+++ b/packages/super-editor/src/core/super-converter/v2/importer/tableImporter.js
@@ -1,4 +1,4 @@
-import { eigthPointsToPixels, halfPointToPoints } from '../../helpers.js';
+import { eighthPointsToPixels, halfPointToPoints } from '../../helpers.js';
 import { generateV2HandlerEntity } from '@converter/v3/handlers/utils.js';
 import { translator as tableTranslator } from '@converter/v3/handlers/w/tbl/tbl-translator.js';
 
@@ -104,7 +104,7 @@ function processTableBorders(borderElements) {
     const color = attributes['w:color'];
     const size = attributes['w:sz'];
     if (color && color !== 'auto') attrs['color'] = color.startsWith('#') ? color : `#${color}`;
-    if (size && size !== 'auto') attrs['size'] = eigthPointsToPixels(size);
+    if (size && size !== 'auto') attrs['size'] = eighthPointsToPixels(size);
 
     const rowBorderNames = ['insideH', 'insideV'];
     if (rowBorderNames.includes(borderName)) rowBorders[borderName] = attrs;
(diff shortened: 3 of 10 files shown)


packages/super-editor/src/core/super-converter/v3/handlers/utils.js
```diff
@@ -181,6 +209,54 @@ export function decodeProperties(translatorsBySdName, properties) {
...
+    decode: (params) => {
+      const currentValue = params.node.attrs?.[sdName];
+
+      // Process property translators
+      const elements = decodeProperties(propertyTranslatorsBySdName, currentValue);
+
+      const newNode = {
+        name: xmlName,
+        type: 'element',
+        attributes: {},
```
chittolinag, line 249: just checking, is this supposed to be empty? @luccas-harbour

packages/super-editor/src/core/super-converter/v3/handlers/w/bottom/bottom-translator.test.js
```diff
@@ -21,7 +21,7 @@ describe('w:bottom border translator', () => {
 
       expect(result).toEqual({
         val: 'single',
-        color: 'FF0000',
+        color: '#FF0000',
```
chittolinag, line 24: It seems like the `decode` function should translate it back to `FF0000`. Does the `decode` function takes care of this, or is it just passing-through as is? Is this expected? 

I see there are some other places where this is happening too.

packages/super-editor/src/core/super-converter/v3/handlers/utils.js
luccas-harbour: yes, that is correct... that's because this function is used only for tags that serve just as aggregators of multiple children but don't have any attributes of their own (you can see that no attributes were declared in [line 232](https://github.com/Harbour-Enterprises/SuperDoc/blob/59ff542025ae46f9ded99111b34297ea89111b5e/packages/super-editor/src/core/super-converter/v3/handlers/utils.js#L232) either).

this function is being used with these tags, and none of them have attributes: `w:trPr`, `w:tcPr`, `w:tblPr`, `w:tblBorders`, `w:tcMar`, `w:tblCellMar`, `w:tcBorders`

packages/super-editor/src/extensions/table-cell/table-cell.js
```diff
@@ -97,8 +135,7 @@ export const TableCell = Node.create({
 
       verticalAlign: {
         renderDOM({ verticalAlign }) {
-          if (!verticalAlign) return {};
-          const style = `vertical-align: ${verticalAlign}`;
+          const style = `vertical-align: ${verticalAlign ?? 'top'}`;
```
chittolinag: out of curiosity: why exactly this needs to be changed?

packages/super-editor/src/core/super-converter/v3/handlers/w/bottom/bottom-translator.test.js
luccas-harbour: Good catch! I actually had implemented removing the `#` if it exists but forgot to modify the tests. Fixed that now.

packages/super-editor/src/extensions/table-cell/table-cell.js
luccas-harbour: The old implementation would set vertical align to `top` during import if that field was not defined in the `docx`. This would cause the export to include that property even if the original document didn't have it. With this change, we only use a default at rendering time so that when exporting the node, it will be the same as in the docx.

packages/super-editor/src/core/super-converter/v3/handlers/utils.js
```diff
@@ -89,8 +91,34 @@ export function createBorderPropertyHandler(xmlName, sdName = null) {
...
+            return '#000000';
+          } else if (v) {
+            return `#${v}`;
+          } else {
+            return undefined;
+          }
+        },
+        (v) => {
+          if (v === '#000000') {
+            return 'auto';
```
harbournick: We may need to handle the special word 'auto' in some special way here. I think that if we convert all 000000 to 'auto' and back we may introduce unintended changes in documents. 'Auto' can actually show other colors depending on the theme file (possibly other rules?) so it _may not_ always be black. any ideas?

luccas-harbour: Gotcha. That makes sense. I will revert this change

luccas-harbour: Actually, do you mean this is a problem during encoding too? I actually did this to be compatible with what was being done [here when handling a table cell](https://github.com/Harbour-Enterprises/SuperDoc/blob/99d4ce77dbe699eeded73531957782fa05c9afc8/packages/super-editor/src/core/super-converter/v3/handlers/w/tc/helpers/legacy-handle-table-cell-node.js#L170-L170)

luccas-harbour: I was having a look at the ECMA and it sounds like when the color is set to `auto` it means it's up to the client application to determine which color should be used. This is what it says:

> Specifies the color for this border.
> This value can be defined as either:
> - A color value using the RGB color model whose red, green, and blue values are
written as numbers in the range 0 to 255, hex encoded, and concatenated. [Example: Full intensity red would be 255 red, 0 green, 0 blue, encoded to FF, 00, 00, and concatenated to FF0000. end example] . RGB colors are specified in the sRGB color space.
> - `auto` to **allow a consumer to automatically determine the border color in order to make the document's text readable**. [Example: A document with white text and a background color of auto might result in the use of a black background, in order to ensure legibility of the content. end example]

I suppose the only situation where we would want to use a color other than black would be if we implemented something like a dark mode? After we implement styles via CSS, we could use a CSS variable reference instead of `#000000` (something like `var(--border-auto-color)`)

luccas-harbour: This logic was removed to favor dealing with `auto` during rendering.

PR #1095, opened 2025-10-08
feat: fix type export on build

caio-pizzol:
- Introduced a new script `ensure-types.cjs` to verify and ensure the presence of the SuperDoc export in the generated type declarations.
- Updated `package.json` to include a `postbuild` script that runs the new type verification script after building the project.
- Adjusted exports in `src/index.js` and `src/core/index.js` to explicitly export `SuperDoc` instead of using wildcard exports.

---

> [!NOTE]
> Add a postbuild type verification script and switch to explicit `SuperDoc` exports to guarantee correct type declarations.
> 
> - **Build/Tooling**:
>   - Add `postbuild` script in `packages/superdoc/package.json` to run `scripts/ensure-types.cjs`.
>   - New `scripts/ensure-types.cjs` verifies `dist/index.d.ts` exists and includes an `SuperDoc` export.
> - **Exports**:
>   - Replace wildcard exports with named `SuperDoc` export in `packages/superdoc/src/core/index.js` and `packages/superdoc/src/index.js`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 439b690462c3c3b10b0015e1115648a2a9a2e9a2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/superdoc/scripts/ensure-types.cjs
+++ b/packages/superdoc/scripts/ensure-types.cjs
@@ -4,47 +4,19 @@ const fs = require('node:fs');
 const path = require('node:path');
 
 const distIndexPath = path.resolve(__dirname, '../dist/index.d.ts');
-const fallbackIndexPath = path.resolve(__dirname, '../types/index.d.ts');
 
-const hasSuperDocExport = (content) => /export\s+\{[^}]*\bSuperDoc\b[^}]*\}/m.test(content);
-
-const readFileIfExists = (filePath) => {
-  if (!fs.existsSync(filePath)) {
-    return null;
-  }
-
-  return fs.readFileSync(filePath, 'utf8');
-};
-
-const ensureDistFile = () => {
-  const fallbackContent = readFileIfExists(fallbackIndexPath);
-
-  if (!fallbackContent) {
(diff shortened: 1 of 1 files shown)


packages/superdoc/scripts/ensure-types.cjs
```diff
@@ -0,0 +1,50 @@
...
+if (!distContent) {
+  distContent = ensureDistFile();
+}
+
+if (!hasSuperDocExport(distContent)) {
+  console.warn('[ensure-types] SuperDoc export missing in generated declarations. Using fallback file.');
+  distContent = ensureDistFile();
+}
+
+if (!hasSuperDocExport(distContent)) {
```
chittolinag: nitpick: since the `if` condition is the same, we could add it in the same `if` above? 🤔

chittolinag: also, just out of curiosity, why exactly we need the fallback path? won't we always deploy whatever is in the `dist` folder?

caio-pizzol: agreed - let me simplify this and remove fallback

caio-pizzol:
cursor review

PR #1100, opened 2025-10-09
fix: allow both leading and trailing spaces

VladaHarbour:


packages/super-editor/src/core/helpers/htmlSanitizer.js
```diff
@@ -37,8 +37,12 @@ export function stripHtmlStyles(html) {
       const name = attr.name.toLowerCase();
 
       if (node.nodeName.toLowerCase() === 'span') {
-        // Preserve trailing spaces
-        node.innerHTML = node.innerHTML.replace(/(\S) (<\/span>)/g, '$1&nbsp;$2');
+        // Preserve leading and trailing spaces inside span tags
+        node.innerHTML = node.innerHTML.replace(/(<span[^>]*>)([\s\S]*?)(<\/span>)/g, (_, open, content, close) => {
```
harbournick: won't this ONLY check for nested spans since node is already span type? we might miss <span>  text    </span> possibly?

Let's also write some tests for this case and nested case to make sure this works as expected. If simpler for testing you could extract this piece into a helper too

VladaHarbour: Client who reported it has html structure with a lot of span wrappers inside paragraph. I changed this approach to process only those spans which has text inside

PR #1101, opened 2025-10-09
refactor: pict handler

artem-harbour:
Linear: SD-232


packages/super-editor/src/core/super-converter/v3/handlers/w/pict/pict-translator.js
```diff
@@ -0,0 +1,73 @@
...
+ * @returns {import('@translator').SCDecoderResult}
+ */
+function decode(params) {
+  const { node } = params;
+  if (!node || !node.type) {
+    return null;
+  }
+
+  const types = {
+    shapeContainer: () => translateShapeTextbox(params),
```
harbournick: are these swapped? looks like shapeContainer is pointing to translateShapeTextBox and shapeTextbox points to translateShapeContainer

artem-harbour: You are right, it's because of the rearrangement of properties. Fixed.

PR #1102, opened 2025-10-09
SD-374 - Implement "onFontsResolved" callback

chittolinag:
## Description

Implements a callback that is triggered when all fonts used in the document are resolved, with two parameters: 

- `documentFonts` - all fonts that are used in the document
- `unsupportedFonts` - all fonts that are used in the document but not supported by the host machine

## Solution

There is an API called [Local Font Access API](https://developer.mozilla.org/en-US/docs/Web/API/Local_Font_Access_API). This works well for some browsers, including Chrome and Edge. However, this is not supported across mobile devices, Firefox and Safari. To overcome that, I've implemented a fallback solution that will be used in such cases. I'll add details about the fallback solution below. But before I do that, here's the overall logic to resolve the unsupported fonts.

1. Check if Local Font Access is available
2. If not, run fallback solution
3. If it is, then
  3.1. Check if permission is already granted
  3.2. If not, ask for permission
      3.2.1. If permission is granted, use API
      3.2.2. If permission is not granted, use fallback solution
  3.3. If it is, use API

### Fallback solution

The fallback solution to determine whether a font can be rendered or not is somewhat hacky, but it seems to good enough for most cases:

1. Render some text using a fallback font in a Canvas, ie: `font: '72px sans-serif'`
5. Save its width and height
6. Render the same text now using the font we're testing for, but adding the same fallback, ie: `font: 72px "${font}", sans-serif`
7. Save its width and height
8. Compare the width and height. If they are different, it means the given font was indeed rendered. If they are the same, it means the browser had to use the fallback font because it could not render the given font.

> [!WARNING]  
> I did notice that this produced a couple false negatives on my tests in a Ubuntu docker image, but it worked for 100% of cases on MacOS fonts. 
> 
> However, I cannot ensure that they won't fail in other cases, so we should let our customers know to use this with caution *if* the Local Font Access API is not available.

## Caveats

Before we request for permission to Local Font Access, there's gotta be some user interaction on the page. For example, if you open a page with SuperDoc and the first thing it does is to ask for permission, it will fail. You can reproduce this on your local environment by running `npm run dev`. We won't be able to ask for permissions because by the first time you open the dev playground, SuperDoc is there already and we would have not interacted with the page at all. In such case, we will use the fallback solution.

However, if you're already using the page and have clicked some buttons, for example, it will simply just ask for permission when you try to load the file, or when you click a button that loads SuperDoc with a given document.

You can find more details here: https://developer.mozilla.org/en-US/docs/Web/Security/User_activation

## Local Font Access support

@luccas-harbour shared this with me (Thanks!). This gives us a good idea of how often we'll be able to use this API.

Screenshot 2025-10-14 at 11 32 35

Screenshot 2025-10-14 at 11 32 23

### DIFF
--- a/packages/super-editor/src/core/Editor.js
+++ b/packages/super-editor/src/core/Editor.js
@@ -183,6 +183,7 @@ export class Editor extends EventEmitter {
 
   /**
    * All the embedded fonts that were imported by the Editor
+   * @type {string[]}
    */
   fontsImported = [];
 


packages/super-editor/src/core/helpers/isFontAvailable.js
```diff
@@ -0,0 +1,51 @@
...
+ * @returns {boolean} True if the font is available, false otherwise.
+ *
+ * @example
+ * if (isFontAvailable('Arial')) {
+ *   // Use Arial font
+ * }
+ */
+
+export function isFontAvailable(fontName, fallbackFont = 'sans-serif') {
+  const _isFontAvailable = (fontName, fallbackFont) => {
```
chittolinag: We can run this fallback solution **if** `window.queryLocalFonts` (see [here](https://developer.mozilla.org/en-US/docs/Web/API/Local_Font_Access_API)) is not available.

packages/super-editor/src/core/Editor.js
```diff
@@ -874,13 +877,25 @@ export class Editor extends EventEmitter {
...
+    const styleString = this.converter.getFontFaceImportString();
 
     if (styleString?.length) {
       const style = document.createElement('style');
       style.textContent = styleString;
       document.head.appendChild(style);
     }
+
+    // FIXME: Check if this is enough to determine whether we're running on NodeJS
+    if (this.options.isHeadless) {
```
chittolinag, line 911: Maybe use `if (process) {}...`

packages/super-editor/src/core/helpers/canRenderFont.js
```diff
@@ -0,0 +1,51 @@
...
+
+  if (_isFontAvailable(fontName, fallbackFont)) {
+    return true;
+  }
+  // This extra verification is for the case where the `fontName` is the actual fallback font.
+  // If the browser renders Helvetica by default when the fallback is `sans-serif`, and the
+  // font being tested here is also Helvetica, this would return `false` because the width
+  // and height wouldn't change at all. To avoid that case, we check it again switching
+  // the fallback font.
+  const oppositeFallbackFont = fallbackFont === 'sans-serif' ? 'serif' : 'sans-serif';
```
luccas-harbour, line 49: Maybe you could check the `fontName` and run this before line 41? Just a small optimization so that we only run the font check once.

Just a quick note that when comparing the `fontName` to anything else you may need to make it lower case first as it could (?) come from the document as `Sans-Serif` for example.

packages/super-editor/src/core/helpers/isFontAvailable.js
luccas-harbour: I think that's a good idea. Not sure how performant it is to create these canvases and measure them, although a document probably doesn't have a very large number of different fonts is what I would guess. I saw that it's [supported by 88% of desktop browser users](https://caniuse.com/?search=queryLocalFonts) so I think it might be worth it.

chittolinag: Yeah exactly, thanks for pointing the stats too, very helpful! I see that ~30% of ALL users can use this as well. The bigger problem is with mobile devices. So I think the canvas solution could be used as a fallback.

I'll add some performance tests for this too, and if this starts to become a problem after X number of fonts (say 50, 100), we can always add a note on the documentation saying that this might impact the performance. And then, we wouldn't run this by default, but only if the user has explicitly passed a callback.

luccas-harbour: > And then, we wouldn't run this by default, but only if the user has explicitly passed a callback.

Oh that's smart. Good idea

packages/super-editor/src/core/helpers/canRenderFont.js
chittolinag: Hey! Not sure if I understand your question correctly, but... the reason I don't think we can run this only once is because we **do not** know what is the font that is rendered by the browser when it determines it should use `sans-serif` or `serif`. For example, for `sans-serif`, some browsers might render an `Arial` first, but in some other browsers it could render a different font.

So if the user gives `Arial` as the font to test in a browser where the default font of `sans-serif` is also `Arial`, this would say that the font is not supported. And hence we need to run it **again** by changing the fallback font to `serif`, hence the browser would render likely a `Times` or something similar, but always something different than `Arial`, and then in this extra call to `_isFontAvailable` it would return the correct value, saying that it is indeed supported.

Does that make sense?


Now, answering the 2nd question about lowercase... the `fontName` parameter is not supposed to be `serif` or `sans-serif`, but instead real font names such as `Arial`, `Helvetica`, etc... They would always come from the parsed `.docx`. If for some reason the `.docx` file has lowercased font families - they are usually capitalized -, we could definitely try to guess them and capitalize, but I'm not sure if it's worth doing this as we might risk breaking other fonts that are indeed lowercased. Or, if it doesn't work with the current font we're testing, we could then capitalize them and run again, but this would be an extra performance consideration here too... so not sure if it's worth it.

luccas-harbour: Ohh gotcha. I totally misunderstood, you're right, nothing to change here

packages/super-editor/src/core/Editor.js
```diff
@@ -874,7 +881,7 @@ export class Editor extends EventEmitter {
    * @returns {void}
    */
   #initFonts() {
-    const styleString = this.converter.getDocumentFonts();
+    const styleString = this.converter.getFontFaceImportString();
```
chittolinag: @artem-harbour @harbournick @VladaHarbour I think we import fonts that are embedded on the document, is that correct? If so, I might need to consider this before checking whether the font is available or not.

VladaHarbour: Exactly. But we only add according styles if font file is available in docx

packages/super-editor/src/core/Editor.js
```diff
@@ -883,6 +890,73 @@ export class Editor extends EventEmitter {
...
+   * @returns {Promise<void>}
+   */
+  async #checkFonts() {
+    // We only want to run the algorithm to resolve the fonts if the user has asked for it
+    if (!this.options.onFontsResolved || typeof this.options.onFontsResolved !== 'function') {
+      return;
+    }
+
+    // Ignore this in NodeJS
+    if (process && process.version && process.versions) {
```
VladaHarbour: I suppose when in Node we have isHeadless set to true. @harbournick am I right?

harbournick: that's true. I do think we need a better standardized way to do this check but @chittolinag for now maybe just do editor.options.isHeadless check instead. Or, maybe implement a simple getter like this and we can start using that instead of other versions of this check (there's a few through the codebase I think!)

```
get isNode() {
    return this.options.isHeadless || (
      typeof process !== 'undefined' && 
      process.versions?.node != null
    );
  }
}
```

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -561,6 +562,30 @@ class SuperConverter {
...
+      return [...new Set(inlineDocumentFonts)];
+    }
+
+    const wFonts = fontTable.elements?.find((element) => element.name === 'w:fonts');
+    if (!wFonts) {
+      return [];
+    }
+
+    if (!wFonts.elements) {
+      return [];
```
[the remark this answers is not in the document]
harbournick: @chittolinag check this one out

packages/super-editor/src/core/Editor.js
```diff
@@ -238,6 +240,7 @@ export class Editor extends EventEmitter {
     onPaginationUpdate: () => null,
     onException: () => null,
     onListDefinitionsChange: () => null,
+    onFontsResolved: null,
```
harbournick, line 249: for consistency should we use () => null?

packages/super-editor/src/core/Editor.js
```diff
@@ -883,6 +890,73 @@ export class Editor extends EventEmitter {
...
+    try {
+      const fonts = await window.queryLocalFonts();
+      const localFonts = [...new Set(fonts.map((font) => font.family))];
+      const unsupportedFonts = fontsUsedInDocument.filter((font) => !localFonts.includes(font));
+
+      this.options.onFontsResolved({
+        documentFonts: fontsUsedInDocument,
+        unsupportedFonts: unsupportedFonts,
+      });
+    } catch (err) {
```
harbournick: if we're not logging 'err' lets just drop it for the linter ie: catch { ... }

packages/superdoc/src/core/SuperDoc.js
```diff
@@ -96,6 +96,7 @@ export class SuperDoc extends EventEmitter {
     onException: () => null,
     onListDefinitionsChange: () => null,
     onTransaction: () => null,
+    onFontsResolved: null,
```
harbournick, line 99: use () => null for consistency, unless we have reason not to?

shared/common/data/blank.docx
harbournick, line 1: what changed about this file?

packages/super-editor/src/core/super-converter/SuperConverter.js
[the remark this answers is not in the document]
chittolinag: oh... great catch by copilot! just pushed a fix.

packages/super-editor/src/core/Editor.js
chittolinag: So I did this on purpose... if I set a default function here, it means we would always run the font check because there's indeed a function to be called. However, if it's `null` I won't execute the font check.

packages/superdoc/src/core/SuperDoc.js
chittolinag: Same as I mentioned in the other comment.

shared/common/data/blank.docx
chittolinag: Ohh great point. I forgot to add what changed here. I turns out the original `blank.docx` file that we use for initializing SuperDoc had some fonts that are usually not supported. So I simply removed the fonts, otherwise we'd get a callback saying that the fonts are not available when you initialize SuperDoc. You can see the difference between them here:

Screenshot 2025-10-13 at 12 38 38

Screenshot 2025-10-13 at 12 38 25

packages/super-editor/src/core/Editor.js
chittolinag: I'll use the `isHeadless` option for now, I'm a bit concerned about making that change since the deadline is a bit tight for this and it would require manual testing in multiple places.

packages/super-editor/src/core/Editor.js
chittolinag: done - added a verification to ensure that we take that into account.

shared/common/data/blank.docx
harbournick: perfect!

packages/super-editor/src/core/Editor.js
harbournick: ohhh, that makes sense actually. cool.

PR #1107, opened 2025-10-10
feat: ai configurable provider

falsaadehh:


packages/super-editor/src/extensions/ai/ai-plugin.js
```diff
@@ -142,6 +158,51 @@ export const AiPlugin = Extension.create({
...
           return true;
         },
+
+      /**
+       * Run an AI-powered content search and insert the results at the current selection.
+       * @param {string} prompt - The natural language prompt to send to the provider.
+       * @param {import('./provider-interface.js').AIProviderInterface} [provider] - Optional provider override.
+       * @returns {Promise<string|undefined>} - Resolves with provider result, if any.
+       */
+      aiFindContent: (prompt, provider) => () => {
```
artem-harbour: All these commands are not following the PM command contract. They should always return a boolean as a result and, for example, change something in the document.

In your case, they are all asynchronous and return a promise, can throw an exception and return some result other than a boolean. I would say that this is not a command at all, but, for example, a helper or something else. So I am confident they shouldn't be added as a command.

These functions should most likely be added as helpers (like, in `addHelpers()`) or follow PM command contract.

packages/superdoc-ai-controller/src/ai-commands.js
```diff
@@ -0,0 +1,294 @@
+/**
+ * AI Commands - Content manipulation commands powered by AI
+ */
+
+import { TextSelection } from 'prosemirror-state';
+
+/**
+ * Find content using AI search
+ */
+export async function aiFindContent(editor, prompt, provider) {
```
artem-harbour: These are not commands and they do not follow PM command contract. Looks more like a regular helper function.

packages/super-editor/package.json
```diff
@@ -32,6 +32,15 @@
     "./ai-writer": {
       "import": "./dist/ai-writer.es.js"
     },
+    "./ai": {
```
artem-harbour: I am not sure why these exports are needed here, I would prefer extensions to remain an internal implementation.

packages/ai/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@harbour-enterprises/ai",
```
artem-harbour: As far as I understand, this package will be published to npm, right?

I would suggest a more relevant package name. Example - `@harbour-enterprises/superdoc-ai`

packages/ai/README.md
```diff
@@ -0,0 +1,410 @@
+# @harbour-enterprises/ai
```
artem-harbour: See previous comment.

packages/superdoc/src/dev/components/SuperdocDev.vue
```diff
@@ -370,10 +370,42 @@ const exportDocxBlob = async () => {
   console.debug(blob);
 };
 
-const onEditorCreate = ({ editor }) => {
+const onEditorCreate = async ({ editor }) => {
   activeEditor.value = editor;
   window.editor = editor;
 
+  const ai = new SuperDocAI(superdoc.value, {
```
artem-harbour: It seems to me that the SD instance can be moved and passed around in the same options object as everything else.

packages/super-editor/src/extensions/track-changes/track-delete.js
```diff
@@ -53,6 +53,17 @@ export const TrackDelete = Mark.create({
         },
       },
 
+      authorImage: {
```
artem-harbour, line 56: What is the purpose of this new attribute?

packages/ai/src/utils.ts
```diff
@@ -0,0 +1,112 @@
+/**
+ * Shared utility functions
+ */
+import {FoundMatch, type Result} from "./types";
+
+export function validateInput(input: string, name: string): void {
+    if (!input?.trim()) {
+        throw new Error(`${name} cannot be empty`);
```
artem-harbour: This is a validation function, I think a boolean should be returned here. And at the level above (where you call it), then you can choose to throw an exception or do something else.

packages/ai/src/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+ */
+export function generateId(prefix: string): string {
+    return `${prefix}-${Date.now()}-${Math.random().toString(36).substr(2, 9)}`;
+}
+
+
+/**
+ * Normalizes find and replace entries from AI response
+ */
+export function normalizeReplacements(response: Result): FoundMatch[] {
```
artem-harbour: This function is hard to read.

packages/ai/src/ai-actions.ts
```diff
@@ -0,0 +1,451 @@
...
+import {validateInput, parseJSON, normalizeReplacements} from './utils';
+
+/**
+ * AI-powered document actions
+ * All methods are pure - they receive dependencies and return results
+ */
+export class AIActions {
+    constructor(
+        private provider: AIProvider,
+        private editor: EditorLike | null,
```
artem-harbour: EditorLike - I think it would be better to provide a specific type of editor instance here.

packages/superdoc-ai/src/ai-actions.ts
```diff
@@ -0,0 +1,451 @@
...
+ */
+export class AIActions {
+    constructor(
+        private provider: AIProvider,
+        private editor: EditorLike | null,
+        private user: AIUser,
+        private documentContext: string,
+        private enableLogging: boolean = false
+    ) {
+    }
```
artem-harbour, line 29: Do you need this class constructor? You don't define anything here.

packages/ai/src/ai-actions.ts
```diff
@@ -0,0 +1,451 @@
...
+        ]);
+
+        const result = parseJSON<Result>(response, {success: false, results: []}, this.enableLogging);
+
+        if (!result.success || !result.results || !this.editor) {
+            return {success: false, results: []};
+        }
+
+        try {
+            const adapter = new EditorAdapter(this.editor);
```
artem-harbour: I think the adapter should be created once when the class is initialized, and not in every method at runtime.

packages/ai/src/ai-actions.ts
```diff
@@ -0,0 +1,451 @@
...
+              "results": [ { 
+                  "originalText": string,
+                }
+              ]
+            }`;
+    }
+
+    private buildReplacePrompt(query: string, replaceAll: boolean): string {
+        const finalQuery = replaceAll
+            ? `apply this query: ${query} if find and replace query then Find and replace the EXACT text of ALL occurrences`
```
artem-harbour: I would suggest moving these prompts into constants or utility functions with arguments.

packages/ai/src/editor-adapter.ts
```diff
@@ -0,0 +1,135 @@
...
+    FoundMatch,
+} from './types';
+import {generateId} from "./utils";
+
+/**
+ * Adapter for SuperDoc editor operations
+ * Encapsulates all editor-specific API calls
+ */
+export class EditorAdapter {
+    constructor(private editor: EditorLike) {}
```
artem-harbour: Make the editor definition explicit in the constructor and change the type to the actual editor type.

packages/ai/src/editor-adapter.ts
```diff
@@ -0,0 +1,135 @@
...
+                    ...match,
+                    positions,
+                };
+            })
+            .filter((entry) => entry.positions.length > 0);
+    }
+
+    // Highlight operations
+    createHighlight(from: number, to: number, inlineColor: string = "#6CA0DC"): void {
+        this.editor.commands.setTextSelection({ from, to })
```
artem-harbour: I think that commands should be chained at least where possible, otherwise it may not always give the expected result in general.

Example
`editor.chain(). setTextSelection().setHighlight().run()`

packages/ai/src/editor-adapter.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    // Comment operations
+    async createComment(
+        from: number,
+        to: number,
+        text: string,
+        author: AIUser,
+    ): Promise<string> {
+        const commentId = generateId('comment');
+        this.editor.commands.enableTrackChanges();
```
artem-harbour: The same, it's better to chain commands when possible.

packages/superdoc-ai/src/providers.ts
```diff
@@ -0,0 +1,762 @@
...
+    };
+}
+
+/**
+ * Extracts text content from Anthropic completion payloads.
+ *
+ * @param payload - Raw Anthropic JSON payload.
+ * @returns Parsed text content, falling back to the default parser when needed.
+ */
+function parseAnthropicCompletion(payload: unknown): string {
```
artem-harbour: Unreadable code for me.

packages/superdoc-ai/src/providers.ts
```diff
@@ -0,0 +1,762 @@
...
+}
+
+/**
+ * Fallback completion parser capable of handling several common payload shapes
+ * (OpenAI-style choices, Anthropic content blocks, or raw strings).
+ *
+ * @param payload - Provider response payload.
+ * @returns Extracted text content suitable for callers.
+ */
+function defaultParseCompletion(payload: unknown): string {
```
artem-harbour: Unreadable code for me.

packages/superdoc-ai/src/providers.ts
```diff
@@ -0,0 +1,762 @@
...
+ *
+ * @param payload - Stream event payload.
+ * @returns Concatenated chunk text or undefined when no content is present.
+ */
+function parseOpenAIStreamChunk(payload: unknown): string | undefined {
+    if (!payload || typeof payload !== 'object') {
+        return undefined;
+    }
+
+    if ('choices' in payload && Array.isArray((payload as { choices?: unknown }).choices)) {
```
artem-harbour: I think the code style of this function could also be improved, as it's currently very AI generated-style (as it seems to me).

packages/superdoc-ai/src/providers.ts
```diff
@@ -0,0 +1,762 @@
...
+ * Streams data from a fetch `Response`, parsing SSE / chunked payloads into
+ * plain text snippets that are yielded back to the caller.
+ *
+ * @param response - Fetch response to consume.
+ * @param parseStreamChunk - Provider-specific chunk parser.
+ * @param fallbackParser - Parser to apply when custom parsing fails.
+ * @returns Async generator yielding parsed string chunks.
+ * @throws Error when the response status indicates failure.
+ */
+async function* readStreamResponse(
```
artem-harbour, line 319: Can we simplify this function?

packages/super-editor/src/extensions/ai/ai-plugin.js
artem-harbour: The comment is no longer relevant.

packages/super-editor/package.json
artem-harbour: The comment is no longer relevant.

packages/super-editor/src/extensions/track-changes/track-delete.js
falsaadehh: this to pass the author Image of the tracked changes if the user is changed

packages/superdoc-ai-controller/src/ai-commands.js
artem-harbour: The comment is no longer relevant.

packages/superdoc-ai/src/ai-actions.ts
falsaadehh: I'm using the constructor to set the API dependencies private provider: AIProvider,
        private editor: Editor | null,
        private user: AIUser,
        private documentContext: string,
        private enableLogging: boolean = false
    isn't the same as class a(r,c){this.r = r; this.c = c} I found we follow this appoach in ts
    Can you advise on a better code style to assign these dependencies?

packages/superdoc-ai/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "document",
+    "collaboration",
+    "llm",
+    "openai",
+    "anthropic"
+  ],
+  "author": "Harbour Enterprises",
+  "license": "AGPL-3.0",
+  "peerDependencies": {
+    "superdoc": "*"
```
artem-harbour, line 25: It probably makes sense to specify some minimum version here, rather than any version.

packages/superdoc-ai/src/ai-actions.ts
```diff
@@ -0,0 +1,394 @@
...
+
+        const prompt = buildFindPrompt(query, this.documentContext, findAll);
+        const response = await this.provider.getCompletion([
+            {role: 'system', content: SYSTEM_PROMPTS.SEARCH},
+            {role: 'user', content: prompt},
+        ]);
+
+        const result = parseJSON<Result>(response, {success: false, results: []}, this.enableLogging);
+
+        if (!result.success || !result.results || !this.adapter) {
```
artem-harbour: `!this.adapter` - does this check make sense? I think an adapter is required for this class, no?

packages/superdoc-ai/src/providers.ts
artem-harbour: Here the code style could be improved and made cleaner.

packages/superdoc-ai/src/providers.ts
```diff
@@ -0,0 +1,734 @@
...
+ */
+function defaultParseCompletion(payload: unknown): string {
+    if (typeof payload === 'string') return payload;
+    if (!payload || typeof payload !== 'object') return '';
+
+    const obj = payload as Record<string, unknown>;
+
+    // Try OpenAI format: choices[0].message.content or choices[0].text
+    const choice = Array.isArray(obj.choices) ? obj.choices[0] : null;
+    if (choice && typeof choice === 'object') {
```
artem-harbour: The code for this function could be improved and made cleaner.

This part in particular looks very unreadable.

packages/superdoc-ai/src/superdoc-ai.ts
```diff
@@ -0,0 +1,328 @@
...
+ */
+export class SuperDocAI<TSuperdoc = SuperDocInstance> {
+    private readonly superdoc: TSuperdoc;
+    private readonly config: SuperDocAIConfig;
+    private callbacks: SuperDocAICallbacks<TSuperdoc>;
+    private isReady = false;
+    private initializationPromise: Promise<void> | null = null;
+    private readonly actions: AIActions;
+
+    public readonly action = {
```
artem-harbour, line 49: I think the creation of the `action` object can be automated, since the same contract (args of the function, async, etc) is used everywhere and the names of the internal functions are repeated. And this needs to be done in a separate class method and called in the constructor.

packages/superdoc-ai/src/superdoc-ai.ts
```diff
@@ -0,0 +1,328 @@
...
+
+        editor?.setOptions({
+            user: {
+                id: this.config.user.userId,
+                name: this.config.user.displayName,
+                image: this.config.user.profileUrl,
+            },
+        });
+
+        this.actions = new AIActions(this.config.provider, editor, this.config.user, context, this.config.enableLogging);
```
artem-harbour: What is the difference between `action` and `actions`? Won't this introduce confusion into the public interface of the class?

packages/superdoc-ai/src/superdoc-ai.ts
```diff
@@ -0,0 +1,328 @@
...
+ *
+ * // With existing provider instance
+ * const provider = createAIProvider({ type: 'openai', ... });
+ * const ai = new SuperDocAI(superdoc, {
+ *   user: { display_name: 'Bot' },
+ *   provider
+ * });
+ * ```
+ */
+export class SuperDocAI<TSuperdoc = SuperDocInstance> {
```
artem-harbour: Not sure why this class has a generic type? It seems to me that it is supposed to be used only with SD instance, no?

artem-harbour:
Did another iteration of the code review, and I still think there are comments/questions that are worth considering.

packages/superdoc-ai/src/superdoc-ai.ts
falsaadehh: actions is internal and refer to the class action is what the user will be calling
await ai.action.insertTrackedChange
I don't think it is confusing because the user won't have access to actions and when you're in the class you can see the difference

packages/superdoc-ai/src/superdoc-ai.ts
falsaadehh: I tried this but didn't like the results, I think I prefer it this way for now

falsaadehh:
@artem-harbour this is ready again. please take a look

packages/superdoc-ai/src/superdoc-ai.ts
artem-harbour: Why is this marked as resolved but there's no explanation? Just clarifying, unresolved for now.

packages/superdoc-ai/src/providers.ts
artem-harbour: No changes here, unresolved the comment for now.

packages/superdoc-ai/src/superdoc-ai.ts
falsaadehh: Hey because I fixed it no need for the type anymore

packages/superdoc-ai/src/providers.ts
falsaadehh: The code has been improved based on your comment that's why it's resolved the code here is outdated

artem-harbour (approved):

caio-pizzol:
@falsaadehh can you please hold on merging this one until Monday?

I'd like to do a throughout review first

@harbournick @edoversb FYI

falsaadehh:
> @falsaadehh can you please hold on merging this one until Monday?
> 
> I'd like to do a throughout review first
> 
> @harbournick @edoversb FYI

sure! please do so

edoversb:
This is such an exciting feature (and really sets the pattern for so much of our AI work moving forward) — let's do a code walkthrough/discussion too collectively, booked for Monday. Should be fun.

packages/superdoc-ai/src/superdoc-ai.ts
```diff
@@ -0,0 +1,335 @@
...
+        fn: () => Promise<T>
+    ): Promise<T> {
+        const editor = this.getEditor();
+        if (!editor) {
+            throw new Error('No active SuperDoc editor available for AI actions');
+        }
+        try {
+            this.callbacks.onStreamingStart?.();
+            const result: any = await fn();
+            this.callbacks.onStreamingEnd?.(result);
```
harbournick: since the shape is onStreamingEnd?: (context: { fullResult: any }) => void; 
should we be passing( { fullResult: result }) instead?

packages/superdoc-ai/src/superdoc-ai.ts
```diff
@@ -0,0 +1,335 @@
...
+    private async executeActionWithCallbacks<T extends Result>(
+        fn: () => Promise<T>
+    ): Promise<T> {
+        const editor = this.getEditor();
+        if (!editor) {
+            throw new Error('No active SuperDoc editor available for AI actions');
+        }
+        try {
+            this.callbacks.onStreamingStart?.();
+            const result: any = await fn();
```
harbournick: is 'any' needed here? shouldn't this be T (or drop it since it should be inferred from Promise<T>)? 

by using 'any' here we disable type checking

packages/superdoc-ai/src/superdoc-ai.ts
```diff
@@ -0,0 +1,335 @@
...
+            this.callbacks.onStreamingEnd?.(result);
+
+            return result;
+        } catch (error: Error | any) {
+            this.callbacks.onError?.(error);
+            this.handleError(error as Error);
+            throw error;
```
harbournick, line 182: since this.handleError already invokes this.callbacks.onError we will get a duplicate here

packages/superdoc-ai/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@harbour-enterprises/superdoc-ai",
```
harbournick: since we are moving away from @harbour-enterprises org we should use something else. Maybe superdoc-dev - @caio-pizzol thoughts? It is _possible_ that one day this gets its own npm publish path separately from main superdoc

packages/superdoc-ai/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "keywords": [
+    "superdoc",
+    "ai",
+    "document",
+    "collaboration",
+    "llm",
+    "openai",
+    "anthropic"
+  ],
+  "author": "Harbour Enterprises",
```
harbournick: Should we use Harbour Enterprises for authorship? @edoversb 
we could just remove this line, too

packages/superdoc-ai/src/editor-adapter.ts
```diff
@@ -0,0 +1,124 @@
...
+                const positions = rawMatches
+                    .map((match: { from?: number; to?: number}) => {
+                        const from = match.from;
+                        const to = match.to;
+                        if (typeof from !== 'number' || typeof to !== 'number') {
+                            return null;
+                        }
+                        return { from, to };
+                    })
+                .filter((value: any): value is { from: number; to: number } => value !== null);
```
harbournick: since the map returns { from, to } we don't need 'any' here I don't think?

packages/superdoc-ai/src/editor-adapter.ts
```diff
@@ -0,0 +1,124 @@
...
+    // Replace operations
+    async replaceText(from: number, to: number, suggestedText: string): Promise<void> {
+        this.editor.commands.setTextSelection({ from, to })
+        const marks = this.editor.commands.getSelectionMarks();
+        this.editor.commands.deleteSelection();
+        if (marks.length > 0) {
+            this.editor.commands.insertContent({
+                type: 'text',
+                text: suggestedText,
+                marks: marks.map((mark: any) => ({
```
harbournick: should we define a local MarkLike type rather than any here?  or use ProseMirror Mark type directly?

packages/superdoc-ai/package.json
caio-pizzol: I agree, let's use `@superdoc-dev` instead

packages/superdoc-ai/package.json
edoversb: Happy to remove author line

packages/superdoc-ai/package.json
edoversb: Agreed

harbournick (approved):
LGTM

edoversb:
Yes!!! Woot

PR #1113, opened 2025-10-11
fix: grab tracked change from inline node for contextMenu

palmer-cl:
- Fix the context for contextMenu returned so that track changes are found


palmer-cl:
@harbournick this used to work in https://github.com/Harbour-Enterprises/SuperDoc/pull/1039 - not quite sure what caused the regression where it stopped working

PR #1121, opened 2025-10-13
fix: hyperlink export causes issues in Word

VladaHarbour:
@harbournick While working on different ticket noticed that documents with links are fail to be opened in Word after export


packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -658,13 +660,19 @@ export function getTextNodeForExport(text, marks, params) {
...
+  const { node, extraParams } = params;
 
   // Separate tracked changes from regular text
   const trackedMarks = [TrackInsertMarkName, TrackDeleteMarkName];
   const isTrackedNode = node.marks?.some((m) => trackedMarks.includes(m.type));
   if (isTrackedNode) return translateTrackedNode(params);
 
+  // Separate links from regular text
+  const isLinkNode = node.marks?.some((m) => m.type === 'link');
+  if (isLinkNode && !extraParams?.linkProcessed) {
```
VladaHarbour, line 672: Links that were created in SD don't go through the processing inside run translator

harbournick: nice catch!

PR #1123, opened 2025-10-13
Feat/autocomplete iliad

faddock:


caio-pizzol:
@faddock thanks for the effort on this — autocomplete/ghost text is an interesting idea! However, this PR has a few structural issues that make it hard to move forward:

1. **Scope** — the PR includes a forked copy of the superdoc package, two full example apps, and an unrelated "Add to Chat" feature. PRs need to be focused on a single feature.
2. **Architecture** — ghost text needs to be implemented as a ProseMirror decoration (via `addPmPlugins()` + `DecorationSet`), not direct DOM manipulation. Look at `src/extensions/search/` for an example of how we do visual overlays.
3. **Regressions** — the AI endpoint default was changed to localhost and some toolbar items were removed, which would break existing users.

If you're interested in contributing autocomplete as a focused, architecture-aligned feature, we'd welcome a new PR that follows the extension pattern in `src/extensions/`. Happy to provide guidance if you want to give it another shot!

PR #1128, opened 2025-10-14
feat: added support for annotating tables in the structured content blocks

andrii-harbour:


packages/super-editor/src/extensions/structured-content/structured-content-table-commands.js
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * @typedef {Object} StructuredContentTableAppendRowsOptions
+ * @property {string} id - Structured content block identifier
+ * @property {number} [tableIndex=0] - Index of the table inside the block
+ * @property {Array<string[]>|Array<string>} rows - Cell values to append
+ * @property {boolean} [copyRowStyle=false] - Clone the last row's styling when true
+ */
+
+export const StructuredContentTableCommands = Extension.create({
+  name: 'structuredContentTableCommands',
```
harbournick: I'm not sure we need a new / unique extension here. I think we should put the commands inside the existing `StructuredContentCommands` extension. Any reason not to?

packages/super-editor/src/extensions/structured-content/structured-content-table-commands.js
```diff
@@ -0,0 +1,111 @@
...
+
+  addHelpers() {
```
harbournick: let's follow the same structure for helpers as in StructuredContentCommands - we can put each helper in its own file and import via import * as structuredContentHelpers from './structuredContentHelpers/index';

packages/super-editor/src/extensions/structured-content/structured-content-table-commands.js
```diff
@@ -0,0 +1,111 @@
...
+    return {
+      /**
+       * Find all tables inside a structured content block identified by id.
+       * @param {string} id - Structured content block id
+       * @param {import('prosemirror-state').EditorState} state - Editor state
+       * @returns {Array<{ node: import('prosemirror-model').Node, pos: number }>} tables with absolute positions
+       */
+      getTablesInStructuredContentBlockById: (id, state) => {
```
harbournick: getTablesInStructuredContentBlockById is really long. Seems we have other helpers already named:
getStructuredContentTags
getStructuredContentInlineTags
getStructuredContentBlockTags
getStructuredContentTagsById

so maybe we just call this one getStructuredContentTablesById

packages/super-editor/src/extensions/structured-content/structured-content-table-commands.js
```diff
@@ -0,0 +1,111 @@
...
+  },
+
+  addCommands() {
+    /**
+     * Normalize append row input into an array of row arrays.
+     * @private
+     * @param {Array} rowsOrValues - Raw row data
+     * @returns {Array<string[]>}
+     */
+    const normalizeRowsInput = (rowsOrValues) => {
```
harbournick: let's move this out of the addCommands() fn and into its own helper file (or even just a standalone function outside of the extension itself)

packages/super-editor/src/extensions/table/table.js
```diff
@@ -420,6 +457,47 @@ export const Table = Node.create({
...
   addCommands() {
     return {
+      /**
+       * Append multiple rows to the end of a table in a single transaction.
+       * @category Command
+       * @param {AppendRowsWithContentToTableOptions} options - Append configuration
+       * @example
+       * editor.commands.appendRowsWithContentToTable({ tablePos, valueRows: [['A','B'], ['C','D']], copyRowStyle: true })
+       */
+      appendRowsWithContentToTable:
```
harbournick: since we're already in the table extension I think we can drop the "toTable" part. Just call it appendRowsWithContent

packages/super-editor/src/extensions/table/table.js
```diff
@@ -1106,3 +1184,164 @@ function getCurrentCellAttrs(state) {
...
+/**
+ * Resolve the table node that should receive appended rows.
+ * Prefers an explicit table node, falling back to a position lookup.
+ * @private
+ * @param {import('prosemirror-state').Transaction} tr - Current transaction
+ * @param {number} [tablePos] - Absolute position of the table in the document
+ * @param {import('prosemirror-model').Node} [tableNode] - Explicit table node reference
+ * @returns {import('prosemirror-model').Node|null} Table node to append rows to, or null if not found
+ */
+function defineTableForRowAppend(tr, tablePos, tableNode) {
```
harbournick: since this function doesn't care about 'row append' at all - could possibly be used in other places where we need to resolve a table, let's just call it 'resolveTable' ?

packages/super-editor/src/extensions/table/table.js
```diff
@@ -1106,3 +1184,164 @@ function getCurrentCellAttrs(state) {
...
+
+/**
+ * Select the template row used to derive structure and attributes for appended rows.
+ * Prefers the last body row (containing table cells) and falls back to the last row in the table.
+ * @private
+ * @param {import('prosemirror-model').Node} tableNode - Table node to inspect
+ * @param {import('prosemirror-model').Schema} schema - Editor schema
+ * @returns {import('prosemirror-model').Node|null} Template row node or null if none exist
+ */
+function pickTemplateRowForAppend(tableNode, schema) {
```
harbournick: lots of helpers follow here. I wonder if we should move them to a helpers file / helpers folder and import? the table extension is getting really long and since some of these are specific to row append maybe we should make a 'row-insertion-helpers.js' file or something like that

packages/super-editor/src/extensions/index.js
```diff
@@ -192,6 +193,7 @@ const getStarterExtensions = () => {
     NodeResizer,
     CustomSelection,
     TextTransform,
+    StructuredContentTableCommands,
```
harbournick: once we move the commands to StructuredContentCommands let's remove this

harbournick (changes requested):
Looks good - just some higher level changes for organization and API names really. Nice work!

harbournick:
Check out the copilot comments too! A couple of important ones there

packages/super-editor/src/extensions/table/table.js
```diff
@@ -420,6 +457,47 @@ export const Table = Node.create({
...
+                copyRowStyle,
+              }),
+            )
+            .filter(Boolean);
+          if (!newRows.length) return false;
+
+          return chain()
+            .command(({ tr, dispatch }) => {
+              if (dispatch) {
+                insertRowsAtTableEnd({ tr, tablePos, tableNode: workingTable, rows: newRows });
```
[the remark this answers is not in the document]
andrii-harbour: I don't like this approach. It is limited to UI only and headless mode is not supported here. I'll do a little bit different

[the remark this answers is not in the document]
harbournick: The bigger issue is that we need to make sure insertRowsAtTableEnd doesn't get an undefined position ever because it _will_ break _that_ function

harbournick (approved):
LGTM

PR #1135, opened 2025-10-15
fix: replace tracked change bubble [SD-484]

palmer-cl:
- When replacing text, pass insertion ID to deletion to link with the same ID
- Add tests for single insertion, single deletion and replace

image


harbournick (approved):
LGTM

PR #1139, opened 2025-10-15
fix: toolbar font-size when multiple selected

artem-harbour:
Increase input size.
Screenshot 2025-10-15 at 17 09 20

Empty input if multiple sizes are selected.
Screenshot 2025-10-15 at 17 09 46

### DIFF
--- a/packages/super-editor/src/components/toolbar/super-toolbar.js
+++ b/packages/super-editor/src/components/toolbar/super-toolbar.js
@@ -790,7 +790,13 @@ export class SuperToolbar extends EventEmitter {
       const activeMark = markNegated ? null : rawActiveMark;
 
       if (activeMark) {
-        item.activate(activeMark.attrs);
+        if (activeMark.name === 'fontSize') {
+          const fontSizes = marks.filter((i) => i.name === 'fontSize').map((i) => i.attrs.fontSize);
+          const isMultiple = [...new Set(fontSizes)].length > 1;
+          item.activate(activeMark.attrs, isMultiple);
+        } else {
+          item.activate(activeMark.attrs);
+        }
       } else {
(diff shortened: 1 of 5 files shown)


packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -790,7 +790,13 @@ export class SuperToolbar extends EventEmitter {
       const activeMark = markNegated ? null : rawActiveMark;
 
       if (activeMark) {
-        item.activate(activeMark.attrs);
+        if (activeMark.name === 'fontSize') {
+          const fontSizeMarks = marks.filter((i) => i.name === 'fontSize');
+          const isMultiple = fontSizeMarks.length > 1;
```
harbournick: isMultple will be true if the selection spans across multiple nodes (ie: across paragraphs) even if the font sizes are the same in this case because each paragraph node would add a mark here. We would probably have to check mark attrs / equality to determine weather the same mark type with different attrs is present here: https://cdn.zappy.app/261d5b8b5143eea0ea51d73e5df9d7be.mp4

harbournick (changes requested):
Just one issue with isMultiple. It strikes me that the same "issue" we're fixing here happens with any other mark and toolbar button: for example, selections that have both bold and not bold text highlighted. Or different fonts. 

I mention it here as I believe there may be a follow-on "generalized" version of this ticket which may affect the approach.

packages/super-editor/src/components/toolbar/super-toolbar.js
artem-harbour: Fixed.

harbournick (approved):
LGTM

PR #1140, opened 2025-10-15
fix(toolbar): allow accept and reject tracked changes from the toolbar [SD-199]

palmer-cl:
- add toolbar specific wrapper functions in the track changes extension
- adds removal by Id, and falls back safely to selection
- add tests


palmer-cl:
> There is some behavior in super-toolbar.js that enables/disables the track change buttons in toolbar if selection is not inside a change fyi. We may need to hook into that as well - just double check behavior? specifically look for collectTrackedChanges

@harbournick sounds good I can fix that in this PR too

palmer-cl:
@harbournick I wasnt able to reproduce any issues with the buttons being enabled or disabled. Can you send me repro steps?

PR #1155, opened 2025-10-17
feat: cache tab decoration measurements

johanneswilm:


artem-harbour:
Added a label so that it wouldn't be merged during the investigation.

johanneswilm:
@artem-harbour thanks. I don't have access rights to allow adding labels to any of my PRs.

PR #1157, opened 2025-10-17
feat: typescript infrastructure for monorepo

caio-pizzol:
Set up root TypeScript infrastructure and update package tsconfigs to extend it, preparing the monorepo for gradual TypeScript migration.

---
Linear Issue: [SD-674](https://linear.app/harbour/issue/SD-674/ts-migration-phase-0-setup-typescript-infrastructure)

<a href="https://cursor.com/background-agent?bcId=bc-2c3298a6-7151-45d4-82c0-dbe15c201cdd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-2c3298a6-7151-45d4-82c0-dbe15c201cdd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">Open in Web</picture></a>

---

> [!NOTE]
> Adds a shared `tsconfig.base.json`, updates package `tsconfig.json` files to extend it with relaxed checks, adds `typescript` and a `type-check` script, updates `.gitignore` for generated d.ts, and adds a migration guide.
> 
> - **TypeScript Infrastructure**:
>   - Add root `tsconfig.base.json` (ESNext module, node resolution, relaxed checks, declaration maps) for monorepo.
>   - Update `packages/*/tsconfig.json` to extend root config, set `checkJs: false`, keep `allowJs: true`, and simplify options (drop `NodeNext` settings).
> - **Scripts & Dependencies**:
>   - Add `typescript` dev dependency and `type-check` script (`tsc --noEmit -p tsconfig.base.json`) in root `package.json`.
> - **Repo Hygiene**:
>   - Update `.gitignore` to ignore generated `**/src/**/*.d.ts` and `**/src/**/*.d.ts.map`, while allowing `!packages/*/src/*.d.ts`.
> - **Docs**:
>   - Add `docs/typescript-migration.md` outlining migration phases, guidelines, and configs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bfa9ac8a28644a7627d982a7d664735077609f78. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/typescript-migration.md
+++ b/docs/typescript-migration.md
@@ -1,93 +1,35 @@
 # TypeScript Migration Guide
 
-This document outlines the guidelines and conventions for gradually migrating the SuperDoc monorepo from JavaScript to TypeScript.
+This guide covers the gradual migration of the SuperDoc monorepo from JavaScript to TypeScript.
 
-## Overview
+## Quick Start
 
-The migration is being done in phases to minimize disruption to ongoing development. This approach allows us to:
+**Infrastructure is ready.** You can now:
 
-- Maintain full backward compatibility during the migration
-- Test incrementally as we convert files
-- Learn and adjust our patterns as we go
-- Avoid blocking other development work
+- Create new files as `.ts` or `.vue` with `<script lang="ts">`
+- Migrate existing `.js` files when making significant changes
+- Mix `.js` and `.ts` files freely - they work together
 
-## Infrastructure Setup (Phase 0 - Complete)
+**Type checking:** `npm run types:check --workspace=packages/super-editor`
 
-### Root Configuration
+---
 
-The monorepo uses a shared TypeScript configuration at the root:
+## When to Migrate a File
 
-- **`tsconfig.base.json`**: Base configuration that all packages extend
-- Key settings:
-  - `allowJs: true` - Allows mixing `.js` and `.ts` files
-  - `checkJs: false` - Does not type-check JavaScript files (gradual migration)
-  - `strict: false` - Relaxed type checking initially
-  - `skipLibCheck: true` - Faster builds, skip checking node_modules types
-  - `module: "ESNext"` with `moduleResolution: "node"` - Compatible with existing codebase
+Convert to TypeScript when:
 
-### Package Configuration
-
-Each package extends the root config and adds package-specific settings:
-
-```json
-{
-  "extends": "../../tsconfig.base.json",
-  "compilerOptions": {
-    // Package-specific overrides
-    "outDir": "dist",
-    "baseUrl": ".",
-    "paths": {
-      // Path aliases
-    }
-  },
-  "include": ["./src/**/*"],
-  "exclude": ["node_modules", "dist", "**/*.test.js"]
-}
-```
-
-### Type Checking
-
-Run type checking across the entire codebase:
-
-```bash
-npm run type-check
-```
-
-This command runs TypeScript in "no emit" mode, checking types without generating files.
-
-**Note**: During Phase 0, this command may show errors in JavaScript files since path aliases are configured per-package. These errors don't affect builds or tests. As files are migrated to TypeScript, this command will become more useful for catching actual type errors in `.ts` files.
-
-For package-specific type checking:
-
-```bash
-# Check types in super-editor
-npm run types:check --workspace=packages/super-editor
-
-# Build type declarations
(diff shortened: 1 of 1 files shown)


caio-pizzol:
@codex review

caio-pizzol:
@artem-harbour anything you'd change on this approach here?

@harbournick let me know your thoughts as well.

artem-harbour:
> @artem-harbour anything you'd change on this approach here?
> 
> @harbournick let me know your thoughts as well.

I don't have a strong opinion on what could be changed here, but this approach works well for me. The types are generated for JSDoc and TS files can also be used (which is great for migration).

I wonder what @chittolinag thinks about this setup.

artem-harbour:
Note:
`packages/super-editor/package.json`

It's probably worth updating the script to this one.
`"types:build": "tsc -p tsconfig.build.json || true",`

Example:
We created a new `.ts` file with this function. 
```typescript
export function testFn(a: number, b: number): number {
  return a + b;
}
```

If we imported it into another `.js` file and used it like that `testFn(1, 'b');`, it wouldn't produce any errors during build, which is good. But if we use it like this in the `.ts` file, it fails the build, which I think is not good at the migration step.

I think even if something is wrong with the types in the `.ts` file, it should be OK to continue creating the build at this stage. This step should be fairly straightforward when transitioning from JS files to TS and should not interfere with the build.

@caio-pizzol @harbournick @chittolinag

caio-pizzol:
> Note: `packages/super-editor/package.json`
> 
> It's probably worth updating the script to this one. `"types:build": "tsc -p tsconfig.build.json || true",`
> 
> Example: We created a new `.ts` file with this function.
> 
> ```ts
> export function testFn(a: number, b: number): number {
>   return a + b;
> }
> ```
> 
> If we imported it into another `.js` file and used it like that `testFn(1, 'b');`, it wouldn't produce any errors during build, which is good. But if we use it like this in the `.ts` file, it fails the build, which I think is not good at the migration step.
> 
> I think even if something is wrong with the types in the `.ts` file, it should be OK to continue creating the build at this stage. This step should be fairly straightforward when transitioning from JS files to TS and should not interfere with the build.
> 
> @caio-pizzol @harbournick @chittolinag

Good point @artem-harbour!

@chittolinag @harbournick any thoughts?

chittolinag:
> Note: `packages/super-editor/package.json`
> 
> It's probably worth updating the script to this one. `"types:build": "tsc -p tsconfig.build.json || true",`
> 
> Example: We created a new `.ts` file with this function.
> 
> ```ts
> export function testFn(a: number, b: number): number {
>   return a + b;
> }
> ```
> 
> If we imported it into another `.js` file and used it like that `testFn(1, 'b');`, it wouldn't produce any errors during build, which is good. But if we use it like this in the `.ts` file, it fails the build, which I think is not good at the migration step.
> 
> I think even if something is wrong with the types in the `.ts` file, it should be OK to continue creating the build at this stage. This step should be fairly straightforward when transitioning from JS files to TS and should not interfere with the build.
> 
> @caio-pizzol @harbournick @chittolinag

That's a good point @artem-harbour . My opinion though, is that we should not ignore the type check in those cases. My reasoning is that if we have a `.ts` file, they should be part of the compile-time check. If that file is still not migrated it would be a `.js` file, meaning that it wouldn't break like you said. However, once we migrate it to `.ts`, it means it should be treated as such, even if this produces a compile error (which then would be a good sign and we have to fix it). If we don't have time to migrate the entire file at once, we either do not migrate or migrate partially using `any`s .

caio-pizzol:
@artem-harbour @chittolinag Tested both approaches.

Key finding: **TypeScript doesn't generate .d.ts files for files with type errors, even with `|| true`**. So we'd get incomplete types without clear feedback.

Agree with @chittolinag - strict checking for `.ts` files is the right call. For gradual migration, the better path is keeping files as `.js` + JSDoc until ready to fully migrate to `.ts`. This gives us both migration flexibility and complete type safety.

PR #1168, opened 2025-10-21
fix: tabs initial loading

artem-harbour:
Linear: SD-561.

See comment in linear.


artem-harbour:
@harbournick, please review the PR.

PR #1170, opened 2025-10-23
perf(list-item): remove list item node view when in headless mode

palmer-cl:
- Adds helper to check if in headless mode 
- Adds unit tests 

Perf metrics:
```yaml
  Average Total Time:
    Before: 122.64ms
    After:  117.58ms
    Improvement: -4.1%

  Average Node Views:
    Before: 7.0
    After:  6.0
```


palmer-cl:
@harbournick Im curious your thoughts on this. After my first commit I noticed a failing test, which I believe to be caused by headless vs not headless rendering. This proposed update takes the cache, puts it on the converter which should work in both headless and rendered mode. Any blind spots here Im not thinking about?

palmer-cl:
@harbournick moving out of Draft, let me know your thoughts on above

PR #1172, opened 2025-10-23
fix: remove legacy handler from tableCell translator

VladaHarbour:


VladaHarbour:
> ### 💡 Codex Review
> https://github.com/Harbour-Enterprises/SuperDoc/blob/7f4d877ef89fed7710b92f147319768afed4853e/packages/super-editor/src/core/super-converter/v3/handlers/w/tc/helpers/legacy-handle-table-cell-node.js#L287-L289
> 
> 
> **P1 Badge Null referencedStyles crashes table cell import**
> The new flow passes `referencedStyles` from the table translator instead of letting the cell handler fetch them. When `_getReferencedTableStyles` returns `null` (e.g., tables without a style or styles missing from `styles.xml`), the row translator still forwards `referencedStyles: null`. Because the destructured default in `handleTableCellNode` only applies when the property is `undefined`, `referencedStyles` remains `null` here and `getTableCellMargins` immediately destructures from it. This throws a `TypeError` (`Cannot destructure property 'cellMargins' of null`), causing the entire DOCX import to fail for such tables. Guard the value before destructuring or ensure undefined is passed instead of null.
> 
> ℹ️ About Codex in GitHub

fixed in second commit

PR #1174, opened 2025-10-24
refactor: create v3 translator for w:t

VladaHarbour:


packages/super-editor/src/core/super-converter/v3/handlers/w/t/t-translator.js
```diff
@@ -0,0 +1,109 @@
...
+    // Handle the removal of a temporary wrapper that we added to preserve empty spaces
+    text = text.replace(/\[\[sdspace\]\]/g, '');
+  }
+  // Word sometimes will have an empty text node with a space attribute, in that case it should be a space
+  else if (!elements.length && encodedAttrs.xmlSpace === 'preserve') {
+    text = ' ';
+  }
+
+  // Ignore others - can catch other special cases here if necessary
+  else return null;
```
artem-harbour: Please don't do this and don't separate parts of the condition.

```javascript
// bad
if (cond) {
//
}

else if (cond) {
//
}

else {
  //
}


// good
if (cond) {
  //
} else if (cond) {
  //
} else {
  //
}
```

PR #1181, opened 2025-10-27
fix: update trackedTransaction to handle 'addToHistory' meta

caio-pizzol:


caio-pizzol:
@codex review

caio-pizzol:
@codex review

harbournick:
I don't think this is needed, and it can have unintended side effects. 

To set addToHistory false but still pass this gate, you'd just need to add an inputType:
```
const tr = editor.state.tr.replaceWith(from, to, node)
    .setMeta('inputType', 'programmaticChange')
    .setMeta('addToHistory', false);```

harbournick:
Closing - see my last comment. If that doesn't work LMK but I don't think we need to make any changes here

PR #1188, opened 2025-10-28
feat: add markdown serialization support to the editor

caio-pizzol:
- Introduced `getMarkdown` method in the `Editor` class to serialize editor content as Markdown.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -561,6 +561,16 @@
         "node": ">= 6"
       }
     },
+    "node_modules/@cypress/request/node_modules/uuid": {
+      "version": "8.3.2",
+      "resolved": "https://registry.npmjs.org/uuid/-/uuid-8.3.2.tgz",
+      "integrity": "sha512-+NYs2QeMWy+GWFOEm9xnn6HCDp0l7QBD7ml8zLUmJ+93Q5NF0NocErnwkTkXVFNiX3/fpC6afS8Dhb/gz7R7eg==",
+      "dev": true,
+      "license": "MIT",
+      "bin": {
+        "uuid": "dist/bin/uuid"
+      }
+    },
     "node_modules/@emotion/hash": {
       "version": "0.8.0",
       "resolved": "https://registry.npmjs.org/@emotion/hash/-/hash-0.8.0.tgz",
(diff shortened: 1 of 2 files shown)


caio-pizzol:
@codex review

packages/super-editor/package.json
```diff
@@ -68,6 +68,7 @@
     "prosemirror-transform": "^1.9.0",
     "prosemirror-view": "^1.33.8",
     "release-it": "19.0.3",
+    "turndown": "^7.2.2",
```
artem-harbour: I would suggest considering these two libraries for converting from HTML to Markdown. As far as I know, this is a more modern option for today.
https://github.com/remarkjs/remark
https://github.com/rehypejs/rehype

Examples
https://unifiedjs.com/learn/recipe/remark-html/#how-to-turn-html-into-markdown

caio-pizzol: This is great suggestion @artem-harbour - I was actually wondering myself :)

Let me change the implementation to use it instead.

artem-harbour: Nice!

caio-pizzol:
@codex review

packages/super-editor/package.json
```diff
@@ -67,6 +67,11 @@
     "prosemirror-tables": "^1.4.0",
     "prosemirror-transform": "^1.9.0",
     "prosemirror-view": "^1.33.8",
+    "rehype-parse": "^9.0.1",
```
artem-harbour, line 70: FYI - We will also need to think about the strategy of what we add to the build and what we add as `peerDependencies` so as not to bloat the SD build.

packages/super-editor/src/core/Editor.js
```diff
@@ -1593,6 +1598,25 @@ export class Editor extends EventEmitter {
...
   }
 
+  /**
+   * Get the editor content as Markdown
+   * @returns {string} Editor content as Markdown
+   */
+  getMarkdown() {
+    const html = this.getHTML();
+    const file = unified()
+      .use(rehypeParse, { fragment: true })
```
artem-harbour, line 1608: I would add `fragment` to options argument, so that the user can choose.

caio-pizzol: Good thought on flexibility! However, `fragment: true` should always be the value here because:

1. `getHTML()` returns `div.innerHTML` (a fragment), never a complete HTML document
https://github.com/Harbour-Enterprises/SuperDoc/blob/e5fc8fd5d8b374ae4419557828cd6eb238e45b5d/packages/super-editor/src/core/Editor.js#L1592
2. Setting `fragment: false` would cause rehype to expect `<!DOCTYPE html>...` which we never provide
3. This would result in parsing errors or incorrect markdown output

Because we are leveraging `getHTML` in the first place, the `fragment` option is an internal implementation detail that matches our HTML structure, not a user-facing choice.

artem-harbour (approved):
LGTM

packages/super-editor/package.json
caio-pizzol: Good point about bundle size!

I think `peerDependencies` doesn't make sense here because, users would need to manually install 5 packages just to use one method + users wouldn't know they need to install anything

What about lazy loading instead? @artem-harbour

artem-harbour: @caio-pizzol As far as I know, all the latest versions of NPM install them automatically, but this is probably not always true. So I think it's ok to keep these deps in the bundle at the moment, but something we need consider about that.

packages/super-editor/src/core/Editor.js
```diff
@@ -1593,6 +1598,25 @@ export class Editor extends EventEmitter {
...
+      .use(rehypeRemark)
+      .use(remarkGfm)
+      .use(remarkStringify, {
+        bullet: '-',
+        fences: true,
+      })
+      .processSync(html);
```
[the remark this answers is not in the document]
caio-pizzol: Great catch! This is correct - rehype-remark v10 is async and processSync will fail.

Updated to use async `.process()` and made the method async. This also means dynamic imports make even more sense since we're already forced to be async anyway.

Usage becomes:
`const markdown = await editor.getMarkdown();`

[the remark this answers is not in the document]
caio-pizzol: This isn't an issue - `processSync` works fine with `rehype-remark@10.0.1`. I tested it and all tests pass without errors.

The unified pipeline only throws the "processSync not supported" error when there are truly async plugins in the chain. Our current setup (rehypeParse → rehypeRemark → remarkGfm → remarkStringify) works synchronously.

Refs:
- [rehype-remark v10 docs](https://github.com/rehypejs/rehype-remark)
- [unified processSync](https://unifiedjs.com/explore/package/unified/#processorprocesssyncfile)

packages/super-editor/package.json
harbournick: @caio-pizzol what is the size increase of our bundle after adding these deps?

caio-pizzol: ### editor.es.js

**w dependencies:**

* Uncompressed: 1,918.48 kB
* Gzipped: 408.35 kB

**wo dependencies:**

* Uncompressed: 1,350.74 kB
* Gzipped: 288.29 kB

The five rehype/remark dependencies (`rehype-parse`, `rehype-remark`, `remark-gfm`, `remark-stringify`, `unified`) add:

* **567.74 kB** uncompressed (**29.6%** of total bundle)
* **120.06 kB** gzipped (**29.4%** of total bundle)

PR #1193, opened 2025-10-30
feat: implement convertSdtContentToRuns function for structured content conversion

andrii-harbour:
- Added convertStdContentToRuns function to convert SDT child elements into Word run elements.
- Updated translateStructuredContent to utilize convertStdContentToRuns for final document exports.
- Enhanced tests for translateStructuredContent and added new tests for convertStdContentToRuns to ensure correct functionality.

### DIFF
--- a/packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-structured-content.js
+++ b/packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-structured-content.js
@@ -1,5 +1,5 @@
 import { translateChildNodes } from '@converter/v2/exporter/helpers/translateChildNodes';
-import { convertStdContentToRuns } from '@converter/exporter';
+import { convertSdtContentToRuns } from '@converter/exporter';
 
 /**
  * @param {Object} params - The parameters for translation.
@@ -15,7 +15,7 @@ export function translateStructuredContent(params) {
     if (node?.type === 'structuredContent') {
       // For final docs, inline structured content is flattened to run elements,
       // removing the SDT wrapper so content renders as plain text in the output DOCX.
-      return convertStdContentToRuns(childElements);
+      return convertSdtContentToRuns(childElements);
     }
 
     if (node?.type === 'structuredContentBlock') {
--- a/packages/super-editor/src/core/super-converter/exporter.js
+++ b/packages/super-editor/src/core/super-converter/exporter.js
@@ -119,7 +119,7 @@ export const isLineBreakOnlyRun = (node) => {
  * @param {Array<Object>|Object} elements
  * @returns {Array<Object>}
  */
-export function convertStdContentToRuns(elements) {
+export function convertSdtContentToRuns(elements) {
   const normalized = Array.isArray(elements) ? elements : [elements];
   const runs = [];
 
@@ -139,13 +139,13 @@ export function convertStdContentToRuns(elements) {
       // Recursively flatten nested SDTs into the surrounding run sequence, skipping property bags.
       const sdtContent = (element.elements || []).find((child) => child?.name === 'w:sdtContent');
       if (sdtContent?.elements) {
-        runs.push(...convertStdContentToRuns(sdtContent.elements));
+        runs.push(...convertSdtContentToRuns(sdtContent.elements));
       }
       return;
     }
 
     if (RUN_LEVEL_WRAPPERS.has(element.name)) {
-      const wrapperElements = convertStdContentToRuns(element.elements || []);
+      const wrapperElements = convertSdtContentToRuns(element.elements || []);
       if (wrapperElements.length) {
         runs.push({
           ...element,
--- a/packages/super-editor/src/tests/export/structured-content-exporter.test.js
+++ b/packages/super-editor/src/tests/export/structured-content-exporter.test.js
@@ -1,7 +1,7 @@
 import { describe, it, expect } from 'vitest';
-import { convertStdContentToRuns } from '@converter/exporter.js';
+import { convertSdtContentToRuns } from '@converter/exporter.js';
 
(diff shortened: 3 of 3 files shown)


packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-structured-content.js
```diff
@@ -1,16 +1,30 @@
...
 
 /**
  * @param {Object} params - The parameters for translation.
  * @returns {Object} The XML representation.
  */
 export function translateStructuredContent(params) {
-  const { node } = params;
+  const { node, isFinalDoc } = params;
 
-  const childContent = translateChildNodes({ ...params, nodes: node.content });
```
artem-harbour, line 10: Why was it changed to this one? And `node` param instead `nodes`?
`const childContent = translateChildNodes({ ...params, node });`

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-structured-content.js
```diff
@@ -1,16 +1,30 @@
...
+  const childContent = translateChildNodes({ ...params, node });
+  const childElements = Array.isArray(childContent) ? childContent : [childContent];
+
+  if (isFinalDoc) {
+    if (node?.type === 'structuredContent') {
+      // Inline structured content is converted into plain runs so the final DOCX renders without SDT wrappers.
+      const runs = convertStdContentToRuns(childElements);
+      return runs.length === 1 ? runs[0] : runs;
```
artem-harbour: This check is not clear to me and the code does not explain it. Can you explain?

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-structured-content.js
```diff
@@ -1,16 +1,30 @@
...
+    }
+
+    if (node?.type === 'structuredContentBlock') {
+      return childElements.length === 1 ? childElements[0] : childElements;
```
artem-harbour, line 24: The same.

harbournick:
> The code looks ok, although it raises some questions for me.
> 
> @harbournick - Why don't we just keep the SDT content/fields as is? In Word it seems to look pretty ok along with the overall content. Please review the PR too.

I think this is a good point. We went this route for consistency (and because our PDF path was having issues with these sdts) but it is definitely tempting not to do this at all if possible. Less code always better! Will discuss with @andrii-harbour before we move forward either way

artem-harbour:
> > The code looks ok, although it raises some questions for me.
> > @harbournick - Why don't we just keep the SDT content/fields as is? In Word it seems to look pretty ok along with the overall content. Please review the PR too.
> 
> I think this is a good point. We went this route for consistency (and because our PDF path was having issues with these sdts) but it is definitely tempting not to do this at all if possible. Less code always better! Will discuss with @andrii-harbour before we move forward either way

Sounds good and I think it makes sense (at least discuss whether this is really necessary)! Thanks!

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-structured-content.js
andrii-harbour: `nodes` in the `translateChildNodes` is never used. https://github.com/Harbour-Enterprises/SuperDoc/blob/andrii/sd-481-add-isfinaldoc-flag-to-new-wsdt-generic-field-export/packages/super-editor/src/core/super-converter/v2/exporter/helpers/translateChildNodes.js#L9 

switching to `node` just makes it more explicit, what's is used instead

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-structured-content.js
andrii-harbour: Sure! 
when `isFinalDoc` is true, we need to strip std blocks. 
There are 2 types: inline and block.
For inline - we convert std's children into Word runs - `convertStdContentToRuns` handles that, preserving hyperlinks or revision marks but leaving us with plain w:r elements.
For block - we just remove the outer SDT and hand back those block elements directly. If there’s only one block we return that node; if there are several, we return the array so the caller can append them in order.


LT;DR: we are pulling out stdContent and converting either to w:r(inline) or just pasting as is (block)

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-structured-content.js
andrii-harbour: answered above for both

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-structured-content.js
artem-harbour: `nodes` - This is part of the params that are passed on to further calls, for example `exportSchemaToJson`, and can probably be used there. Or am I missing something here? fyi @harbournick

andrii-harbour: yes, but we take `nodes` not from our `{..., nodes...}` that is passed, but from `const { content: nodes } = params.node;`
 https://github.com/Harbour-Enterprises/SuperDoc/blob/andrii/sd-481-add-isfinaldoc-flag-to-new-wsdt-generic-field-export/packages/super-editor/src/core/super-converter/v2/exporter/helpers/translateChildNodes.js#L10 
 
 as for my understanding we just have a confused naming here and `nodes` has been never actually used.

harbournick (approved):
LGTM

PR #1199, opened 2025-10-31
fix: force track changes if input type is programmatic

artem-harbour:
Linear: SD-710

### DIFF
--- a/packages/super-editor/src/extensions/track-changes/track-changes.js
+++ b/packages/super-editor/src/extensions/track-changes/track-changes.js
@@ -24,7 +24,8 @@ export const TrackChanges = Extension.create({
           //   to += 1;
           // }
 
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/extensions/track-changes/track-changes.js
```diff
@@ -24,7 +24,8 @@ export const TrackChanges = Extension.create({
           //   to += 1;
           // }
 
-          tr.setMeta('acceptReject', true);
+          // tr.setMeta('acceptReject', true);
```
artem-harbour, line 27: Since we only have this check `notAllowedMeta.includes(tr.getMeta('inputType'))`, then it was probably an error, although it doesn't seem to affect anything.

PR #1200, opened 2025-10-31
fix: delete decoration widget blocks parent list item

VladaHarbour:


packages/super-editor/src/extensions/track-changes/plugins/trackChangesBasePlugin.js
```diff
@@ -139,10 +139,12 @@ const getTrackChangesDecorations = (state, onlyOriginalShown, onlyModifiedShown)
           () => {
             const span = document.createElement('span');
             span.classList.add('track-delete-widget');
-            span.contentEditable = false;
```
VladaHarbour, line 142: @harbournick widget has [contenteditable=false] on its own

PR #1211, opened 2025-11-03
feat: streaming and superdoc-ai demo

falsaadehh:
In this PR I'm adding a new option to our superdoc-ai where the results can be streamed back if the user pass streamResults: true. this is not just making the provider stream the results but receiving the results format it and stream it to the UI as well.  
When streamResults is true SuperDoc AI will pass stream: true to the provider and the actions: insertContent, and summarize will stream the results back

### DIFF
--- a/examples/superdoc-ai-quickstart/.env.example
+++ b/examples/superdoc-ai-quickstart/.env.example
@@ -1,0 +1,3 @@
+# Copy this file to ".env" and fill in your API key before running the demo.
+VITE_OPENAI_API_KEY=sk-your-openai-key
+VITE_OPENAI_MODEL=gpt-4o-mini
--- a/examples/superdoc-ai-quickstart/README.md
+++ b/examples/superdoc-ai-quickstart/README.md
@@ -1,0 +1,57 @@
+# SuperDoc AI Quickstart
+
+Launch a live SuperDoc editor with AI helpers. This example shows how to wire up
+`@superdoc-dev/superdoc-ai` with a fresh SuperDoc instance, register a few buttons, and stream results back into the
+document.
+
+## Prerequisites
+
+- Node.js 18+
+- An OpenAI API key (or update the provider section in `src/main.js` for another provider)
+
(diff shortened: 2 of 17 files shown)


harbournick (changes requested):
- the 'theme' button doesn't work. Let's remove or fix (for simplicity, I'd suggest remove): https://cdn.zappy.app/129cccfbbd690db7b674aca8bbcef8c3.png
- there is no info on the README in the example for how to set up the backend expected: http://127.0.0.1:5002/api/ai - when I just follow the readme to run the example, insights fail with "failed to fetch" and I'm not clear on what I'm missing
- When insights fail to fetch, we still populate insights and say "insights ready" even if they all failed: https://cdn.zappy.app/c7863e26cf32cbf5d3dc74dd18c5df1e.png
- let's not abbreviate to "SD" here. Use the logo or full name SuperDoc: https://cdn.zappy.app/a5b0e921100e6f0b58b1d6b4d393083f.png

falsaadehh:
> * the 'theme' button doesn't work. Let's remove or fix (for simplicity, I'd suggest remove): https://cdn.zappy.app/129cccfbbd690db7b674aca8bbcef8c3.png
> * there is no info on the README in the example for how to set up the backend expected: http://127.0.0.1:5002/api/ai - when I just follow the readme to run the example, insights fail with "failed to fetch" and I'm not clear on what I'm missing
> * When insights fail to fetch, we still populate insights and say "insights ready" even if they all failed: https://cdn.zappy.app/c7863e26cf32cbf5d3dc74dd18c5df1e.png
> * let's not abbreviate to "SD" here. Use the logo or full name SuperDoc: https://cdn.zappy.app/a5b0e921100e6f0b58b1d6b4d393083f.png

Ready for review again
Changes made:
- Added streaming support to insertContent and summarize methods
- Ensured all actions fetch the latest document text for context rather than using stale text from instantiation time
- Added a quick start demo with improved readme

PR #1213, opened 2025-11-03
feat: add comments structure demo

mattConnHarbour:
Demo is live here: https://superdoc-comments-export.pages.dev/
Pre-built demo = faster load than stackblitz


harbournick:
Tagging @chittolinag here since this will be useful intel for upcoming comments projects

caio-pizzol:
Discussed with @mattConnHarbour we will move this demo component as an embed experience in the documentation

PR #1215, opened 2025-11-04
fix: superdoc-ai package.json

falsaadehh:


falsaadehh:
@harbournick had a problem with using the published version, this fixed it. I already published it for testing purposes. please check and let me know

PR #1219, opened 2025-11-05
feat: improve style scoping in the editor to prevent external styles affecting content

harbournick:
### Summary
  Prevents external/customer CSS from affecting editor styling. All editor instances now have the `.sd-editor-scoped` class with `all: revert` reset, protecting
  against global styles like `span { border: 1px solid red; }` that would otherwise leak into editor content.

### Why
  Customers' global CSS rules were bleeding into the editor, causing visual corruption. We can't control their stylesheets, so we need defensive isolation.

  ### How
  - Uses `all: revert` wrapped in `:where()` for zero-specificity reset
  - Applied to all editor containers and portals (main editor, resizers, popovers, pagination helpers)
  - Editor styles reapply cleanly after reset
  - Critical properties (fonts, sizes) set via inline styles, so they survive
  
### Example
**Before:**
  ```css
  /* Before: */
  .track-insert-dec { background-color: lightblue; }

  /* After: */
  .sd-editor-scoped .track-insert-dec { background-color: lightblue; }
```

### Notes
Testing will be added to visual testing repo

### DIFF
--- a/packages/super-editor/src/extensions/pagination/pagination.js
+++ b/packages/super-editor/src/extensions/pagination/pagination.js
@@ -14,6 +14,7 @@ import { LinkedStylesPluginKey } from '@extensions/linked-styles/index.js';
 import { findParentNodeClosestToPos } from '@core/helpers/findParentNodeClosestToPos.js';
 import { generateDocxRandomId } from '../../core/helpers/index.js';
 import { computePosition, autoUpdate, hide } from '@floating-ui/dom';
+import { applyStyleIsolationClass } from '@/utils/styleIsolation.js';
 
 const SEPARATOR_CLASS = 'pagination-separator';
 const SEPARATOR_FLOATING_CLASS = 'pagination-separator-floating';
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by harbournick** [comment]
> ### 💡 Codex Review
> https://github.com/Harbour-Enterprises/SuperDoc/blob/e1c7d9d3bed525d6861ecc92a204ca503788a228/packages/super-editor/src/extensions/pagination/pagination.js#L384-L396
> 
> 
> **P1 Badge Retain style isolation on hidden pagination container**
> The page-break calculation builds a hidden clone of the editor element, but `tempContainer.className = 'temp-container super-editor';` removes the new `sd-editor-scoped` class that applies the `all: revert` reset. When customer CSS overrides elements (e.g. `p { line-height: 2 }`), the visible editor is now isolated while the hidden container still picks up the external styles, so its measured heights will no longer match the real editor and pagination will insert breaks at wrong positions. Reapply the isolation class (or call `applyStyleIsolationClass`) after cloning to keep the layout calculations consistent.
> 
> ℹ️ About Codex in GitHub

fixed in 2758d880cbf9dd63d2c98a5b3f6e543227ab5d1b

File: packages/super-editor/src/assets/styles/helpers/isolation.css [hunk 1]
```diff
@@ -0,0 +1,11 @@
+.sd-editor-scoped,
+.sd-editor-scoped :where(*),
+.sd-editor-scoped :where(*::before),
+.sd-editor-scoped :where(*::after) {
+  all: revert;
```
johanneswilm, line 5: @harbournick This line is affecting the display of SVGs within the editor, at least in Firefox. They are not being rendered. Currently looking for a solution.

PR #1221, opened 2025-11-05
fix: floating comments not appearing on first load

harbournick:
## Summary
  Fixes comment positions not appearing on initial document load when only the active thread changes (without document content changes).

  ## Changes
  - Added `hasEverEmitted` flag to track first position emission
  - Fixed initial load detection logic to properly handle `prevDoc === null`
  - Enhanced position comparison to check for `bounds` existence before comparing
  - Ensures positions emit on first update even when only active thread changes

  ## Test Plan
  - ✅ Added regression test verifying positions emit on initial load with active thread change
  - ✅ Added integration test for imported comments + position loading flow
  - ✅ Verify floating comments appear immediately on document load

### DIFF
--- a/packages/superdoc/src/SuperDoc.test.js
+++ b/packages/superdoc/src/SuperDoc.test.js
@@ -2,6 +2,11 @@ import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { mount } from '@vue/test-utils';
 import { h, defineComponent, ref, reactive, nextTick } from 'vue';
 import { DOCX } from '@harbour-enterprises/common';
+import { Schema } from 'prosemirror-model';
+import { EditorState, TextSelection } from 'prosemirror-state';
+import { Extension } from '../../super-editor/src/core/Extension.js';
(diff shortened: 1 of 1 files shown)


packages/superdoc/src/SuperDoc.test.js
```diff
@@ -433,4 +433,48 @@ describe('SuperDoc.vue', () => {
     expect(superdocStoreStub.activeSelection.value).toBeNull();
     expect(wrapper.find('.superdoc__tools').exists()).toBe(false);
   });
+
+  it('shows floating comments after imported threads and positions load', async () => {
```
chittolinag, line 502: Hey @harbournick the fix seems to be working well on my tests here!

Quick thing I noticed here... this test passes on the current `main` branch, meaning that it would pass even with the actual broken version. I'd double check this test to ensure we're testing what we're expected to correctly.

harbournick: good catch! fixed in 2f11bb434fdaf4b83be36e84636afedb9e8adb4f

PR #1235, opened 2025-11-07
SD-688 - Move automated tests to main repo

chittolinag:


.github/workflows/pr-validation.yml
```diff
@@ -34,6 +34,6 @@ jobs:
   test:
     uses: ./.github/workflows/test-suite.yml
     with:
-      run-visual: true
-      run-e2e: ${{ github.base_ref == 'main' }}
+      run-visual-tests-examples: true
```
caio-pizzol: run-example-visual-tests?

packages/super-editor/src/extensions/font-family/font-family.js
```diff
@@ -54,7 +54,7 @@ export const FontFamily = Extension.create({
             parseDOM: (el) => el.style.fontFamily?.replace(/['"]+/g, ''),
             renderDOM: (attrs) => {
               if (!attrs.fontFamily) return {};
-              return { style: `font-family: ${attrs.fontFamily}` };
+              return { style: `font-family: Arial` };
```
caio-pizzol: Not able to set font attr anymore?

chittolinag: I did this on purpose to force an error (so I can validate the tests are working). I removed it already!

.github/workflows/pr-validation.yml
chittolinag: yeah I like it! just changed

chittolinag:
> Looks good,
> 
> Just one comment - why don't we place e2e inside `tests` folder (later we could move everything to it - so tests would be organized together)

Thanks @caio-pizzol ! I like that idea, but the reason it's not there is because e2e tests are supposed to test SuperDoc end-to-end. Right now we have separate `tests` folders since we have different packages. So the `e2e-tests` on the root folder is supposed to be used to test everything.

.github/workflows/test-suite.yml
```diff
@@ -29,138 +29,120 @@
...
-          gh pr create \
-            --base ${{ github.ref_name }} \
-            --title "chore: update visual snapshots" \
-            --body "Automated visual test updates" \
-            --label "visual-update"
-
-  e2e-tests:
-    if: inputs.run-e2e
+
+  run-e2e-tests:
```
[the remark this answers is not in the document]
chittolinag: applied

.github/workflows/release.yml
```diff
@@ -24,8 +24,8 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     uses: ./.github/workflows/test-suite.yml
```
harbournick: test-suite.yml has been deleted so this step will fail

chittolinag: great catch - just fixed!

PR #1237, opened 2025-11-07
Ai methods docs support

falsaadehh:
We're changing the package name and main class based on internal feedback. The new name will make the code easier to integrate, use, and extend

### DIFF
--- a/packages/ai/src/ai-actions.ts
+++ b/packages/ai/src/ai-actions.ts
@@ -9,7 +9,7 @@ import type {
     SuperDocInstance,
     SuperDoc,
 } from './types';
-import {CommandsEngine} from './commands-engine';
+import {AIActionsService} from './ai-actions-service';
 import {createAIProvider, isAIProvider} from './providers';
 
 /**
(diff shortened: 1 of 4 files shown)


examples/collaboration-token-example/src/main.js
```diff
@@ -0,0 +1,128 @@
+import 'superdoc/style.css';
```
caio-pizzol: @falsaadehh let's open a separate PR for those changes

packages/ai/src/ai-actions.ts
```diff
@@ -32,19 +32,19 @@ import {createAIProvider, isAIProvider} from './providers';
...
+export class AIActions {
     private readonly superdoc: SuperDocInstance;
-    private readonly config: SuperDocAIConfig;
-    private callbacks: SuperDocAICallbacks;
+    private readonly config: AIActionsConfig;
+    private callbacks: AIActionsCallbacks;
     private isReady = false;
     private initializationPromise: Promise<void> | null = null;
-    private readonly actions: AIActions;
+    private readonly actions: CommandsEngine;
```
caio-pizzol: now that we renamed to `CommandsEngine` should we still call it `actions`? Or maybe `commands`?

falsaadehh:
> Looks good!
> 
> Just 2 comments

done

caio-pizzol (approved):
LGTM

examples/superdoc-ai-quickstart/src/main.js
caio-pizzol, line 77: @falsaadehh going back here, I think this is not a very good example to have (since it can't be used in `production` wo a backend)

What do you think of adding an example using customHttp + minimal backend?

falsaadehh: @caio-pizzol this is the example, all examples requires some setup. If you're talking about the demo (something in the home page) that's a different thing we need to consider. like using a proxy similar to what we have for the insights endpoint. for now this is important to show the eng how to set it up from scratch

caio-pizzol: Right! I was actually talking about the examples for SuperDoc AI package - that I think should be an end2end working example with SuperDoc.<br><br>I agree we can keep it simple but I think it is important to add a comment about not using in production since it would leak the `API key` in the frontend

falsaadehh: yea we can add that note to the docs as well.

PR #1245, opened 2025-11-10
fix: update toolbar item label when linked style selected

VladaHarbour:
Change base branch to main after https://github.com/Harbour-Enterprises/SuperDoc/pull/1235 is merged


e2e-tests/tests/toolbar/toolbar.spec.js
```diff
@@ -233,6 +233,37 @@ test.describe('toolbar', () => {
       expect(hello).toBeVisible();
       expect(await hello.evaluate((el) => window.getComputedStyle(el).backgroundColor)).toBe('rgb(210, 0, 63)');
     });
+
+    test.use({ viewport: { width: 1920, height: 1080 } });
```
chittolinag: Hey @VladaHarbour ! It's worth confirming if this should be added inside the `test` block to avoid applying this globally (unless that's what we want).

e2e-tests/tests/toolbar/toolbar.spec.js
```diff
@@ -233,6 +233,37 @@ test.describe('toolbar', () => {
...
+
+      // Type "Hello"
+      await page.keyboard.type('Hello');
+
+      // Ensure the text is Arial
+      const hello = await superEditor.getByText('Hello');
+      expect(hello).toBeVisible();
+
+      const styleAttribute = await hello.getAttribute('styleid');
+      expect(styleAttribute).toBe('Heading2');
```
chittolinag: I wonder if we have other options to ensure that the style has been applied to the text. Checking by the `styleid` works for now but this doesn't ensure that the the styles are indeed applied. If someone breaks the logic that applies the style based on the attribute the test would still pass even though it's broken. 

What if we check whether the styles have been applied to the text by getting the computed styles of the element? We have some examples where we do this on this file.

This way we are 100% sure that the styles were indeed applied, and even if someone refactors this logic (ie, remove the `styleid` for example), the test would still be valid. Then we are relying on expected behavior rather than implementation details.

We can keep the `styleid` test too if that attribute has any importance besides applying the styles to the text.

Please let me know what you think here/if you have another opinion!

e2e-tests/tests/toolbar/toolbar.spec.js
```diff
@@ -910,5 +941,43 @@ test.describe('toolbar', () => {
...
+
+      // Wait for the toolbar to update
+      await sleep(500);
+
+      // Click back on the text
+      await superEditor.getByText('Hello').click();
+
+      // Wait for the toolbar to update
+      await sleep(500);
+      const styleButtonText = await styleButton.getByText('heading 2');
```
chittolinag: This one is great IMO because it doesn't rely on the implementation, it's simply checking whether the applied style label is visible on the screen.

VladaHarbour: That's what i was fixing actually. I think I can just rename this testcase to make it more clear but still check is label is applied since implementation is tested in different test case

e2e-tests/tests/toolbar/toolbar.spec.js
VladaHarbour: I agree! Let me add applied styles check as well

e2e-tests/tests/toolbar/toolbar.spec.js
```diff
@@ -911,4 +911,96 @@ test.describe('toolbar', () => {
...
+
+      // Select "heading 2"
+      await page.locator('div[aria-label="Linked style - Heading2"]').click();
+
+      // Click on the editor
+      await superEditor.click();
+
+      // Type "Hello"
+      await page.keyboard.type('Hello');
+      await page.keyboard.press('Enter');
```
VladaHarbour, line 948: in order for decorations to be applied

harbournick:
@chittolinag tests passed locally here but not in CICD. mind checking it out?

PR #1255, opened 2025-11-11
feat: upgrade list importer to v3 unify rendering with paragraph (SD-600)

luccas-harbour:
- Modifies rendering and parsing of lists to use paragraphs instead of separate nodes


harbournick:
@codex review

PR #1257, opened 2025-11-12
feat: in memory collaboration in SuperdocDev

falsaadehh:
Not Prod ready, this PR is for engineering collaboration and early feedback. please don't review or test unless assigned to you by ticket


packages/superdoc/src/core/collaboration/collaboration.js
```diff
@@ -66,9 +66,7 @@ function createProvider({ config, user, documentId, socket, superdocInstance })
 function createSuperDocProvider({ config, user, documentId, superdocInstance }) {
   const ydoc = new YDoc({ gc: false });
   const options = {
-    params: {
-      ...config.params,
-    },
+    params: config.params,
```
harbournick: Just curious: why the change?

packages/superdoc/src/core/collaboration/helpers.js
```diff
@@ -65,9 +65,24 @@ export const initSuperdocYdoc = (superdoc) => {
   if (!superdoc.config.superdocId) return;
 
   const documentId = isInternal ? baseName : `${baseName}-external`;
+
+  // Use user from collaboration config if provided, otherwise use top-level user
+  const user = superdoc.config.modules.collaboration?.user || superdoc.config.user;
```
harbournick: Curious: why the dual-user setup?

packages/superdoc/src/core/collaboration/helpers.js
```diff
@@ -65,9 +65,24 @@ export const initSuperdocYdoc = (superdoc) => {
...
+
+  // Add user info to params so it's sent to the server
+  const collabConfig = {
+    ...superdoc.config.modules.collaboration,
+    params: {
+      ...superdoc.config.modules.collaboration?.params,
+      name: user?.name,
+      email: user?.email,
+      role: user?.role,
```
harbournick: Why not keep user object rather than merging it with params?

packages/superdoc/src/core/collaboration/helpers.js
```diff
@@ -65,9 +65,24 @@ export const initSuperdocYdoc = (superdoc) => {
...
+    },
+  };
+
   const superdocCollaborationOptions = {
-    config: superdoc.config.modules.collaboration,
-    user: superdoc.config.user,
+    config: collabConfig,
+    user: user,
```
harbournick: Since the user is now in the configuration too seems we're doubling up. I'm not clear on the intent here

packages/superdoc/src/core/collaboration/helpers.js
```diff
@@ -88,10 +103,27 @@ export const initSuperdocYdoc = (superdoc) => {
...
+    // Use user from collaboration config if provided, otherwise use top-level user
+    const user = superdoc.config.modules.collaboration?.user || superdoc.config.user;
+
+    if (user && superdoc.colors[0]) {
+      user.color = superdoc.colors[0];
+    }
+
+    // Add user info to params so it's sent to the server
+    const collabConfig = {
+      ...superdoc.config.modules.collaboration,
```
harbournick: If we need to do this twice, once per fn, let's extract to a helper

harbournick (commented):
I'd probably skip the user stuff here. See my comment in the api side.

PR #1258, opened 2025-11-12
SD-379 - Bug: docx renders overlapping text

chittolinag:


packages/super-editor/src/core/super-converter/v3/handlers/w/p/helpers/w-p-helpers.js
```diff
@@ -129,7 +129,13 @@ export const getParagraphSpacing = (node, docx, styleId = '', marks = [], option
   // 2. Default style spacing
   // 3. Default paragraph spacing
   const lineSpacing = inLineSpacing?.['w:line'] || line || pDefaultSpacing?.['w:line'];
-  if (lineSpacing) spacing.line = twipsToLines(lineSpacing);
+  if (lineSpacing) {
```
chittolinag: On the documents that we reproduced this issue, the calculated line height was way too small (ie, 0.05, 0.005, etc). This makes all the text to be squashed together. 

The reason this happened there was because the style applied to the text says that the line height should be **at least** 1 twip (which after conversion would result in a super small value), but it doesn't mean that we should apply that value. So as a quick fix, we can simply make sure that the line height is reasonable.

PR #1269, opened 2025-11-14
fix: prevent counter sharing between lists with same abstract definition

harbournick:


luccas-harbour:
As explained in [this comment](https://github.com/Harbour-Enterprises/SuperDoc/pull/1265#discussion_r2526733801), this fix isn't necessary

harbournick:
not needed

PR #1278, opened 2025-11-18
feat: add proxy to ai example

mattConnHarbour:


falsaadehh:
@harbournick  I feel the proxy scripts shouldn't be here and should be moved to another repo for the team reference what do you think?

mattConnHarbour:
Closing since we don't want proxy code or deploy scripts in our examples

PR #1283, opened 2025-11-19
[SD-732] Maintain table grid widths across import/export

afn:
Give the tblGrid widths, if present, precedence over the first row's cell widths.


afn:
@luccas-harbour Would like your thoughts on whether this looks reasonable, since it inverts the logic you had previously explicitly tested in `tblGrid-translator.test.js`

harbournick (approved):
LGTM

PR #1288, opened 2025-11-20
fix: comment export to merge fragmented marks per commentId

financialvice:
## Summary
- merge all commentMark slices per commentId (and child comments) into one min..max range before export
- insert a single commentRangeStart/End pair per id, using transaction mapping for correct positions after prior inserts
- add a unit test covering fragmented marks exporting as one contiguous range

## Background

Previously, prepareCommentsForExport would emit range nodes using only the first encountered slice for a commentId. When a commentMark spanned multiple adjacent text nodes, the exported DOCX/round-trip highlight was truncated. This change collects min..max spans and emits a single start/end pair, preventing truncated anchors.

## Testing
- npm run build:super-editor
- npm run test:editor

---

disclaimer, generated primarily by gpt-5.1-codex-max

### DIFF
--- a/packages/super-editor/src/extensions/comment/comments-helpers.js
+++ b/packages/super-editor/src/extensions/comment/comments-helpers.js
@@ -56,74 +56,64 @@ export const getCommentPositionsById = (commentId, doc) => {
  * @returns {void}
  */
 export const prepareCommentsForExport = (doc, tr, schema, comments = []) => {
-  // Collect all pending insertions in an array
-  const startNodes = [];
-  const endNodes = [];
-  const seen = new Set();
+  // Collect a min..max range for every commentId (and children), then insert a
+  // single start/end pair per id. This prevents fragmented highlights when marks
+  // span multiple adjacent nodes.
+  const ranges = new Map();
+
+  const addRange = (id, from, to, attrs = {}) => {
+    if (!id || id === 'pending') return;
+    const existing = ranges.get(id);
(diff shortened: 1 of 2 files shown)


harbournick:
@chittolinag related to our discussion today. Mind checking this one out?

chittolinag:
> @chittolinag related to our discussion today. Mind checking this one out?

yes - I'll review this one today.

chittolinag:
Hey @financialvice - thanks for submitting the PR here! Can you please share the document you reproduced the issue? Just by reading the PR description + code, I can't really tell what _exactly_ is being fixed - it will be easier with a real example. Also, I'm working on something possibly related so just wanna double check whether your PR fixes my issue or my PR fixes yours 🙂

Thank you!

caio-pizzol:
@financialvice thanks for the contribution! the core issue you identified (fragmented comment marks exporting as truncated ranges) was a real bug.

however, this fix has already been shipped on main - `prepareCommentsForExport` was rewritten since this PR was opened and now collects `min..max` ranges per `commentId` in the same way your `addRange` helper does (see https://github.com/superdoc-dev/superdoc/blob/main/packages/super-editor/src/extensions/comment/comments-helpers.js#L252-L289).

the current version of the function also handles tracked change integration, Google Docs threading, and parent/child nesting order, so merging this PR as-is would regress those features.

closing this one out - the bug is fixed. appreciate you flagging it!

PR #1289, opened 2025-11-20
feat: add tag-based operations for structured content management

caio-pizzol:


packages/super-editor/src/extensions/structured-content/structured-content-commands.js
```diff
@@ -230,6 +231,68 @@ export const StructuredContentCommands = Extension.create({
           return true;
         },
 
+      /**
+       * Updates all structured content fields that share the same tag.
+       * Multiple fields can have the same tag, so this will update all matching fields.
+       * If any updated node does not match the schema, none of the fields will be updated.
+       * @category Command
+       * @param {string} tag - Tag value shared by multiple fields
```
harbournick: I think we should enforce a 'tag' shape rather than plain-text. This version should be OK but its more prone to user error. Eventually other strings might be required inside tags that are NOT meant to link two fields together and we'll run into trouble. By enforcing a shape for this specific case we can guarantee that plain strings will still be safe tags in the future. 

Something simple, like { linkedField: { key, value } } or something that is uniquely identifiable as intended to be linked.

harbournick (changes requested):
See my comment and LMK if you disagree

andrii-harbour (approved):
LGTM

PR #1294, opened 2025-11-20
feat: add automated regex search test and corresponding document

elliot-harbour:
Added regex search test from the superdoc-tests repo that didn't make it into the switchover


elliot-harbour:
@chittolinag I moved the test document into a new folder that doesn't run the visual diffs. Also added a fix for a race condition I experienced while testing.

chittolinag (approved):

PR #1295, opened 2025-11-21
updated collab example

mattConnHarbour:


packages/collaboration-yjs/examples/collaboration-in-production/server/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    } catch(error) {
+      const err = new Error('Failed to load document: ' + error);
+      err.name = 'LoadError';
+      throw err;
+    }
+  }) as LoadFn)
+  .onAutoSave((async (params) => {
+    try {
+      await saveDocument(params);
+    } catch (error) {
```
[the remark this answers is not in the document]
caio-pizzol: @mattConnHarbour take a look here

packages/collaboration-yjs/examples/collaboration-in-production/server/tiptap.ts
```diff
@@ -1,48 +1,44 @@
-import { encodeStateAsUpdate } from 'yjs';
-import type { CollaborationParams } from '@superdoc-dev/superdoc-yjs-collaboration';
+import type { StorageFunction } from './storage-types.js';
 
 const baseUrl = `https://${process.env.TIPTAP_APP_ID}.collab.tiptap.cloud/api/documents`;
 const existingDocuments = new Set<string>();
 
-export const loadFromTiptapCloud = async (documentId: string): Promise<Uint8Array | null> => {
-  const response = await fetch(`${baseUrl}/${documentId}`, {
+export const loadFromTiptapCloud: StorageFunction = async (id: string) => {
```
caio-pizzol: @mattConnHarbour, are you sure that what Cam was asking here was related to using TipTap Cloud to store the document or to actually use the TipTap WebSocket Cloud solution?

packages/collaboration-yjs/examples/collaboration-in-production/server/disk.ts
```diff
@@ -0,0 +1,125 @@
+import { promises as fs } from 'fs';
+import path from 'path';
+import type { StorageFunction } from './storage-types.js';
+import { encodeStateAsUpdate, applyUpdate, Doc as YDoc } from 'yjs';
+import JSZip from 'jszip';
+
+const DOCX_STORAGE_DIR = path.join(process.cwd(), 'docx-storage');
+
+// Conversion functions for DOCX to/from byte array
+export const yjsToDocxBlob = async (yjsByteArray: Uint8Array): Promise<Uint8Array | null> => {
```
caio-pizzol: Also, I'm not really sure why we have the save-to-disk example here. I think if we're talking about collaboration in production, we should really just show a flow end-to-end using a production structure.

If we want to show people how to save things to disk, I think we can create another example. Let's try to keep the examples really specific and only one flow.

If we need to show them more options, we should create a new example folder.

packages/collaboration-yjs/examples/collaboration-in-production/server/index.ts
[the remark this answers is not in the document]
mattConnHarbour: Fixing in latest commit

packages/collaboration-yjs/examples/collaboration-in-production/server/tiptap.ts
mattConnHarbour: Not sure here, actually removing all storage implementations. This example will still run, but now the developer will have to plug in their own storage code if they want file persistence. https://github.com/Harbour-Enterprises/SuperDoc/blob/SD-733/packages/collaboration-yjs/examples/collaboration-in-production/server/storage.ts

packages/collaboration-yjs/examples/collaboration-in-production/server/storage.ts
```diff
@@ -1,17 +1,18 @@
...
+metaMap.set('docx', []);
 
-if (storageType === 'tiptap') {
-  loadDocument = loadFromTiptapCloud;
-  saveDocument = saveToTiptapCloud;
-} else { // default to postgres
-  loadDocument = loadFromPostgres;
-  saveDocument = saveToPostgres;
-}
+export const loadDocument: StorageFunction = async (id: string) => {
```
mattConnHarbour, line 10: @caio-pizzol these load and save functions are essentially no-ops that the developer can use to make requests to their own storage implementation. I imagine them using an API layer/RESTful API here.

packages/collaboration-yjs/examples/collaboration-in-production/server/disk.ts
mattConnHarbour: Removing the disk.ts file along with postgres.ts and tiptap.ts

mattConnHarbour:
@caio-pizzol I removed the storage strategies here to keep things simple. Basically there are now two mock storage functions that a developer can plug their own storage implementation into, with some types as guidance.

From server/storage.ts
```
export const loadDocument: StorageFunction = async (id: string) => {
  // Return an empty Y.js document with minimal DOCX structure
  return encodeStateAsUpdate(blankDocxYdoc);
};

export const saveDocument: StorageFunction = async (id: string, file?: Uint8Array) => {
  // No-op - just return success
  return true;
};
```

Let me know if this is a bit too heavy handed. For demos and internal usage, I have a working example with file persistence here: https://github.com/mattConnHarbour/superdoc-collaboration-demo
That will be less of an example and, at some point, turn into a fully-functional application.

PR #1300, opened 2025-11-21
fix: import and export tagUtils forstructured content

caio-pizzol:
…ement


harbournick:
@caio-pizzol please revert this one and see my comments. We should not export this on its own. It should be part of the extension.

PR #1303, opened 2025-11-21
fix: add tagUtils as helpers

harbournick:
- usage: editor.helpers.structuredContent.<utilname>


harbournick:
nevermind - already exported via barrel

PR #1310, opened 2025-11-22
[SD-978] Fix bug: removing borders did not persist to exported docx

afn:
The underlying issue is that table nodes have both a `borders` attribute and a `tableProperties` attribute, which in turn also contains `borders`. This redundancy led to a discrepancy between how tables are rendered within SuperDoc and in Word.

This PR fixes the bug relating to table borders but doesn't address the lack of a single source-of-truth for table borders, so there may be other similar bugs wherein the two border specs are out-of-sync. This redundancy should be eliminated in a future PR.

Commits:

- **feat: change "Transparent borders" menu/toolbar item to "Remove borders"**
- **fix: removing borders did not persist to exported docx**


harbournick:
@afn mind checking the export case from linear, and resolving conflicts here?

afn:
Rebased to main and addressed the issue (see details in Linear)

PR #1318, opened 2025-11-24
fix: remove unused files, deps + root package.json file

caio-pizzol:


mattConnHarbour:
@caio-pizzol thanks for this. Great README, too!

PR #1323, opened 2025-11-24
feat: editor api upgrades for summarizing current schema and

harbournick:
- Reverts schema freezing fns
- Removes frozen schemas
- Keeps editor.getSchemaSummaryJSON and editor.validateJSON
- adds 'summary' and 'excludeFromSummaryJSON' config keys to extensions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -563,6 +563,16 @@
         "node": ">= 6"
       }
     },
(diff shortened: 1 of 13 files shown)


caio-pizzol:
@harbournick did we revert the schema management?

PR #1348, opened 2025-11-26
feat: add automated tests for importing tracked changes from Google D…

elliot-harbour:
Adds two tests and supporting documents for verifying that track changes redlines are retained on import to Superdoc


caio-pizzol:
Closing — the test infrastructure has changed since this was opened. e2e-tests/ is no longer in use; behavior tests now live in tests/behavior/using SuperDocFixture. See `tracked-change-existing-doc.spec.ts` as a reference for the current pattern.

The test intent is still valuable and can be revisited in a new PR targeting the current framework.

PR #1350, opened 2025-11-27
fix: close toolbar dropdowns with esc key

luciorubeens:
Fixes Escape key behavior to properly close toolbar dropdowns and the overflow menu.

This is a known issue: https://github.com/Harbour-Enterprises/SuperDoc/pull/579#issue-3141224985


packages/super-editor/src/components/toolbar/OverflowMenu.vue
```diff
@@ -32,19 +33,37 @@ const toggleOverflowMenu = () => {
...
+      emit('close');
+    }
+  }
+};
+
+onMounted(() => {
+  document.addEventListener('keydown', handleKeyDown, true);
+});
+
+onBeforeMount(() => {
```
harbournick: This should be onBeforeUnmount instead (make sure to update the import too)

packages/super-editor/src/components/toolbar/OverflowMenu.vue
```diff
@@ -1,11 +1,11 @@
 <script setup>
-import { getCurrentInstance, ref, computed } from 'vue';
+import { getCurrentInstance, ref, computed, onMounted, watch, onBeforeMount } from 'vue';
```
harbournick: watch is imported but not used - let's remove it

harbournick (changes requested):
So great to see this PR! Thank you for submitting it / catching it / contributing! 
Super useful fix.
Just two minor tweaks please!

packages/super-editor/src/components/toolbar/OverflowMenu.vue
luciorubeens: oh, good catch!

luciorubeens:
@harbournick Thanks for the quick review! Good catch. I've added two e2e tests to verify the behavior.

harbournick (approved):
LGTM - thank you for the contribution! It is such a nice addition.

PR #1371, opened 2025-12-01
feat: standardize usage of run node (SD-907)

luccas-harbour:
- Makes it so that every text node is always wrapped by a run node
- Fixes run-related issues identified in SD-783
- Fixes insertion of tab nodes from within run nodes
- Fixes arrow key navigation issues around run nodes


harbournick (approved):
LGTM

PR #1392, opened 2025-12-02
Preserving annotation field marks for text sdts

andrii-harbour:
- Introduced `keepTextNodeStyles` property to preserve text node styles during updates.
- Enhanced `updateStructuredContentById` and `updateStructuredContentByGroup` commands to utilize this new option.
- Implemented validation checks before applying updates to ensure content integrity. (SD-897)
- Added comprehensive tests for the new functionality, covering various scenarios including validation failures and style preservation.

### DIFF
--- a/packages/super-editor/src/extensions/structured-content/structured-content-commands.js
+++ b/packages/super-editor/src/extensions/structured-content/structured-content-commands.js
@@ -400,10 +400,12 @@ export const StructuredContentCommands = Extension.create({
           const { schema } = editor;
 
           if (dispatch) {
-            structuredContentTags.forEach((structuredContent) => {
+            // First pass: prepare and validate all updates before making any changes
+            // This ensures all-or-nothing behavior - either all nodes update or none do
+            const updates = [];
+
+            for (const structuredContent of structuredContentTags) {
               const { pos, node } = structuredContent;
-              const posFrom = tr.mapping.map(pos);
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/extensions/structured-content/structured-content-commands.js
```diff
@@ -413,6 +432,14 @@ export const StructuredContentCommands = Extension.create({
 
               const updatedNode = node.type.create({ ...node.attrs, ...options.attrs }, content, node.marks);
 
+              try {
+                const nodeForValidation = editor.validateJSON(updatedNode.toJSON());
+                nodeForValidation.check();
+              } catch (error) {
+                console.error('Invalid content.', 'Passed value:', content, 'Error:', error);
+                return false;
```
harbournick, line 443: since this return false is inside the forEach loop it doesn't actually propagate the failure state fyi. return here will simply continue to the next iteration of the loop ignoring the failure.

harbournick (changes requested):

packages/super-editor/src/extensions/structured-content/structured-content-commands.js
andrii-harbour: yeap, you're right! Thanks! Just updated

harbournick (approved):
LGTM

PR #1413, opened 2025-12-04
feat: AI Planner

falsaadehh:


packages/ai/src/ai-actions/planner.ts
```diff
@@ -0,0 +1,586 @@
...
+        };
+    }
+
+    /**
+     * Builds the system prompt with available tools and guidelines
+     * Uses the centralized prompt template from prompts module
+     * @private
+     * @returns System prompt string
+     */
+    private buildSystemPrompt(): string {
```
caio-pizzol: As I mentioned previously, I don't think passing tool descriptions and building the plan in a system prompt is the way we should do it.

If we're taking this approach, I would rather not call it AI Builder and keep calling it AI Actions with open prompt approach.

packages/ai/src/tools/builtin/find-tools.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { AIBuilderToolDefinition } from '../types';
+import { ERROR_MESSAGES } from '../../shared';
+
+/**
+ * Creates the findAll tool for locating content
+ * 
+ * @param actions - AI actions service instance
+ * @returns Tool definition with handler
+ */
+export function createFindAllTool(actions: any): AIBuilderToolDefinition {
```
caio-pizzol: Following up on the other comment, this is also not how providers define tools.

For example, this wouldn't work if you provide this tool as a tool parameter when calling the AI directly. This would be the case for OpenAI or Anthropic. We should find a common schema that we could use with any provider.

https://platform.openai.com/docs/guides/tools?tool-type=function-calling
https://platform.claude.com/docs/en/agents-and-tools/tool-use/overview

packages/ai/src/ai-actions/planner.ts
```diff
@@ -0,0 +1,586 @@
...
+    // PRIMARY: Pass SuperDoc instance + user config (AIActions created internally)
+    superdoc?: SuperDocInstance;
+    user?: AIUser;
+
+    // ADVANCED: Pass editor + pre-built AIActions instance
+    editor?: Editor;
+    aiActions?: AIActions;
+
+    // Provider configuration (required)
+    provider: AIProviderInput;
```
caio-pizzol: Also, ideally, we wouldn't run anything for the end user. We would just provide tool definitions and some helpers so the user could run it themselves together with other tools and third-party providers inside their workflow.

```
const response = await anthropic.beta.messages.create({
  model: 'claude-sonnet-4-5',
  system: systemPrompt, //<< their system prompt, not ours
  tools: [...superDocTools(), web_search, file_uploader], //<< our tools def combined with others
});
```

packages/ai/src/ai-actions/planner.ts
```diff
@@ -0,0 +1,586 @@
...
+                onStreamingEnd: config.onStreamingEnd,
+                onError: config.onError,
+            });
+            this.actions = this.aiActionsInstance.action as AIToolActions;
+        } else {
+            this.aiActionsInstance = null;
+            const service = new AIActionsService(
+                this.provider,
+                this.editor,
+                () => this.getFullDocumentContext(),
```
caio-pizzol: Even though we're using plain text, there could be cases where we don't support huge documents and we probably reach the context limit.

We should make that really clear in docs

caio-pizzol (commented):
Hey @falsaadehh,

I'm just adding the comments we talked about during our call yesterday. I think if we go with this approach, we shouldn't call it AI Builder because this is not really providing the low-level primitives people would look for when using AI at a more advanced level.

Also, I understand the benefits of plaintext, but I don't know how we would solve the formatting without leveraging the JSON schema definition.

packages/ai/README.md
```diff
@@ -21,18 +21,20 @@ npm install @superdoc-dev/ai
 
 ## Quick Start
 
+> ⚠️ **SECURITY WARNING**: This example is for **development only**. Never expose API keys in production browser code. See [Production Deployment](#production-deployment) for secure patterns.
```
caio-pizzol: I think this comment is enough no need to add it in 3 different places

packages/ai/README.md
```diff
@@ -72,11 +74,12 @@ new AIActions(superdoc: SuperDocInstance, options: AIActionsOptions)
 
 - `user` (required): User/bot information
   - `displayName`: Display name for AI-generated changes
-  - `userId?`: Optional user identifier
+  - `userId`: Unique identifier for the AI user (required)
   - `profileUrl?`: Optional profile image URL
 - `provider` (required): AI provider configuration or instance
 - `systemPrompt?`: Custom system prompt for AI context
 - `enableLogging?`: Enable debug logging (default: false)
+- `maxContextLength?`: Maximum number of characters from the document that will accompany AI prompts (default: 8,000)
```
caio-pizzol, line 79: Nice

packages/ai/README.md
```diff
@@ -412,6 +469,24 @@ AGPL-3.0 - see [LICENSE](../../LICENSE) for details.
 - 🐛 [Issue Tracker](https://github.com/harbour-enterprises/superdoc/issues)
 - 📧 [Email Support](mailto:support@superdoc.dev)
 
-## Changelog
+## Version & Compatibility
+
+**Current Version**: 0.1.7-next (Pre-release)
+
+**Supported SuperDoc Versions**: >=1.0.0-beta.17 <2.0.0
```
caio-pizzol: let's refer to @next channel only - @beta shouldn't be considered ehre

packages/ai/src/ai-actions/planner.ts
```diff
@@ -0,0 +1,613 @@
...
+  }
+
+  /**
+   * Builds the system prompt with available tools and guidelines
+   * Uses the centralized prompt template from prompts module
+   * @private
+   * @returns System prompt string
+   */
+  private buildSystemPrompt(): string {
+    const toolDescriptions = getToolDescriptions(this.tools);
```
caio-pizzol, line 507: As mentioned previously, this embeds tools as text in the system prompt rather than using native tool APIs

OpenAI and Anthropic both support structured tool definitions that provide:
- Guaranteed JSON schema validation
- Native `tool_choice` parameter
- Tool call streaming events
- Better token efficiency (schemas outside context window)

packages/ai/src/ai-actions/planner.ts
```diff
@@ -0,0 +1,613 @@
...
+    // Get editor - prioritize from superdoc if provided, otherwise use config.editor
+    const superdoc = config.superdoc ? (config.superdoc as unknown as SuperDoc) : null;
+    this.editor = superdoc?.activeEditor ?? config.editor!;
+    if (!isEditorReady(this.editor)) {
+      throw new Error(ERROR_MESSAGES.NO_ACTIVE_EDITOR);
+    }
+
+    this.enableLogging = Boolean(config.enableLogging);
+    this.logger = new Logger(this.enableLogging);
+    this.maxContextLength = config.maxContextLength ?? DEFAULT_CONTEXT_LENGTH;
```
caio-pizzol, line 169: `maxContextLength` uses characters, not tokens. 8,000 chars ≈ 2,000-3,000 tokens depending on content. Consider adding token estimation or documenting this clearly.

packages/ai/src/ai-actions/planner.ts
```diff
@@ -0,0 +1,613 @@
...
+            step,
+            context: { editor: this.editor, actions: this.actions },
+            previousResults: previousResults.map((pr) => ({
+              stepId: pr.stepId,
+              tool: pr.tool as AIToolName,
+              result: pr.result,
+            })),
+          });
+
+          if (!result?.success) {
```
caio-pizzol, line 314: Execution stops on first tool failure. For multi-step plans, consider adding a `continueOnError` option to allow independent steps to complete.

packages/ai/package.json
```diff
@@ -25,7 +44,7 @@
     "prosemirror-model": "^1.25.2",
     "prosemirror-state": "^1.4.3",
     "prosemirror-view": "^1.33.8",
-    "superdoc": "*"
+    "superdoc": "1.0.0-next.1"
```
caio-pizzol, line 47: Use `^1.0.0-next.1` to allow compatible versions (patch or minor updates)

packages/ai/src/ai-actions/tools/registry.ts
```diff
@@ -0,0 +1,102 @@
...
+  return new Map(entries.map((tool) => [tool.name, tool]));
+}
+
+/**
+ * Gets an array of tool descriptions for use in system prompts
+ *
+ * @param toolRegistry - Map of registered tools
+ * @returns Formatted string of tool descriptions
+ */
+export function getToolDescriptions(toolRegistry: Map<AIToolName, AIToolDefinition>): string {
```
caio-pizzol, line 78: Tool descriptions become plain text bullets here. Consider adding JSON schema `parameters` to each tool definition so they can be converted to native OpenAI/Anthropic tool formats later

packages/ai/src/ai-actions/planner.ts
```diff
@@ -0,0 +1,613 @@
...
+          });
+
+          if (step.tool === 'respond') {
+            assistantResponse = result.message ?? instruction;
+          } else if (!assistantResponse && result.message) {
+            assistantResponse = result.message;
+          }
+        } catch (error) {
+          const detail = getErrorMessage(error);
+          this.logger.error('Tool execution error', error, {
```
caio-pizzol, line 357: Good that instruction is truncated to 100 chars. Just ensure no sensitive document content gets logged (PII)

packages/ai/src/ai-actions/planner.ts
falsaadehh: @caio-pizzol the reasons I'm not using the native tool APIs 
1-The AIProvider interface only exposes getCompletion/streamCompletion with messages, not tool definitions and is expecting openAI, anthropic and customHttp which might not support the tool definition. the goal of this package to go beyond openAI and anthropic. 
2- Multi-step planning: The planner needs to generate a plan structure, not execute tools immediately
3- Step dependencies: Steps can reference previous step outputs, which native tool APIs don't handle directly, for example "create a summary and insert it as a comment" the comment step needs the output from the previous step and that's what we are handling.

packages/ai/src/ai-actions/planner.ts
falsaadehh: this will cause complication and many special situations handling, for example what if the next step depends on the previous step explicitly using $previous or implicitly using highlight or find all as first step. so for now we don't want to continue on error

packages/ai/src/ai-actions/tools/registry.ts
falsaadehh: adding this now is non trivial refactor, as we aren't supporting the native tool calling for the reasons mentioned above I'd consider this as a future enhancement

packages/ai/src/ai-actions/planner.ts
falsaadehh: done

packages/ai/package.json
falsaadehh: done

packages/ai/src/ai-actions/planner.ts
falsaadehh: done

falsaadehh:
> Great work on this refactor @falsaadehh
> 
> _(some of my comments may not apply due to the expected goal of the AIActions package)_
> 
> My main concern is that tools are passed as text in system prompts rather than using native OpenAI/Anthropic tool APIs. This affects reliability (manual JSON parsing vs guaranteed schema) and misses provider features like `tool_choice` and parallel tool calls.

@caio-pizzol checked all the comments, please check out my reply updates and let me know

caio-pizzol (approved):
LGTM

note: since tools defs are not following the standard approach - we won't be able to use them in the future

PR #1432, opened 2025-12-05
fix: pasting lists from docx (SD-975)

luccas-harbour:
- Fixes style resolution issues when pasting lists from docx


packages/super-editor/src/core/inputRules/docx-paste/docx-paste.js
```diff
@@ -81,6 +92,7 @@ export const handleDocxPaste = (html, editor, view) => {
         start = startMap[numId];
       }
 
+      item.setAttribute('data-marker-font-family', markerFontFamily);
```
VladaHarbour, line 95: seems that it may be set to undefined when no font family defined nor in numberingStyles neither in  normalStyles

luccas-harbour: that's ok... in that case it will be ignored when the numbering definition is created and it will eventually be set to the document's default font family

PR #1433, opened 2025-12-05
(stable) SD-984 - Bug: resolved comments marked as active

chittolinag:
Comments might have more than one sub-element, each with their own `paraId`. We were ignoring all the elements except for the first and using that element's `paraId` as the default. However, we use `paraId` to determine whether the comment is resolved or not so if the `commentsExtended` xml uses another sub-element's `paraId` instead (which is the case for the bug originally reported), we wouldn't identify that this comment is actually resolved.

### DIFF
--- a/packages/super-editor/src/extensions/comment/comments-helpers.test.js
+++ b/packages/super-editor/src/extensions/comment/comments-helpers.test.js
@@ -17,8 +17,8 @@ describe('prepareCommentsForImport', () => {
   const schema = new Schema({
     nodes: {
       doc: { content: 'inline*' },
-      commentRangeStart: { name: 'commentRangeStart', groups: ['inline'], isBlock: false },
-      commentRangeEnd: { name: 'commentRangeEnd', groups: ['inline'], isBlock: false },
+      commentRangeStart: { groups: ['inline'] },
+      commentRangeEnd: { groups: ['inline'] },
       text: { group: 'inline' },
     },
   });


packages/super-editor/src/core/super-converter/v2/importer/documentCommentsImporter.js
```diff
@@ -57,7 +57,8 @@ export function importCommentData({ docx, editor, converter }) {
       creatorName: authorName,
       creatorEmail: authorEmail,
       createdTime: unixTimestampMs,
-      textJson: parsedComment[0],
+      textJson: parsedElements[0],
```
chittolinag, line 60: Not to be tackled on this PR because it's out of scope, but this also cause issues because we're getting only the first element's content, but for cases where we have more than one, we'll be missing relevant text from the comment. I'll create a ticket to tackle this separately.

packages/super-editor/src/extensions/comment/comments-helpers.js
```diff
@@ -167,12 +167,14 @@ export const prepareCommentsForImport = (doc, tr, schema, converter) => {
...
-        start: pos,
-      });
+      if (!matchingImportedComment.isDone) {
+        toMark.push({
+          commentId: resolvedCommentId,
+          importedId,
+          internal,
+          start: pos,
+        });
+      }
```
luccas-harbour, line 177: Would this prevent the `w:commentRangeStart` tag from being imported if the comment is resolved? If that's the case, when exporting the document wouldn't we fail to include this XML element?

chittolinag: That's a great point. I wondered that myself as well, but on the current `stable` branch it seems like resolved comments are not exporting `commentRangeStart` and `commentRangeEnd` tags anyways (which is a bug by itself). So this change above is to specifically avoid highlight the comments if they are resolved.

I do have a ticket to tackle the `commentRangeStart` and `commentRangeEnd` not being exported on resolved comments, though!

But I'll make some more tests here before merging, I might be able to fix both on the same PR because I agree this might seem weird.

harbournick: @chittolinag based on teh above, is this ready to merge?

chittolinag: yes!

PR #1456, opened 2025-12-08
fix: correct pre-processing of w:fldChar (SD-1070)

luccas-harbour:
- Fixes an issue where during processing of `w:fldChar` fields, the raw nodes used when the field isn't handled were not the actual raw ones since they would be modified during processing.


harbournick (approved):
LGTM

PR #1457, opened 2025-12-08
feat: add logging to ydoc docx update function

mattConnHarbour:


harbournick (changes requested):
We shouldn't add logging without some kind of debug flag for this. I'd prefer not to add the logging at all but if its helpful to customers then please add a flag

mattConnHarbour:
@harbournick so far it's only been of use once. If it comes up again I'll make this PR with a flag.

PR #1459, opened 2025-12-08
feat: make collab provider-agnostic

caio-pizzol:


packages/superdoc/src/core/collaboration/collaboration.js
```diff
@@ -130,4 +134,34 @@ const onDestroy = (superdocInstance, documentId) => {
...
+function setupAwarenessHandler(provider, superdocInstance, user) {
+  const awareness = provider.awareness;
+  if (!awareness) {
+    console.warn('[superdoc] External provider missing awareness property');
+    return;
+  }
+
+  // Set local user state using standard Yjs awareness API
+  if (user && awareness.setLocalStateField) {
+    awareness.setLocalStateField('user', user);
```
harbournick, line 155: does this work in all the examples? double check that awareness works as expected. For some reason I seem to recall differenced between providers here.

packages/superdoc/src/core/collaboration/collaboration.js
```diff
@@ -130,4 +134,34 @@ const onDestroy = (superdocInstance, documentId) => {
...
+  }
+
+  // Listen to standard Yjs awareness 'change' event
+  awareness.on('change', (changes = {}) => {
```
harbournick, line 159: I think different providers emit different events. For instiance hocuspocus has awarenessUpdate no? does change capture what the end user would want to implement their own copresence view?

shouldn't it be 'update' more generally? I could be wrong though just double check

packages/superdoc/src/core/collaboration/collaboration.js
caio-pizzol: It does work!!

  Currently testing with `hocuspocus`, `y-sweet`, `tiptap-cloud` (same as hocuspocus) and 
  `liveblocks`.

 `setLocalStateField` is the s[tandard Yjs awareness API](https://docs.yjs.dev/api/about-awareness#awareness-crdt-api) from y-protocols - all 
  providers expose the same awareness object with this method.

  You might be thinking of `provider.setAwarenessField()` which is a 
  Hocuspocus-specific convenience method (we use it in the deprecated 
  createHocuspocusProvider), but that's different from 
  `awareness.setLocalStateField()`.

packages/superdoc/src/core/collaboration/collaboration.js
caio-pizzol: Change is actually the right choice here.

  The difference:
  - `change` → fires only when state actually changes (user joins/leaves/updates)
  - `update` → fires on every awareness message including heartbeats (~30s intervals)

  For co-presence views, change avoids unnecessary re-renders from heartbeats.

  You're right that Hocuspocus has awarenessUpdate/awarenessChange events, but those 
  are provider-level events (provider.on(...)). Here we're listening on the standard 
  Yjs awareness object (awareness.on(...)), which all providers expose with the same 
  change/update events.

PR #1465, opened 2025-12-08
fix: collab selection for multiple users clean up

harbournick:


harbournick:
@caio-pizzol just FYI

PR #1496, opened 2025-12-12
[SD-609] Switch to pnpm and various other build improvements

afn:
## Overview

This PR switches the package manager from`npm` to `pnpm` along with a number of other improvements/fixes to the development environment.

### pnpm

The repository is set up as a [pnpm workspace](https://pnpm.io/workspaces) (aka monorepo), defined by [pnpm-workspace.yaml](https://github.com/Harbour-Enterprises/SuperDoc/blob/81c1d76ce2c857eefda2af2e06d5ff91b6dbbeb9/pnpm-workspace.yaml). This file:
* enumerates the packages in the repository
* defines a [catalog of dependencies](https://pnpm.io/catalogs) to ensure version consistency among dependencies required by multiple packages (e.g. vite, eslint, etc.)

Because of the transition to pnpm, there is now a top-level `pnpm-lock.yaml`, replacing `package-lock.json`.

### Node version

The node version set in `.nvmrc` has been changed from 20.18.0 to 22.21.1. Node 20 will be EOL soon, and the `@types/node` package was already at v22.13.1.

We may want to consider further upgrading to Node 24, unless there are concerns with compatibility for users.

### Conditional exports

Several internal packages have been modified to include a conditional export with a condition named `source`, e.g.:

https://github.com/Harbour-Enterprises/SuperDoc/blob/07488b8af898c1545b66d255d22ca5908104f1b9/packages/super-editor/package.json#L9-L18

This condition is set in [vite.config.js](https://github.com/Harbour-Enterprises/SuperDoc/blob/81c1d76ce2c857eefda2af2e06d5ff91b6dbbeb9/packages/superdoc/vite.config.js#L155) so that imports from workspace packages will import the source, rather than the build package from `dist`, thus eliminating the need to (for example) build the `super-editor` package before building `superdoc`. It's also set in [tsconfig.json](https://github.com/Harbour-Enterprises/SuperDoc/blob/81c1d76ce2c857eefda2af2e06d5ff91b6dbbeb9/tsconfig.base.json#L28) to allow IDEs to navigate to dependent packages.

### Explicit package boundaries

With the change to conditional exports, there's no longer a need to define aliases that cross package boundaries. These have been removed:

https://github.com/Harbour-Enterprises/SuperDoc/blob/6b947eb4d500fb0970d9bf9dc919739609bb81f3/packages/superdoc/vite.config.js#L25-L36

and instead, the `superdoc` package can reference dependent packages like `@superdoc/contracts` directly.

### Separation of example code

In order to ensure that example code works in a standalone fashion (e.g. a user can copy the contents of an example directory like `examples/getting-started/vanilla` and build it as its own package), the `examples` directory has its own `pnpm-workspace.yaml` and is excluded from the top-level workspace. There's also no pnpm catalog for the examples, since including catalog references in individual example `package.json` files would prevent them from being used as-is outside of the project. Instead, `examples/pnpm-workspace.yaml` has:

```
overrides:
  superdoc: "file:../packages/superdoc/superdoc.tgz"
```

This allows developers to develop examples with a locally-packed `superdoc` package (built with `pnpm run --filter superdoc pack`) overriding the specific version referenced in the example's own `package.json`.

## Usage

For most commands, simply use `pnpm` instead of `npm` and `pnpx` instead of `npx`.

You may have to run `corepack enable` if the `pnpm` command is not found.

Ensure that you are using the node version from `.nvmrc`; this can be automated: [Auto-loading Node.js version with zsh and nvm](https://www.zeljkovic.sh/blog/2023/auto-loading-nodejs-version-with-zsh-and-nvm).

## Future work

There are a number of further improvements that can be addressed in future PRs:

1. Move packages from `shared/*` into `packages`
2. Use consistent naming for packages, reflecting the directory structure. For example, packages under `layout-engine` should probably have `layout-engine` in their names (`@superdoc/layout-engine-measuring-dom`? Or `@superdoc-layout-engine/measuring-dom`?)
3. The `superdoc` package still drills into `super-editor` by exposing aliases for `@core`, `@extensions`, etc. Consider breaking these up into separate packages. This may be challenging due to circular dependencies among the submodules of super-editor.
4. The `superdoc` package uses an alias `@superdoc` for its source, which conflicts with packages in the `@superdoc` scope. This requires some acrobatics in the [vite.config.js](https://github.com/Harbour-Enterprises/SuperDoc/blob/38b5c890bbce18b0783c10f88ff3d4f09e8b4d73/packages/superdoc/vite.config.js#L21-L31), which is error-prone and could be avoided by changing the alias to something like `@superdoc/superdoc`.

## Commits

- **feat: bump node to v22.21.1 (latest v22) in .nvmrc**
- **feat: switch to pnpm workspace**
- **feat: move package versions to pnpm catalog (via codemod)**
- **feat: move more dependencies to pnpm catalog**
- **feat: add "source" conditional export to super-editor package**
- **feat: move more dependencies to pnpm catalog**
- **feat: eliminate aliases for drilling into external packages**
- **feat: eliminate aliases from tsconfig and add "source" condition**
- **fix: add missing dependency: measuring-dom depends on word-layout**
- **fix: use pnpm instead of npm in install-rollup-binary.mjs**
- **feat: approve canvas, esbuild, and vue-demi to run build scripts**
- **fix: update path printed in ensure-types output**
- **fix: run superdoc tsc without "source" custom condition**
- **fix: use pnpm instead of npm in package.json scripts**
- **fix: import nextTick from vue instead of process**


afn:
Note: still working on fixing unit tests --- the majority of them seem to be related to changes in vitest (specifically with regards to how mocks are defined -- https://vitest.dev/api/vi#vi-spyon)

scripts/install-rollup-binary.mjs
```diff
@@ -87,6 +87,6 @@ if (existsSync(packageJsonPath)) {
 }
 
 console.log(`Installing Rollup native binary ${packageName}@${rollupVersion}...`);
-execSync(`npm install --no-save --no-package-lock ${packageName}@${rollupVersion}`, {
+execSync(`pnpm install --no-save --no-package-lock ${packageName}@${rollupVersion}`, {
   stdio: 'inherit'
```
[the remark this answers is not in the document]
afn: @harbournick I've removed the `install-rollup-binary.mjs` because it doesn't seem to be needed; see https://github.com/Harbour-Enterprises/SuperDoc/pull/1496/commits/d11d366b7b0ae96886c23d93f86d99e453aa2894. Please correct me if I'm wrong.

.github/workflows/manual-patch-release.yml
```diff
@@ -28,10 +28,12 @@ jobs:
           ref: ${{ github.event.inputs.ref || github.ref }}
           token: ${{ steps.generate_token.outputs.token }}
 
+      - uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # Pinned commit hash for @v4
```
caio-pizzol, line 31: why not use:

```
      - uses: pnpm/action-setup@v4
        with:
          version: XX
```

.github/workflows/pr-validation.yml
```diff
@@ -20,10 +20,12 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.sha }}
 
+      - uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # Pinned commit hash for @v4
```
caio-pizzol, line 23: Same here

.github/workflows/release.yml
```diff
@@ -61,18 +63,18 @@ jobs:
...
+        run: pnpm run build
 
       - name: Release
         env:
           GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
           LINEAR_TOKEN: ${{ secrets.LINEAR_TOKEN }}
-        run: npx semantic-release
+        run: pnpx semantic-release
```
caio-pizzol, line 77: not sure if applicable, but we could also release packages separately (if they are dependent on each other).

e.g. superdoc uses super-editor as dep

```
        env:
        ...
        run: pnpm --filter @superdoc-dev/super-editor exec semantic-release
```

e2e-tests/templates/vue/package.json
```diff
@@ -10,11 +10,11 @@
   },
   "dependencies": {
     "@fontsource/inter": "^5.2.8",
-    "superdoc": "file:../../superdoc.tgz",
-    "vue": "^3.5.13"
+    "superdoc": "workspace:*",
```
caio-pizzol, line 13: Nice!

Note that
`"superdoc": "workspace:*"` -> installs all versions
`"superdoc": "workspace:^"` -> only install patches and minors

pnpm-workspace.yaml
```diff
@@ -0,0 +1,95 @@
+packages:
+  - examples/*
```
caio-pizzol: 🔥

shared/common/package.json
```diff
@@ -34,10 +37,10 @@
     "typecheck": "tsc --noEmit"
   },
   "devDependencies": {
-    "@types/node": "^22.13.1",
-    "@vitejs/plugin-vue": "^5.0.4",
-    "typescript": "^5.7.3",
-    "vitest": "^3.0.8",
-    "vue": "^3.5.20"
+    "@types/node": "catalog:",
```
caio-pizzol, line 41: shoudn't `shared` be also inside `packages`?

packages/super-editor/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@harbour-enterprises/super-editor",
+  "name": "@superdoc/super-editor",
```
caio-pizzol, line 2: this is private right? we are not publishing to npm

p.s. npm org would be `@superdoc-dev/**`

packages/super-editor/package.json
```diff
@@ -50,63 +68,67 @@
     "test:debug": "vitest --inspect-brk --pool threads --poolOptions.threads.singleThread",
     "preview": "vite preview",
     "clean": "rm -rf dist types",
-    "pack": "rm *.tgz 2>/dev/null || true && npm run build && npm pack && mv harbour-enterprises-super-editor-0.0.1-alpha.0.tgz ./super-editor.tgz"
+    "pack": "rm *.tgz 2>/dev/null || true && pnpm run build && pnpm pack && mv harbour-enterprises-super-editor-0.0.1-alpha.0.tgz ./super-editor.tgz"
```
caio-pizzol, line 71: we might not need `pack` anymore since pnpm will handle deps build

.github/workflows/manual-patch-release.yml
afn: The commit hash specifies the version of the `pnpm/action-setup` action. For non-GitHub-managed actions like this one, it's safer to use a commit hash rather than a version tag, since the tag could be changed in the future. (This issue was flagged by CodeQL.)

The `version` parameter to the `pnpm/action-setup`, on the other hand, tells it which version of pnpm to install.

shared/common/package.json
afn: Yes 🙂  (I've noted this under "Future work" in the PR description --- I figured this was a big enough change that it would make sense to merge this first before reorganizing the directory structure and dealing with potential merge conflicts as a result)

packages/super-editor/package.json
afn: It is, yes. Something we should address separately is cleaning up our presence on npmjs:
* There's an old version of `@harbour-enterprises/super-editor` that should be pulled or at least marked deprecated
* The superdoc package is published as both `superdoc` and `@harbour-enterprises/superdoc`
* Other packages are under `@superdoc-dev`, not `@harbour-enterprises`

Should we open a Linear ticket to address / discuss this? It seemed out of scope for this PR... but when we *do* normalize the package names on npmjs, we should make sure the internal packages are named consistently.

packages/super-editor/package.json
afn: Thanks! I'll check and remove it if unnecessary...

packages/super-editor/package.json
harbournick: all future packages will be under @superdoc-dev (since we don't have the @superdoc organization) with the exception of 'superdoc' which is unscoped

for now we're still publishing to both the soon-to-be deprecated @harbour-enterprises scope and the unscoped package for superdoc. But the '@harbour-enterprises' scope won't be used for SuperDoc packages in the future fyi!

I removed the unused @harbour-enterprises/super-editor packaged fyi

afn:
> This looks good to me! Nice work here. Seems to build/work nicely. FYI I opened a PR to fix the pm-adapter and image test issues

Thanks! Merged in those fixes. Once all tests are passing, I want to check a few more things -- specifically, confirming that nothing's been inadvertently dropped from the superdoc package, and that all of the examples build correctly. I'll request a re-review once I'm satisfied that everything is correct.

packages/super-editor/package.json
chittolinag: Harbour's CLM is still using @harbour-enterprises, we'll need to update there as soon as we deprecate it. Do we have an estimate of when we will deprecate it? I'll log at ticket to start using `superdoc` instead sooner than later though.

chittolinag: Just as a side note, since every package (except `superdoc` itself) will be scoped, doesn't it make sense to have `superdoc` as a scoped package under `@superdoc-dev` as well? For consistency. Then users could install it with `@superdoc-dev/superdoc`. I'm afraid it might be counter-intuitive that every single package is available under @superdoc-dev but not `superdoc` itself.

afn:
@harbournick I haven't dug too deeply into it, but `test/performance.test.ts` is failing pretty consistently. Any ideas on where to start looking?

Also, I've changed the way that the visual tests in `examples/tests` are run so that it uses the superdoc package from the repo rather than pulling it from the npm registry (apart from the CDN example), so there are a lot of failures there. Some of them are probably expected, but I'd appreciate a set of eyes on this to verify which diffs are potential bugs.

chittolinag:
> @harbournick I haven't dug too deeply into it, but `test/performance.test.ts` is failing pretty consistently. Any ideas on where to start looking?
> 
> Also, I've changed the way that the visual tests in `examples/tests` are run so that it uses the superdoc package from the repo rather than pulling it from the npm registry (apart from the CDN example), so there are a lot of failures there. Some of them are probably expected, but I'd appreciate a set of eyes on this to verify which diffs are potential bugs.

performance were failing indeed, they were working at some point but some things got merged and we decided to skip them back then. we were supposed to revisit them once layout engine was deployed because this would impact their results.

for visual tests on examples, I can check it out.

.github/workflows/release.yml
afn: I didn't modify this (seemed to be reasonable to maintain a top-level `.releaserc` which can manage multiple packages if needed) but note that I added the `semantic-release-pnpm` plugin in fac1fc9331bf1edbf7184e4e98422a5c6bd3140e.

packages/super-editor/package.json
afn: Could we have a `superdoc` package that's just a stub that imports & re-exports everything from `@superdoc-dev/superdoc`?

Either way, I've created a ticket in Linear (I couldn't find an existing one) - https://linear.app/superdocworkspace/issue/SD-1194/normalize-npm-package-names

packages/super-editor/package.json
afn: We actually do need `pack` because of how visual tests are run: the `examples` directory is a separate pnpm workspace which overrides `superdoc` imports to use the generated tarball, so that visual tests are run on a package that resembles as closely as possible the published package:

https://github.com/Harbour-Enterprises/SuperDoc/blob/fac1fc9331bf1edbf7184e4e98422a5c6bd3140e/examples/pnpm-workspace.yaml#L16-L17

afn:
> performance were failing indeed, they were working at some point but some things got merged and we decided to skip them back then. we were supposed to revisit them once layout engine was deployed because this would impact their results.

I was thinking it might make sense to run performance tests in the same test suite as visual tests, so that it reflects performance of the production build (in case there's any difference compared to the dev build).

> for visual tests on examples, I can check it out.

Thanks! Please lmk if any of them look like they're failures due to newly introduced bugs/regressions as opposed to expected changes...

chittolinag:
> > performance were failing indeed, they were working at some point but some things got merged and we decided to skip them back then. we were supposed to revisit them once layout engine was deployed because this would impact their results.
> 
> I was thinking it might make sense to run performance tests in the same test suite as visual tests, so that it reflects performance of the production build (in case there's any difference compared to the dev build).
> 
> > for visual tests on examples, I can check it out.
> 
> Thanks! Please lmk if any of them look like they're failures due to newly introduced bugs/regressions as opposed to expected changes...

@afn I was checking out what went wrong with visual tests on the example apps and it seems like they are all expected changes. Nothing broken, simply UI updates. We'll have to update the baseline screenshots, but we don't have a way to do that now (I think we had it at some point but somethings changed and it stopped working), so I can think of two options:

1. Create an action that updates the screenshots on a given branch (and then I would trigger it manually passing your branch here).
2. Setup docker so we can update the screenshots locally (for consistency, like we discussed in the past, we can't update it on our local machines because we use MacOS)

I could simply "copy and paste" the current action that executes the tests and add a flag `--update-snapshots` on another github action that receives a branch as a parameter and that should be super simple, but I wonder if you have any other ideas? (of course, this would be just a workaround until we implement Storybook which should make this process a lot easier, so I wouldn't spend too much time here).


Now, about the performance tests, not sure I get what you mean, are the performance tests being executed separately now?

afn:
> @afn I was checking out what went wrong with visual tests on the example apps and it seems like they are all expected changes. Nothing broken, simply UI updates. We'll have to update the baseline screenshots, but we don't have a way to do that now (I think we had it at some point but somethings changed and it stopped working), so I can think of two options:
> 
> 1. Create an action that updates the screenshots on a given branch (and then I would trigger it manually passing your branch here).
> 2. Setup docker so we can update the screenshots locally (for consistency, like we discussed in the past, we can't update it on our local machines because we use MacOS)

Perfect, thanks for checking @chittolinag! I've added a Dockerfile and a few package.json scripts in d872b1a3311bd5735932ba231620072a31c206a9 and updated the screenshots in 1688bd14a040cab969a40b67fcd4bd10055e7b15.

afn:
> Now, about the performance tests, not sure I get what you mean, are the performance tests being executed separately now?

@chittolinag The performance tests are run in the `run-unit-tests` job, which builds the test code against the source (see note on "Conditional exports" in the PR description above) rather than against the production build of the package, which is how the e2e-tests and visual tests are run. I'm not sure if we have any performance-impacting functionality that's different in the test build versus production, but if so, we might see different results by moving performance tests into e2e-tests.

afn:
> > Now, about the performance tests, not sure I get what you mean, are the performance tests being executed separately now?
> 
> @chittolinag The performance tests are run in the `run-unit-tests` job, which builds the test code against the source (see note on "Conditional exports" in the PR description above) rather than against the production build of the package, which is how the e2e-tests and visual tests are run. I'm not sure if we have any performance-impacting functionality that's different in the test build versus production, but if so, we might see different results by moving performance tests into e2e-tests.

Turns out the failing performance tests have to do with the upgrade to jsdom --- this PR upgrades jsdom from 25.0.1 to 27.3.0 (the latest), and testing each incremental version of jsdom, the regression appears to come between 27.0.0-beta.2 and 27.0.0-beta.3. Looks like other folks have noted the regression as well, e.g. see https://github.com/jsdom/jsdom/issues/3936, https://github.com/jsdom/jsdom/issues/3984, https://github.com/jsdom/jsdom/issues/3985. For now I've [relaxed the latency expectations in performance.test.ts](https://github.com/Harbour-Enterprises/SuperDoc/commit/4c4258d7f7a23760e46bbdc9109db931d10a0835). cc @harbournick

chittolinag:
> > Now, about the performance tests, not sure I get what you mean, are the performance tests being executed separately now?
> 
> @chittolinag The performance tests are run in the `run-unit-tests` job, which builds the test code against the source (see note on "Conditional exports" in the PR description above) rather than against the production build of the package, which is how the e2e-tests and visual tests are run. I'm not sure if we have any performance-impacting functionality that's different in the test build versus production, but if so, we might see different results by moving performance tests into e2e-tests.

Got it! I thought you were talking about the performance tests on e2e. It might be worth checking this with @harbournick , this seems like a new thing that was introduced in the layout engine, but in general I agree that it's probably better to use the prod package for consistency/more accurate results.

And thanks for updating the screenshots on the example apps!

PR #1526, opened 2025-12-15
SD-1096 - Bug: tracked changes deletion is tracked char by char instead

chittolinag:
Problem: now that text nodes are wrapped inside a `run` node, the `nodesBetween` call that finds exiting marks between two positions (to check whether the newly added tracked changed is considered the same, to link them toegether) will no longer return the text node.

Example: if you compare this document in `main` vs `stable`:

Screenshot 2025-12-15 at 11 41 57

In `stable` its position is `4` whether in `main` it's `7`, because we now have one `run` node wrapping `tes` and another one wrapping the last `t`.


harbournick (approved):
LGTM

PR #1532, opened 2025-12-15
SD-1141, SD-1140 - Bug: cant delete content in suggesting mode between 2 others

chittolinag:
Fixes SD-1141 and SD-1140.

**Problem**:  the issue seems to be the same for both. We were correctly adding the marks, but the content was not being re-painted. Also, I noticed that sometimes the deletions were not being merged together correctly, so I made some changes on the `findTrackedMarkBetween.js` file, too.

### DIFF
--- a/packages/super-editor/src/extensions/track-changes/trackChangesHelpers/findTrackedMarkBetween.js
+++ b/packages/super-editor/src/extensions/track-changes/trackChangesHelpers/findTrackedMarkBetween.js
@@ -14,10 +14,10 @@ export const findTrackedMarkBetween = ({
   const startPos = Math.max(from - offset, 0); // $from.start()
   const endPos = Math.min(to + offset, doc.content.size); // $from.end()
 
-  let markFound;
+  let markFound = null;
 
   const tryMatch = (node, pos) => {
-    if (!node) {
+    if (!node || node?.nodeSize === undefined) {
       return;
     }
 
@@ -31,15 +31,13 @@ export const findTrackedMarkBetween = ({
         to: pos + node.nodeSize,
         mark,
       };
+      // Return false to stop the search
+      return false;
     }
   };
 
   doc.nodesBetween(startPos, endPos, (node, pos) => {
-    if (!node || node?.nodeSize === undefined) {
-      return;
-    }
-
-    tryMatch(node, pos);
+    return tryMatch(node, pos);
   });
 
   const inspectAroundPosition = (pos) => {
@@ -66,7 +64,10 @@ export const findTrackedMarkBetween = ({
     }
   };
 
-  inspectAroundPosition(startPos);
-  inspectAroundPosition(endPos);
+  if (!markFound) {
+    inspectAroundPosition(startPos);
+    inspectAroundPosition(endPos);
+  }
+
   return markFound;
 };


packages/super-editor/src/extensions/track-changes/trackChangesHelpers/findTrackedMarkBetween.js
```diff
@@ -32,27 +32,41 @@ export const findTrackedMarkBetween = ({
         mark,
       };
     }
+  };
+
+  doc.nodesBetween(startPos, endPos, (node, pos) => {
+    if (!node || node?.nodeSize === undefined) {
+      return;
+    }
```
afn: Move this condition inside `tryMatch`

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/findTrackedMarkBetween.js
```diff
@@ -32,27 +32,41 @@ export const findTrackedMarkBetween = ({
...
+
+    tryMatch(node, pos);
```
afn: Note that `nodesBetween` stops iteration early if the callback returns false, so we can make this more efficient if we change `tryMatch` above to return false when a match has already been found and propagate its return value here:

```suggestion
  doc.nodesBetween(startPos, endPos, (node, pos) => {
    return tryMatch(node, pos);
```

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/findTrackedMarkBetween.js
```diff
@@ -16,12 +16,12 @@ export const findTrackedMarkBetween = ({
 
   let markFound;
 
-  doc.nodesBetween(startPos, endPos, (node, pos) => {
-    if (!node || node?.nodeSize === undefined) {
+  const tryMatch = (node, pos) => {
+    if (!node) {
       return;
     }
```
afn, line 22:
```suggestion
  const tryMatch = (node, pos) => {
    if (!node || node?.nodeSize === undefined) {
      return !markFound;
    }
```

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/findTrackedMarkBetween.js
```diff
@@ -32,27 +32,41 @@ export const findTrackedMarkBetween = ({
         mark,
       };
     }
```
afn, line 36:
```suggestion
    }
    return !markFound;
```

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/findTrackedMarkBetween.js
```diff
@@ -32,27 +32,41 @@ export const findTrackedMarkBetween = ({
...
+    if (after?.type?.name === 'run') {
+      const node = after.content?.content?.[0];
+      if (node?.type?.name === 'text') {
+        tryMatch(node, pos);
+      }
+    }
+  };
+
+  inspectAroundPosition(startPos);
+  inspectAroundPosition(endPos);
```
afn: Minor nit, but neither of these is necessary if `markFound` is already set.

afn (commented):
LGTM - just a minor suggestion for improving performance

afn (approved):

PR #1554, opened 2025-12-17
fix: prevent naive ui from injecting styles globally, add uiDisplayFontFallback config option [main]

harbournick:
## Summary

- adds uiDisplayFallbackFont config key for UI element font
- SD UI will use 'Arial' as default font but can be overridden via uiDisplayFallbackFont
- prevents naive UI from injecting global styles


edoversb (approved):
LGTM -- great fix

PR #1555, opened 2025-12-17
feat: add displayMarginsOverride, used only when pagination is off [stable]

harbournick:
- add displayMarginsOverride config key (only active if pagination: false is also set)
- prevents min-height of editor when displayMarginsOverride
- used for "semantic" display of content

### DIFF
--- a/packages/super-editor/src/components/SuperEditor.vue
+++ b/packages/super-editor/src/components/SuperEditor.vue
@@ -46,12 +46,11 @@ const editor = shallowRef(null);
 const message = useMessage();
 
 /**
- * Computed property that determines if displayMarginsOverride is active.
- * The override is only active when pagination is disabled and an override is provided.
- * @returns {boolean} True if displayMarginsOverride should be applied
+ * Computed property that determines if responsive layout mode is active.
+ * @returns {boolean} True if layoutMode is 'responsive'
  */
-const hasDisplayMarginsOverride = computed(() => {
-  return !props.options.pagination && !!props.options.displayMarginsOverride;
+const isResponsiveMode = computed(() => {
+  return props.options.layoutMode === 'responsive';
 });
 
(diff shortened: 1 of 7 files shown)


caio-pizzol (commented):
Looks great @harbournick!

I was thinking we should rename this config:

`displayMarginsOverride` is a bit confusing - the name suggests it's about margins, but it actually enables responsive layout (100% width, text reflow). Someone reading the API won't know this is the solution for mobile/WCAG issues.

Proposal:
- `layoutMode: 'responsive' | 'paginated'` - clearly says what it does
- `layoutMargins: { top, bottom, left, right }` - optional, only in responsive mode

This way you just set `layoutMode: 'responsive'` and it works. No need to also set `pagination: false`.

Lmk what you think (I have the changes done locally already)

harbournick:
> Looks great @harbournick!
> 
> I was thinking we should rename this config:
> 
> `displayMarginsOverride` is a bit confusing - the name suggests it's about margins, but it actually enables responsive layout (100% width, text reflow). Someone reading the API won't know this is the solution for mobile/WCAG issues.
> 
> Proposal:
> 
> * `layoutMode: 'responsive' | 'paginated'` - clearly says what it does
> * `layoutMargins: { top, bottom, left, right }` - optional, only in responsive mode
> 
> This way you just set `layoutMode: 'responsive'` and it works. No need to also set `pagination: false`.
> 
> Lmk what you think (I have the changes done locally already)

Works for me. Have to be careful making the changes in stable vs main as they are very different. Please make a PR with this one as the base (or push your commits here). Some of the logic of _how_ and _when_ we actually use "layoutMargins" would actually change then too. 

We also have to make sure "layoutMode" is optional in v1 (main) as it should default to layout engine

caio-pizzol (approved):
LGTM

PR #1558, opened 2025-12-18
fix: pass user data in collaborative comment helpers, remove global superdoc

mattConnHarbour:
interesting: originally getting superdoc from window via DOM element. If missing DOM element, would get superdoc undefined

### DIFF
--- a/packages/superdoc/src/core/collaboration/collaboration-comments.js
+++ b/packages/superdoc/src/core/collaboration/collaboration-comments.js
@@ -1,6 +1,6 @@
 import { Map as YMap } from 'yjs';
 
-export const addYComment = (yArray, ydoc, event) => {
+export const addYComment = (yArray, ydoc, event, user) => {
   const { comment } = event;
   const yComment = new YMap(Object.entries(comment));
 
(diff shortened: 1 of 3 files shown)


packages/superdoc/src/core/collaboration/collaboration-comments.js
```diff
@@ -1,18 +1,18 @@
...
+export const addYComment = (yArray, ydoc, event, user) => {
   const { comment } = event;
   const yComment = new YMap(Object.entries(comment));
 
   ydoc.transact(
     () => {
       yArray.push([yComment]);
     },
-    { user: superdoc.user },
+    { user },
```
mattConnHarbour, line 11: This was originally reading from window, which was set by DOM element with id superdoc.
If user did not have that element on the page, this would fail.

packages/superdoc/src/core/collaboration/helpers.js
```diff
@@ -118,19 +118,20 @@ export const syncCommentsToClients = (superdoc, event) => {
   if (!superdoc.isCollaborative || !superdoc.config.modules.comments) return;
 
   const yArray = superdoc.ydoc.getArray('comments');
+  const user = superdoc.config.user;
```
mattConnHarbour, line 121: Now getting user from config.

harbournick (approved):
LGTM

PR #1559, opened 2025-12-18
SD-1098 - fix: properly re-render comment dialogs when tracked changes are added

chittolinag:
We are currently not updating the comment & tc dialog positions correctly. We are missing a dependency on the `watchEffect` hook. 

Note that the current e2e-test screenshots were wrong too, probably an error during some migration. I've update it so we have the correct baseline screenshots.


packages/superdoc/src/components/CommentsLayer/FloatingComments.vue
```diff
@@ -110,21 +110,15 @@ watchEffect(() => {
...
   nextTick(processLocations);
 });
 
-const resetLayout = async () => {
-  firstGroupRendered.value = false;
-  renderedSizes.value = [];
-  commentsRenderKey.value++;
-  verticalOffset.value = 0;
-};
-
```
chittolinag, line 127: Not used

packages/superdoc/src/components/CommentsLayer/FloatingComments.vue
```diff
@@ -110,21 +110,15 @@ watchEffect(() => {
   }
 
   const totalComments = getFloatingComments.value.length;
+  const measuredComments = renderedSizes.value.length;
```
chittolinag, line 113: We're basically just adding this so `renderedSizes` becomes a dependency of the `watchEffect`.

PR #1581, opened 2025-12-22
SD-58 (part 1) - feat: added permStart/permEnd translators and schema

chittolinag:
Add initial support for `w:permStart` and `w:permEnd` tags. These tags are used to enable edits in protected documents (ie, view only).  This is the initial part of it. With this PR, we still do not support the full feature, but at least the tags should be encoded/decoded properly. Next step is to implement the locking feature itself.

If you want more info about how this works in Word: https://support.microsoft.com/en-us/office/allow-changes-to-parts-of-a-protected-word-document-187ed01c-8795-43e1-9fd0-c9fca419dadf


packages/super-editor/src/extensions/perm-start/perm-start.js
```diff
@@ -0,0 +1,51 @@
...
+  name: 'permStart',
+  group: 'inline',
+  inline: true,
+
+  renderDOM() {
+    return ['span', { style: 'display: none;' }];
+  },
+
+  parseDOM() {
+    return [{ tag: 'span' }];
```
chittolinag: Now that I am looking at it again, I think this could cause problems. I will refine this a little bit.

packages/super-editor/src/core/super-converter/v3/handlers/w/perm-start/attributes/w-col-first.js
```diff
@@ -0,0 +1,29 @@
...
+};
+
+/**
+ * Decoder for the 'colFirst' attribute in SuperDoc.
+ * Maps to the 'w:colFirst' attribute in OOXML.
+ * @param {Object} attrs
+ * @returns {string|undefined}
+ */
+export const decode = (attrs) => {
+  return attrs.colFirst;
```
luccas-harbour: This is a bit nit-picky but I noticed that this `colFirst` and the `colLast` attributes are actually integers. You could use [these `parseInteger` and `integerToString` functions](https://github.com/Harbour-Enterprises/SuperDoc/blob/84d40f8a22d894d92445b74e3f858a597d42359b/packages/super-editor/src/core/super-converter/v3/handlers/utils.js#L438-L453) so that they can be handled as integers by the rest of the app.

packages/super-editor/src/core/super-converter/v3/handlers/w/perm-end/attributes/index.js
```diff
@@ -0,0 +1,8 @@
+// @ts-check
+import { attrConfig as wId } from './w-id.js';
```
luccas-harbour, line 2: I noticed that `w:permEnd` also has a [`w:displacedByCustomXml`](https://github.com/Harbour-Enterprises/superdoc-devtools/blob/1e23d8b9c6b640c7dce99a005ae3594b2fd37f8e/ooxml-oracle/specification/OfficeOpenXML-XMLSchema-Transitional/wml.xsd#L1722-L1725) attribute. I think we'll need to encode/decode it too

chittolinag: good catch, thanks!!

PR #1591, opened 2025-12-24
fix: can't add deletion track change when selected content has link/another track change

VladaHarbour:


harbournick:
@chittolinag fyi

harbournick (approved):
LGTM

PR #1594, opened 2025-12-25
fix: keep resolved comment anchor for export

VladaHarbour:


harbournick:
@chittolinag tagging just FYI since it has to do with comments.

PR #1606, opened 2025-12-29
fix: min header footer height

caio-pizzol:


VladaHarbour (approved):

harbournick:
After discussion with @caio-pizzol I'm proposing an alternate approach here that doesn't enforce a minimum height of 48px which deviates from expected behavior. It seems we need a deeper fix to the pagination loop in cases where top margin = header margin (so header in theory would have 0 available space) but MS word still uses header content as the source of truth and will adjust main content size accordingly even in this case.

PR #1611, opened 2025-12-30
[SD-1131] Collapse adjacent table-cell borders

afn:
* Fix bug with border size units: eliminate eighthPointsToPixels conversion in processInlineCellBorders (legacy-handle-table-cell-node.js)
* Eliminate redunant logic in renderTableRow: have a single case handle any combination of explicit borders & table borders
* Collapse borders by only painting right and bottom borders (omitting left and top) unless the cell is the first column or the first row (or treated as the first row, because it is first on the page)

**TODO:** Needs visual tests, which will be added once #1422 is done.

### DIFF
--- a/packages/layout-engine/measuring/dom/src/table-utils.ts
+++ b/packages/layout-engine/measuring/dom/src/table-utils.ts
@@ -1,0 +1,11 @@
+import { BorderSpec } from '@superdoc/contracts';
+
+export function getBorderWidth(border: BorderSpec | null | undefined): number {
+  if (!border || border.style === 'none') {
+    return 0;
+  }
+
+  // TODO: Is there a reason the default is different from the default border weight in defaultTableCellAttrs.borders?
+  // (This is the same default value used in applyBorder() in painters/dom/src/table/border-utils.ts)
+  return border.width ?? 1;
+}
--- a/packages/layout-engine/measuring/dom/src/table-utils.ts
+++ b/packages/layout-engine/measuring/dom/src/table-utils.ts
@@ -5,7 +5,5 @@ export function getBorderWidth(border: BorderSpec | null | undefined): number {
     return 0;
   }
 
-  // TODO: Is there a reason the default is different from the default border weight in defaultTableCellAttrs.borders?
-  // (This is the same default value used in applyBorder() in painters/dom/src/table/border-utils.ts)
   return border.width ?? 1;
 }
--- a/packages/layout-engine/painters/dom/src/table/renderTableRow.ts
+++ b/packages/layout-engine/painters/dom/src/table/renderTableRow.ts
@@ -9,7 +9,7 @@ import type {
   TableMeasure,
 } from '@superdoc/contracts';
 import { renderTableCell } from './renderTableCell.js';
-import { resolveTableCellBorders, borderValueToSpec } from './border-utils.js';
+import { borderValueToSpec } from './border-utils.js';
 import type { FragmentRenderContext } from '../renderer.js';
 
 type TableRowMeasure = TableMeasure['rows'][number];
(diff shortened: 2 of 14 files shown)


packages/layout-engine/painters/dom/src/table/renderTableRow.ts
```diff
@@ -242,8 +242,9 @@ export const renderTableRow = (deps: TableRowRenderDependencies): void => {
     let resolvedBorders;
     if (hasBordersAttribute && !hasExplicitBorders) {
       // Cell explicitly has borders={} meaning "no borders"
+      // TODO: Is this actually accurate? Shouldn't borders={} mean "don't override any borders"?
```
afn, line 246: @harbournick I wanted to check this with you: wouldn't "no borders" be represented as `{ top: { style: 'none', ... }, ... }` rather than `{}`?

packages/layout-engine/painters/dom/src/table/renderTableRow.ts
```diff
@@ -257,47 +258,16 @@ export const renderTableRow = (deps: TableRowRenderDependencies): void => {
...
+        bottom: cellBordersAttr?.bottom ?? borderValueToSpec(treatAsLastRow ? tableBorders?.bottom : undefined),
         // For left: use cell's if defined, otherwise use table's left for first col
-        left: cellBordersAttr.left ?? borderValueToSpec(isFirstCol ? tableBorders.left : tableBorders.insideV),
+        left: isFirstCol
+          ? (cellBordersAttr?.left ?? borderValueToSpec(isFirstCol ? tableBorders?.left : tableBorders?.insideV))
+          : undefined,
         // For right: use cell's if defined, otherwise use table's right for last col only
-        right: cellBordersAttr.right ?? borderValueToSpec(isLastCol ? tableBorders.right : undefined),
-      };
-    } else if (hasExplicitBorders) {
```
afn, line 274: @harbournick I wanted to call this out to make sure I didn't miss anything subtle in combining the three cases (tableBorders and hasExplicitBorders, only tableBorders, only hasExplicitBorders) into a single one.

packages/layout-engine/measuring/dom/src/table-utils.ts
```diff
@@ -0,0 +1,9 @@
+import { BorderSpec } from '@superdoc/contracts';
+
+export function getBorderWidth(border: BorderSpec | null | undefined): number {
```
afn, line 3: Does `measuring/dom` seem like the right place for this? It means that `painters/dom` is now dependent on `measuring/dom`, which it wasn't before. Open to other suggestions of which package to put it in.

afn (commented):
@harbournick Couple of questions/notes for you in the comments above

packages/layout-engine/painters/dom/src/table/renderTableRow.ts
```diff
@@ -257,47 +258,16 @@ export const renderTableRow = (deps: TableRowRenderDependencies): void => {
...
         // For top: use cell's if defined, otherwise use table's top border for first row OR continuation
-        top: cellBordersAttr.top ?? borderValueToSpec(treatAsFirstRow ? tableBorders.top : tableBorders.insideH),
+        top: treatAsFirstRow ? (cellBordersAttr?.top ?? borderValueToSpec(tableBorders?.top)) : undefined,
         // For bottom: use cell's if defined, otherwise use table's bottom border for last row OR before continuation
-        bottom: cellBordersAttr.bottom ?? borderValueToSpec(treatAsLastRow ? tableBorders.bottom : undefined),
+        bottom: cellBordersAttr?.bottom ?? borderValueToSpec(treatAsLastRow ? tableBorders?.bottom : undefined),
         // For left: use cell's if defined, otherwise use table's left for first col
-        left: cellBordersAttr.left ?? borderValueToSpec(isFirstCol ? tableBorders.left : tableBorders.insideV),
+        left: isFirstCol
+          ? (cellBordersAttr?.left ?? borderValueToSpec(isFirstCol ? tableBorders?.left : tableBorders?.insideV))
```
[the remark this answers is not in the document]
afn: 👎

afn:
Note that the failing visual tests are also failing in main (looks like _expected_ is missing text that's present in _actual_). To that end, see #1622 which changes the PR validation workflow to run on every commit, rather than only on open PRs.

PR #1613, opened 2025-12-30
fix: fallback to converter titlePg if no metadata [stable]

caio-pizzol:


harbournick:
Merging main into stable today so not needed

PR #1618, opened 2025-12-30
SD-58 (part 2) - Editable ranges

chittolinag:
Implements an edit permission plugin based on `w:permStart` and `w:permEnd`.

In short, every node inside a `w:permStart` and `w:permEnd` range, should be editable when in `viewing` mode. Also adds support for user-based permission.

On this example, only the highlighted section is editable by the current user:

Screenshot 2026-01-02 at 11 58 10

For more info about these tags, see [here](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.wordprocessing.permstart?view=openxml-3.0.1).


packages/super-editor/src/core/PresentationEditor.ts
```diff
@@ -6331,4 +6420,18 @@ export class PresentationEditor extends EventEmitter {
...
+
+  /**
+   * Determines whether the current viewing mode should block edits.
+   * When documentMode is viewing but the active editor has been toggled
+   * back to editable (e.g. permission ranges), we treat the view as editable.
+   */
+  #isViewLocked(): boolean {
+    if (this.#documentMode !== 'viewing') return false;
+    const hasPermissionOverride = Boolean(
+      (this.#editor as Editor & { storage?: Record<string, any> })?.storage?.permissionRanges?.hasAllowedRanges,
```
chittolinag: I personally don't love this typecast here, but the `storage` type is defined as `Record<string, unknown>` which requires type refinement before accessing (we have other examples of this on this file). One idea would be to change it to `Record<string, any>`, but that would make the code more dangerous. 

Ideally... we'd better define the `storage` type in a way that it's completely type-safe though.

packages/super-editor/src/core/header-footer/EditorOverlayManager.ts
```diff
@@ -94,9 +94,6 @@ export class EditorOverlayManager {
   /** Full-width border line element (MS Word style) */
   #borderLine: HTMLElement | null = null;
 
-  /** Dimming overlay element (for dimming body content during editing) */
-  #dimmingOverlay: HTMLElement | null = null;
-
   /**
```
chittolinag, line 97: Not used

packages/super-editor/src/core/header-footer/EditorOverlayManager.ts
```diff
@@ -466,18 +463,6 @@ export class EditorOverlayManager {
...
-   * @internal Reserved for future implementation of body dimming during header/footer editing.
-   */
-  // eslint-disable-next-line no-unused-private-class-members
-  #hideDimmingOverlay(): void {
-    if (this.#dimmingOverlay) {
-      this.#dimmingOverlay.remove();
-      this.#dimmingOverlay = null;
-    }
-  }
-
```
chittolinag, line 480: Not used

packages/layout-engine/painters/dom/src/styles.ts
```diff
@@ -87,6 +87,7 @@ export const lineStyles = (lineHeight: number): Partial<CSSStyleDeclaration> =>
   // provides defense-in-depth against any remaining sub-pixel rendering
   // differences between measurement and display.
   overflow: 'visible',
+  zIndex: '10',
```
chittolinag, line 90: at this point this is expected to be a str.

packages/super-editor/src/core/PresentationEditor.ts
```diff
@@ -742,6 +743,17 @@ export class PresentationEditor extends EventEmitter {
...
     this.#selectionSync.on('render', () => this.#updateSelection());
+    this.#selectionSync.on('render', () => this.#updatePermissionOverlay());
+
+    this.#permissionOverlay = doc.createElement('div');
+    this.#permissionOverlay.className = 'presentation-editor__permission-overlay';
+    Object.assign(this.#permissionOverlay.style, {
+      position: 'absolute',
+      inset: '0',
+      pointerEvents: 'none',
+      zIndex: '5',
```
chittolinag, line 775: min value of `zIndex` that ensures the permission overlay won't make the text obfuscated.

packages/super-editor/src/extensions/permission-ranges/permission-ranges.js
```diff
@@ -0,0 +1,382 @@
...
+ * @returns {{ ranges: Array<{ id: string, from: number, to: number }>, hasAllowedRanges: boolean }}
+ */
+const buildPermissionState = (doc, allowedIdentifiers = EMPTY_IDENTIFIER_SET) => {
+  const ranges = [];
+  /** @type {Map<string, { from: number, attrs: any }>} */
+  const openRanges = new Map();
+
+  doc.descendants((node, pos) => {
+    if (node.type?.name === 'permStart') {
+      const id = getPermissionNodeId(node, pos, 'permStart');
```
chittolinag, line 58: Check if this can be removed (it probably can)

packages/super-editor/src/extensions/permission-ranges/permission-ranges.js
```diff
@@ -0,0 +1,382 @@
...
+        // Filters transactions to ensure only allowed edits are applied.
+        filterTransaction(tr, state) {
+          if (!tr.docChanged) return true;
+          if (!editor || editor.options.documentMode !== 'viewing') return true;
+          const pluginState = PERMISSION_PLUGIN_KEY.getState(state);
+          if (!pluginState?.hasAllowedRanges) {
+            return true;
+          }
+          const changedRanges = collectChangedRanges(tr);
+          if (!isSelectionAllowed(state, pluginState.ranges)) {
```
chittolinag: Check if this can be removed (probably yes)

luccas-harbour (approved):

PR #1653, opened 2026-01-06
SD-1373 - Bug: NBSP being stripped from document even whey they should be present

chittolinag:
For text nodes that only contain non-breaking spaces (`U+00A0`), the spaces are being removed incorrectly.

**Behavior before:**

In Word:

Screenshot 2026-01-06 at 10 38 29

In SuperDoc: 
Screenshot 2026-01-06 at 10 38 19

### DIFF
--- a/packages/layout-engine/contracts/src/index.ts
+++ b/packages/layout-engine/contracts/src/index.ts
@@ -1011,6 +1011,36 @@ export type DropCapDescriptor = {
 };
 
 /**
+ * Marker metadata for word-layout lists.
+ * Contains styling and positioning information for list markers.
+ */
+export type WordLayoutMarker = {
+  /** The text content of the marker (e.g., "1.", "a)", "•"). */
+  markerText?: string;
+  /** Horizontal alignment of the marker within its allocated space. */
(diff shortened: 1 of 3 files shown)


packages/super-editor/src/core/super-converter/v3/handlers/w/t/t-translator.js
```diff
@@ -72,7 +72,8 @@ const encode = (params, encodedAttrs = {}) => {
     // drop the node entirely. This prevents creating empty/whitespace text nodes that ProseMirror
     // may treat as invalid. The placeholder was only needed to prevent xml-js from dropping
     // the node during parsing - but if xml:space doesn't require preservation, we should drop it.
-    if (xmlSpace !== 'preserve' && typeof text === 'string' && !text.trim()) {
+    const isWhitespaceOnly = /^[ \t\n\r]*$/.test(text);
```
chittolinag, line 75: For NBSPs, at this point we have it as expected, but since we were using the `trim()` function (which removes NBSPs), we were completely skipping this node and returning `null`, hence the space was not rendered on the document.

PR #1662, opened 2026-01-06
fix: prevent crash when deleting tracked changes with empty nodes [stable]

caio-pizzol:


harbournick:
merged to main and will move main->stable today

PR #1664, opened 2026-01-06
SD-1328 - fix: images behind text not fully stretched

chittolinag:
- Anchored images with "behind text" were being clamped to the column width during measurement and forced into positive offsets in the editor. It seems like Word expects those wrapNone/behindDoc images to ignore the column constraints, stretch to their real size, and use negative offsets (if provided) so they can extend into margins while remaining behind the text.
- The fix updates measureImageBlock to bypass width constraints when `wrap.type === 'None'`. On the rendering side, we now allow negative top offsets for those anchors, apply them via top when absolutely positioned, and strip the default max-width:100%.

### DIFF
--- a/packages/super-editor/src/extensions/image/image.js
+++ b/packages/super-editor/src/extensions/image/image.js
@@ -511,6 +511,8 @@ export const Image = Node.create({
       const appliedTopViaStyle = isAbsolutelyPositioned && allowNegativeTopOffset && !relativeFromMarginV;
       if (appliedTopViaStyle) {
         style += `top: ${top}px;`;
+      // Don't apply vertical offset as margin-top for images positioned relative to margin
+      // as this causes double-counting of the offset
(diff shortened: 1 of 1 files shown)


packages/layout-engine/layout-engine/src/index.ts
```diff
@@ -741,257 +741,6 @@ export function layoutDocument(blocks: FlowBlock[], measures: Measure[], options
...
-          return { forcePageBreak: false, forceMidPageRegion: true };
-        }
-        // For continuous without column changes, schedule for next page
-        if (block.columns) {
-          pendingColumns = { count: block.columns.count, gap: block.columns.gap };
-        }
-        return { forcePageBreak: false, forceMidPageRegion: false };
-    }
-  };
-
```
chittolinag, line 994: Not used

PR #1689, opened 2026-01-08
SD-1405 - fix: image being downsized when it should be stretched

chittolinag:
We were wrongly downsizing the image even though the image was supposed to be stretched (and pushing it to the left even when block is positioned relatively to the page). 

**Before:**

Screenshot 2026-01-08 at 17 10 13

**After:**
Screenshot 2026-01-08 at 17 10 52

### DIFF
--- a/packages/layout-engine/measuring/dom/src/index.test.ts
+++ b/packages/layout-engine/measuring/dom/src/index.test.ts
@@ -2505,16 +2505,17 @@ describe('measureBlock', () => {
         expect(measure.width).toBe(100);
       });
 
-      it('bypasses maxHeight when objectFit is set to fill', async () => {
+      it('bypasses maxHeight when objectFit is set to cover', async () => {
         const block: FlowBlock = {
           kind: 'image',
-          id: 'img-fill-fit',
+          id: 'img-cover-fit',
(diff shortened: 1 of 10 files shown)


packages/layout-engine/layout-engine/src/floating-objects.ts
```diff
@@ -354,17 +354,11 @@ export function computeAnchorX(
   let baseX: number;
   let availableWidth: number;
   if (relativeFrom === 'page') {
-    // Word's page-relative origin is the physical page edge. For single-column
-    // documents, authors typically expect column-relative behavior; if we detect
-    // a single column layout, treat page-relative as margin-relative to align
-    // with practical expectations and DOCX samples.
-    if (columns.count === 1) {
-      baseX = contentLeft;
-      availableWidth = contentWidth;
```
chittolinag, line 363: @harbournick I am not entirely sure whether this is still needed. This piece of code was pushing the image to the left, even though the image was positioned relatively to the page edge. Why exactly do we need this when there's only a single column? Please let me know if you think this is going to break something.

harbournick (approved):
LGTM

PR #1725, opened 2026-01-13
fix: popover menu renders at correct position in presentation mode

caio-pizzol:


caio-pizzol:
@harbournick this is still in draft - but do you think this is the "correct" path for fixing Popover Extension? (when @ mentioning someone)

PR #1738, opened 2026-01-14
feat: annotation rendering

artem-harbour:
Linear - SD-1354.

- Update annotation when content/attributes are updated.
- Render image/signature/html annotation.
- Style fixes.


harbournick (approved):
LGTM

PR #1742, opened 2026-01-14
fix: anchor images in table cells

VladaHarbour:


packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -3392,8 +3371,8 @@ export class DomPainter {
     let contentContainer: HTMLElement = groupEl;
```
VladaHarbour, line 3392: @harbournick removed this scale because we scale shapes when process it on import. This caused double scaling

packages/layout-engine/painters/dom/src/utils/anchor-helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+      continue;
+    }
+    merged.push(current);
+    current = next;
+  }
+  merged.push(current);
+  return merged;
+};
+
+export const applySquareWrapExclusionsToLines = (
```
VladaHarbour, line 25: @harbournick I'm not sure this is a good way to apply additional indent to the line since it cause this on long text

Image

PR #1746, opened 2026-01-15
fix: guarding against component init when ref is null

mattConnHarbour:
On quick component mount/unmount (like with HMR), the element ref may be null, but the init function may still be waiting to fire using a null ref. This one-liner guards against that with an early return.


harbournick (approved):
LGTM

PR #1752, opened 2026-01-15
fix: annotation issues

artem-harbour:
- Fix drag drop of annotations with the same fieldId.
- Fix merging annotation runs (rendering annotations next to each other).


harbournick (approved):
LGTM

PR #1759, opened 2026-01-16
chore: cleaning unused functions

andrii-harbour:


packages/super-editor/src/core/Editor.ts
```diff
@@ -2056,13 +2043,6 @@ export class Editor extends EventEmitter<EditorEventMap> {
     this.#dispatchTransaction(tr);
   }
 
-  /**
-   * Get document identifier for telemetry (async - may generate hash)
-   */
-  async getDocumentIdentifier(): Promise<string | null> {
```
harbournick, line 2064: let's not remove this one, part of the public API which would make this a breaking change

andrii-harbour: done

PR #1761, opened 2026-01-16
feat: add insertTrackedChange programmatic shortcut

harbournick:
- Add insertTrackedChange programmatic suggestion shortcut with user override

### DIFF
--- a/packages/super-editor/src/extensions/track-changes/track-changes.js
+++ b/packages/super-editor/src/extensions/track-changes/track-changes.js
@@ -248,6 +248,13 @@ export const TrackChanges = Extension.create({
         ({ state, dispatch, editor }) => {
           const { from = state.selection.from, to = state.selection.to, text = '', user, comment } = options;
 
+          // Validate bounds to prevent RangeError
+          const docSize = state.doc.content.size;
+          if (from < 0 || to > docSize || from > to) {
+            console.warn('insertTrackedChange: invalid range', { from, to, docSize });
+            return false;
+          }
+
           // Check if there's actually a change to make
           const originalText = state.doc.textBetween(from, to, '', '');
           if (originalText === text) {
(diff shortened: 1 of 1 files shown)


packages/super-editor/src/extensions/track-changes/track-changes.js
```diff
@@ -240,6 +243,114 @@ export const TrackChanges = Extension.create({
           return commands.rejectTrackedChangesBetween(from, to);
         },
 
+      insertTrackedChange:
+        (options = {}) =>
+        ({ state, dispatch, editor }) => {
+          const { from = state.selection.from, to = state.selection.to, text = '', user, comment } = options;
```
caio-pizzol, line 249: Shouldn't we validate from/to bounds before using them? Invalid positions may throw RangeError.

Check how it was implemented in `applyPatch.ts`

packages/super-editor/src/extensions/track-changes/track-changes.js
```diff
@@ -240,6 +243,114 @@ export const TrackChanges = Extension.create({
...
+          const originalText = state.doc.textBetween(from, to, '', '');
+          if (originalText === text) {
+            return false;
+          }
+
+          if (!dispatch) {
+            return true;
+          }
+
+          const resolvedUser = user ?? editor?.options?.user ?? {};
```
caio-pizzol, line 268: Consider early return with warning if `resolvedUser` lacks name/email - marks will be created with undefined author which may cause display issues

packages/super-editor/src/extensions/track-changes/track-changes.js
harbournick: I like it, good call. Fixed in [eb3481b](https://github.com/superdoc-dev/superdoc/pull/1761/commits/eb3481ba03dbeb7d0388599be67f4d60f657f227)

packages/super-editor/src/extensions/track-changes/track-changes.js
harbournick: Fixed in [eb3481b](https://github.com/superdoc-dev/superdoc/pull/1761/commits/eb3481ba03dbeb7d0388599be67f4d60f657f227)

PR #1769, opened 2026-01-16
refactor(presentation-editor): consolidate input handling into dedicated managers

harbournick:
Refactor presentation editor input handling into dedicated managers


artem-harbour:
@harbournick - check my comment in linear (SD-1436).

PR #1777, opened 2026-01-19
feat(comments): add scrollToComment API

financialvice:
Exposes a public scrollToComment method for custom comment UIs.\n\nTests: pnpm -C packages/superdoc test -- SuperDoc.test.js

### DIFF
--- a/packages/superdoc/src/core/SuperDoc.js
+++ b/packages/superdoc/src/core/SuperDoc.js
@@ -668,6 +668,27 @@ export class SuperDoc extends EventEmitter {
   }
 
   /**
+   * Scroll the document to a given comment or tracked change by thread id.
+   *
+   * @param {string} commentId The comment or tracked change id
+   * @param {{ behavior?: ScrollBehavior, block?: ScrollLogicalPosition }} [options]
+   * @returns {boolean} Whether a matching element was found
+   */
+  scrollToComment(commentId, options = {}) {
+    const commentsConfig = this.config?.modules?.comments;
+    if (!commentsConfig || commentsConfig === false) return false;
+    if (!commentId || typeof commentId !== 'string') return false;
+
+    const element = document.querySelector(`[data-thread-id="${commentId}"]`);
+    if (!element) return false;
+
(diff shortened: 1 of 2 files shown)


packages/superdoc/src/core/SuperDoc.js
```diff
@@ -667,6 +667,27 @@ export class SuperDoc extends EventEmitter {
...
+   * @param {{ behavior?: ScrollBehavior, block?: ScrollLogicalPosition }} [options]
+   * @returns {boolean} Whether a matching element was found
+   */
+  scrollToComment(commentId, options = {}) {
+    const commentsConfig = this.config?.modules?.comments;
+    if (!commentsConfig || commentsConfig === false) return false;
+    if (!commentId || typeof commentId !== 'string') return false;
+
+    const element = document.querySelector(`[data-thread-id="${commentId}"]`);
+    if (!element) return false;
```
caio-pizzol, line 683: if `commentId` contains characters like `"`, `\`, or `]`, this will throw a `DOMException` instead of returning `false`. the codebase already uses `CSS.escape` for this (e.g. `SuperEditor.vue:803`), so worth applying here too.

```suggestion
    const element = document.querySelector(`[data-thread-id="${CSS.escape(commentId)}"]`);
```

packages/superdoc/src/core/SuperDoc.test.js
```diff
@@ -252,6 +252,30 @@ describe('SuperDoc core', () => {
     expect(instance.user).toEqual(expect.objectContaining({ name: 'Default SuperDoc user', email: null }));
   });
 
+  it('scrolls to a comment and sets it active', async () => {
+    const { commentsStore } = createAppHarness();
```
caio-pizzol, line 256: the tests only cover the happy path. a case where the element isn't in the DOM (expecting `false`) would catch the most likely real-world failure mode — worth adding?

caio-pizzol:
@financialvice thanks for this! We're moving your work to a new PR so we can wrap up the remaining changes and get it merged. Closing in favor of #2440

PR #1790, opened 2026-01-20
SD-1525 - comment export with empty array

mattConnHarbour:
- Calling exportDocx with empty comments array should now export docx without any comments.
- Previously, was exporting with previous list of comments.
example:
```
const newFile = editor.exportDocx({ comments: [], commentsType: 'external'  });
```

### DIFF
--- a/packages/super-editor/src/core/Editor.ts
+++ b/packages/super-editor/src/core/Editor.ts
@@ -2543,28 +2543,28 @@ export class Editor extends EventEmitter<EditorEventMap> {
         updatedDocs['word/_rels/footnotes.xml.rels'] = String(footnotesRelsXml);
       }
 
-      // Check if comment files exist in convertedXml (they're removed when cleaning or empty array)
+      // Serialize comment files if they exist, or mark them for removal from the zip
+      const commentFileKeys = [
+        'word/comments.xml',
+        'word/commentsExtended.xml',
+        'word/commentsExtensible.xml',
+        'word/commentsIds.xml',
+      ] as const;
+
       const commentsFile = this.converter.convertedXml['word/comments.xml'];
       if (commentsFile?.elements?.[0]) {
-        const commentsXml = this.converter.schemaToXml(commentsFile.elements[0]);
-        updatedDocs['word/comments.xml'] = String(commentsXml);
+        updatedDocs['word/comments.xml'] = String(this.converter.schemaToXml(commentsFile.elements[0]));
 
-        const commentsExtended = this.converter.convertedXml['word/commentsExtended.xml'];
-        if (commentsExtended?.elements?.[0]) {
-          const commentsExtendedXml = this.converter.schemaToXml(commentsExtended.elements[0]);
-          updatedDocs['word/commentsExtended.xml'] = String(commentsExtendedXml);
+        for (const key of commentFileKeys.slice(1)) {
+          const file = this.converter.convertedXml[key];
+          if (file?.elements?.[0]) {
+            updatedDocs[key] = String(this.converter.schemaToXml(file.elements[0]));
+          }
         }
-
-        const commentsExtensible = this.converter.convertedXml['word/commentsExtensible.xml'];
-        if (commentsExtensible?.elements?.[0]) {
-          const commentsExtensibleXml = this.converter.schemaToXml(commentsExtensible.elements[0]);
-          updatedDocs['word/commentsExtensible.xml'] = String(commentsExtensibleXml);
-        }
-
-        const commentsIds = this.converter.convertedXml['word/commentsIds.xml'];
-        if (commentsIds?.elements?.[0]) {
-          const commentsIdsXml = this.converter.schemaToXml(commentsIds.elements[0]);
-          updatedDocs['word/commentsIds.xml'] = String(commentsIdsXml);
+      } else {
+        // Comments were removed — tell DocxZipper to strip these files from the zip
+        for (const key of commentFileKeys) {
+          updatedDocs[key] = null as unknown as string;
         }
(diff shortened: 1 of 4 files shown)


packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -998,17 +998,16 @@ class SuperConverter {
...
-    }
+    const { documentXml, relationships } = this.#prepareCommentsXmlFilesForExport({
+      defs: params.exportedCommentDefs,
+      exportType: commentsExportType,
+      commentsWithParaIds,
+    });
+    updatedXml = { ...documentXml };
+    commentsRels = relationships;
 
     this.convertedXml = { ...this.convertedXml, ...updatedXml };
```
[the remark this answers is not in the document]
caio-pizzol: @mattConnHarbour check comment please

packages/super-editor/src/core/super-converter/v2/exporter/commentsExporter.js
```diff
@@ -354,7 +354,8 @@ export const prepareCommentsXmlFilesForExport = ({
 }) => {
   const relationships = [];
 
-  if (exportType === 'clean') {
+  // Remove comment files if explicitly cleaning OR if no comments to export (empty array)
```
caio-pizzol: This is a breaking change. Consider adding a migration note or separate option like `preserveComments: true` for the old behavior.

Also make sure to add covering tests on `commentsExporter.test.js` for `prepareCommentsXmlFilesForExport`

packages/super-editor/src/core/super-converter/SuperConverter.js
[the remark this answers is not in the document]
mattConnHarbour: addressed

packages/super-editor/src/core/super-converter/v2/exporter/commentsExporter.js
mattConnHarbour: I added in flag `replaceExistingComments`, which is false by default.
Adding in tests next!

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -1019,20 +1020,34 @@ class SuperConverter {
     );
 
     // Update content types and comments files as needed
+    // Empty array preserves existing comments unless preserveCommentsOnEmpty is true
```
caio-pizzol: Comment is backwards.
 
 Says "unless preserveCommentsOnEmpty is true" but should be "false".
 `true` = preserve
 `false` = remove

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -1019,20 +1020,34 @@ class SuperConverter {
...
     let updatedXml = { ...this.convertedXml };
```
caio-pizzol: `let updatedXml = { ...this.convertedXml }` is immediately overwritten on [line 1031](https://github.com/superdoc-dev/superdoc/pull/1790/files#diff-05992586622d81798e474614e76bb863dd9c4549600cb785e7f76322c8eca44eR1031)

packages/super-editor/src/core/Editor.ts
```diff
@@ -2537,8 +2543,10 @@ export class Editor extends EventEmitter<EditorEventMap> {
         updatedDocs['word/_rels/footnotes.xml.rels'] = String(footnotesRelsXml);
       }
 
-      if (preparedComments.length) {
-        const commentsXml = this.converter.schemaToXml(this.converter.convertedXml['word/comments.xml'].elements[0]);
+      // Check if comment files exist in convertedXml (they're removed when cleaning or empty array)
```
caio-pizzol: skipping comment files in `updatedDocs` isn't enough — `DocxZipper` keeps the originals from the source file unless you explicitly set them to `null`. the old comments survive in the exported docx. fix: set `updatedDocs['word/comments.xml'] = null` (and the other 3) when they should be gone. `DocxZipper` already handles `null` as "remove from zip" (line 330).

packages/super-editor/src/core/super-converter/SuperConverter.js
caio-pizzol: still open from last round — comment says "true" but means "false". and `let updatedXml = { ...this.convertedXml }` is overwritten two lines later, so the initial copy is wasted.

```suggestion
    // Empty array preserves existing comments unless preserveCommentsOnEmpty is false
    let updatedXml;
```

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -1019,20 +1020,34 @@ class SuperConverter {
...
+    const { documentXml, relationships } = this.#prepareCommentsXmlFilesForExport({
+      defs: params.exportedCommentDefs,
+      exportType: commentsExportType,
+      commentsWithParaIds,
+      preserveCommentsOnEmpty,
+    });
+    updatedXml = { ...documentXml };
+    commentsRels = relationships;
 
     this.convertedXml = { ...this.convertedXml, ...updatedXml };
```
caio-pizzol: the spread here puts the old comment keys back — so the deletion that happened inside `prepareCommentsXmlFilesForExport` gets undone. only the `delete` loop below (lines 1041-1048) actually removes them. same logic in two places, one of them doesn't work — worth picking one owner?

caio-pizzol:
hey @mattConnHarbour - this one's already been fixed on main in #2123. that PR made it so passing an empty comments array always removes comments, which is exactly what SD-1525 needed.

PR #1793, opened 2026-01-20
feat: view options print and web

caio-pizzol:


packages/super-editor/src/components/SuperEditor.vue
```diff
@@ -1141,6 +1155,35 @@ onBeforeUnmount(() => {
   flex-direction: column;
 }
 
+/* Web layout mode (OOXML ST_View 'web'): content reflows to fit container */
+.super-editor-container.web-layout {
```
caio-pizzol, line 1159: @harbournick lmk if this should be removed

packages/superdoc/src/core/SuperDoc.js
```diff
@@ -142,6 +147,14 @@ export class SuperDoc extends EventEmitter {
       this.config.trackChanges.visible = false;
     }
 
+    // Web layout mode requires layout engine to be disabled (content reflows vs pagination)
+    if (this.config.viewOptions?.layout === 'web' && this.config.useLayoutEngine) {
+      console.warn(
+        '[SuperDoc] Web layout mode requires useLayoutEngine: false. Automatically disabling layout engine.',
+      );
+      this.config.useLayoutEngine = false;
```
caio-pizzol, line 155: layout engine disabled - eventually we should support web view on LE as well

packages/superdoc/src/dev/components/SuperdocDev.vue
```diff
@@ -1249,4 +1285,80 @@ if (scrollTestMode.value) {
   line-height: 1.5;
   color: #fde68a;
 }
+
+/* Mobile responsive styles */
+@media screen and (max-width: 768px) {
```
caio-pizzol, line 1290: make dev app responsive so we can test "web" view

packages/super-editor/src/components/SuperEditor.vue
harbournick: let's leave it - I added a note to SD-1531 to add this to theming as well so we don't forget to upgrade this then

PR #1808, opened 2026-01-22
SD-1588 - Added `setDocumentProtection` command

chittolinag:
Introduced a new `setProtectionMode` command that lets SuperDoc callers toggle DOCX protection settings directly from the editor. The command normalizes supported modes (no protection, revisions-only, comments-only, form-fields-only, and read-only), patches the `w:documentProtection` node inside `word/settings.xml`.

 Examples:

```ts
editor.commands.setProtectionMode('allowOnlyRevisions'); // tracked-changes only
editor.commands.setProtectionMode('allowOnlyComments'); // comment-only edits
editor.commands.setProtectionMode('allowOnlyReading'); // read-only document
editor.commands.setProtectionMode('allowOnlyFormFields'); // form fields only
editor.commands.setProtectionMode('noProtection'); // remove protection
```

### DIFF
--- a/packages/super-editor/src/core/commands/setProtectionMode.js
+++ b/packages/super-editor/src/core/commands/setProtectionMode.js
@@ -19,9 +19,8 @@ const DEFAULT_MODE = 'noProtection';
  * @param {string} mode
  * @returns {keyof typeof PROTECTION_VALUE_MAP}
  */
-function normalizeMode(mode = '') {
+function normalizeMode(mode) {
   const normalized = typeof mode === 'string' ? mode.trim() : '';
-  console.log('normalized', normalized);
   return /** @type {keyof typeof PROTECTION_VALUE_MAP} */ (normalized || DEFAULT_MODE);
 }
 
@@ -46,14 +45,13 @@ function createDocProtectionNode(editValue) {
  * @param {'noProtection' | 'allowOnlyRevisions' | 'allowOnlyComments' | 'allowOnlyFormFields' | 'allowOnlyReading'} mode
  * @returns {import('./types').Command}
  */
-export const setProtectionMode = (mode = 'noProtection') => {
+export const setProtectionMode = (mode) => {
   return ({ editor }) => {
+    if (!mode || typeof mode !== 'string') return false;
     const convertedXml = editor?.converter?.convertedXml;
     if (!convertedXml) return false;
 
     const normalizedMode = normalizeMode(mode);
-    console.log('normalizedMode', normalizedMode);
-    console.log('PROTECTION_VALUE_MAP', PROTECTION_VALUE_MAP, normalizedMode in PROTECTION_VALUE_MAP);
     if (!(normalizedMode in PROTECTION_VALUE_MAP)) return false;
     const settingsXml = convertedXml[SETTINGS_PATH];
     const settingsRoot = settingsXml?.elements?.[0];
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/core/commands/setProtectionMode.js
```diff
@@ -0,0 +1,87 @@
...
+const DEFAULT_MODE = 'noProtection';
+
+/**
+ * Normalize the caller provided document protection mode.
+ * @param {string} mode
+ * @returns {keyof typeof PROTECTION_VALUE_MAP}
+ */
+function normalizeMode(mode = '') {
+  const normalized = typeof mode === 'string' ? mode.trim() : '';
+  console.log('normalized', normalized);
```
caio-pizzol: Looks like that's debug - remove?

packages/super-editor/src/core/commands/setProtectionMode.js
```diff
@@ -0,0 +1,87 @@
...
+    const elements = updatedRoot.elements;
+    const existingIndex = elements.findIndex((node) => node?.name === DOC_PROTECTION_NODE);
+    if (existingIndex !== -1) {
+      elements.splice(existingIndex, 1);
+    }
+
+    const mappedValue = PROTECTION_VALUE_MAP[normalizedMode];
+    if (mappedValue) {
+      const protectionNode = createDocProtectionNode(mappedValue);
+      const insertIndex = existingIndex >= 0 ? existingIndex : 0;
```
caio-pizzol, line 77: When no existing protection node exists, this inserts at index 0 which puts it before `w:zoom`.

OOXML spec says `documentProtection` should come at position 35, after zoom/trackRevisions/etc which might cause schema validation issues.

packages/super-editor/src/core/commands/setProtectionMode.js
```diff
@@ -0,0 +1,87 @@
...
+
+const DEFAULT_MODE = 'noProtection';
+
+/**
+ * Normalize the caller provided document protection mode.
+ * @param {string} mode
+ * @returns {keyof typeof PROTECTION_VALUE_MAP}
+ */
+function normalizeMode(mode = '') {
+  const normalized = typeof mode === 'string' ? mode.trim() : '';
```
caio-pizzol, line 23: Passing `null/undefined` here becomes empty string, which then falls back to `noProtection` via the default.

Means invalid input silently removes protection instead of erroring.

Maybe we should return false for non-strings instead?

packages/super-editor/src/core/commands/setProtectionMode.test.js
```diff
@@ -0,0 +1,109 @@
...
+  it('supports AllowOnlyFormFields casing and replaces existing node', () => {
+    const base = createBaseSettings();
+    base.elements[0].elements.unshift({
+      type: 'element',
+      name: 'w:documentProtection',
+      attributes: { 'w:edit': 'comments', 'w:enforcement': '1' },
+    });
+
+    const editor = buildEditor(base);
+    const command = setProtectionMode('allowOnlyFormFields');
```
caio-pizzol: Test name says it tests casing but only uses one case

chittolinag: oops, outdated test. thanks!

packages/super-editor/src/core/commands/setProtectionMode.js
chittolinag: Interesting! Can you please send the link (or relevant documentation) where you found this?

caio-pizzol: ECMA-376 ref - Section 17.15.1 (p.960) - Document Settings.

Since `CT_Settings` uses `xsd:sequence` (not `xsd:all`), the order is mandatory. Inserting `documentProtection` at index 0 (before `writeProtection`) would violate the sequence constraint and produce schema-invalid XML.

PR #1813, opened 2026-01-22
fix: align floating comments by page

pallavibakale:
Refactored the comments layout to be page-aware: comments are aligned to their corresponding document pages, positioned relative to the page container, and rendered with independent per-page scrolling to avoid overlaps and maintain stable layouts.

- Align floating comments per page using PresentationEditor page bounds and per-page containers.
- Normalize comment positions on each page to avoid overlap and render with scoped scrolling.
- Expand tracked-change and regular comment bodies when active; keep floating comments column full height.
Screenshot

### DIFF
--- a/packages/superdoc/src/SuperDoc.vue
+++ b/packages/superdoc/src/SuperDoc.vue
@@ -1001,6 +1001,7 @@ watch(getFloatingComments, () => {
 .floating-comments {
   min-width: 300px;
   width: 300px;
+  height: 100%;
 }
 
 .superdoc__layers {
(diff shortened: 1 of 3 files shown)


caio-pizzol:
@pallavibakale thanks for putting this together! the page-aware layout idea was solid, and the collision avoidance per page was a good direction.

however, #2195 landed on main since this was opened and it completely rewrote `FloatingComments.vue` and reworked `CommentDialog.vue` — covering all the same concerns (collision avoidance, comment body truncation, active comment scrolling) with a more robust approach (virtualization, bidirectional collision resolution, height caching). this PR would be 100% merge conflicts at this point.

closing this one out — the problems it set out to fix are resolved. thanks again for the work here!

PR #1822, opened 2026-01-23
feat: context-sensitive perm tags creation

chittolinag:
Implements a context-aware creation of `permStart` and `permEnd` tags. We're reusing the same logic that we use in the `passthroughNodeImporter` file, which detects the context based on the node's path.

If the context accepts only `block`, we'll create a block permission tag. Otherwise we'll create the inline version.


packages/super-editor/src/core/super-converter/helpers/node-context.js
```diff
@@ -0,0 +1,53 @@
...
+    if (INLINE_NODE_NAMES.has(ancestorName) || INLINE_PARENT_NAMES.has(ancestorName)) {
+      return true;
+    }
+    if (BLOCK_BOUNDARY_NAMES.has(ancestorName)) {
+      return false;
+    }
+  }
+
+  return false;
+};
```
chittolinag, line 53: Just moved this out of `passthroughNodeImporter` so it can be reused.

PR #1825, opened 2026-01-23
SD-1621 - add table insertion example

mattConnHarbour:


mattConnHarbour:
@caio-pizzol I will address the codex remarks so we can merge/close this

caio-pizzol:
@mattConnHarbour I don't think we should add an example about this tbh - it is a very specific use case and more example to manage.

PR #1847, opened 2026-01-27
fix: annotation and interaction issues

artem-harbour:
Linear: SD-1609.

This PR fixes several issues:
- Insert text before an annotation if it's at the beginning of a line.
- Render caret after annotation if there is no text.
- Incorrect emit of annotation delete event.


harbournick (approved):
LGTM

PR #1848, opened 2026-01-27
fix: patch broken numbering definitions

luccas-harbour:
- Implements a function that will patch a document's numbering definition to make sure it's valid

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -130,23 +130,10 @@ jobs:
         run: |
           pnpm run build
         
-      - name: Cache Playwright browsers
-        uses: actions/cache@v4
-        id: playwright-cache
-        with:
-          path: ~/.cache/ms-playwright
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/core/super-converter/v2/importer/patchNumberingDefinitions.test.js
```diff
@@ -0,0 +1,73 @@
...
+  elements: [
+    {
+      type: 'element',
+      name: 'w:abstractNumId',
+      attributes: { 'w:val': String(abstractId) },
+    },
+  ],
+});
+
+describe('patchNumberingDefinitions', () => {
```
caio-pizzol, line 43: minor: should we add a no-op test case (all abstract IDs exist) and one with multiple missing IDs?

luccas-harbour: good call! done

harbournick (approved):
LGTM

PR #1850, opened 2026-01-27
SD-1460 - fix: don't drop track format on run nodes

chittolinag:
Tracked formatting changes were being dropped after export.

The reason is that we were not re-generating the `rPrChange` element before decoding the run node. We do encode them correctly, but we were just missing doing it in the opposite direction.


packages/super-editor/src/core/super-converter/v3/handlers/helpers.js
```diff
@@ -4,9 +4,17 @@ import { processOutputMarks } from '@converter/exporter.js';
  * @param {Array} marks SD node marks.
  * @returns {Object|undefined} Properties element for trackFormat change or undefined.
  */
-export const createTrackStyleMark = (marks) => {
+export const decodeTrackFormatMark = (marks) => {
```
chittolinag, line 7: Renamed this - the current name seems counter-intuitive, it's not creating a mark, but instead decoding it back to OOXML.

Thoughts?

packages/super-editor/src/core/super-converter/v3/handlers/w/r/r-translator.js
```diff
@@ -176,6 +179,21 @@ const decode = (params, decodedAttrs = {}) => {
     node: { attrs: { runProperties: runProperties } },
   });
 
+  const trackFormatMark = trackingMarksByType.get(TrackFormatMarkName);
+  if (trackFormatMark) {
+    const rPrChangeElement = decodeTrackFormatMark([trackFormatMark]);
```
chittolinag: @luccas-harbour I see you've created this function called `decodeRPrFromMarks` on `super-converter/styles.js` a while ago. I guess this is what we should be using here?

packages/super-editor/src/core/super-converter/v3/handlers/w/r/r-translator.js
```diff
@@ -176,6 +179,21 @@ const decode = (params, decodedAttrs = {}) => {
...
+      if (!runPropertiesElement) {
+        runPropertiesElement = { name: 'w:rPr', elements: [] };
+      }
+      if (!Array.isArray(runPropertiesElement.elements)) {
+        runPropertiesElement.elements = [];
+      }
+      runPropertiesElement.elements = runPropertiesElement.elements.filter((el) => el?.name !== 'w:rPrChange');
+      runPropertiesElement.elements.push(rPrChangeElement);
+    }
+  }
```
luccas-harbour, line 263: Just a small suggestion. Maybe this logic could go in a separate function at the bottom of the file just so we keep the `decode` function smaller (it's already quite long tbh)?

packages/super-editor/src/core/super-converter/v3/handlers/w/r/r-translator.js
luccas-harbour: You are right @chittolinag. You can use the existing function, that would be safer as we wouldn't have the same logic in multiple places

luccas-harbour: Good catch

chittolinag: It turns out that that function wouldn't really work for this case here, it's used in some other places and would require some refactoring to make it work because it doesn't serve the exact same purpose (and we risk breaking the existing behavior). I decided to simply create a separate function for this use case.

caio-pizzol (commented):
approach looks right — adding `w:rPrChange` generation at the `r-translator.decode` level is the correct place for this. left one inline comment worth addressing before merge: `trackFormat` marks aren't excluded from `preservedMarks` the same way `trackInsert` and `trackDelete` are, which is inconsistent with the function's intent. the fix is a one-liner. everything else looks good.

packages/super-editor/src/core/super-converter/v3/handlers/w/r/helpers/track-change-helpers.js
```diff
@@ -24,8 +24,9 @@ const cloneRuns = (runs = []) => runs.map((run) => cloneNode(run));
 
 export const prepareRunTrackingContext = (node = {}) => {
   const marks = Array.isArray(node.marks) ? node.marks : [];
-  const trackingMarks = marks.filter(
-    (mark) => mark?.type === TrackInsertMarkName || mark?.type === TrackDeleteMarkName,
+
+  const trackingMarks = marks.filter((mark) =>
+    [TrackInsertMarkName, TrackDeleteMarkName, TrackFormatMarkName].includes(mark?.type),
```
caio-pizzol, line 29: `trackFormat` is now collected into `trackingMarksByType` here, but the `preservedMarks` filter on line ~42 still only removes `trackInsert` and `trackDelete` — so `trackFormat` ends up in both places and gets pushed onto children too. nothing downstream processes it today, so no duplicate output — but it's inconsistent with the function's intent. worth aligning the `preservedMarks` filter too?

```js
const preservedMarks = marks
  .filter((mark) => ![TrackInsertMarkName, TrackDeleteMarkName, TrackFormatMarkName].includes(mark?.type))
  .map((mark) => cloneMark(mark));
```

caio-pizzol: @tupizz 👀

harbournick:
merging via  https://github.com/superdoc-dev/superdoc/pull/2395 instead

PR #1853, opened 2026-01-27
SD-1557 - Comment text after enter break is dropped on render and export

chittolinag:
Comments that have multiple lines were being parsed in the wrong way. We were using only the first text element, but comments that span multiple lines will have multiple text elements. 

Changed the logic to parse all the text elements for a given comment on import _and_ export.

### DIFF
--- a/packages/super-editor/src/core/super-converter/styles.js
+++ b/packages/super-editor/src/core/super-converter/styles.js
@@ -509,7 +509,6 @@ export function decodeRPrFromMarks(marks) {
     return runProperties;
   }
 
-  console.log('[debug] marks:', marks);
   marks.forEach((mark) => {
     const type = mark.type.name ?? mark.type;
     switch (type) {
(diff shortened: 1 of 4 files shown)


packages/super-editor/src/core/super-converter/v2/importer/documentCommentsImporter.js
```diff
@@ -71,8 +72,8 @@ export function importCommentData({ docx, editor, converter }) {
       creatorName: authorName,
       creatorEmail: authorEmail,
       createdTime: unixTimestampMs,
-      textJson: parsedElements[0],
-      elements: parsedElements,
+      textElements,
+      elements: textElements,
```
luccas-harbour: I can see that both these keys have the same data. Could we keep just one of them?

chittolinag: great catch, I didn't notice it. Will change!

harbournick (approved):
LGTM

PR #1860, opened 2026-01-28
chore: dev collab mode

caio-pizzol:


caio-pizzol:
@harbournick merging this one (since it doesn't impact the package - dev app specific)

PR #1863, opened 2026-01-28
fix: use stable comment ids for imported comments

luccas-harbour:
- SD used to generate a different UUID as a comment's internal id every time a document was imported. 
- This PR changes that so that it uses a hash of the comment's imported id and its date for imported comments so that they are stable.


packages/super-editor/src/core/super-converter/v2/importer/documentCommentsImporter.js
```diff
@@ -1,5 +1,6 @@
 import { v4 as uuidv4 } from 'uuid';
 import { defaultNodeListHandler } from './docxImporter';
+import crypto from 'crypto';
```
caio-pizzol: Should we use Web Crypto API or a simple hash function instead? `crypto` could increase significant bundle size

packages/super-editor/src/core/super-converter/v2/importer/documentCommentsImporter.js
```diff
@@ -633,3 +635,16 @@ const applyParentRelationships = (comments, parentMap, trackedChangeParentMap =
...
+/**
+ * Resolve a stable comment ID for imported comments.
+ * - Prefer the explicit internal ID when present.
+ * - If the comment has an imported ID, derive a stable hash from imported ID + created time.
+ * - Otherwise, fall back to a new UUID.
+ */
+const getCommentId = (internalId, importedId, createdTime) => {
+  if (internalId != null) return internalId;
+  if (importedId == null) return uuidv4();
+  const hash = crypto.createHash('md5').update(`${importedId}-${createdTime}`).digest('hex');
```
caio-pizzol: Per ECMA-376 §17.13.4.2, the `w:date` attribute on comments is optional ("If this attribute is omitted, then no date information shall be associated with the parent annotation type").

When `w:date` is missing, `unixTimestampMs` becomes NaN, causing the hash to be ${importedId}-NaN. Two comments with the same importedId but different missing dates would collide.

Should fall back to UUID or use a different stable identifier when date is unavailable.

packages/super-editor/src/tests/import/documentCommentsImporter.unit.test.js
```diff
@@ -131,10 +132,12 @@ describe('importCommentData metadata parsing', () => {
     });
 
     const [comment] = importCommentData({ docx });
-    expect(comment.commentId).toBe('00000000-0000-4000-8000-000000000001');
-    expect(uuidv4).toHaveBeenCalledTimes(1);
+    const createdTime = new Date('2024-02-10T12:30:00Z').getTime();
+    const expectedHash = crypto.createHash('md5').update(`1-${createdTime}`).digest('hex');
```
caio-pizzol: Minor one: the test duplicates the `md5` logic - if the hash construction is wrong, both would be wrong. Maybe hardcode an expected value or test the stability property (same input → same ID)?

packages/super-editor/src/core/super-converter/v2/importer/documentCommentsImporter.js
luccas-harbour: Good catch! I implemented a simpler hashing function instead

packages/super-editor/src/core/super-converter/v2/importer/documentCommentsImporter.js
luccas-harbour: Fixed!

packages/super-editor/src/tests/import/documentCommentsImporter.unit.test.js
luccas-harbour: Fixed!

harbournick (approved):
LGTM

PR #1875, opened 2026-01-29
fix: horizontal rule

artem-harbour:


packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -3056,7 +3056,10 @@ export class DomPainter {
     sanitize(element);
   }
 
-  private getEffectExtentMetrics(block: VectorShapeDrawingWithEffects): {
+  private getEffectExtentMetrics(
```
caio-pizzol, line 3059: `effectExtent` is still read from block while width/height can now come from a different geometry - if these sources have different scales, the innerWidth/innerHeight calculation may be incorrect.

should we also allow passing effectExtent, or scale it proportionally?

harbournick: This one seems OK I don't think there's anything that would cause this now. I added a test to guard this case in case that changes.

harbournick (approved):
LGTM

PR #1876, opened 2026-01-29
fix: table width

artem-harbour:
Linear: SD-1581

Fixes tables in the header/footer with "AutoFit to Window" formatting applied.

### DIFF
--- a/packages/super-editor/src/core/super-converter/v3/handlers/w/tbl/tbl-translator.test.js
+++ b/packages/super-editor/src/core/super-converter/v3/handlers/w/tbl/tbl-translator.test.js
@@ -178,6 +178,54 @@ describe('w:tbl translator', () => {
       expect(result.attrs.tableProperties).toEqual({ cellMargins: {} });
       expect(trTranslator.encode).not.toHaveBeenCalled();
     });
+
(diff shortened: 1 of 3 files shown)


packages/super-editor/src/core/super-converter/v3/handlers/w/tbl/tbl-translator.test.js
```diff
@@ -147,7 +147,7 @@ describe('w:tbl translator', () => {
       });
       expect(result.attrs.grid).toEqual([{ col: 2000 }]);
       expect(result.attrs.tableStyleId).toBe('TableGrid');
-      expect(result.attrs.tableWidth).toEqual({ width: 250, type: 'pct' });
+      expect(result.attrs.tableWidth).toEqual({ value: 5000, type: 'pct' });
```
caio-pizzol, line 150: should we add a test case for the mixed pct table width with dxa grid columns scenario from SD-1581?

also the auto type fallback path changed - maybe worth a dedicated test?

harbournick: added tests

PR #1877, opened 2026-01-29
fix: paragraph autospacing calculation (SD-1653)

luccas-harbour:
- Fixes logic behind before/after paragraph spacing when auto spacing is enabled
- Fixes line height measurement so that it doesn't re-apply default single spacing multiplier on top of value calculated by `pm-adapter`

### DIFF
--- a/packages/layout-engine/contracts/src/engines/paragraph.ts
+++ b/packages/layout-engine/contracts/src/engines/paragraph.ts
@@ -8,7 +8,8 @@
 export interface ParagraphSpacing {
   before: number; // pt
   after: number; // pt
-  line: number; // pt or multiplier (depends on lineRule)
+  line: number; // pt or multiplier (depends on lineUnit)
+  lineUnit?: 'px' | 'multiplier'; // unit for line spacing value
   lineRule: 'auto' | 'exact' | 'atLeast';
 }
 


packages/layout-engine/measuring/dom/src/index.ts
```diff
@@ -3268,28 +3261,13 @@ const appendSegment = (
...
 
-const resolveLineHeight = (spacing: ParagraphSpacing | undefined, baseLineHeight: number): number => {
-  if (!spacing || spacing.line == null || spacing.line <= 0) {
-    return baseLineHeight;
+const resolveLineHeight = (spacing: ParagraphSpacing | undefined, fontSize: number, maxHeight: number = -1): number => {
+  const computedHeight = fontSize * (spacing?.line ?? WORD_SINGLE_LINE_SPACING_MULTIPLIER);
+  const lineRule = spacing?.lineRule ?? 'auto';
+  if (['atLeast', 'auto'].includes(lineRule)) {
+    return Math.max(computedHeight, maxHeight, WORD_SINGLE_LINE_SPACING_MULTIPLIER * fontSize);
   }
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/layout-engine/pm-adapter/src/attributes/spacing-indent.ts
```diff
@@ -108,16 +112,23 @@ export const normalizeParagraphSpacing = (
...
   const afterAutospacing = value.afterAutospacing;
+  const line = normalizeLineValue(lineRaw, lineRule);
 
-  if (beforeAutospacing && isList) {
-    before = undefined;
+  if (beforeAutospacing) {
+    if (isList) {
+      before = undefined;
+    } else {
+      before = (lineRaw ?? AUTO_SPACING_LINE_DEFAULT) * AUTO_SPACING_DEFAULT_MULTIPLIER;
```
caio-pizzol, line 110: when `lineRaw` is a small multiplier like 2, the autospacing calculation produces 2 * 1.15 = 2.3 twips instead of 2 * 240 * 1.15 = 552 twips - should we apply the same multiplier-to-twips conversion that `normalizeLineValue` does?

packages/layout-engine/pm-adapter/src/attributes/spacing-indent.ts
```diff
@@ -128,18 +139,25 @@ export const normalizeParagraphSpacing = (
...
+ * @returns Normalized line spacing value as a multiplier, or undefined
+ */
+export const normalizeLineValue = (
   value: number | undefined,
   lineRule: ParagraphSpacing['lineRule'] | undefined,
-): number | undefined => {
-  if (value == null) return undefined;
+): number => {
+  if (value == null) return AUTO_SPACING_DEFAULT_MULTIPLIER;
+  if (value > 0 && value <= MIN_AUTO_LINE_TWIPS) {
```
caio-pizzol: just confirming - converting exact line spacing to a multiplier makes it font-size-dependent now.

if a paragraph has 24px exact spacing with mixed 12px/20px text, the old behavior was 24px for all, but now it scales with the run's `fontSize`. is that the intended behavior?

luccas-harbour: That's a good call. I modified the logic so that the unit used for the line height is explicitly stated in the spacing properties accessed by the measuring layer. Now line heights with rules `exact` or `atLeast` will always have the value in pixels and not multipliers.

Thanks!

packages/layout-engine/pm-adapter/src/attributes/spacing-indent.ts
luccas-harbour: I was fixing this but then decided to run a test in Word to make sure it was the correct behavior. It turns out that Word never considers small line heights as multipliers. The values are always taken to be in twips.

I see that this logic was implemented to fix an issue in the `tiny-spacing.docx` test file but the problem was actually different from what was fixed. 

The original issue was that that document had a line height with value `1` which generally would make it very tiny but the thing is that the line rule was `atLeast`. So the actual problem was that we had been handling the `atLeast` rule the same way as `exact`, making the text very small during rendering.

The handling of the `atLeast` rule has been fixed now so this workaround of handling small values as multipliers is not needed anymore. I removed that logic so that we are compatible with Microsoft Word. The `tiny-spacing.docx` test document still renders correctly.

Here's the document I used for checking this by the way: [test-spacing.docx](https://github.com/user-attachments/files/24966299/test-spacing.docx)

caio-pizzol: This is awesome!! Thanks

harbournick (approved):
LGTM - nice work!

harbournick:
:tada: This PR is included in **superdoc** v1.11.0

PR #1892, opened 2026-01-30
fix: normalize bookmarks in tables

VladaHarbour:


packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -841,6 +842,169 @@ export function filterOutRootInlineNodes(content = []) {
...
+  return {
+    ...tableNode,
+    content: rebuiltContent,
+  };
+}
+
+function insertInlineIntoRow(rowNode, inlineNodes, editor, position) {
+  if (!rowNode || !inlineNodes?.length || !Array.isArray(rowNode.content)) return rowNode;
+
+  const targetIndex = position === 'end' ? rowNode.content.length - 1 : 0;
```
caio-pizzol: should we add a guard for rowNode.content.length === 0?

if `rowNode.content` is empty, `targetIndex` becomes -1 and bookmarks get silently dropped

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.test.js
```diff
@@ -1,5 +1,9 @@
 import { describe, it, expect } from 'vitest';
-import { collapseWhitespaceNextToInlinePassthrough, filterOutRootInlineNodes } from './docxImporter.js';
+import {
```
caio-pizzol, line 2: should we add a test for empty rows (no cells)? that case would expose the -1 index issue mentioned above

VladaHarbour: @caio-pizzol added a workaround for an empty row and covered this case with test

caio-pizzol (approved):
LGTm

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -841,6 +842,160 @@ export function filterOutRootInlineNodes(content = []) {
...
+      rowCursor += 1;
+      seenRow = true;
+      return acc;
+    }
+
+    if (isBookmarkNode(child)) {
+      if (seenRow) {
+        trailing.push(child);
+      } else {
+        leading.push(child);
```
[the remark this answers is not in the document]
caio-pizzol: @VladaHarbour does this make sense still?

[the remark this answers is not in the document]
VladaHarbour: @caio-pizzol This one was fixed

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
VladaHarbour: fixed and covered with test

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -841,6 +842,177 @@ export function filterOutRootInlineNodes(content = []) {
...
+      }
+      return acc;
+    }
+
+    acc.push(child);
+    return acc;
+  }, []);
+
+  updatedRows.forEach((row, index) => {
+    if (rowStartInlines[index]?.length) {
```
caio-pizzol: So I checked the spec (§17.13.6.2) actually shows `bookmarkEnd` as a sibling of table cells, not inside them:

```                                                                                                                                                                                                                
  <w:tr>                                                                                                                                                                                                                                           
    <w:tc><w:p/></w:tc>                                                                                                                                                                                                                            
    <w:bookmarkEnd w:id="0" />  <!-- sits here, outside the cell -->                                                                                                                                                                               
  </w:tr>                                                                                                                                                                                                                                          
```
But we're putting it inside a paragraph. This is probably fine given ProseMirror's schema constraints - just worth double-checking and documenting that we're intentionally normalizing the structure.

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -841,6 +842,177 @@ export function filterOutRootInlineNodes(content = []) {
...
+/**
+ * Normalize bookmark nodes that appear as direct table children.
+ * Moves bookmarkStart/End into the first/last cell textblock of the table.
+ *
+ * @param {Array<{type: string, content?: any[], attrs?: any}>} content
+ * @param {Editor} [editor]
+ * @returns {Array}
+ */
+export function normalizeTableBookmarksInContent(content = [], editor) {
+  if (!Array.isArray(content) || content.length === 0) return content;
```
caio-pizzol, line 863: Table bookmarks can have `colFirst` and `colLast` attributes that specify which columns they cover: 
  
  `<w:bookMarkStart w:colFirst="0" w:colLast="1" w:id="0" w:name="table"/>`
  
I didn't see these being preserved when we relocate the bookmarks. If someone has a column-range bookmark, we might lose that info. Worth checking if this matters for our use cases, or at least adding a comment about it.

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -841,6 +842,177 @@ export function filterOutRootInlineNodes(content = []) {
   return result;
 }
 
+/**
```
caio-pizzol, line 848: ECMA spec doesn't actually show bookmarks as valid direct children of `<w:tbl>`. So this whole normalization is handling documents that aren't strictly conformant.

Would be nice to add a comment explaining that - something like:
```
/**                                                                                                                                                                                                                                              
   * Some non-conformant DOCX producers place bookmarks as direct table children.                                                                                                                                                                  
   * Per ECMA-376 §17.13.6.2, they should be inside cells (bookmarkStart) or                                                                                                                                                                       
   * as children of rows (bookmarkEnd). We relocate them for PM compatibility.                                                                                                                                                                     
   */
```

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
VladaHarbour: will check it

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
VladaHarbour: The issue here is that Prosemirror schema doesn't accept anything but tableCell inside tableRow. We can intentionally change this but the compatibility with table plugins may be broken

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
VladaHarbour: @caio-pizzol aded support for colFirst/colLast attributes when calculating the cell for normalisation

VladaHarbour:
@caio-pizzol please let me know If I have to do something with the last spec review. https://ooxml.dev/ page 900 tells that colFirst/colLast are valid attributes for bookmarkStart

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -844,6 +845,210 @@ export function filterOutRootInlineNodes(content = []) {
...
+
+  if (node.type === 'table') {
+    node = normalizeTableBookmarksInTable(node, editor);
+  }
+
+  if (Array.isArray(node.content)) {
+    node.content = normalizeTableBookmarksInContent(node.content, editor);
+  }
+
+  return node;
```
caio-pizzol, line 879: this reassigns `node` (line 872) then mutates `node.content` (line 876) -- half new object, half mutation. the other normalizers in this file pick one or the other. not blocking, just worth being consistent.

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -844,6 +845,210 @@ export function filterOutRootInlineNodes(content = []) {
...
+  });
+
+  // Rebuild content with updated rows (preserve any non-row nodes in place).
+  let updatedRowIndex = 0;
+  const rebuiltContent = normalizedContent.map((child) => {
+    if (child?.type === 'tableRow') {
+      const replacement = updatedRows[updatedRowIndex] ?? child;
+      updatedRowIndex += 1;
+      return replacement;
+    }
```
caio-pizzol, line 969: the reduce builds `normalizedContent` with refs to `updatedRows`, but those get modified after (in the forEach), so there's a second pass here to swap them in. would be simpler to collect bookmark positions first, then build the content array once. not blocking either.

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.test.js
```diff
@@ -178,3 +182,114 @@ describe('collapseWhitespaceNextToInlinePassthrough', () => {
...
+
+    const firstCellContent = rowContent[0].content[0].content;
+    expect(firstCellContent).toEqual([{ type: 'text', text: 'A', marks: [] }]);
+
+    const secondCellContent = rowContent[1].content[0].content;
+    expect(secondCellContent[0]).toMatchObject({ type: 'bookmarkStart', attrs: { id: 'b1', colFirst: '1' } });
+    expect(secondCellContent[1]).toMatchObject({ type: 'text', text: 'B', marks: [] });
+    expect(secondCellContent[2]).toMatchObject({ type: 'bookmarkEnd', attrs: { id: 'b1' } });
+  });
+});
```
caio-pizzol, line 318: tests cover flat tables well, but there's no test for a nested table (table inside a cell with bookmarks as direct children) -- which is the actual crash from the ticket. worth adding that case.

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
VladaHarbour: changed to respect previous approach

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.test.js
VladaHarbour: added test case for this

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
VladaHarbour: rewrote this logic. should be more clear now

Issue #1675, opened 2026-01-07
Bug: `createNumbering` crashes when `lvlText` is null - "Cannot read properties of null (reading 'replace')"


rccoe:
## Description

When processing documents with incomplete or corrupted list numbering definitions (where `lvlText` is `null`), SuperDoc throws an unhandled error:

```
TypeError: Cannot read properties of null (reading 'replace')
    at createNumbering (SuperConverter-D7wt-eeZ.cjs:19862)
    at Array.reduce (<anonymous>)
    at createNumbering (SuperConverter-D7wt-eeZ.cjs:19861)
    at generateNumbering (SuperConverter-D7wt-eeZ.cjs:19867)
    at handleDecimal (SuperConverter-D7wt-eeZ.cjs:19827)
    at Object.generateOrderedListIndex (SuperConverter-D7wt-eeZ.cjs:19858)
```

## Root Cause

In `shared/common/list-numbering/index.ts`, the `createNumbering` function does not handle null `lvlText`:

```typescript
const createNumbering = (values, lvlText) => {
  return values.reduce((acc, value, index2) => {
    return Number(value) > 9 
      ? acc.replace(/^0/, "").replace(`%${index2 + 1}`, value) 
      : acc.replace(`%${index2 + 1}`, value);
  }, lvlText);  // <-- lvlText can be null here
};
```

The call chain is:
1. `numberingPlugin.appendTransaction()` calls `ListHelpers.getListDefinitionDetails()`
2. `lvlText` is destructured from the result (may be `null` or `undefined`)
3. For non-bullet lists, `generateOrderedListIndex({ listLevel, lvlText, ... })` is called
4. This calls `handleDecimal(path, lvlText)` → `generateNumbering(path, lvlText, formatter)` → `createNumbering(formattedValues, lvlText)`
5. `createNumbering` attempts `lvlText.replace(...)` which throws when `lvlText` is null

## Suggested Fix

Option 1: Add null check in `createNumbering`:
```typescript
const createNumbering = (values: number[], lvlText: string | null): string | null => {
  if (!lvlText) return null;
  return values.reduce((acc, value, index) => {
    return Number(value) > 9 
      ? acc.replace(/^0/, "").replace(`%${index + 1}`, String(value)) 
      : acc.replace(`%${index + 1}`, String(value));
  }, lvlText);
};
```

Option 2: Add null check in `generateOrderedListIndex`:
```typescript
[...]

PR #1904, opened 2026-02-02
fix: add null checks to list numbering functions

lloyd-lighthouse:
## Summary

Added null safety to the list numbering functions in `shared/common/list-numbering/index.ts` to prevent runtime errors when `lvlText` is null or undefined.

## Changes

- Updated `createNumbering` to accept `lvlText: string | null` and return early if null
- Updated `generateNumbering` to accept `lvlText: string | null` with null guard
- Updated `generateFromCustom` to accept `lvlText: string | null` with null guard
- Updated return types to `string | null` to reflect nullable returns

## Why

When `lvlText` is null, calling `.replace()` on it causes a runtime error. This fix ensures graceful handling by returning null early.

## Test Plan

- [X] Existing unit tests pass
- [ ] Verify list numbering still works correctly with valid `lvlText` values
- [ ] Verify null `lvlText` returns null without throwing errors

## Breaking Changes

None - this is a backwards-compatible fix. Functions that previously crashed on null input will now return null.

### DIFF
--- a/shared/common/list-numbering/index.ts
+++ b/shared/common/list-numbering/index.ts
@@ -49,7 +49,8 @@ export const generateOrderedListIndex = ({
   return handler ? handler(listLevel, lvlText, customFormat) : null;
 };
 
-const createNumbering = (values: string[], lvlText: string): string => {
+const createNumbering = (values: string[], lvlText: string | null): string | null => {
+  if (lvlText == null) return null;
   return values.reduce<string>((acc, value, index) => {
     return Number(value) > 9
       ? acc.replace(/^0/, '').replace(`%${index + 1}`, value)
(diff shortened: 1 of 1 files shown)


lloyd-lighthouse:
Hi team.

Its strange that a version was released to fix numbering issues recently (https://github.com/superdoc-dev/superdoc/issues/1675)

However this is a similar issue, perhaps some sort of an edge case?

--- 

This document works
[broken-numbering-missing-abstract.docx](https://github.com/user-attachments/files/25000494/broken-numbering-missing-abstract.docx)

This document however, does not work
[document-2026-02-02 (2).docx](https://github.com/user-attachments/files/25000631/document-2026-02-02.2.docx)

--- 

We've tried our best to trace this issue down and think the suggested fix in this PR is a way for us to make sure we never encounter this issue again, but we are obviously new to this codebase and can only do a best guess realistically.

PR #1912, opened 2026-02-02
feat(react): add @superdoc-dev/react wrapper package

tupizz:
## Summary

Adds `@superdoc-dev/react` - a first-party React wrapper for SuperDoc with proper lifecycle management, SSR safety, and React Strict Mode compatibility.

Closes SD-1716

## Key Design Decisions

| Decision | Choice |
|----------|--------|
| **Type system** | Types extracted from `superdoc` (no duplication) |
| **Ref API** | Single `getInstance()` method |
| **Style import** | `@superdoc-dev/react/style.css` |
| **React version** | 16.8+ (uses `useRef` for ID generation) |
| **SSR behavior** | Client-only (renders `renderLoading` if provided, otherwise null) |
| **documentMode** | Prop changes handled without rebuild |

## Package Structure

```
packages/react/
├── src/
│   ├── index.ts              # Exports
│   ├── SuperDocEditor.tsx    # Main component
│   ├── SuperDocEditor.test.tsx
│   ├── types.ts              # Types extracted from superdoc
│   └── utils.ts              # ID generation (internal)
├── style.css                 # Re-exports superdoc/style.css
├── package.json
├── tsconfig.json
├── vite.config.ts
└── vitest.config.ts
```

## Usage

```tsx
import { SuperDocEditor } from '@superdoc-dev/react';
import '@superdoc-dev/react/style.css';

function App() {
  const [mode, setMode] = useState<DocumentMode>('editing');
  const editorRef = useRef<SuperDocRef>(null);
  
  return (
    <>
      <button onClick={() => setMode('viewing')}>View</button>
      <SuperDocEditor
        ref={editorRef}
        document={file}
        documentMode={mode}
        onReady={({ superdoc }) => console.log('Ready!')}
      />
    </>
  );
}

// Access instance via ref for methods
await editorRef.current?.getInstance()?.export({ triggerDownload: true });
```

## Props

All props from `SuperDocConfig` are supported, plus React-specific props:

| Prop | Type | Default | Description |
|------|------|---------|-------------|
| `document` | `File \| Blob \| string \| object` | - | Document to load |
| `documentMode` | `'editing' \| 'viewing' \| 'suggesting'` | `'editing'` | Editing mode |
| `role` | `'editor' \| 'viewer' \| 'suggester'` | `'editor'` | User role |
| `id` | `string` | auto-generated | Custom container ID |
| `user` | `{ name, email?, image? }` | - | Current user |
| `users` | `Array<{ name, email, image? }>` | - | All users |
| `modules` | `object` | - | Modules config |
| `hideToolbar` | `boolean` | `false` | Hide the toolbar |
| `renderLoading` | `() => ReactNode` | - | Loading UI (shown during SSR and init) |
| `className` | `string` | - | Wrapper CSS class |
| `style` | `CSSProperties` | - | Wrapper inline styles |

### Props That Trigger Rebuild

| Prop | Reason |
|------|--------|
| `document` | New document to load |
| `user` | User identity changed |
| `users` | Users list changed |
| `modules` | Module configuration changed |
| `role` | Permission level changed |
| `hideToolbar` | Toolbar visibility changed |

Other props like `documentMode` and callbacks are handled efficiently without rebuild.

## Ref API

```typescript
interface SuperDocRef {
  getInstance(): SuperDocInstance | null;
}
```

## Examples

| Example | Description |
|---------|-------------|
| `examples/getting-started/react` | React + TypeScript example with full type safety |
| `examples/integrations/nextjs-ssr` | Next.js App Router with SSR support |

Run examples from repo root:
```bash
pnpm -C examples/getting-started/react dev
pnpm -C examples/integrations/nextjs-ssr dev
```

## Documentation Updated

- `apps/docs/getting-started/frameworks/react.mdx` - Full React integration guide
- `apps/docs/getting-started/frameworks/nextjs.mdx` - Next.js integration guide
- `packages/react/README.md` - Package README
- `packages/react/CLAUDE.md` - Implementation details for AI assistants

## Test Plan

- [x] `pnpm --filter @superdoc-dev/react build` succeeds
- [x] `pnpm --filter @superdoc-dev/react test` passes (14 tests)
- [x] Example apps run and load documents
- [x] Types correctly extracted from superdoc
- [x] Ref API works as expected
- [x] Docs consistent with implementation

### DIFF
--- a/apps/docs/getting-started/frameworks/nextjs.mdx
+++ b/apps/docs/getting-started/frameworks/nextjs.mdx
@@ -53,7 +53,7 @@ export default function EditorPage() {
 ```
 
 <Tip>
-The React wrapper is client-only (returns null on server, renders after hydration). For custom loading UI during SSR, use Next.js dynamic imports with a `loading` component.
+The React wrapper is SSR-safe — it renders container divs on the server (hidden until initialized) and starts SuperDoc after hydration. Use `renderLoading` for custom loading UI, or Next.js dynamic imports if you prefer to skip SSR entirely.
 </Tip>
 
 ---
--- a/apps/docs/getting-started/frameworks/react.mdx
+++ b/apps/docs/getting-started/frameworks/react.mdx
@@ -46,7 +46,7 @@ That's it! You now have a fully functional DOCX editor in your React app.
 - **Mounting** - Creates a SuperDoc instance when the component mounts
 - **Updates** - Rebuilds automatically when the `document` prop changes
 - **Cleanup** - Properly destroys the instance on unmount
-- **Client-Only** - Returns null on server, renders only after hydration
+- **SSR-Safe** - Renders container structure on server, initializes SuperDoc after hydration
 
 ### Document Modes
 


packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,269 @@
...
+      onEditorUpdate,
+      onContentError,
+      onException,
+    };
+  }, [onReady, onEditorCreate, onEditorDestroy, onEditorUpdate, onContentError, onException]);
+
+  // Track documentMode changes and apply imperatively
+  const prevDocumentModeRef = useRef(documentMode);
+  useEffect(() => {
+    if (prevDocumentModeRef.current !== documentMode && instanceRef.current) {
```
caio-pizzol: Good catch. Added `pendingModeRef` and `isInitializingRef` to handle this case - the mode change gets queued and applied in the `onReady` callback.

See lines 74-91 and 144-147 in `SuperDocEditor.tsx`

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,269 @@
...
+    [],
+  );
+
+  // Main effect: create and destroy SuperDoc instance
+  useEffect(() => {
+    // Skip on server-side
+    if (typeof window === 'undefined') return;
+
+    let destroyed = false;
+    let instance: SuperDocInstance | null = null;
```
caio-pizzol, line 132: should we call `setIsLoading(true)` here at the start of init?

`isLoading` starts as `true` but never resets when the document changes

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,269 @@
...
+        // Check if we were destroyed while loading
+        if (destroyed) return;
+
+        // Build configuration
+        const superdocConfig = {
+          selector: `#${containerId}`,
+          ...(toolbar && toolbarContainerRef.current ? { toolbar: `#${toolbarId}` } : {}),
+          documentMode,
+          role,
+          ...(documentProp !== undefined ? { document: documentProp } : {}),
```
caio-pizzol: this passes `null` through but the type disallows it - should we use `documentProp != null` instead?

packages/react/src/types.ts
```diff
@@ -0,0 +1,415 @@
...
+}
+
+/**
+ * SuperDoc instance interface (subset of public methods)
+ */
+export interface SuperDocInstance {
+  version: string;
+  superdocId: string;
+  config: SuperDocConfig;
+  activeEditor: object | null;
```
caio-pizzol: could we re-export types from the main superdoc package for better type safety?

packages/react/src/utils.ts
```diff
@@ -0,0 +1,24 @@
+import { useRef } from 'react';
+
+/**
+ * Counter for generating unique IDs
+ */
+let idCounter = 0;
```
caio-pizzol: if multiple requests share the module we could cause SSR hydration mismatches, e.g.

1. Server renders: `<div id="superdoc-1">...</div>` (counter starts at 0, increments to 1)                                                                                                                                                         
2. Client renders: `<div id="superdoc-3">...</div>` (counter might be at 2 from other components, increments to 3)

React sees the mismatch (superdoc-1 vs superdoc-3) and throws a hydration warning

packages/react/tsup.config.ts
```diff
@@ -0,0 +1,16 @@
+import { defineConfig } from 'tsup';
```
caio-pizzol: question: why `tsup` instead of matching the main superdoc package's build setup? adds a new dependency and different tooling to maintain

packages/react/src/types.ts
```diff
@@ -0,0 +1,415 @@
+import type { CSSProperties, ReactNode } from 'react';
```
caio-pizzol, line 1: can we re-export types from `superdoc/types` instead of duplicating them here? would reduce maintenance burden and keep types in sync

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,269 @@
...
+  const prevDocumentModeRef = useRef(documentMode);
+  useEffect(() => {
+    if (prevDocumentModeRef.current !== documentMode && instanceRef.current) {
+      instanceRef.current.setDocumentMode(documentMode);
+    }
+    prevDocumentModeRef.current = documentMode;
+  }, [documentMode]);
+
+  // Expose ref methods
+  useImperativeHandle(
```
caio-pizzol, line 116: 15 methods is a large API surface - would it be simpler to just expose `getInstance()` and let users access the full API that way? keeps the wrapper thin and reduces maintenance

packages/react/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "dev": "tsup --watch",
+    "test": "vitest run",
+    "lint": "eslint src --ext .ts,.tsx",
+    "prepublishOnly": "pnpm run build",
+    "version:next": "pnpm version prerelease --preid=next --no-git-tag-version",
+    "version:beta": "pnpm version prerelease --preid=beta --no-git-tag-version",
+    "version:patch": "pnpm version patch --no-git-tag-version",
+    "version:minor": "pnpm version minor --no-git-tag-version",
+    "version:major": "pnpm version major --no-git-tag-version",
+    "publish:next": "pnpm run build && pnpm run test && pnpm publish --tag next --access public",
```
caio-pizzol: the version:*/publish:*/release:* scripts feel like CI concerns - can we remove these and handle versioning/publishing through the monorepo's release workflow?

packages/react/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
```
caio-pizzol, line 1: the monorepo has tsconfig.base.json - this should extend from it like other packages

packages/react/src/SuperDocEditor.tsx
tupizz: Done - added `setIsLoading(true)` at the start of the effect (line 108). Also reset `isInitializingRef` there.

packages/react/src/SuperDocEditor.tsx
tupizz: Yep, changed to `documentProp != null` - line 133.

packages/react/src/types.ts
tupizz: Refactored the whole types.ts - now we extract types directly from the SuperDoc constructor using `ConstructorParameters<typeof SuperDoc>[0]` and `NonNullable<>`. No more duplication, stays in sync automatically.

See `types.ts` lines 1-34.

packages/react/tsup.config.ts
tupizz: Replaced with Vite to match the rest of the monorepo. See `vite.config.ts`.

packages/react/src/types.ts
tupizz: Same as above - types are now extracted from superdoc, not duplicated.

packages/react/src/SuperDocEditor.tsx
tupizz: Agreed. Simplified to just `getInstance()` - users can access the full SuperDoc API through that. Keeps the wrapper thin.

packages/react/package.json
tupizz: Removed all of them. Now just has build, dev, test, lint, prepublishOnly.

packages/react/tsconfig.json
tupizz: Done - now extends `../../tsconfig.base.json`.

packages/react/src/SuperDocEditor.tsx
tupizz: Added `pendingModeRef` and `isInitializingRef` to handle this - the mode change gets queued and applied in the `onReady` callback. See lines 74-91 and 144-147.

packages/react/src/utils.ts
tupizz: The real fix: the component is now client-only. Returns `null` (or `renderLoading()` if provided) on server, only renders after client hydration. Since SuperDoc needs browser APIs anyway, this makes sense.

See latest commit - added `isClient` state check before render.

examples/getting-started/react/src/main.jsx
```diff
@@ -1,9 +1,7 @@
...
-import ReactDOM from 'react-dom/client';
+import { createRoot } from 'react-dom/client';
 import App from './App';
 
-ReactDOM.createRoot(document.getElementById('root')).render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>
-);
+// Note: StrictMode is disabled because SuperDoc's toolbar Vue component
```
caio-pizzol: the test verifies no throw in StrictMode, but this example disables it

do we know what is happening in toolbar?

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,248 @@
...
+    return () => {
+      destroyed = true;
+      isInitializingRef.current = false;
+      pendingModeRef.current = null;
+      if (instance) {
+        instance.destroy();
+        instanceRef.current = null;
+      }
+    };
+    // Only these props trigger a full rebuild. Other props (rulers, etc.) are
```
caio-pizzol, line 223: the comment explains this, but should we document in the README which props trigger rebuilds (document, user, users, modules, role, hideToolbar) vs which are initial-only and need `getInstance()` methods?

packages/react/src/SuperDocEditor.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+      const wrapper = container.querySelector('.superdoc-wrapper');
+      expect(wrapper?.classList.contains('custom-class')).toBe(true);
+      expect((wrapper as HTMLElement)?.style.backgroundColor).toBe('red');
+    });
+
+    it('should handle unmount without throwing', async () => {
+      const { unmount } = render(<SuperDocEditor />);
+
+      // Wait a bit for async initialization
+      await new Promise((resolve) => setTimeout(resolve, 100));
```
caio-pizzol: minor: could use `waitFor` with a predicate instead of fixed 100ms delay for more reliable CI behavior

examples/getting-started/react-with-typescript/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "react-with-typescript-example",
```
caio-pizzol: let's keep it simple only 1 example, called `react` that uses typescript

packages/react/.releaserc.cjs
```diff
@@ -0,0 +1,44 @@
+/* eslint-env node */
+const branch = process.env.GITHUB_REF_NAME || process.env.CI_COMMIT_BRANCH;
+
+const config = {
+  branches: [
+    { name: 'stable', channel: 'latest' },
+    { name: 'main', prerelease: 'next', channel: 'next' },
+    // TEMPORARY: Remove before merging to main
```
caio-pizzol: +1

examples/getting-started/react-with-typescript/package.json
caio-pizzol: ok to remove `typescript` folder

.github/workflows/ci-react.yml
```diff
@@ -0,0 +1,45 @@
+name: CI React
```
caio-pizzol, line 1: nice!! thanks

.github/workflows/release-react.yml
```diff
@@ -0,0 +1,63 @@
+# Auto-releases on push to main (@next channel)
+# For stable (@latest): cherry-pick commits to stable branch, then manually dispatch this workflow
+name: 📦 Release react
```
caio-pizzol, line 3: nice!! 💯

CLAUDE.md
```diff
@@ -37,17 +38,20 @@ packages/
   collaboration-yjs/ Collaboration server
 shared/              Internal utilities
 e2e-tests/           Playwright tests
+examples/
```
caio-pizzol: don't need examples here

CLAUDE.md
```diff
@@ -60,3 +64,26 @@ e2e-tests/           Playwright tests
 - `pnpm build` - Build all packages
 - `pnpm test` - Run tests
 - `pnpm dev` - Start dev server (from examples/)
+
+## Git Commit Rules
```
caio-pizzol: let's remove this part as well

README.md
```diff
@@ -53,7 +53,30 @@ Or install with CDN
 <script type="module" src="https://unpkg.com/superdoc/dist/superdoc.umd.js"></script>
 ```
 
-### Basic usage
+### React
+
+```bash
+npm install @superdoc-dev/react
```
caio-pizzol, line 59: 🥳

apps/docs/getting-started/frameworks/nextjs.mdx
```diff
@@ -3,124 +3,228 @@ title: Next.js
...
 ```
 
-## App Router (Next.js 13+)
+<Tip>
+The React wrapper is client-only (returns null on server, renders after hydration). For custom loading UI during SSR, use Next.js dynamic imports with a `loading` component.
+</Tip>
+
+---
+
+## With Dynamic Import (Optional)
```
caio-pizzol: is this really necessary? (i'm not really a Next.js person)

feel free to remove any over complication from the docs + let's make sure it works with a real app (maybe add a examples/nextjs?)

apps/docs/getting-started/frameworks/react.mdx
```diff
@@ -1,268 +1,580 @@
 ---
 title: React Integration
 sidebarTitle: React
-keywords: "react docx editor, react word component, superdoc react, microsoft word react, document editor react hooks"
+keywords: "react docx editor, react word component, superdoc react, microsoft word react, document editor react hooks, @superdoc-dev/react"
 ---
-SuperDoc works with React 16.8+ (hooks) and React 18+ (concurrent features).
 
-## Install
+SuperDoc provides `@superdoc-dev/react` - a first-party wrapper with proper lifecycle management, SSR safety, and React Strict Mode compatibility.
```
caio-pizzol, line 7: ok, it is more of a question here.

those guides are inside of a "getting started" section - so we should keep things simple and straight to the point.

but I agree that we should document this package somewhere in the docs - let me think about it and get back to you on this.

caio-pizzol (commented):
nice! almost there

apps/docs/getting-started/frameworks/nextjs.mdx
tupizz: I'll add an example with nextjs and make this part here simpler

examples/getting-started/react-with-typescript/package.json
tupizz: removed

CLAUDE.md
tupizz: removed

CLAUDE.md
tupizz: removed

packages/react/src/SuperDocEditor.test.tsx
tupizz: Fixed in commit 90d7b764. Replaced the fixed 100ms `setTimeout` with `waitFor` using a predicate that waits for `onReady` to be called. See `SuperDocEditor.test.tsx` lines 39-53.

packages/react/src/SuperDocEditor.tsx
tupizz: Fixed in commit 90d7b764. Added a new section "Props That Trigger Rebuilds" to the README documenting:

- Props that trigger rebuilds: `document`, `user`, `users`, `modules`, `role`, `hideToolbar`
- Props handled efficiently: `documentMode` (uses `setDocumentMode()` internally)
- Initial-only props: other SuperDoc options need `getInstance()` for runtime changes

See `README.md` lines 80-99.

examples/getting-started/react/src/main.jsx
tupizz: This was already fixed in an earlier commit. The example at `examples/getting-started/react/src/main.tsx` now has StrictMode enabled:

```tsx
import { StrictMode } from 'react';
import { createRoot } from 'react-dom/client';
import App from './App';
import './index.css';

createRoot(document.getElementById('root')!).render(
  <StrictMode>
    <App />
  </StrictMode>
);
```

The old `main.jsx` (which had StrictMode disabled) was replaced when consolidating the examples.

packages/react/.releaserc.cjs
tupizz: removed

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+    // Reset states when document changes
+    setIsLoading(true);
+    setHasError(false);
+    isInitializingRef.current = true;
+
+    // Capture the current documentMode for this effect run
+    const effectDocumentMode = initialDocumentModeRef.current;
+    initialDocumentModeRef.current = documentMode;
+
```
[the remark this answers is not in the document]
caio-pizzol: maybe just use documentMode directly?

```js
  const effectDocumentMode = initialDocumentModeRef.current;  // reads OLD value                                                                                                                                                                 
  initialDocumentModeRef.current = documentMode;              // then updates for next time
```

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,277 @@
...
+        // Check if we were destroyed while loading
+        if (destroyed) return;
+
+        // Build configuration - pass through all props
+        const superdocConfig = {
+          ...restProps,
+          selector: `#${CSS.escape(containerId)}`,
+          // Use internal toolbar container unless hideToolbar is true
+          ...(!hideToolbar && toolbarContainerRef.current ? { toolbar: `#${CSS.escape(toolbarId)}` } : {}),
+          documentMode: effectDocumentMode,
```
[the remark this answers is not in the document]
caio-pizzol: check if relevant

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,277 @@
...
+          callbacksRef.current.onException?.({ error: error as Error });
+        }
+      }
+    };
+
+    initSuperDoc();
+
+    // Cleanup function
+    return () => {
+      destroyed = true;
```
caio-pizzol: should we call `onEditorDestroy` before setting the flag, or remove the guard for destroy specifically?

`destroyed = true` is set before `instance.destroy()`, and the `onEditorDestroy` wrapper checks !destroyed (so it something happened and `instance.destroy()` was never called - we would still consider it destroyed from the flag

packages/react/src/SuperDocEditor.test.tsx
```diff
@@ -0,0 +1,183 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, cleanup, waitFor } from '@testing-library/react';
+import { createRef, StrictMode } from 'react';
+import { SuperDocEditor } from './SuperDocEditor';
+import type { SuperDocRef } from './types';
+
+describe('SuperDocEditor', () => {
```
caio-pizzol, line 7: missing tests for: `onEditorDestroy` on unmount, `documentMode` changes, error states.

.github/workflows/release-react.yml
```diff
@@ -0,0 +1,66 @@
...
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'packages/react/**'
+      - 'packages/layout-engine/**'
```
caio-pizzol, line 13: `superdoc` updates wouldn't trigger it?

examples/getting-started/nextjs/.gitignore
```diff
@@ -0,0 +1,41 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+/.pnp
```
caio-pizzol: let's keep it really concise, just the necessary files to ignore

examples/getting-started/nextjs/README.md
```diff
@@ -0,0 +1,36 @@
+This is a [Next.js](https://nextjs.org) project bootstrapped with [`create-next-app`](https://nextjs.org/docs/app/api-reference/cli/create-next-app).
+
+## Getting Started
```
caio-pizzol: should we remove this README?

caio-pizzol: or make it SuperDoc specific

examples/getting-started/nextjs/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "eslint"
+  },
+  "dependencies": {
+    "@superdoc-dev/react": "workspace:*",
+    "next": "16.1.6",
+    "react": "19.2.3",
+    "react-dom": "19.2.3"
```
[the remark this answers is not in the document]
caio-pizzol: @tupizz > worth using catalog version here

packages/react/src/index.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * @superdoc-dev/react - Official React wrapper for SuperDoc
+ * @packageDocumentation
+ * @version 1.0.0
+ */
```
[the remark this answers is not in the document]
caio-pizzol: remove it or make sure to update it from `package.json`

apps/docs/getting-started/frameworks/react.mdx
caio-pizzol: we can tackle this one later - no worries

examples/getting-started/nextjs/README.md
tupizz: yes, we should remove, good catch, were generated by next create boilerplate

packages/react/src/SuperDocEditor.tsx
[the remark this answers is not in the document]
tupizz: Fixed in commit 2971d57f. Removed `initialDocumentModeRef` entirely — now using `documentMode` directly in the effect body. Since `documentMode` is not in the dependency array (it's handled separately via `setDocumentMode()`), the closure already captures the correct value at effect run time. See `SuperDocEditor.tsx` lines 157-163.

packages/react/src/SuperDocEditor.tsx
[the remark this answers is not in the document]
tupizz: Checked — this is fine. The toolbar ref is populated by the time the effect runs because `useEffect` fires after paint, and the toolbar div is rendered in the same render cycle (gated by `isClient`). Also using `CSS.escape()` for safety per your earlier suggestion.

packages/react/src/SuperDocEditor.tsx
tupizz: Fixed in commit 2971d57f. Moved `destroyed = true` to after `instance.destroy()` so the `onEditorDestroy` callback can fire during teardown. See `SuperDocEditor.tsx` lines 227-234. Also added a test that verifies `onEditorDestroy` is called on unmount.

packages/react/src/SuperDocEditor.test.tsx
tupizz: Fixed in commit 2971d57f. Added tests for:
- `onEditorDestroy` on unmount — verifies the callback fires when the component unmounts
- Error states — verifies error UI or `onException` fires when given an invalid document

See `SuperDocEditor.test.tsx` lines 122-155. Now at 16 tests total, all passing.

.github/workflows/release-react.yml
tupizz: Fixed in commit 2971d57f. Added `packages/superdoc/**` to the trigger paths in `release-react.yml` (line 11).

examples/getting-started/nextjs/.gitignore
tupizz: Fixed in commit 2971d57f. Trimmed to just the essentials:
```
node_modules
/.next/
/out/
next-env.d.ts
*.tsbuildinfo
.DS_Store
```

examples/getting-started/nextjs/package.json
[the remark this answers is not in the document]
tupizz: Fixed in commit 2971d57f. The `examples/` directory is a separate pnpm workspace (with its own `pnpm-workspace.yaml`) that doesn't have access to the root catalog, so `catalog:` can't be used here. Changed to `^19.2.0` semver range instead of hardcoded version. See `examples/getting-started/nextjs/package.json` line 14-15.

packages/react/src/index.ts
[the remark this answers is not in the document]
tupizz: Fixed in commit 2971d57f. Removed the `@version 1.0.0` JSDoc tag — `package.json` is the source of truth. See `index.ts`.

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+  // Apply defaults
+  const documentMode = props.documentMode ?? 'editing';
+  const role = props.role ?? 'editor';
+
+  const instanceRef = useRef<SuperDocInstance | null>(null);
+  const toolbarContainerRef = useRef<HTMLDivElement | null>(null);
+
+  // Generate stable IDs once per component instance (use provided id if available)
+  const idsRef = useRef<{ containerId: string; toolbarId: string } | null>(null);
```
afn: Recommend using [`useId()`](https://react.dev/reference/react/useId) instead. (And if we want to support React versions older than 18, we can implement a polyfill in a separate file.)

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,270 @@
...
+    () => ({
+      getInstance: () => instanceRef.current,
+    }),
+    [],
+  );
+
+  // Main effect: create and destroy SuperDoc instance
+  useEffect(() => {
+    // Wait for client-side render so the container div exists in DOM
+    if (!isClient) return;
```
afn: Is this needed? `useEffect` will only be called on the client side.

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+/** Callback props type for the ref */
+type CallbacksType = {
+  onReady?: (event: SuperDocReadyEvent) => void;
+  onEditorCreate?: (event: SuperDocEditorCreateEvent) => void;
+  onEditorDestroy?: () => void;
+  onEditorUpdate?: (event: SuperDocEditorUpdateEvent) => void;
+  onContentError?: (event: SuperDocContentErrorEvent) => void;
+  onException?: (event: SuperDocExceptionEvent) => void;
+};
```
afn: This looks like a duplicate of CallbackProps in `src/types.ts`. Can we import from there & avoid duplication?

packages/react/src/types.ts
```diff
@@ -0,0 +1,163 @@
...
+  | 'onEditorDestroy'
+  | 'onEditorUpdate'
+  | 'onContentError'
+  | 'onException';
+
+/**
+ * Explicitly typed callback props to ensure proper TypeScript inference.
+ * These override any loosely-typed callbacks from SuperDocConfig.
+ */
+interface CallbackProps {
```
afn: This could be a follow-on chore outside the scope of this work, but ideally the event types (`SuperDocReadyEvent` etc) and the callback props of the config object (`onReady` etc) would be defined and exported from `superdoc` and imported by the `superdoc-react` package.

packages/react/src/SuperDocEditor.tsx
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  return (
+    <div className={wrapperClassName} style={style}>
+      {!hideToolbar && <div ref={toolbarContainerRef} id={toolbarId} className='superdoc-toolbar-container' />}
+      <div id={containerId} className='superdoc-editor-container' />
+      {isLoading && !hasError && renderLoading && <div className='superdoc-loading-container'>{renderLoading()}</div>}
+      {hasError && <div className='superdoc-error-container'>Failed to load editor. Check console for details.</div>}
+    </div>
+  );
```
afn: Recommend avoid duplicating the wrapper container --
```suggestion
    return renderLoading ? (
      <div className={wrapperClassName} style={style}>
        {renderLoading()}
      </div>
    ) : null;
  }

  return (
    <div className={wrapperClassName} style={style}>
      {isClient && <>
        {!hideToolbar && <div ref={toolbarContainerRef} id={toolbarId} className='superdoc-toolbar-container' />}
        <div id={containerId} className='superdoc-editor-container' />
      </>}
      {isLoading && !hasError && renderLoading && <div className='superdoc-loading-container'>{renderLoading()}</div>}
      {hasError && <div className='superdoc-error-container'>Failed to load editor. Check console for details.</div>}
    </div>
  );
```

afn: Also, for the above: I'm not sure if there's a reason not to render the toolbar and container divs on the server side. If not, it could be simplified further to remove the `isClient` condition on that fragment, i.e.:
```suggestion
  return (
    <div className={wrapperClassName} style={style}>
      {!hideToolbar && <div ref={toolbarContainerRef} id={toolbarId} className='superdoc-toolbar-container' />}
      <div id={containerId} className='superdoc-editor-container' />
      {isLoading && !hasError && renderLoading && <div className='superdoc-loading-container'>{renderLoading()}</div>}
      {hasError && <div className='superdoc-error-container'>Failed to load editor. Check console for details.</div>}
    </div>
  );
```

packages/react/src/SuperDocEditor.tsx
tupizz: Done — replaced `generateId()` with a `useStableId()` hook in `utils.ts`. It uses `React.useId()` when available (React 18+) and falls back to a `useRef`-based polyfill for React 16.8+/17. See [a247622](https://github.com/superdoc-dev/superdoc/commit/a247622c).

packages/react/src/SuperDocEditor.tsx
tupizz: The `isClient` guard here isn't about `useEffect` itself — it's about the container `<div id={containerId}>` existing in the DOM before SuperDoc tries to mount into it via `selector: '#id'`. Without the guard, the effect runs on the first client render when the div hasn't been committed to the DOM yet (since the SSR path returns `renderLoading` or null). The `setIsClient(true)` in the first effect triggers a re-render that adds the container div, and then this effect runs on the next cycle with the div present. Keeping it as-is for safety.

packages/react/src/SuperDocEditor.tsx
tupizz: Done — removed the duplicate `CallbacksType` and now importing `CallbackProps` from `types.ts` (also exported it from there). See [a247622](https://github.com/superdoc-dev/superdoc/commit/a247622c).

packages/react/src/types.ts
tupizz: Agreed — tracked as a follow-on chore. The core `superdoc` package currently types callbacks loosely in its config object, so exporting proper event types from there would benefit both this React wrapper and any future framework wrappers.

packages/react/src/SuperDocEditor.tsx
tupizz: Keeping the current structure intentionally. The two render paths serve different purposes — the SSR path returns only the loading placeholder (or null), while the client path renders the full container with toolbar, editor div, loading overlay, and error state. Consolidating them would mean rendering toolbar/container divs during SSR, but SuperDoc mounts into these divs imperatively — having empty container divs in SSR HTML that briefly exist before hydration could cause layout flicker. The `isClient` gate ensures the container divs only appear when SuperDoc is ready to mount into them.

tupizz: See reply above — rendering these divs server-side would mean empty container elements in the SSR HTML that SuperDoc can't use until client-side JS loads. Keeping the `isClient` guard avoids the brief flash of empty containers and ensures a clean SSR → hydration transition.

packages/react/src/SuperDocEditor.tsx
tupizz: This is a great tooling for nextjs

packages/react/src/SuperDocEditor.tsx
afn: ~Hmm, I don't see this change in a247622 (nor in the current `feat/react-wrapper` branch)...~

Never mind, I missed the change because I didn't expect `idsRef` to still be there.

You can simplify this by eliminating the `useRef` and just calling:

```tsx
  const generatedId = useStableId();
  const baseId = id ?? `superdoc${generatedId}`;
  const containerId = baseId;
  const toolbarId = `${baseId}-toolbar`;
```

packages/react/src/SuperDocEditor.tsx
afn: > The `isClient` guard here isn't about `useEffect` itself — it's about the container `<div id={containerId}>` existing in the DOM before SuperDoc tries to mount into it via `selector: '#id'`.
Ah, I see. Then this is actually related to my comment below about the conditional rendering in the final `return` statement. If you were to unconditionally render the `superdoc-editor-container` div as I suggested there, then the div would be present after the first render, avoiding the need for this additional guard (and the immediate rerender triggered by calling `setIsClient` from `useEffect`). (Generally speaking, [setting state in useEffect is undesirable](https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-effect).)

Regarding your concern about rendering empty container divs in the original SSR response, there are several options here; for example, you could have the divs be hidden until SuperDoc has been initialized on the client side:

```tsx
  const hideWhenLoading = isLoading ? { display: 'none' } : {};
  return (
    <div className={wrapperClassName} style={style}>
      {!hideToolbar && <div ref={toolbarContainerRef} id={toolbarId} className='superdoc-toolbar-container' style={hideWhenLoading} />}
      <div id={containerId} className='superdoc-editor-container' style={hideWhenLoading} />
      {isLoading && !hasError && renderLoading && <div className='superdoc-loading-container'>{renderLoading()}</div>}
      {hasError && <div className='superdoc-error-container'>Failed to load editor. Check console for details.</div>}
    </div>
  );
```

So in sum, I'd recomment:
* Eliminating the `isClient` state and corresponding useEffect
* Changing the `return` statement to always include the containers, so that they are present in the DOM after the first render

packages/react/src/SuperDocEditor.tsx
tupizz: Done — removed `idsRef`, now just:

```tsx
const generatedId = useStableId();
const baseId = id ?? `superdoc${generatedId}`;
const containerId = baseId;
const toolbarId = `${baseId}-toolbar`;
```

Much cleaner since `useStableId()` already returns a stable value.

packages/react/src/SuperDocEditor.tsx
tupizz: Great point, done. Eliminated `isClient` state and the corresponding `useEffect` entirely.

harbournick (approved):
LGTM

PR #1916, opened 2026-02-03
Create README.md

missysuperdoc:
README for the feature ticket template I just added.


templates/linear/README.md
```diff
@@ -0,0 +1,101 @@
+# Rendering Ticket Template (Word / DOCX)
```
caio-pizzol, line 1: should this live in github repo?

missysuperdoc: Great call! I will move to private repo.

PR #1932, opened 2026-02-04
feat: telemetry

andrii-harbour:
- Added telemetry support to track document opens for usage-based billing.
- Introduced `initTelemetry` method in the Editor class to initialize telemetry based on configuration.
- Created tests for telemetry functionality, ensuring correct behavior when enabled/disabled and with various configurations.
- Updated Editor and SuperDoc configurations to include telemetry options and license key handling.
- Enhanced SuperConverter to manage document creation timestamps for unique identification.

### DIFF
--- a/packages/superdoc/src/dev/components/SuperdocDev.vue
+++ b/packages/superdoc/src/dev/components/SuperdocDev.vue
@@ -180,7 +180,7 @@ const init = async () => {
     documentMode: 'editing',
     licenseKey: 'community-and-eval-agplv3',
     telemetry: {
-      enabled: true
+      enabled: false,
     },
     comments: {
       visible: true,


packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -245,6 +245,9 @@ class SuperConverter {
     this.documentHash = null; // Temporary hash for unmodified documents
     this.documentModified = false; // Track if document has been edited
 
+    // Track if this is a new file created from template
+    this.isNewFile = params?.isNewFile || false;
```
caio-pizzol: `NewFile` means never opened in SuperDoc previously?

packages/super-editor/src/core/Editor.ts
```diff
@@ -447,12 +473,61 @@ export class Editor extends EventEmitter<EditorEventMap> {
...
+      const documentCreatedAt = this.converter?.getDocumentCreatedTimestamp?.() || null;
+      const telemetryPayload = {
+        documentId: documentId || null,
+        documentCreatedAt,
+        isNewFile: this.options.isNewFile,
+      };
+      console.debug('[super-editor] Document info:', telemetryPayload);
+      this.#telemetry.trackDocumentOpen(documentId || null, documentCreatedAt);
+    } catch {
+      // Fail silently - telemetry should never break the app
```
caio-pizzol: 👍

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -669,43 +721,36 @@ class SuperConverter {
...
-   * Generate document hash (async, lazy)
+   * Generate document identifier hash from docId and created timestamp
+   * Uses CRC32 of the combined string for a compact identifier
+   * @returns {string} Hash identifier in format "HASH-XXXXXXXX"
    */
-  async #generateDocumentHash() {
-    if (!this.fileSource) return `HASH-${Date.now()}`;
+  #generateIdentifierHash() {
+    const docId = this.documentGuid || uuidv4();
+    const created = this.getDocumentCreatedTimestamp() || new Date().toISOString();
```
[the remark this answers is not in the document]
caio-pizzol: @andrii-harbour this is really important - we can't overwrite createdTimestamp everytime doc is opened

shared/common/Telemetry.ts
```diff
@@ -255,267 +89,68 @@ export class Telemetry {
...
-    await this.sendDataToTelemetry(report);
-  }
 
-  /**
-   * Sends data to the service
-   * @param data - Payload to send
-   * @returns Promise that resolves when data is sent
-   */
-  async sendDataToTelemetry(data: TelemetryPayload): Promise<void> {
+    console.log('[Telemetry] Sending payload:', payload);
```
caio-pizzol: should be removed - they'll log every payload to the console in production

packages/super-editor/src/core/Editor.ts
```diff
@@ -988,6 +1063,7 @@ export class Editor extends EventEmitter<EditorEventMap> {
     setTimeout(() => {
       if (this.isDestroyed) return;
       this.emit('create', { editor: this });
+      this.#trackDocumentOpen();
```
caio-pizzol: `#trackDocumentOpen()` is called from both `#emitCreateAsync()` and `mount()`

if an editor gets created and then mounted later, this could double-count. might want a guard flag.

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -640,8 +685,28 @@ class SuperConverter {
...
    */
   resolveDocumentGuid() {
+    const originalTimestamp = this.getDocumentCreatedTimestamp();
+
+    // If this is a new file created from template, update dcterms:created to current time
+    // This ensures documents created from the same template get unique identifiers
+    if (this.isNewFile) {
+      const newTimestamp = new Date().toISOString();
+      this.setDocumentCreatedTimestamp(newTimestamp);
+      console.debug('[super-converter] New file: updated dcterms:created', {
```
caio-pizzol: these debug statements should be removed before merging - unless you want to monitor them during internal testing?

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -669,43 +734,39 @@ class SuperConverter {
...
-      return this.documentGuid;
-    }
-
-    if (!this.documentHash && this.fileSource) {
-      this.documentHash = await this.#generateDocumentHash();
+  getDocumentIdentifier() {
+    // Generate and cache the hash if not already done
+    if (!this.documentHash) {
+      this.documentHash = this.#generateIdentifierHash();
+      console.debug('[super-converter] Generated document identifier:', {
```
caio-pizzol: same here

shared/common/Telemetry.ts
```diff
@@ -255,267 +89,68 @@ export class Telemetry {
...
+        body: JSON.stringify(payload),
+        credentials: 'omit',
       });
-
-      if (!response.ok) {
-        throw new Error(`Upload failed: ${response.statusText}`);
-      } else {
-        this.resetStatistics();
-      }
+      console.log('[Telemetry] Response status:', response.status);
```
caio-pizzol: these logs should be removed or wrapped in a dev check - especially the payload log which includes all telemetry data (final user would see them when inspecting)

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -588,6 +591,48 @@ class SuperConverter {
     return SuperConverter.setStoredCustomProperty(docx, 'SuperdocVersion', version, false);
   }
 
+  /**
+   * Get the dcterms:created timestamp from the already-parsed core.xml
+   * @returns {string|null} The created timestamp in ISO format, or null if not found
+   */
+  getDocumentCreatedTimestamp() {
```
caio-pizzol, line 616: `getDocumentCreatedTimestamp()` and `setDocumentCreatedTimestamp()` don't have dedicated tests - might be worth adding since they're key to document identity

packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -588,6 +591,48 @@ class SuperConverter {
...
+    );
+
+    return createdElement?.elements?.[0]?.text || null;
+  }
+
+  /**
+   * Set the dcterms:created timestamp in the already-parsed core.xml
+   * @param {string} timestamp - The timestamp to set (ISO format)
+   */
+  setDocumentCreatedTimestamp(timestamp) {
```
caio-pizzol, line 636: the `dcterms:created` element requires `xsi:type="dcterms:W3CDTF"` attribute per `ECMA-376 Part 2 §11.4 [M4.5]` -> verify this

current code only sets text content but doesn't verify the required attribute exists in document and/or create the element if missing

this might cause issues with strict OOXML validators

shared/common/Telemetry.ts
```diff
@@ -255,267 +89,68 @@ export class Telemetry {
...
-      size: file.size,
-      lastModified: file.lastModified ? new Date(file.lastModified).toISOString() : null,
-      type: file.type || 'docx',
-      internalId: options.internalId, // Microsoft's GUID if present
+  private async sendEvent(event: DocumentOpenEvent): Promise<void> {
+    const payload: TelemetryPayload = {
+      superdocVersion: this.superdocVersion,
+      browserInfo: this.getBrowserInfo(),
+      ...(this.metadata && { metadata: this.metadata }),
+      events: [event],
```
caio-pizzol, line 117: are we batching those events?

andrii-harbour:
@caio-pizzol forgot to mention. I intentionally left console logs and livetest url here for easy testing. Once code reviewed - I'll remove that of course

packages/super-editor/src/core/super-converter/SuperConverter.js
andrii-harbour: new file means there were no imported document provided and we are using our default blank.docx template

shared/common/Telemetry.ts
andrii-harbour: I'm leaving space for this. As for now it's one event at a time, but backend API is designed to be able to accept multiple events. Who knows what else we would like to add to the array

packages/super-editor/src/core/super-converter/SuperConverter.js
caio-pizzol: why not call it blankDoc or something? name it with something explicit and easy to understand

`blankDoc`?

shared/common/Telemetry.ts
```diff
@@ -1,252 +1,86 @@
...
+  browserInfo: BrowserInfo;
+  metadata?: Record<string, unknown>;
+  events: DocumentOpenEvent[];
 }
 
-export interface UnknownElement {
-  readonly elementName: string;
-  count: number;
-  attributes?: TelemetryAttributes;
+interface TelemetryOptions {
```
caio-pizzol: could simplify to `constructor(config: TelemetryConfig)` directly?

shared/common/Telemetry.ts
```diff
@@ -255,267 +89,68 @@ export class Telemetry {
...
-   * @private
+   * Disable telemetry
    */
-  generateId(): string {
-    const timestamp = Date.now();
-    const random = Array.from(randomBytes(4))
-      .map((b) => b.toString(16).padStart(2, '0'))
-      .join('');
-    return `${timestamp}-${random}`;
+  disable(): void {
```
caio-pizzol: are these methods used anywhere outside tests?

```
    disable(): void {                                                                                                                                                                                                                            
      this.enabled = false;                                                                                                                                                                                                                      
    }

    enable(): void {
      this.enabled = true;
    }

    isEnabled(): boolean {
      return this.enabled;
    }
```

caio-pizzol (approved):
LGTM - feel free to merge after addressing all comments and fixing unit-test

packages/super-editor/src/core/Editor.telemetry.test.ts
```diff
@@ -0,0 +1,229 @@
...
+}));
+
+// Test the telemetry initialization logic in isolation
+// This mirrors the #initTelemetry method in Editor.ts
+function initTelemetry(options: {
+  telemetry?: { enabled: boolean; endpoint?: string; metadata?: Record<string, unknown> } | null;
+  licenseKey?: string;
+}): Telemetry | null {
+  // Use default license key if not provided (mirrors Editor.options default)
+  const licenseKey = options.licenseKey ?? Editor.COMMUNITY_LICENSE_KEY;
```
harbournick: I feel like this makes this test slightly misleading because we're being opinionated by inserting the Editor community license key. This would hide the current issue with null key when undefined?

packages/superdoc/src/core/SuperDoc.js
```diff
@@ -72,6 +72,12 @@ export class SuperDoc extends EventEmitter {
     modules: {}, // Optional: Modules to load. Use modules.ai.{your_key} to pass in your key
     permissionResolver: null, // Optional: Override for permission checks
 
+    // License key for organization identification
+    licenseKey: null,
```
harbournick: Shouldn't we use the community key as default here? if licenseKey is passed in as undefined that might override the default in Editor.ts.

packages/super-editor/src/core/Editor.ts
```diff
@@ -449,13 +480,62 @@ export class Editor extends EventEmitter<EditorEventMap> {
...
+    // Skip if telemetry is not enabled
+    if (!telemetryConfig?.enabled) {
+      return;
+    }
+
+    try {
+      this.#telemetry = new Telemetry({
+        enabled: true,
+        endpoint: telemetryConfig.endpoint,
+        licenseKey,
```
harbournick: Could guard against the undefined thing (see other comments) here with  licenseKey ?? Editor.COMMUNITY_LICENSE_KEY

harbournick (changes requested):
Just the null licenseKey/default stuff otherwise looks good!

packages/super-editor/src/core/Editor.ts
andrii-harbour: what if the key is null or undefined? I think we still should send telemetry data. The only reason not to - is explicitly set `telemetry: false`

packages/superdoc/src/core/SuperDoc.js
andrii-harbour: you're right. Community key it is

packages/super-editor/src/core/Editor.ts
andrii-harbour: discussed. Agree about the comment

harbournick:

PR #1939, opened 2026-02-04
feat(super-editor): add w:lock support for StructuredContent nodes (SD-1616)

tupizz:
## Summary

Add ECMA-376 `w:lock` support for `StructuredContent` (inline) and `StructuredContentBlock` (block) nodes, enabling template variables to be read-only per the OOXML specification (§17.5.2.23).

Also adds **grouped hover** for multi-fragment block SDTs, **stale fragment rebuild detection**, **inline SDT wrapping for geometry rendering path**, and **SDT utility extraction** for better code organization.

## Lock Modes

| Lock Mode | SDT Wrapper | Content | Use Case |
|-----------|-------------|---------|----------|
| `unlocked` | Can delete | Can edit | Default, no restrictions |
| `sdtLocked` | Cannot delete | Can edit | Protect field structure |
| `contentLocked` | Can delete | Cannot edit | Read-only value, removable |
| `sdtContentLocked` | Cannot delete | Cannot edit | Fully protected field |

## Visual Comparison (Word vs SuperDoc)

CleanShot 2026-02-04 at 13 57 37@2x

**Border behavior:** Blue border (#629be7) appears on hover only, matching Word's behavior.

---

### 1. Plugin-Only Defense (No `contentEditable='false'`)

**Decision:** Block edits through the lock plugin instead of setting `contentEditable='false'` on locked content.

**Why:** Setting `contentEditable='false'` completely prevents cursor placement inside the node. This is poor UX because users:
- Can't select text to copy it
- Can't navigate through the document smoothly
- Experience jarring cursor behavior around locked regions

**Solution:** Keep content editable at the DOM level and rely on the plugin to block actual edits. Users can move cursor and select text freely while being prevented from modifying content.

### 2. Three-Layer Defense Strategy

The lock plugin uses three mechanisms to enforce locks:

| Layer | Hook | Purpose |
|-------|------|---------|
| 1 | `handleKeyDown` | Block Delete/Backspace/Cut **before** transaction is created |
| 2 | `handleTextInput` | Block typing in content-locked nodes |
| 3 | `filterTransaction` | Safety net for paste, drag-drop, programmatic changes |

**Why `handleKeyDown` instead of just `filterTransaction`?**

ProseMirror's transaction flow is:
```
User action → Browser event → handleKeyDown → Transaction created → filterTransaction → Applied
```

When `filterTransaction` blocks a transaction, the browser may have already modified DOM selection, causing the cursor to jump unexpectedly. By intercepting in `handleKeyDown` and calling `event.preventDefault()`, we prevent the browser from processing the event at all.

### 3. Step Relationship Analysis

To detect lock violations, we analyze the geometric relationship between a step's affected range and each SDT node:

```
Document: [----SDT----]
               pos    end

1. containsSDT: Would fully delete the SDT
   Step:    [================]
   SDT:        [--------]
   → Blocked for sdtLocked/sdtContentLocked

2. insideSDT: Modification entirely within SDT (but not deleting it)
   Step:          [--]
   SDT:        [--------]
   → Blocked for contentLocked/sdtContentLocked

3. crossesStart/crossesEnd: Step crosses SDT boundary
   Step:    [=======]        or        [=======]
   SDT:        [--------]         [--------]
   → Blocked for sdtLocked/sdtContentLocked (damages wrapper)
```

### 4. NodeView Provides Visual Feedback Only

The NodeView adds CSS classes for styling but does NOT set `contentEditable`:

```javascript
updateContentEditability() {
  if (this.dom) {
    this.dom.classList.toggle('sd-structured-content--content-locked', this.isContentLocked());
    this.dom.classList.toggle('sd-structured-content--sdt-locked', this.isSdtLocked());
  }
}
```

---

### Import/Export
- Parse `w:lock` element from `w:sdtPr` on DOCX import
- Export `w:lock` element to `w:sdtPr` on DOCX save
- Round-trip preservation of lock mode values

### Editor Behavior
- Add `lockMode` attribute to `StructuredContent` and `StructuredContentBlock` extensions
- Add `structuredContentLockPlugin` with 3-layer defense:
  - `handleKeyDown`: Block Delete/Backspace/Cut before transaction
  - `handleTextInput`: Block typing in content-locked nodes
  - `filterTransaction`: Safety net for programmatic changes
- NodeView adds CSS classes for visual feedback (allows cursor movement)

### Visual Styling (Presentation Mode)
- Add `StructuredContentLockMode` type to contracts
- Add `lockMode` to `StructuredContentMetadata` in style-engine
- Add `data-lock-mode` attribute rendering in DomPainter
- Add CSS styles matching Word's appearance for each lock mode
- Unified hover styles for block (`.sdt-hover`) and inline (`:hover`) SDTs with consistent `!important` overrides

### Inline SDT Wrapping for Geometry Rendering Path
When a line contains tabs, the DomPainter uses a geometry-based rendering path where elements are absolutely positioned. Previously, inline SDT wrapping only worked in the run-based path, causing tabs inside inline SDTs to lose their visual wrapper. Now:

- **`appendToLineGeo` helper** routes all geometry-path elements (tabs, images, field annotations, text segments) through SDT wrapping logic
- The wrapper uses `position: absolute` with computed left/width to span all contained children
- PM position tracking (`data-pm-start`/`data-pm-end`) is maintained on the wrapper element
- `TabRun` contract type extended with optional `sdt` metadata; pm-adapter passes SDT metadata through to tab runs

### Grouped Hover for Multi-Fragment Block SDTs
When a block SDT spans multiple paragraphs, each renders as a separate DOM fragment. Previously, CSS `:hover` only highlighted the individual fragment under the cursor. Now all fragments of the same SDT highlight simultaneously via JavaScript event delegation:

- **`SdtGroupedHover` class** (`utils/sdt-hover.ts`): Event delegation on mount container using `mouseover`/`mouseleave`. Finds all fragments sharing the same `data-sdt-id` and toggles `.sdt-hover` CSS class on all of them.
- **Post-render re-apply**: After DomPainter re-renders pages (e.g. during editing), newly created DOM elements lose the `.sdt-hover` class. `SdtGroupedHover.reapply()` is called at the end of `updateVirtualWindow()` to restore the class on all matching fragments.
- **CSS**: Block SDT rules changed from `:hover` to `.sdt-hover` selector. Inline SDTs remain `:hover`-based (single element, no grouping needed).

### Stale Fragment Rebuild Detection
- **`shouldRebuildForSdtBoundary`** (moved to `utils/sdt-helpers.ts`): Detects when a fragment's SDT boundary attributes (`data-sdt-container-start/end`) are stale and need rebuilding. Now correctly handles the case where a fragment leaves an SDT group (attributes present but should be removed).
- **`mappingUnreliable` check**: After a full-document paste of identical content, ProseMirror's transaction mapping becomes degenerate — it maps all old positions to the insertion end. This corrupted `data-pm-start`/`data-pm-end` attributes on reused spans, breaking cursor/click navigation. The new check verifies `mapping.map(oldPmStart) === fragment.pmStart` and forces a full fragment rebuild when the mapping is unreliable.

### Code Organization
- Extracted `SdtGroupedHover` class from renderer.ts into `utils/sdt-hover.ts`
- Moved `shouldRebuildForSdtBoundary` from renderer.ts into `utils/sdt-helpers.ts`
- Reduced renderer.ts SDT hover footprint from ~65 lines to 4 one-liner calls
- Removed unnecessary try/catch around `removeEventListener` in SdtGroupedHover
- Re-export `StructuredContentLockMode` from contracts instead of duplicating in node-attributes.ts

### Visual Testing Story
- **`sdt-lock-modes.ts`**: New interaction story demonstrating all lock modes:
  - Creates inline SDTs with `unlocked`, `sdtLocked`, `contentLocked` modes via `insertStructuredContentInline`
  - Creates block SDT with `sdtContentLocked` via `insertStructuredContentBlock`
  - Demonstrates typing inside `sdtLocked` SDT (content editable)
  - Demonstrates typing blocked inside `contentLocked` SDT
  - Demonstrates `updateStructuredContentById` to change lock modes programmatically
  - 8 milestones capturing each interaction step

---

## Test Coverage

**35 tests** covering all lock modes and operations:

- Wrapper deletion tests (4 modes × 2 node types = 8 tests)
- Content modification tests (4 modes × 2 node types = 8 tests)
- Boundary crossing tests (crosses start/end of SDT)
- Edge cases (empty document, multiple SDTs, collapsed selection)

```
Test Files  1 passed (1)
     Tests  35 passed (35)
```

---

## Files Changed

| Package | Files |
|---------|-------|
| `contracts` | `src/index.ts` — Add `StructuredContentLockMode` type, `sdt` on `TabRun` |
| `style-engine` | `src/index.ts` — Add `lockMode` normalization |
| `pm-adapter` | `inline-converters/tab.ts` — Pass SDT metadata to tab runs |
| `painter-dom` | `renderer.ts` — Grouped hover, post-render reapply, mapping reliability check, geometry-path SDT wrapping |
| `painter-dom` | `styles.ts` — Block SDT hover uses `.sdt-hover` class, unified hover styles with `!important` |
| `painter-dom` | `utils/sdt-helpers.ts` — Add `shouldRebuildForSdtBoundary`, fix stale attribute handling |
| `painter-dom` | `utils/sdt-hover.ts` — `SdtGroupedHover` class for multi-fragment hover |
| `super-editor` | Extensions — `lockMode` attribute for inline and block nodes |
| `super-editor` | NodeViews — CSS class toggling for lock state visual feedback |
| `super-editor` | Lock plugin — `structured-content-lock-plugin.js` (169 lines) |
| `super-editor` | Lock tests — `structured-content-lock-plugin.test.js` (416 lines, 35 tests) |
| `super-editor` | Import/Export — Parse and generate `w:lock` element |
| `super-editor` | Import/Export tests — Round-trip tests for lock modes |
| `super-editor` | Types — `node-attributes.ts` — Re-export `StructuredContentLockMode` from contracts |
| `visual-testing` | `stories/structured-content/sdt-lock-modes.ts` — Visual testing story for lock modes |

---

## Test Plan

- [x] Unit tests for import parsing (all 4 lock modes)
- [x] Unit tests for export generation
- [x] Unit tests for lock plugin (35 tests covering all scenarios)
- [x] Cursor movement works in locked content
- [x] Manual test with DOCX containing locked SDTs
- [x] Verify visual styling matches Word
- [x] Grouped hover highlights all fragments of multi-line block SDTs
- [x] Hover clears correctly on mouse leave
- [x] Editing inside hovered SDT preserves hover on new fragments
- [x] Full-document paste (Cmd+A, Cmd+C, Cmd+V) preserves cursor navigation
- [x] Tab inside inline SDT preserves visual wrapper (both rendering paths)
- [x] Visual testing story captures 8 milestones for all lock mode interactions

## Related

- IT-336

## Demo

https://github.com/user-attachments/assets/96b32892-d327-4a0f-8e29-8e9a378b7f9d

### DIFF
--- a/packages/super-editor/src/extensions/structured-content/structured-content.js
+++ b/packages/super-editor/src/extensions/structured-content/structured-content.js
@@ -1,6 +1,7 @@
 import { Node, Attribute } from '@core/index';
 import { StructuredContentInlineView } from './StructuredContentInlineView';
 import { createStructuredContentLockPlugin } from './structured-content-lock-plugin';
+import { createStructuredContentSelectPlugin } from './structured-content-select-plugin';
 
 export const structuredContentClass = 'sd-structured-content';
 export const structuredContentInnerClass = 'sd-structured-content__content';
@@ -115,7 +116,7 @@ export const StructuredContent = Node.create({
   },
 
   addPmPlugins() {
-    return [createStructuredContentLockPlugin()];
+    return [createStructuredContentLockPlugin(), createStructuredContentSelectPlugin()];
   },
 
   addNodeView() {
--- a/packages/super-editor/src/extensions/structured-content/structured-content-select-plugin.js
+++ b/packages/super-editor/src/extensions/structured-content/structured-content-select-plugin.js
@@ -1,0 +1,58 @@
+import { Plugin, TextSelection  } from 'prosemirror-state';
+
+/**
+ * Select-all-on-click plugin for inline StructuredContent nodes.
+ *
+ * When a click places a collapsed cursor inside an inline SDT and the previous
+ * selection was outside that SDT, the entire SDT content is selected. This
+ * matches Word's content control behavior: first click selects all for easy
+ * replacement, second click (cursor already inside) allows normal positioning.
+ *
+ * Uses appendTransaction so it works in both editing mode (PM DOM clicks) and
+ * presentation mode (PresentationEditor dispatched selections).
+ */
+export function createStructuredContentSelectPlugin() {
+  return new Plugin({
+    appendTransaction(transactions, oldState, newState) {
+      const { selection } = newState;
+
+      // Only for collapsed selections (cursor placement, not range selections)
+      if (!selection.empty) return null;
+
+      // Only when selection actually changed
+      if (oldState.selection.eq(newState.selection)) return null;
+
+      // Only for selection-only transactions (no doc changes — filters out
+      // typing, paste, etc. that also move the cursor)
+      if (transactions.some((tr) => tr.docChanged)) return null;
+
+      const $pos = selection.$from;
+
+      // Walk up to find an enclosing inline structuredContent node
+      for (let d = $pos.depth; d > 0; d--) {
+        const node = $pos.node(d);
+        if (node.type.name === 'structuredContent') {
+          const sdtStart = $pos.before(d);
+          const contentFrom = $pos.start(d);
+          const contentTo = $pos.end(d);
+
+          // Don't select empty content
+          if (contentFrom === contentTo) return null;
+
+          // If old selection was already inside this same SDT, allow normal
+          // cursor placement (second click / arrow navigation within SDT)
+          const old$pos = oldState.selection.$from;
+          for (let od = old$pos.depth; od > 0; od--) {
+            if (old$pos.node(od).type.name === 'structuredContent' && old$pos.before(od) === sdtStart) {
+              return null;
+            }
+          }
+
+          return newState.tr.setSelection(TextSelection.create(newState.doc, contentFrom, contentTo));
+        }
+      }
+
+      return null;
+    },
+  });
+}


packages/layout-engine/painters/dom/src/utils/sdt-hover.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (this.mount) {
+      if (this.onMouseOver) {
+        try {
+          this.mount.removeEventListener('mouseover', this.onMouseOver);
+        } catch {}
+      }
+      if (this.onMouseLeave) {
+        try {
+          this.mount.removeEventListener('mouseleave', this.onMouseLeave);
+        } catch {}
```
andrii-harbour: are those `catch {}` intended to be empty? 
If no - please add logic there, if yes - please leave the comment.

packages/super-editor/src/extensions/structured-content/structured-content-lock-plugin.js
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * Strategy:
+ * 1. handleKeyDown - Intercept keys BEFORE transaction to prevent browser selection issues
+ * 2. filterTransaction - Safety net to catch programmatic changes
+ */
+
+/**
+ * Collect all SDT nodes from the document
+ */
+function collectSDTNodes(doc) {
```
andrii-harbour, line 22: Every handleKeyDown, handleTextInput, and filterTransaction call walks the full document tree because of this function. For documents with many nodes this could be noticeable. Not now, but once we have this question being raised by customers consider:
    - Caching SDT nodes in plugin state (rebuild on docChanged transactions only)
    - Early-exit optimization: skip unlocked nodes during collection
    - For handleKeyDown/handleTextInput, you only need to check SDTs near the selection, not all of them

packages/super-editor/src/extensions/types/node-attributes.ts
```diff
@@ -593,6 +593,8 @@ export interface HardBreakAttrs extends InlineNodeAttributes {
 // STRUCTURED CONTENT
 // ============================================
 
+export type StructuredContentLockMode = 'unlocked' | 'sdtLocked' | 'contentLocked' | 'sdtContentLocked';
```
andrii-harbour: this is the same as in the `packages/layout-engine/contracts/src/index.ts` . Can we move this to some shared file?

andrii-harbour (changes requested):
overall very good! Nice job! 
I have only minor comments/suggestions there

andrii-harbour:
@tupizz it can also be a separate ticket as it's more about rendering than sdt logic itself - https://cdn.zappy.app/1bf739f84f32971a4be16dcc57ea71d7.mp4 

clicking `tab` inside the sdt container - brakes the UI of the sdt

packages/layout-engine/painters/dom/src/utils/sdt-hover.ts
tupizz: fixed!

packages/super-editor/src/extensions/structured-content/structured-content-lock-plugin.js
tupizz: fixed!

packages/super-editor/src/extensions/types/node-attributes.ts
tupizz: fixed!

demos/cdn/demo-config.json
```diff
@@ -1,7 +1,4 @@
 {
-    "tags": [
-        "editing",
-        "viewing"
-    ],
-    "title": "CDN"
-}
\ No newline at end of file
+  "tags": ["editing", "viewing"],
```
caio-pizzol: @tupizz install the pre-commit hook locally - so you avoid pushing formatting changes

packages/super-editor/src/extensions/structured-content/structured-content-lock-plugin.js
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * Strategy:
+ * 1. handleKeyDown - Intercept keys BEFORE transaction to prevent browser selection issues
+ * 2. filterTransaction - Safety net to catch programmatic changes
+ */
+
+/**
+ * Collect all SDT nodes from the document.
+ *
+ * TODO: For large documents, consider caching SDT nodes in plugin state
```
caio-pizzol: The plugin traverses every node in the document on every keydown, text input, and transaction.

This will impact our performance - I'd consider this a blocker. @harbournick

packages/super-editor/src/extensions/structured-content/structured-content-lock-plugin.js
```diff
@@ -0,0 +1,173 @@
...
+        return true;
+      }
+
+      const sdtNodes = collectSDTNodes(state.doc);
+      if (sdtNodes.length === 0) {
+        return true;
+      }
+
+      for (const step of tr.steps) {
+        if (step.from === undefined || step.to === undefined) {
```
caio-pizzol, line 170: `AttrStep`, `AddNodeMarkStep`, and `RemoveNodeMarkStep` use pos instead of from/to and get skipped here. these are niche (attribute changes, node-level marks) so the gap is minor, but a comment explaining why the skip is safe would help future readers.

packages/super-editor/src/extensions/structured-content/structured-content-lock-plugin.js
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        // Calculate the range that would be affected
+        let affectedFrom = from;
+        let affectedTo = to;
+
+        // If selection is collapsed, backspace/delete affects adjacent position
+        if (from === to) {
+          if (isBackspace && from > 0) {
+            affectedFrom = from - 1;
```
caio-pizzol, line 118: `from - 1` approximates backspace as single-character deletion, but ProseMirror's `joinBackward` can span node boundaries (e.g., joining paragraphs across an SDT boundary). 

`filterTransaction` catches these as a safety net, but the user may see a brief cursor jump before the transaction is rejected. Maybe not part of this feature initially, but worth noting.

packages/super-editor/src/extensions/structured-content/structured-content.js
```diff
@@ -104,6 +114,10 @@ export const StructuredContent = Node.create({
     ];
   },
 
+  addPmPlugins() {
```
caio-pizzol, line 118: the lock plugin handles both `structuredContent` and `structuredContentBlock` (line 29) but is only registered here in the inline extension.

if someone only loads `StructuredContentBlock`, locks won't be enforced. should the plugin be registered in both (using a guard against duplicate `PluginKeys`)?

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -5167,6 +5270,7 @@ export class DomPainter {
       this.setDatasetString(el, 'sdtScope', metadata.scope);
       this.setDatasetString(el, 'sdtTag', metadata.tag);
       this.setDatasetString(el, 'sdtAlias', metadata.alias);
+      this.setDatasetString(el, 'lockMode', metadata.lockMode || 'unlocked');
```
caio-pizzol, line 5273: existing lock attributes use `setDatasetBoolean` and only write when defined (`sdtSectionLocked`, `sdtFieldLocked`). this always writes `data-lock-mode="unlocked"`, which triggers `[data-lock-mode] { border-color: transparent }` on all SDTs.

is the border-hidden-by-default change for all SDTs intentional?

packages/layout-engine/painters/dom/src/styles.ts
```diff
@@ -394,7 +394,14 @@ const SDT_CONTAINER_STYLES = `
   text-overflow: ellipsis;
 }
 
-.superdoc-structured-content-block:hover .superdoc-structured-content__label {
+/* Hover effect for block structured content (via event delegation class) */
```
caio-pizzol: `[data-lock-mode].sdt-hover` and `[data-lock-mode]:hover` repeat the same 3 properties as the base hover rules above. since data-lock-mode is set on all SDTs including unlocked, the base rules never apply.

packages/layout-engine/painters/dom/src/utils/sdt-helpers.ts
```diff
@@ -257,6 +260,12 @@ export function applySdtContainerStyling(
   container.dataset.sdtContainerEnd = String(isEnd);
   container.style.overflow = 'visible'; // Allow label to show above
 
+  if (isStructuredContentMetadata(sdt)) {
+    container.dataset.lockMode = sdt.lockMode || 'unlocked';
```
caio-pizzol, line 264: both `applySdtContainerStyling` and `applySdtDataset` set data-lock-mode with the same `|| 'unlocked'` fallback. pick one place to own this.

packages/super-editor/src/extensions/structured-content/structured-content-lock-plugin.test.js
```diff
@@ -0,0 +1,416 @@
...
+  }
+
+  describe('wrapper deletion (sdtLocked behavior)', () => {
+    const wrapperDeletionCases = [
+      // [lockMode, nodeType, shouldBlock, description]
+      ['unlocked', 'structuredContent', false, 'allows deletion of unlocked inline SDT'],
+      ['unlocked', 'structuredContentBlock', false, 'allows deletion of unlocked block SDT'],
+      ['sdtLocked', 'structuredContent', true, 'blocks deletion of sdtLocked inline SDT'],
+      ['sdtLocked', 'structuredContentBlock', true, 'blocks deletion of sdtLocked block SDT'],
+      ['contentLocked', 'structuredContent', false, 'allows deletion of contentLocked inline SDT'],
```
caio-pizzol, line 95: `getNodeFromEditor` in tests/helpers/editor-test-utils.js:62 does the same thing.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -4628,6 +4650,80 @@ export class DomPainter {
...
+          geoSdtWrapper = null;
+          geoSdtId = null;
+        }
+      };
+
+      /**
+       * Append an element to the line, routing through an inline SDT wrapper
+       * when the run has inline structuredContent metadata.
+       */
+      const appendToLineGeo = (elem: HTMLElement, runForSdt: Run, elemLeftPx: number, elemWidthPx: number) => {
```
caio-pizzol, line 4675: the open/close/track-positions/append-label pattern is ~70 lines that mirrors the run-based path at ~4930-4970

if wrapper behavior changes later, both need updating. worth extracting a shared helper.

packages/layout-engine/painters/dom/src/utils/sdt-hover.ts
```diff
@@ -0,0 +1,75 @@
...
+  /** Attach hover listeners to the mount element. Safe to call again on remount. */
+  bind(mount: HTMLElement): void {
+    this.destroy();
+    this.mount = mount;
+
+    this.onMouseOver = (e: Event) => {
+      const target = (e.target as HTMLElement).closest?.(SDT_BLOCK_SELECTOR) as HTMLElement | null;
+      const sdtId = target?.dataset.sdtId ?? null;
+
+      if (this.hoveredSdtId && this.hoveredSdtId !== sdtId) {
```
caio-pizzol: when `sdtId === this.hoveredSdtId`, the handler still runs `querySelectorAll` for nothing.

maybe should we add `if (sdtId === this.hoveredSdtId) return;`.

packages/layout-engine/painters/dom/src/utils/sdt-hover.ts
```diff
@@ -0,0 +1,75 @@
+/**
+ * Grouped hover for multi-fragment SDT blocks.
+ *
+ * When a block SDT spans multiple paragraphs, each renders as a separate DOM element.
+ * This class uses event delegation to highlight ALL fragments of the same SDT
+ * simultaneously via the `.sdt-hover` CSS class.
+ */
+
+const SDT_BLOCK_SELECTOR = '.superdoc-structured-content-block[data-sdt-id]';
+const HOVER_CLASS = 'sdt-hover';
```
caio-pizzol: 'sdt-hover' not in DOM_CLASS_NAMES `constants.ts`

tupizz:
@caio-pizzol 

>Also, because we are adding visual interfaces, this PR will also require visual testing for rendering and interaction stories. Let me know if you have any questions on how to write those.

This is already defined in the PR, please check:
https://github.com/superdoc-dev/superdoc/pull/1939/changes#diff-93ff67f20943e2b47a11c268cf8642caa55bd4bebb7fd3ec83abb7bfab817ff2

packages/super-editor/src/extensions/structured-content/structured-content.js
tupizz: `getStarterExtensions()` always loads both. No current code path loads only the block extension. A guard against duplicate PluginKeys or  extracting to a shared plugin would be cleaner, but this is a defensive improvement for a scenario that doesn't exist today. Minor enhancement, not blocking.

tupizz: @caio-pizzol

packages/super-editor/src/extensions/structured-content/structured-content-lock-plugin.js
tupizz: Fixed in 97f28e1b6. The lock plugin now caches SDT nodes in ProseMirror plugin state (`state: { init, apply }`) and rebuilds only when `tr.docChanged`. All three handlers read from the cache via `STRUCTURED_CONTENT_LOCK_KEY.getState(state)` instead of calling `collectSDTNodes()` directly. This reduces document traversals from 2-3 per keystroke to at most 1 per document-changing transaction.

tupizz: Followed the same plugin state caching pattern used in `permission-ranges.js` (lines 246-264).

packages/super-editor/src/extensions/structured-content/structured-content-lock-plugin.js
tupizz: Added a comment in 1d15b4b49 explaining the skip: these steps (AttrStep, AddNodeMarkStep, RemoveNodeMarkStep) change metadata, not content, so they can't violate lock rules.

packages/super-editor/src/extensions/structured-content/structured-content-lock-plugin.js
tupizz: Added a comment in 1d15b4b49 noting the single-character approximation and that filterTransaction catches the real step range as a safety net. The three-layer defense was designed for this tradeoff the cursor jump risk is limited to paragraph boundaries immediately adjacent to locked SDTs.

packages/layout-engine/painters/dom/src/renderer.ts
tupizz: Intentional, this matches Word behavior where SDT borders are hidden by default and only appear on hover. Added comments in both `renderer.ts` (at the `lockMode || 'unlocked'` fallback) and `styles.ts` (at the `[data-lock-mode]` rule) explaining this intent. Fixed in 49aaaab5d.

packages/layout-engine/painters/dom/src/styles.ts
tupizz: Good catch, those base hover rules were dead code since `[data-lock-mode]` matches all SDTs. Removed the dead base rules and consolidated the block/inline lock-mode hover rules into a single grouped selector. Fixed in 49aaaab5d.

packages/layout-engine/painters/dom/src/utils/sdt-helpers.ts
tupizz: These are separate rendering paths for different DOM elements, `applySdtContainerStyling()` sets it on block SDT container divs, while `applySdtDataset()` sets it on inline SDT span elements. They can't be consolidated into one place since each function owns its own element type. The parallel logic is intentional.

packages/layout-engine/painters/dom/src/utils/sdt-hover.ts
tupizz: Good catch, added the early return in 0369553c7. `mouseover` fires on every pixel of movement, so this avoids unnecessary `querySelectorAll` + `classList.add` when hovering within the same SDT.

packages/super-editor/src/extensions/structured-content/structured-content-lock-plugin.test.js
tupizz: The local \`findSDTNode\` returns \`{node, pos, end}\` while \`getNodeFromEditor\` returns \`{node, pos}\` (no \`end\`). The tests need \`end\` for \`tr.delete(pos, end)\` — using the shared helper would require adding \`+ node.nodeSize\` boilerplate at every call site. The local helper is more ergonomic for this test context.

packages/layout-engine/painters/dom/src/renderer.ts
tupizz: Extracted the shared logic in a70f784a1. Three new private methods on DomPainter:
- `resolveRunSdtId(run)` — extracts inline SDT id from a run
- `createInlineSdtWrapper(sdt)` — creates the `<span>` with className, layoutEpoch, dataset, and label
- `expandSdtWrapperPmRange(wrapper, pmStart, pmEnd)` — updates min/max PM position tracking

Both paths now use these, removing ~30 lines of duplication while keeping their positioning logic separate (absolute vs inline flow).

packages/layout-engine/painters/dom/src/utils/sdt-hover.ts
tupizz: Added `SDT_HOVER` to `DOM_CLASS_NAMES` in `constants.ts` and updated `sdt-hover.ts` to use it (along with `DOM_CLASS_NAMES.BLOCK_SDT` for the selector). Fixed in a70f784a1.

harbournick:

PR #1941, opened 2026-02-04
bug: formatting lost when toggled off from toolbar (SD-1726 and SD-1727)

luccas-harbour:
- Problem (SD-1726): Toolbar formatting wasn’t applied to newly typed text when the selection was empty because active formatting was derived from the selection instead of stored marks.
- Problem (SD-1727): Toggling a mark off in the toolbar was unsetting that mark across existing text due to extending empty mark ranges, so previously formatted text lost styling.
- Fix 1: Use stored marks for empty selections in getActiveFormatting so typing inherits toolbar state, and disable extendEmptyMarkRange in toggleMarkCascade so “toggle off” only affects future input.
- Fix 2: Reworked inline run property calculation and run segmentation to split runs by inline overrides, preserve run attributes/marks when splitting, keep paragraph default runProperties in sync for the first run, and correctly map changed ranges across multi- transaction updates to avoid stale runProperties.
- Fix 3: prevented text nodes inside inline structured content nodes from being wrapped by runs
- Range utilities cleanup: Consolidated range helpers into @utils/rangeUtils.js, updated run/field-annotation/block-node plugins to the new { from, to } + docSize API, and aligned clamping behavior.
- Tests: Added/updated coverage for toolbar behavior, inline run splitting/mapping, selection preservation, and unified range utilities.

### DIFF
--- a/devtools/visual-testing/tests/interactions/stories/formatting/toggle-formatting-off.ts
+++ b/devtools/visual-testing/tests/interactions/stories/formatting/toggle-formatting-off.ts
@@ -1,0 +1,47 @@
+import { defineStory } from '@superdoc-testing/helpers';
+
+export default defineStory({
+  name: 'toggle-formatting-off',
+  description: 'Test that toggling bold off from toolbar only affects selected text',
+  tickets: ['SD-1727'],
+  startDocument: 'styles/sd-1727-formatting-lost.docx',
+  toolbar: 'full',
+  layout: true,
+  hideCaret: false,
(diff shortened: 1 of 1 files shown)


harbournick (approved):
LGTM. FYI I added an interactions story for both cases: https://github.com/superdoc-dev/superdoc/pull/1941/changes/ccd4dbc4a623e6af4200197ff66a83d5dd301c07

harbournick:

PR #1949, opened 2026-02-05
fix: persist comments on reload in collab mode

palmer-cl:
• Fixed collaboration refresh issues for comments and document content.

  ### Summary

  - Added comment hydration from Yjs on init/rejoin so sidebar comments auto-populate without
    manual comments injection.
  - Ensured collaboration comment sync initializes once and re-hydrates safely when the comments store
    becomes available.
  - Prevented duplicate comment sync listeners/observers during repeated init paths.

  ### Tests

  - Added focused collaboration tests for comment population:
      - Hydration from ydoc (including importedId normalization + dedupe).
      - Re-init behavior that re-hydrates store without duplicating listeners.

### DIFF
--- a/packages/superdoc/src/core/collaboration/helpers.js
+++ b/packages/superdoc/src/core/collaboration/helpers.js
@@ -13,12 +13,28 @@ export const loadCommentsFromYdoc = (superdoc) => {
   if (!superdoc?.ydoc || !superdoc?.commentsStore) return false;
   const commentsArray = superdoc.ydoc.getArray('comments');
   const comments = commentsArray.toJSON();
-  const seen = new Set();
+  const seenCommentIdByKey = new Map();
   const filtered = [];
   comments.forEach((c) => {
     const key = c?.importedId ?? c?.commentId;
-    if (!key || seen.has(key)) return;
-    seen.add(key);
+    if (!key) return;
+    if (seenCommentIdByKey.has(key)) {
+      const existingCommentId = seenCommentIdByKey.get(key);
+      const currentCommentId = c?.commentId;
+
+      if (existingCommentId && currentCommentId && existingCommentId !== currentCommentId) {
+        console.warn(
+          `[SuperDoc] Duplicate collaboration comment key "${key}" detected with conflicting commentId values. Keeping first entry and dropping duplicate.`,
+          {
+            key,
+            keptCommentId: existingCommentId,
+            droppedCommentId: currentCommentId,
+          },
+        );
+      }
+      return;
+    }
+    seenCommentIdByKey.set(key, c?.commentId);
     if (!c?.commentId) {
       filtered.push({ ...c, commentId: key });
       return;


palmer-cl:
@harbournick @caio-pizzol - let me know what the best way is to test this. I tested in multiple tabs, but Im not sure if theres a more robust way to test this PR?

caio-pizzol (commented):
Tested locally with `?collab=1` — the core fix works: comments hydrate from ydoc on rejoin. Two items to discuss.

caio-pizzol:
> @harbournick @caio-pizzol - let me know what the best way is to test this. I tested in multiple tabs, but Im not sure if theres a more robust way to test this PR?

Hey, Cole. Yeah, the only way to test this is if you have two collaboration sessions opened and then you only refresh one of them. The collaboration service will keep persistence on the document, so you can see the comments being populated on refresh.

image

packages/superdoc/src/SuperDoc.vue
```diff
@@ -537,6 +542,7 @@ const onEditorCommentLocationsUpdate = (doc, { allCommentIds: activeThreadId, al
     commentsStore.clearEditorCommentPositions?.();
     return;
   }
+  if (!allCommentPositions || Object.keys(allCommentPositions).length === 0) return;
```
caio-pizzol: This unconditionally drops empty position updates. If a user deletes all comments, the editor will emit {} for positions - this guard silently ignores it, so stale position data remains in the store.

The store-level guard in `comments-store.js:827` is better since it has the `commentsList.value.length > 0 condition`. Should we remove this Vue-level check and let the store guard handle it?

packages/superdoc/src/core/collaboration/helpers.js
```diff
@@ -3,6 +3,35 @@ import useComment from '../../components/CommentsLayer/use-comment';
...
+ * @returns {boolean} True if comments were loaded into the store
+ */
+export const loadCommentsFromYdoc = (superdoc) => {
+  if (!superdoc?.ydoc || !superdoc?.commentsStore) return false;
+  const commentsArray = superdoc.ydoc.getArray('comments');
+  const comments = commentsArray.toJSON();
+  const seen = new Set();
+  const filtered = [];
+  comments.forEach((c) => {
+    const key = c?.importedId ?? c?.commentId;
```
caio-pizzol, line 19: Is that the expected behavior for re-imported documents?

Using `importedId ?? commentId` as the dedup key - if two ydoc entries share the same `importedId` but have different `commentIds`, one gets silently dropped. 

Rest of the codebase is consistent with this (`getCommentPositionKey` and `getComment` both prefer `importedId`), just want to confirm.

packages/superdoc/src/SuperDoc.vue
palmer-cl: Yes. Absolutely we can drop and just push this guard to the store.  Should be updated.

packages/superdoc/src/core/collaboration/helpers.js
palmer-cl: Yeah I was just matching the other behaviors.. but silent dropping is not ideal. We could do a console.warn, or we could check timestamps for same importId but different commentIds and take the newest? 

What do you think?

caiopizzol: Let's do a console.warn for now!

I think it is not likely to happen tbh

palmer-cl: Awesome. added a console.warn

harbournick:

PR #1957, opened 2026-02-05
feat: (esign) introduce viewOptions for document layout and deprecate layoutMode and layoutMargins

andrii-harbour:
- Added `viewOptions` to control document layout with options for 'web' and 'print'.
- Deprecated `layoutMode` and `layoutMargins`, recommending CSS for margin control.
- Updated documentation and examples to reflect the new configuration.

### DIFF
--- a/packages/esign/src/index.tsx
+++ b/packages/esign/src/index.tsx
@@ -209,8 +209,8 @@ const SuperDocESign = forwardRef<Types.SuperDocESignHandle, Types.SuperDocESignP
         modules: {
           comments: false,
         },
-        viewOptions: document.viewOptions ?? {
-          layout: document.layoutMode === 'responsive' ? 'web' : 'print',
+        viewOptions: {
+          layout: document.viewOptions?.layout ?? (document.layoutMode === 'responsive' ? 'web' : 'print'),
         },
         onReady: () => {
           // Guard callback execution if cleanup already ran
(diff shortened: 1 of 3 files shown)


packages/esign/src/index.tsx
```diff
@@ -238,16 +238,7 @@ const SuperDocESign = forwardRef<Types.SuperDocESignHandle, Types.SuperDocESignP
...
-    document.source,
-    document.mode,
-    document.layoutMode,
-    document.layoutMargins?.top,
-    document.layoutMargins?.bottom,
-    document.layoutMargins?.left,
-    document.layoutMargins?.right,
-    discoverAndApplyFields,
-  ]);
+  }, [document.source, document.mode, document.layoutMode, document.viewOptions, discoverAndApplyFields]);
```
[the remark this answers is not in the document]
caio-pizzol: ☝️ 

The old code intentionally used primitives to avoid this (note the "Compare margin primitives" comment that was here).

Use `document.viewOptions?.layout` instead.

packages/esign/src/types.ts
```diff
@@ -72,15 +72,22 @@ export interface DocumentConfig {
     };
   };
   /**
+   * Document view options (recommended for v1.x):
```
caio-pizzol: nit: these say "v1.x" but the package is v2.0.0 now -- should update the version references.

packages/esign/src/index.tsx
```diff
@@ -209,9 +209,9 @@ const SuperDocESign = forwardRef<Types.SuperDocESignHandle, Types.SuperDocESignP
         modules: {
           comments: false,
         },
-        // @ts-expect-error - layoutMode is a valid SuperDoc option
-        layoutMode: document.layoutMode,
-        layoutMargins: document.layoutMargins,
+        viewOptions: document.viewOptions ?? {
```
caio-pizzol: minor: if a consumer passes `viewOptions: {}` (not null), the ?? won't trigger, so {} gets forwarded

Unlikely in practice but might be worth a check.

caio-pizzol (approved):
looking good! make sure to test it locally and make sure it works with the new options

after addressing relevant comments feel free to merge

packages/esign/src/types.ts
andrii-harbour: it is about SD v1.x, but I agree it's not clear enough

PR #1959, opened 2026-02-06
fix(table): resolve column resize only working on first page (SD-1772)

tupizz:
## Summary

Fixes table column resize handles not appearing on tables beyond the first page of a document.

- **Issue**: [SD-1772 — Table Columns Only Resizable on First Page of Document](https://linear.app/superdocworkspace/issue/SD-1772/table-columns-only-resizable-on-first-page-of-document)

## Demo

https://github.com/user-attachments/assets/d5f805b7-5887-49a2-803e-95d0ce555e50

## Root Cause

`updateOverlayRect()` in `TableResizeOverlay.vue` was computing the overlay position relative to the **table element's `offsetParent`** (the `.superdoc-page` containing the table), but the overlay itself is rendered as a child of `div.super-editor` — whose `position: absolute` resolves against `.super-editor-container`, a completely different coordinate space.

**DOM hierarchy:**
```
div.super-editor-container  ← overlay's actual offset parent
  div.super-editor
    div.editor-element
      div.superdoc-page[page=1]  ← table's offset parent (page 1)
        div.superdoc-table-fragment
      div.superdoc-page[page=2]  ← table's offset parent (page 2)
        div.superdoc-table-fragment
    <TableResizeOverlay />  ← position: absolute, sibling of editor-element
```

The old code did:
```js
const parent = props.tableElement.offsetParent; // → .superdoc-page
left = tableRect.left - parentRect.left;  // position within the page
top = tableRect.top - parentRect.top;     // position within the page
```

This produced **page-relative** coordinates (e.g., `top: 150px` within page 2), but the overlay's `position: absolute` resolved against `super-editor-container`. On page 1, the numbers coincidentally aligned since page 1 starts near the top of the container. On page 2+, the overlay was placed near the top of the editor — completely off-screen from the actual table.

## Fix

Changed `updateOverlayRect()` to compute coordinates relative to the **overlay's own `offsetParent`** (`overlayEl.value.offsetParent`) instead of the table's offset parent. This ensures the overlay is correctly positioned regardless of which page the table is on.

```js
// Before (broken): used table's offset parent (.superdoc-page)
const parent = props.tableElement.offsetParent;

// After (fixed): use overlay's own offset parent (.super-editor-container)
const referenceEl = overlayEl.value?.offsetParent;
```

## Why This Doesn't Cause Side Effects

- The fix only changes which element is used as the coordinate reference — the math (`getBoundingClientRect` subtraction) is identical
- On page 1, both the old and new reference elements produce the same result (the offset difference is negligible)
- The overlay's hover detection (`isNearColumnBoundary` in `SuperEditor.vue`) is unaffected — it uses viewport-relative `getBoundingClientRect` coordinates which work correctly on all pages
- Drag behavior is unaffected — it uses delta calculations (`event.clientX - initialX`) that are position-independent

## Test Plan

- [x] Open a multi-page document with tables on different pages
- [x] Verify column resize handles appear when hovering column borders on page 1
- [x] Verify column resize handles appear when hovering column borders on page 2+
- [x] Verify column drag-to-resize works correctly on all pages
- [x] Verify resize handles track correctly during scroll
- [x] Test with imported DOCX files containing tables on multiple pages

### DIFF
--- a/packages/super-editor/src/components/TableResizeOverlay.vue
+++ b/packages/super-editor/src/components/TableResizeOverlay.vue
@@ -1,6 +1,12 @@
 <template>
   <!-- Prevent mousedown from propagating to editor - critical for clean resize handle drags -->
-  <div v-if="visible && tableMetadata" class="superdoc-table-resize-overlay" :style="overlayStyle" @mousedown.stop>
+  <div
+    ref="overlayEl"
+    v-if="visible && tableMetadata"
+    class="superdoc-table-resize-overlay"
+    :style="overlayStyle"
+    @mousedown.stop
+  >
     <!-- Resize handles for each column boundary segment -->
     <template
       v-for="(boundary, resizableBoundaryIndex) in resizableBoundaries"
@@ -54,6 +60,7 @@ const props = defineProps({
 
 const emit = defineEmits(['resize-start', 'resize-move', 'resize-end', 'resize-success', 'resize-error']);
 
+const overlayEl = ref(null);
 const overlayRect = ref(null);
 /**
  * Parsed table metadata from data-table-boundaries attribute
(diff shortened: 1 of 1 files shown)


harbournick:

PR #1963, opened 2026-02-07
SD-1830 - use fragment.lines for click position in multi-column layouts

mattConnHarbour:


packages/layout-engine/layout-bridge/src/index.ts
```diff
@@ -968,7 +985,23 @@ export function clickToPosition(
         });
         return null;
       }
-      const line = measure.lines[lineIndex];
+
+      const line = lines[lineIndex];
+      // Convert to absolute index when using fragment-local lines
+      if (lines === fragment.lines) {
```
caio-pizzol: minor: using reference equality (`lines === fragment.lines`) for branch detection is fragile -- if someone copies the array, this silently breaks

a `const isRemeasured` boolean flag set at the branch point would be more explicit.

low risk but either way worth double checking

packages/layout-engine/layout-bridge/test/clickToPosition.test.ts
```diff
@@ -100,3 +100,243 @@ describe('hitTestPage with pageGap', () => {
...
+    //              x=350
+    //
+    // Without the fix: TypeError because measure.lines[1] is undefined
+    // With the fix: uses fragment.lines to find line, returns valid position
+    // ───────────────────────────────────────────────────────────────────────
+    const result = clickToPosition(twoColumnLayout, [twoColumnBlock], [originalMeasure], { x: 350, y: 50 });
+
+    expect(result).not.toBeNull();
+    expect(result?.blockId).toBe('two-column-para');
+    expect(result?.pos).toBeGreaterThanOrEqual(1);
```
caio-pizzol, line 278: the range assertions (1-12) accept any position in the paragraph -- they prove the crash is fixed but wouldn't catch a regression where clicking line 2 ("world") maps to a line 1 position.

tightening to `expect(result?.pos).toBeGreaterThanOrEqual(7)` for the second-line test would verify the correct line was hit.

packages/layout-engine/layout-bridge/src/index.ts
```diff
@@ -959,7 +959,24 @@ export function clickToPosition(
     const { fragment, block, measure, pageIndex, pageY } = fragmentHit;
     // Handle paragraph fragments
     if (fragment.kind === 'para' && measure.kind === 'paragraph' && block.kind === 'paragraph') {
-      const lineIndex = findLineIndexAtY(measure, pageY, fragment.fromLine, fragment.toLine);
+      // Use fragment-specific lines when available (remeasured fragments in multi-column layouts).
+      // When a paragraph is remeasured for column width, fragment.lines contains the actual
+      // rendered lines which may differ from measure.lines.
+      let lines: Line[];
```
caio-pizzol: nit: the renderer already handles this in one line (renderer.ts:2034):
```                                                                           
  const lines = fragment.lines ?? measure.lines.slice(fragment.fromLine, fragment.toLine);
```
this would replace the 3 let declarations + if/else block, and `findLineIndexAtY` can just take `(lines, pageY, 0, lines.length)`.

the slice cost is negligible for a click handler. also removes the need for the reference equality check at line 991.

packages/layout-engine/layout-bridge/src/index.ts
```diff
@@ -968,7 +985,23 @@ export function clickToPosition(
...
       }
-      const line = measure.lines[lineIndex];
+
+      const line = lines[lineIndex];
+      // Convert to absolute index when using fragment-local lines
+      if (lines === fragment.lines) {
+        lineIndex = fragment.fromLine + lineIndex;
+      }
+
+      // Guard against undefined line (defensive check)
```
caio-pizzol: `findLineIndexAtY` already returns `null` when `lines[i]` is `undefined` (line 2127-2128), and if it returns a non-null index, every line in the range was verified to exist during iteration.

this guard can never trigger -- removing it avoids making a reader wonder "wait, can this actually be `undefined`?"

packages/layout-engine/layout-bridge/src/index.ts
caio-pizzol: stale jsdoc: `@param`measure - The paragraph measure containing line data -- the parameter was renamed to lines: `Line[]` but the doc still references measure.

mattConnHarbour:
@caio-pizzol Thanks for the review! Just addressed your feedback - could you merge when you get a chance?

harbournick:

PR #1966, opened 2026-02-07
fix(super-editor): allow Backspace to delete empty paragraphs in suggesting mode

tupizz:
## Demo

https://github.com/user-attachments/assets/f3d9fd12-1682-4418-9547-8737f02d6b27

## Summary

Fixes [SD-1810](https://linear.app/superdocworkspace/issue/SD-1810/backspace-doesnt-delete-empty-paragraph-in-suggesting-mode) — Backspace doesn't delete empty paragraph in suggesting mode.

**MVP blocker for Ontra** (Feb 17 release).

## Problem

When a user creates a new paragraph (Enter) in suggesting mode and immediately presses Backspace, nothing happens — the empty paragraph stays.

### Root cause

The track changes system intercepts every transaction in suggesting mode via `trackedTransaction` → `replaceStep()` → `markDeletion()`. The flow:

1. Backspace on an empty paragraph creates a `ReplaceStep(from, to, Slice.empty)` to remove it
2. `replaceStep()` inverts the original step and tries to re-apply it as a tracked change
3. `markDeletion()` iterates `nodesBetween(from, to)` looking for **inline nodes** to mark as deleted
4. An empty paragraph has **no inline content** — the loop finds nothing
5. The deletion is silently swallowed as a no-op

## Fix

Added an early return guard at the top of `replaceStep()`. Before entering the full tracked-change flow, it checks:

1. **Is this a pure deletion?** — `step.slice.content.size === 0` (no replacement content)
2. **Is the range empty of inline content?** — `nodesBetween` finds no inline nodes

When both are true, the step is applied directly since there's nothing to track as a change. This matches Word behavior — deleting an empty paragraph has no visible "tracked change" to show.

For all other cases (non-empty paragraphs, text content, etc.), the existing tracked-change flow runs unchanged.

## Changes

- **`replaceStep.js`** — 6-line guard clause at function entry (lines 24–42)
- **`replaceStep.test.js`** — new test: creates `<p>Hello</p><p></p>`, simulates Backspace on empty paragraph, asserts it's removed

## Test plan

- [x] New unit test passes: `deletes empty paragraph on Backspace in suggesting mode`
- [x] All 5,499 existing tests pass (622 test files, 0 failures)
- [x] Manual browser verification: Enter → Backspace cycle works correctly in suggesting mode
- [x] Test with multi-paragraph documents (empty paragraph between content paragraphs)
- [x] Test that non-empty paragraph deletion still gets tracked normally
- [x] Test undo (Ctrl+Z) after deleting empty paragraph

### DIFF
--- a/packages/super-editor/src/core/extensions/keymap-history.test.js
+++ b/packages/super-editor/src/core/extensions/keymap-history.test.js
@@ -1,7 +1,7 @@
 import { describe, it, expect, afterEach } from 'vitest';
 import { closeHistory, undoDepth } from 'prosemirror-history';
 import { initTestEditor } from '@tests/helpers/helpers.js';
-import { handleEnter } from './keymap.js';
+import { handleEnter, handleBackspace, handleDelete } from './keymap.js';
 
 describe('keymap history grouping', () => {
   let editor;
@@ -117,5 +117,86 @@ describe('keymap history grouping', () => {
     // Undo should restore the deleted word
     editor.commands.undo();
     expect(editor.state.doc.textContent).toBe('hello world');
+  });
+
+  it('Backspace creates a new undo group boundary', () => {
+    ({ editor } = initTestEditor({ mode: 'text', content: '<p></p>' }));
+
+    // Create two paragraphs: type, Enter, type
+    insertText(editor, 'hello');
+    handleEnter(editor);
+    insertText(editor, 'world');
+    const depthBeforeBackspace = undoDepth(editor.state);
+
+    // Move cursor to start of second paragraph so joinBackward succeeds
+    let secondParaStart = null;
+    editor.state.doc.forEach((_node, offset, index) => {
+      if (index === 1) secondParaStart = offset + 1;
+    });
+    editor.view.dispatch(
+      editor.state.tr.setSelection(editor.state.selection.constructor.create(editor.state.doc, secondParaStart)),
+    );
+
+    // Backspace at start of second paragraph → joins paragraphs
+    handleBackspace(editor);
+
+    insertText(editor, ' after');
+    const depthAfterBackspace = undoDepth(editor.state);
+
+    expect(depthAfterBackspace).toBeGreaterThan(depthBeforeBackspace);
+  });
+
+  it('undo after Backspace join restores paragraph break', () => {
+    ({ editor } = initTestEditor({ mode: 'text', content: '<p></p>' }));
+
+    insertText(editor, 'hello');
+    handleEnter(editor);
+    insertText(editor, 'world');
+
+    expect(editor.state.doc.childCount).toBe(2);
+
+    // Move cursor to start of second paragraph
+    let secondParaStart = null;
+    editor.state.doc.forEach((_node, offset, index) => {
+      if (index === 1) secondParaStart = offset + 1;
+    });
+    editor.view.dispatch(
+      editor.state.tr.setSelection(editor.state.selection.constructor.create(editor.state.doc, secondParaStart)),
+    );
+
+    // Backspace joins paragraphs
+    handleBackspace(editor);
+    expect(editor.state.doc.childCount).toBe(1);
+
+    // Undo should restore the paragraph break
+    editor.commands.undo();
+    expect(editor.state.doc.childCount).toBe(2);
+  });
+
+  it('Delete creates a new undo group boundary', () => {
+    ({ editor } = initTestEditor({ mode: 'text', content: '<p></p>' }));
+
+    // Create two paragraphs
+    insertText(editor, 'hello');
+    handleEnter(editor);
+    insertText(editor, 'world');
+    const depthBeforeDelete = undoDepth(editor.state);
+
+    // Move cursor to end of first paragraph so joinForward succeeds
+    let firstParaEnd = null;
+    editor.state.doc.forEach((node, offset, index) => {
+      if (index === 0) firstParaEnd = offset + node.nodeSize - 1;
+    });
+    editor.view.dispatch(
+      editor.state.tr.setSelection(editor.state.selection.constructor.create(editor.state.doc, firstParaEnd)),
+    );
+
+    // Delete at end of first paragraph → joins paragraphs
+    handleDelete(editor);
+
+    insertText(editor, ' after');
+    const depthAfterDelete = undoDepth(editor.state);
+
+    expect(depthAfterDelete).toBeGreaterThan(depthBeforeDelete);
   });
 });
--- a/packages/super-editor/src/core/extensions/keymap.js
+++ b/packages/super-editor/src/core/extensions/keymap.js
@@ -5,6 +5,10 @@ import { isMacOS } from '../utilities/isMacOS.js';
 
 export const handleEnter = (editor) => {
   const { view } = editor;
+  // Close the current undo group so this structural action becomes its own undo step.
+  // Note: this fires before the command chain, so if no command succeeds (rare — e.g.
+  // Enter with no valid split target) an empty undo boundary is created. Acceptable
+  // trade-off vs. the complexity of post-hoc closeHistory after commands.first.
   view?.dispatch?.(closeHistory(view?.state?.tr));
 
   return editor.commands.first(({ commands }) => [
@@ -18,6 +22,7 @@ export const handleEnter = (editor) => {
 
 export const handleBackspace = (editor) => {
   const { view } = editor;
+  // Close undo group — see comment in handleEnter.
   view?.dispatch?.(closeHistory(view?.state?.tr));
 
   return editor.commands.first(({ commands, tr }) => [
@@ -38,7 +43,8 @@ export const handleBackspace = (editor) => {
 
 export const handleDelete = (editor) => {
   const { view } = editor;
-  editor?.view?.dispatch?.(closeHistory(view?.state?.tr));
+  // Close undo group — see comment in handleEnter.
+  view?.dispatch?.(closeHistory(view?.state?.tr));
 
   return editor.commands.first(({ commands }) => [
     () => commands.deleteSkipEmptyRun(),
--- a/packages/super-editor/src/extensions/block-node/block-node.js
+++ b/packages/super-editor/src/extensions/block-node/block-node.js
@@ -242,6 +242,26 @@ export const BlockNode = Extension.create({
       tr.setNodeMarkup(pos, undefined, nextAttrs, node.marks);
     };
 
+    /**
+     * Ensures a block node has a unique sdBlockId, assigning a new UUID if the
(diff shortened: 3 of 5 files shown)


harbournick (changes requested):
I think there's a bug with: When a user presses Backspace at the start of a non-empty paragraph, the paragraphs are joined, but the delete is applied directly and no deletion mark is created. That makes the edit invisible (not reviewable/rejectable) in Suggesting mode. The direct-delete fast path should only apply to removing a truly empty block node.

see 7ebce03b51a0111bf2483ab6931cd8dd8818e943 (I added a test that fails here for this case)

tupizz:
> I think there's a bug with: When a user presses Backspace at the start of a non-empty paragraph, the paragraphs are joined, but the delete is applied directly and no deletion mark is created. That makes the edit invisible (not reviewable/rejectable) in Suggesting mode. The direct-delete fast path should only apply to removing a truly empty block node.
> 
> see [7ebce03](https://github.com/superdoc-dev/superdoc/commit/7ebce03b51a0111bf2483ab6931cd8dd8818e943) (I added a test that fails here for this case)

We have an existing bug that we don't track new line breaks inside a paragraph, or enters in a document, or backspace in paragraphs.  its a separate issue altogether. In that case let's leave it alone in 1966 we can proceed without it.

.tupizz/docs/sd-1810-backspace-empty-paragraph-suggesting-mode.md
```diff
@@ -0,0 +1,178 @@
+# SD-1810: Backspace Doesn't Delete Empty Paragraph in Suggesting Mode
```
caio-pizzol: remove this one before merging

packages/super-editor/src/core/extensions/keymap.js
```diff
@@ -13,6 +17,9 @@ export const handleEnter = (editor) => {
 };
 
 export const handleBackspace = (editor) => {
+  const { view } = editor;
+  view?.dispatch?.(closeHistory(view?.state?.tr));
```
caio-pizzol, line 26: `closeHistory` runs before the command chain, so it fires even when nothing happens (e.g., Backspace at the start of the doc, Delete at the end). In that case the user gets an invisible undo step -- they press Ctrl+Z, nothing changes, they have to press it again to undo the real action.                                                                 

Two options:
1. Keep as-is and accept the edge case (it's rare in practice)
2. Move closeHistory to run only when a command actually succeeds -- though that's harder since `commands.first` is fire-and-forget

packages/super-editor/src/extensions/block-node/block-node.js
```diff
@@ -331,6 +334,9 @@ export const BlockNode = Extension.create({
 
             const docSize = newState.doc.content.size;
             const mergedRanges = mergeRanges(rangesToCheck, docSize);
+            // Track seen sdBlockIds across all ranges to detect duplicates
+            // (e.g., when tr.split() copies the original paragraph's sdBlockId to the new one).
+            const seenBlockIds = new Set();
```
caio-pizzol, line 353: the `seenIds.has(currentId)` + `seenIds.add(currentId)` pattern is repeated in 3 traversal paths.

duplication predates this PR, but now it's 3 places to update if dedup logic changes.

packages/super-editor/src/core/extensions/keymap.js
```diff
@@ -30,6 +37,9 @@ export const handleBackspace = (editor) => {
 };
 
 export const handleDelete = (editor) => {
+  const { view } = editor;
+  editor?.view?.dispatch?.(closeHistory(view?.state?.tr));
```
caio-pizzol: `editor?.view?.dispatch` here vs `view?.dispatch` in the other two handlers.

same thing since view is destructured above -- just inconsistent.

packages/super-editor/src/core/extensions/keymap-history.test.js
```diff
@@ -0,0 +1,121 @@
+import { describe, it, expect, afterEach } from 'vitest';
```
caio-pizzol, line 1: tests only cover `handleEnter`. `handleBackspace`/`handleDelete` got the same `closeHistory` but no tests.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
```diff
@@ -21,6 +21,26 @@ import { findMarkPosition } from './documentHelpers.js';
  * @param {number} options.originalStepIndex Original step index.
  */
 export const replaceStep = ({ state, tr, step, newTr, map, user, date, originalStep, originalStepIndex }) => {
+  // Handle structural deletions with no inline content (e.g., empty paragraph removal,
+  // paragraph joins). When there's no content being inserted and no inline content in
+  // the deletion range, markDeletion has nothing to mark — apply the step directly.
+  if (step.from !== step.to && step.slice.content.size === 0) {
```
caio-pizzol, line 33: a paragraph with only TrackDelete-marked text passes the `hasInlineContent` check, falls through to `markDeletion` which skips already-deleted nodes.

delete gets silently swallowed. is that intentional?

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
```diff
@@ -21,6 +21,26 @@ import { findMarkPosition } from './documentHelpers.js';
...
+    let hasInlineContent = false;
+    newTr.doc.nodesBetween(step.from, step.to, (node) => {
+      if (node.isInline) {
+        hasInlineContent = true;
+        return false;
+      }
+    });
+
+    if (!hasInlineContent) {
+      if (!newTr.maybeStep(step).failed) {
```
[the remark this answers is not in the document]
caio-pizzol: @tupizz check if relevant, add a comment of the final action it was taken

packages/super-editor/src/core/extensions/keymap.js
```diff
@@ -1,8 +1,12 @@
+import { closeHistory } from 'prosemirror-history';
 import { Extension } from '../Extension.js';
 import { isIOS } from '../utilities/isIOS.js';
 import { isMacOS } from '../utilities/isMacOS.js';
 
 export const handleEnter = (editor) => {
+  const { view } = editor;
+  view?.dispatch?.(closeHistory(view?.state?.tr));
```
caio-pizzol, line 12: `closeHistory` fires even when the command chain does nothing (e.g., Backspace at doc start) - user gets an invisible undo step -- Ctrl+Z does nothing, have to press again.

not blocking but worth checking.

.tupizz/docs/sd-1810-backspace-empty-paragraph-suggesting-mode.md
tupizz: Done — removed in commit dcd5c9a.

packages/super-editor/src/core/extensions/keymap.js
tupizz: Good catch. Added a comment explaining the trade-off: this fires before the command chain, so if no command succeeds (rare — e.g. Backspace at doc start) an empty undo boundary is created. The alternative would be calling closeHistory *after* `commands.first` succeeds, but that adds complexity to detect which commands actually produce structural changes. Accepted as a minor trade-off with documentation. See 8f90140.

packages/super-editor/src/extensions/block-node/block-node.js
tupizz: Agreed — extracted into an `ensureUniqueSdBlockId(node, nextAttrs, seenIds)` helper that encapsulates the has/add/uuid logic. All 3 traversal paths now call it. See 8f90140.

packages/super-editor/src/core/extensions/keymap.js
tupizz: Fixed — changed to `view?.dispatch?.(closeHistory(view?.state?.tr))` to match the other two handlers. See 8f90140.

packages/super-editor/src/core/extensions/keymap-history.test.js
tupizz: Added tests for both. Three new test cases:
- `Backspace creates a new undo group boundary` — joins two paragraphs via Backspace at start of second paragraph, verifies undo depth increases
- `undo after Backspace join restores paragraph break` — joins then undoes, verifies childCount goes back to 2
- `Delete creates a new undo group boundary` — joins via Delete at end of first paragraph, verifies undo depth increases

See 8f90140.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
tupizz: This is actually handled correctly — added a comment explaining the edge case. When a paragraph has only TrackDelete-marked text, `hasInlineContent` returns true so we skip the early return and fall through to the normal tracking flow. `markDeletion` skips the already-deleted nodes (as expected), but the structural join still happens through the replace machinery — it's not swallowed. The join merges the blocks while preserving existing deletion marks on the text content. See the comment in 8f90140.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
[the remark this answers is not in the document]
tupizz: Addressed — the Copilot/CodeX suggestions about atom blocks and broader scoping don't apply here. The early-return guard specifically targets structural deletions (empty paragraph removal, paragraph joins) where the deletion range has no inline content at all. Atom blocks like `passthroughBlock` are not inline nodes — they wouldn't be found by the `node.isInline` check inside `nodesBetween`. The guard correctly identifies ranges with only block boundary tokens (opening/closing tags) and applies them directly since `markDeletion` would have nothing to mark. Added a comment documenting the TrackDelete-only paragraph edge case as well. See 8f90140.

packages/super-editor/src/core/extensions/keymap.js
tupizz: Same as above — added a comment documenting the trade-off. The complexity of post-hoc closeHistory (detecting which commands are structural, calling it only on success) outweighs the minor UX impact of an occasional invisible undo step. See 8f90140.

packages/super-editor/src/core/extensions/keymap.js
caio-pizzol: @harbournick @luccas-harbour do you think this could cause any issue later on?

tupizz:
thanks @caio-pizzol 
I added visual interaction tests already 🙏🏻

harbournick:

PR #1968, opened 2026-02-08
fix(layout-bridge): defer table fragment click mapping to geometry fallback

tupizz:
## Summary

This PR contains three fixes for table interaction in the layout engine:

### 1. Fix cursor placement in table cells (SD-1788)

**Problem:** Clicking on table cell padding, borders, or areas where `elementsFromPoint` doesn't return a line element caused the cursor to land in the wrong column.

**Root cause:** `clickToPositionDom` in `dom-mapping.ts` used `processFragment` for table fragments even when no line was in the hit chain. `processFragment` searches all lines across all cells using only Y-coordinate matching, so it picks the first cell with a matching Y — which is often the wrong column.

**Fix:** Added a guard that returns `null` for table fragments without a line in the hit chain, deferring to the geometry fallback (`hitTestTableFragment`) which correctly uses both X and Y coordinates to resolve the target cell.

**File:** `packages/layout-engine/layout-bridge/src/dom-mapping.ts`

---

### 2. Fix table resize handles overflowing on split tables

**Problem:** When a table splits across page boundaries, the column resize handles (blue vertical lines) extended far beyond the table fragment — overflowing into the page gap and neighboring content below.

**Root cause:** In `renderTableFragment.ts`, the `data-table-boundaries` segment computation iterated over **all rows** in the table block (`0 to rowCount`), but each fragment only renders a subset of rows (`fromRow` to `toRow` + optional repeated headers). This caused every fragment to get segments with the full table's height.

**Fix:** Changed the segment computation to build a `renderedRows` list that mirrors the exact row rendering logic.

**File:** `packages/layout-engine/painters/dom/src/table/renderTableFragment.ts`

---

### 3. Fix table cell click on page 2+ (geometry fallback coordinate fix)

**Problem:** Clicking in empty space below text in a table cell worked on page 1 but failed on page 2+. The cursor either didn't appear or landed in the wrong position.

**Root cause:** Two issues compounding:

1. **Snap-to-nearest picks wrong target:** When clicking empty table cell space (below text lines), the DOM-based mapping returns `null`. The geometry fallback's snap-to-nearest logic then picks a nearby paragraph instead of letting `hitTestTableFragment` handle the click.

2. **Coordinate double-subtraction on page 2+:** `normalizeClientPoint` adjusts Y by the page's DOM offset (`pageOffsetY`), making `containerPoint` page-relative. On page 1 this offset is ~0, but on page 2+ it's significant. Then `clickToPosition` subtracts `pageTopY` again, producing incorrect negative coordinates that cause `hitTestPage` to find the wrong page and `hitTestTableFragment` to fail entirely.

**Fix:**
- Skip snap-to-nearest when the click is within a table fragment's bounds (let `hitTestTableFragment` handle it)
- When DOM info is available, determine the page directly from `elementsFromPoint` (same technique `normalizeClientPoint` uses) and treat `containerPoint` as already page-relative, avoiding the double subtraction

**File:** `packages/layout-engine/layout-bridge/src/index.ts`

## Test plan

- [x] Added unit tests for table fragment click fallback behavior
- [x] Added unit test for split-table segment scoping
- [x] Added unit tests for table cell empty space clicks (page 1)
- [x] Added unit tests for table cell clicks on page 2 (multi-page geometry path)
- [x] All 1205 layout-bridge tests pass
- [x] All 9105 tests pass across the monorepo
- [x] Manual: verify cursor placement in table cells on page 1 and page 2+
- [x] Manual: verify clicking below text in a table cell places cursor correctly
- [x] Manual: open advanced-tables.docx, verify resize handles stay within each fragment

### DIFF
--- a/devtools/visual-testing/tests/interactions/stories/basic-commands/table-cell-click-positioning.ts
+++ b/devtools/visual-testing/tests/interactions/stories/basic-commands/table-cell-click-positioning.ts
@@ -1,0 +1,119 @@
+import { defineStory } from '@superdoc-testing/helpers';
+
+const WAIT_MS = 400;
+const WAIT_LONG_MS = 600;
+
+/**
+ * SD-1788: Cursor cannot be reliably placed in table cell
+ *
+ * Bug: Clicking in empty space inside a table cell (below the text line)
+ * would snap the cursor to a nearby paragraph outside the table instead
+ * of placing it in the correct cell. This happened because:
+ *
+ * 1. DOM-based click mapping (clickToPositionDom) searched all lines across
+ *    all cells using only Y matching, picking the wrong column
+ * 2. The snap-to-nearest fallback chose a nearby paragraph over the table cell
+ *
+ * This test verifies:
+ * 1. Clicking inside a table cell places cursor in that cell
+ * 2. Clicking in different cells moves cursor to the correct cell
+ * 3. Clicking below text in a tall cell still resolves to that cell
+ * 4. Table works correctly on page 2+ (multi-page coordinate mapping)
+ */
+export default defineStory({
+  name: 'table-cell-click-positioning',
+  description: 'Test that clicks in table cells map to correct cursor positions',
+  tickets: ['SD-1788'],
+  startDocument: null,
+  layout: true,
+  hideCaret: false,
+  hideSelection: false,
+
+  async run(page, helpers): Promise<void> {
+    const { step, focus, type, press, clickAt, waitForStable, milestone, executeCommand } = helpers;
+
+    // Step 1: Create content above the table (snap-to-nearest candidate)
+    await step('Type paragraph above table', async () => {
+      await focus();
+      await type('Paragraph above the table');
+      await waitForStable(WAIT_MS);
+      await press('Enter');
+      await press('Enter');
+      await waitForStable(WAIT_MS);
+      await milestone('paragraph-above', 'Paragraph above where table will be inserted');
+    });
+
+    // Step 2: Insert a table
+    await step('Insert table', async () => {
+      await executeCommand('insertTable', { rows: 3, cols: 3, withHeaderRow: false });
+      await waitForStable(WAIT_LONG_MS);
+      await milestone('table-inserted', 'Inserted 3x3 table');
+    });
+
+    // Step 3: Type content in first cell
+    await step('Type in first cell', async () => {
+      await type('Cell A1');
+      await waitForStable(WAIT_MS);
+      await milestone('cell-a1-text', 'Typed in first cell (A1)');
+    });
+
+    // Step 4: Navigate to second cell and type
+    await step('Type in second cell', async () => {
+      await press('Tab');
+      await waitForStable(WAIT_MS);
+      await type('Cell B1');
+      await waitForStable(WAIT_MS);
+      await milestone('cell-b1-text', 'Typed in second cell (B1)');
+    });
+
+    // Step 5: Navigate to third cell and type
+    await step('Type in third cell', async () => {
+      await press('Tab');
+      await waitForStable(WAIT_MS);
+      await type('Cell C1');
+      await waitForStable(WAIT_MS);
+      await milestone('cell-c1-text', 'Typed in third cell (C1)');
+    });
+
+    // Step 6: Navigate to second row, first cell
+    await step('Type in second row', async () => {
+      await press('Tab');
+      await waitForStable(WAIT_MS);
+      await type('Cell A2');
+      await waitForStable(WAIT_MS);
+      await milestone('cell-a2-text', 'Typed in row 2 first cell (A2)');
+    });
+
+    // Step 7: Click back into cell A1 — the key test
+    // This tests that clicking in a table cell with content places the cursor
+    // in that cell, not in the paragraph above.
+    // Table starts at ~y=127, rows are ~17px each:
+    //   Row 1: y≈127-144, Row 2: y≈144-161, Row 3: y≈161-178
+    // Use short verification text "!" to avoid wrapping which shifts row heights.
+    await step('Click into first cell (A1)', async () => {
+      await clickAt(100, 135);
+      await waitForStable(WAIT_MS);
+      await type('!');
+      await waitForStable(WAIT_MS);
+      await milestone('verify-click-a1', 'Clicked and typed "!" in A1 — should appear in first cell');
+    });
+
+    // Step 8: Click into cell B1 (middle column, same row)
+    await step('Click into middle cell (B1)', async () => {
+      await clickAt(280, 135);
+      await waitForStable(WAIT_MS);
+      await type('!');
+      await waitForStable(WAIT_MS);
+      await milestone('verify-click-b1', 'Clicked and typed "!" in B1 — should appear in middle cell');
+    });
+
+    // Step 9: Click into cell A2 (second row)
+    await step('Click into second row cell (A2)', async () => {
+      await clickAt(100, 175);
+      await waitForStable(WAIT_MS);
+      await type('!');
+      await waitForStable(WAIT_MS);
+      await milestone('verify-click-a2', 'Clicked and typed "!" in A2 — should appear in second row');
+    });
+  },
+});


packages/layout-engine/layout-bridge/src/index.ts
```diff
@@ -873,23 +873,61 @@ export function clickToPosition(
...
+  // offset, making containerPoint page-relative rather than container-space. On page 1
+  // the offset is ~0 so it doesn't matter, but on page 2+ this causes hitTestPage to
+  // find the wrong page and pageRelativePoint to be doubly subtracted.
+  //
+  // Fix: when DOM info is available, determine the page from elementsFromPoint (same
+  // technique normalizeClientPoint uses) and treat containerPoint as already page-relative.
+  let pageHit: PageHit | null = null;
+  let isContainerPointPageRelative = false;
+
+  if (domContainer != null && clientX != null && clientY != null) {
```
caio-pizzol, line 954: `findPageElement` in `dom-mapping.ts` and `normalizeClientPoint` in `PointerNormalization.ts` already do this.

`findPageElement` is in the same package and has error handling the new code lacks. worth reusing or extracting.

packages/layout-engine/layout-bridge/src/dom-mapping.ts
```diff
@@ -210,6 +210,15 @@ export function clickToPositionDom(domContainer: HTMLElement, clientX: number, c
     return result;
   }
 
+  // For table fragments without a direct line hit, return null to let the geometry
+  // fallback (hitTestTableFragment) handle cell lookup. processFragment searches all
+  // lines across all cells using only Y matching, which picks the wrong column when
+  // multiple cells share the same row height.
+  if (fragmentEl.classList.contains('superdoc-table-fragment')) {
```
caio-pizzol: `CLASS_NAMES`already has page, fragment, line, etc. this class should be there too, since it's now checked across packages.

packages/layout-engine/layout-bridge/src/index.ts
```diff
@@ -898,61 +936,76 @@ export function clickToPosition(
   let fragmentHit = hitTestFragment(layout, pageHit, blocks, measures, pageRelativePoint);
 
   // If no fragment was hit (e.g., whitespace), snap to nearest hit-testable fragment on the page.
+  // But skip snap-to-nearest when the click is within a table fragment — otherwise the snap
+  // picks a nearby paragraph and returns its position, preventing clicks in empty table cell
+  // space (below text lines) from reaching hitTestTableFragment below.
   if (!fragmentHit) {
```
caio-pizzol, line 1003: the table guard wraps 30 lines in another if. fine for now, but `clickToPosition` is getting long.

extracting the snap logic into a small helper would keep it manageable.

packages/layout-engine/layout-bridge/test/clickToPosition.test.ts
```diff
@@ -100,3 +100,303 @@ describe('hitTestPage with pageGap', () => {
     expect(result?.pageIndex).toBe(1);
   });
 });
+
```
caio-pizzol, line 104: the two table test suites copy-paste nearly identical `tableBlock` / `tableMeasure` / `tableCellPara` objects. a shared builder would cut ~80 lines.

packages/layout-engine/painters/dom/src/table/renderTableFragment.ts
```diff
@@ -182,21 +182,45 @@ export const renderTableFragment = (deps: TableRenderDependencies): HTMLElement
...
     // boundarySegments[colIndex] = array of {fromRow, toRow, y, height} segments where this boundary exists
     const boundarySegments: Array<Array<{ fromRow: number; toRow: number; y: number; height: number }>> = [];
     for (let i = 0; i < columnCount; i++) {
       boundarySegments.push([]);
     }
 
-    // For each row, determine which grid columns have cell boundaries
+    // Build the list of rows actually rendered in this fragment, matching the
+    // rendering order: repeated headers first, then body rows.
+    const renderedRows: Array<{ rowIndex: number; height: number }> = [];
```
caio-pizzol, line 202: same header-then-body iteration exists at line ~291. they need to stay in sync. a cross-reference comment would help.

packages/layout-engine/layout-bridge/src/index.ts
```diff
@@ -873,23 +873,61 @@ export function clickToPosition(
...
+  // technique normalizeClientPoint uses) and treat containerPoint as already page-relative.
+  let pageHit: PageHit | null = null;
+  let isContainerPointPageRelative = false;
+
+  if (domContainer != null && clientX != null && clientY != null) {
+    const doc = domContainer.ownerDocument ?? (typeof document !== 'undefined' ? document : null);
+    if (doc && typeof doc.elementsFromPoint === 'function') {
+      const hitChain = doc.elementsFromPoint(clientX, clientY);
+      const pageEl = Array.isArray(hitChain)
+        ? (hitChain.find((el) => (el as HTMLElement)?.classList?.contains('superdoc-page')) as HTMLElement | null)
```
caio-pizzol: `dom-mapping.ts` in the same file's sibling uses `CLASS_NAMES.page`. this should too.

packages/layout-engine/layout-bridge/src/index.ts
tupizz: Done, exported `findPageElement` from `dom-mapping.ts` and replaced the duplicated page lookup in `clickToPosition` with a call to it. The inline `elementsFromPoint` + `classList.contains('superdoc-page')` block is gone. See 5b3523e.

packages/layout-engine/layout-bridge/src/dom-mapping.ts
tupizz: Agreed, added `TABLE_FRAGMENT: 'superdoc-table-fragment'` to `DOM_CLASS_NAMES` in `constants.ts`. Updated both `dom-mapping.ts` (uses `CLASS_NAMES.tableFragment`) and `renderTableFragment.ts` (uses `DOM_CLASS_NAMES.TABLE_FRAGMENT`) to reference the constant instead of hardcoding. See 5b3523e.

packages/layout-engine/layout-bridge/src/index.ts
tupizz: Done — extracted into `snapToNearestFragment(pageHit, blocks, measures, pageRelativePoint)` helper. The inline block in `clickToPosition` is now just the table-guard check + a single helper call. See 5b3523e.

packages/layout-engine/layout-bridge/test/clickToPosition.test.ts
tupizz: Added `buildTableFixtures(opts)` to `mock-data.ts` , takes optional `cellWidth`, `cellHeight`, `lineHeight`, `pmStart`, `pmEnd`, `text`, `blockId`. The `clickToPosition.test.ts` table test now uses it instead of inline definitions, cutting ~60 lines. See 5b3523e.

packages/layout-engine/painters/dom/src/table/renderTableFragment.ts
tupizz: Added cross-reference comments in both directions: the metadata segment builder (~line 199) references the rendering loop (~line 317), and vice versa. See 5b3523e.

packages/layout-engine/layout-bridge/src/index.ts
tupizz: Fixed, the duplicated page lookup in `index.ts` is now replaced by `findPageElement()` from `dom-mapping.ts`, which already uses `CLASS_NAMES.page`. No more hardcoded `'superdoc-page'` string in `clickToPosition`. See 5b3523e.

packages/layout-engine/layout-bridge/src/dom-mapping.ts
```diff
@@ -210,6 +210,15 @@ export function clickToPositionDom(domContainer: HTMLElement, clientX: number, c
...
+    log('Table fragment without line in hit chain, deferring to geometry fallback');
+    return null;
+  }
```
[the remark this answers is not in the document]
tupizz: Updated the comment to make the call-chain explicit: now says "return null so the caller (`clickToPosition` in index.ts) falls back to geometry-based hit testing via `hitTestTableFragment`". See 5b3523e.

packages/layout-engine/layout-bridge/src/dom-mapping.ts
```diff
@@ -210,6 +210,15 @@ export function clickToPositionDom(domContainer: HTMLElement, clientX: number, c
     return result;
   }
 
+  // For table fragments without a direct line hit, return null to let the geometry
+  // fallback (hitTestTableFragment) handle cell lookup. processFragment searches all
+  // lines across all cells using only Y matching, which picks the wrong column when
+  // multiple cells share the same row height.
+  if (fragmentEl.classList.contains('superdoc-table-fragment')) {
+    log('Table fragment without line in hit chain, deferring to geometry fallback');
+    return null;
```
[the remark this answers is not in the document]
tupizz: Done — added `TABLE_FRAGMENT` to the shared `DOM_CLASS_NAMES` constant in `@superdoc/painter-dom` and referenced it in both packages. See 5b3523e.

packages/layout-engine/layout-bridge/test/dom-mapping.test.ts
```diff
@@ -322,6 +322,81 @@ describe('DOM-based click-to-position mapping', () => {
...
+            </div>
+          </div>
+        </div>
+      `;
+
+      const line = container.querySelector('.superdoc-line') as HTMLElement;
+      const lineRect = line.getBoundingClientRect();
+
+      // Click directly on the line — elementsFromPoint should include the line
+      const result = clickToPositionDom(container, lineRect.left + 5, lineRect.top + 5);
```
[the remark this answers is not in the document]
tupizz: Good catch — updated the test to polyfill `document.elementsFromPoint` with a hit chain that includes the line element (matching real browser behavior for direct text clicks). The polyfill returns `[span, line, tableFragment, page, container, ...]` and restores the original after assertion. See 5b3523e.

packages/layout-engine/layout-bridge/src/dom-mapping.ts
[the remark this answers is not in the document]
tupizz: This was exactly the bug. Fixed by detecting if the click is within a table fragment's bounding box before running snap-to-nearest — when it is, we skip the snap entirely and let `hitTestTableFragment` handle it via geometry. The snap logic is now extracted into `snapToNearestFragment()` and only runs when `\!isWithinTableFragment`. See 5b3523e.

tupizz:
@caio-pizzol I'll be adding the visual testing

harbournick (approved):
LGTM

harbournick:

PR #1969, opened 2026-02-08
fix(tracked-changes): fix suggested insertions from paste failures

palmer-cl:
- insertion using a more flexible method (replaceRange) in a temporary transaction
- handle complex pasted content correctly
- ensure paste is inline and on the line we expect

### DIFF
--- a/packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
+++ b/packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
@@ -40,11 +40,15 @@ export const replaceStep = ({ state, tr, step, newTr, map, user, date, originalS
     }
   }
 
+  // When pasting into a textblock, try the open slice first so content merges inline
+  // instead of creating new paragraphs (prevents inserting block nodes into non-textblocks).
   const baseParentIsTextblock = trTemp.doc.resolve(positionTo).parent?.isTextblock;
   const shouldPreferInlineInsertion = step.from === step.to && baseParentIsTextblock;
 
   const tryInsert = (slice) => {
     const tempTr = state.apply(newTr).tr;
+    // Empty slices represent pure deletions (no content to insert).
+    // Detecting them ensures deletion tracking runs even if `tempTr` doesn't change.
     const isEmptySlice = slice?.content?.size === 0;
     try {
       tempTr.replaceRange(positionTo, positionTo, slice);
--- a/packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.test.js
+++ b/packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.test.js
@@ -300,6 +300,45 @@ describe('trackChangesHelpers replaceStep', () => {
     expect(insertedText).toContain('Plain Paste');
   });
 
+  it('tracks paste replacement over selected existing text', () => {
+    const doc = schema.nodes.doc.create(
+      {},
+      schema.nodes.paragraph.create({}, schema.nodes.run.create({}, [schema.text('Hello World')])),
+    );
+    let state = createState(doc);
+
+    const worldPos = findTextPos(state.doc, 'Hello World');
+    expect(worldPos).toBeTypeOf('number');
+    const from = worldPos + 'Hello '.length;
+    const to = from + 'World'.length;
+    state = state.apply(state.tr.setSelection(TextSelection.create(state.doc, from, to)));
+
+    const tempDiv = document.createElement('div');
+    tempDiv.innerHTML = '<p>Pasted</p>';
+    const parsedDoc = PMDOMParser.fromSchema(schema).parse(tempDiv);
+    const slice = new Slice(parsedDoc.content, 0, 0);
+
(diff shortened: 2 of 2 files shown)


packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
```diff
@@ -40,16 +48,40 @@ export const replaceStep = ({ state, tr, step, newTr, map, user, date, originalS
     }
   }
 
+  const baseParentIsTextblock = trTemp.doc.resolve(positionTo).parent?.isTextblock;
+  const shouldPreferInlineInsertion = step.from === step.to && baseParentIsTextblock;
+
   const tryInsert = (slice) => {
```
caio-pizzol, line 48: the old `tryInsert` was pretty lean - just "try the insertion, give me the positions." the new one does insertion + marking + slice extraction + result building all in one go.

i think we can keep it closer to the old shape: have `tryInsert` just return `{ tempTr, insertedFrom, insertedTo }`, then do the marking and slice extraction after we know which attempt won.

that also gets rid of the `buildInsertionResult` helper at the top since it's only used twice.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
```diff
@@ -40,16 +48,40 @@ export const replaceStep = ({ state, tr, step, newTr, map, user, date, originalS
     }
   }
 
+  const baseParentIsTextblock = trTemp.doc.resolve(positionTo).parent?.isTextblock;
+  const shouldPreferInlineInsertion = step.from === step.to && baseParentIsTextblock;
```
caio-pizzol, line 46: this is the actual fix for the bug but there's no comment explaining what it does.

something like "when pasting into a textblock, try the open slice first so content merges inline instead of creating new paragraphs" would go a long way for anyone reading this later.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
```diff
@@ -40,16 +48,40 @@ export const replaceStep = ({ state, tr, step, newTr, map, user, date, originalS
...
+
   const tryInsert = (slice) => {
-    const insertionStep = new ReplaceStep(positionTo, positionTo, slice, false);
-    if (trTemp.maybeStep(insertionStep).failed) return null;
-    return {
-      insertedFrom: insertionStep.from,
-      insertedTo: insertionStep.getMap().map(insertionStep.to, 1),
-    };
+    const tempTr = state.apply(newTr).tr;
+    const isEmptySlice = slice?.content?.size === 0;
```
caio-pizzol, line 52: this looks like a small optimization but it's actually important for correctness - without it, pure deletion steps (empty slice) would fail both `tryInsert` attempts, skip the deletion tracking at line 131, and the delete wouldn't get tracked at all.

worth a quick comment explaining that.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.test.js
```diff
@@ -1,5 +1,6 @@
 import { beforeEach, afterEach, describe, expect, it, vi } from 'vitest';
```
caio-pizzol, line 1: the three new tests all paste at a cursor position (no selection).

could we add one where you select some text and paste over it? that would cover the other branch where `shouldPreferInlineInsertion` is false and the try order flips.

caio-pizzol (approved):
looks good!

after addressing the comments feel free to merge it

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
```diff
@@ -40,16 +48,40 @@ export const replaceStep = ({ state, tr, step, newTr, map, user, date, originalS
...
+    });
+
+    return buildInsertionResult(tempTr, insertedFrom, insertedTo, insertedMark);
   };
 
-  const insertion = tryInsert(step.slice) || tryInsert(Slice.maxOpen(step.slice.content, true));
+  const openSlice = Slice.maxOpen(step.slice.content, true);
+  const insertion = shouldPreferInlineInsertion
+    ? tryInsert(openSlice) || tryInsert(step.slice)
+    : tryInsert(step.slice) || tryInsert(openSlice);
```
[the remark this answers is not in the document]
palmer-cl: This is valid, and updated.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
palmer-cl: sounds good! reverted to the old style `tryInsert` and removed the helper for simplicity.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceStep.js
palmer-cl: Added comments!

harbournick:

PR #1970, opened 2026-02-08
fix(tracked-changes): colors should be restored when format rejected

palmer-cl:
- Refactored track-change mark snapshot logic into shared helpers
- Fixed format suggestion reversibility using exact mark attr snapshot matching and consistent before/after upsert behavior, so rejecting a color suggestion restores original styling correctly.
 - Added regression coverage: an interaction test for “suggest color -> reject -> no inline color left” and new unit tests for markSnapshotHelpers (exact match, type fallback, upsert behavior, and range lookup).


packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
```diff
@@ -0,0 +1,67 @@
...
+    return true;
+  }
+
+  return attrsExactlyMatch(mark.attrs || {}, snapshot.attrs || {});
+};
+
+export const findMarkInRangeBySnapshot = ({ doc, from, to, snapshot }) => {
+  let exactMatch = null;
+  let typeOnlyMatch = null;
+  const shouldFallbackToTypeOnly = !snapshot?.attrs || Object.keys(snapshot.attrs).length === 0;
```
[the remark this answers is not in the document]
caio-pizzol: @palmer-cl worth taking a look, just add comment if relevant or not and close it

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  if (!exact) {
+    return true;
+  }
+
+  return attrsExactlyMatch(mark.attrs || {}, snapshot.attrs || {});
+};
+
+export const findMarkInRangeBySnapshot = ({ doc, from, to, snapshot }) => {
```
caio-pizzol, line 63: When `findMarkInRangeBySnapshot` returns `null`, we silently skip the `RemoveMarkStep`.

Consider adding a `console.warn` here so we can diagnose cases where rejection leaves stale formatting, e.g. `console.warn('[track-changes] could not find live mark for snapshot', newMark)`

caio-pizzol (changes requested):
Very clean implementation!

Unfortunately, it is not fully fixing it

Font color is "rejected" but the font family is not returning to the original one "Times New Roman" but to "Arial"

Original:
Image

After reject:
Image

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
palmer-cl: Updated!

palmer-cl:
> Very clean implementation!
> 
> Unfortunately, it is not fully fixing it
> 
> Font color is "rejected" but the font family is not returning to the original one "Times New Roman" but to "Arial"
> 
> Original: Image
> 
> After reject: Image

@caio-pizzol this should be fixed now. Refactored to make it more robust when tracking formatting

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
```diff
@@ -0,0 +1,75 @@
...
+  return Object.fromEntries(Object.entries(attrs).filter(([, value]) => value !== null && value !== undefined));
+};
+
+export const attrsExactlyMatch = (left = {}, right = {}) => {
+  const normalizedLeft = normalizeAttrs(left);
+  const normalizedRight = normalizeAttrs(right);
+  return objectIncludes(normalizedLeft, normalizedRight) && objectIncludes(normalizedRight, normalizedLeft);
+};
+
+export const markSnapshotMatchesStepMark = (snapshot, stepMark, exact = true) => {
```
caio-pizzol: three matching functions that are the same logic with different parameter types `markSnapshotMatchesStepMark(snapshot, stepMark)`, `markMatchesSnapshot(mark, snapshot)`, and `hasMatchingMark(marks, stepMark)` all do "does type match + do attrs match?"

but each handles different input shapes (snapshot.type vs mark.type.name vs mark.type). A single function with a type-name extractor would eliminate the duplication:

```
  const getTypeName = (m) => m.type?.name ?? m.type;
  const marksMatch = (a, b, exact = true) => {
    if (getTypeName(a) !== getTypeName(b)) return false;
    return !exact || attrsExactlyMatch(a.attrs || {}, b.attrs || {});
  };
```

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/getLiveInlineMarksInRange.js
```diff
@@ -0,0 +1,21 @@
+export const getLiveInlineMarksInRange = ({ doc, from, to }) => {
+  const marks = [];
+  const seen = new Set();
+
+  doc.nodesBetween(from, to, (node) => {
```
caio-pizzol, line 5: Both iterate `doc.nodesBetween`, check `node.isInline`, then deduplicate marks with `mark.type.name` + separator + `JSON.stringify(mark.attrs)` into a Set.

The `unsetAllMarks.js` version uses `\0` separator, this one uses `:`.

Not blocking since they're in different domains, but if more consumers appear this should be a shared utility in `core/helpers/.`

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.js
```diff
@@ -36,39 +44,36 @@ export const addMarkStep = ({ state, step, newTr, doc, user, date }) => {
...
+            attrs: { ...step.mark.attrs },
+          });
         }
       } else {
-        // before = [];
-        before = node.marks.map((mark) => ({
-          type: mark.type.name,
-          attrs: { ...mark.attrs },
-        }));
+        before = liveMarks
```
caio-pizzol, line 71: `liveMarks` is fetched from `newTr.doc` (which already includes the `addMark` from line 42), so the new mark attrs are captured in the snapshot.

Then track-change marks are filtered out. This means the before snapshot already reflects the incoming style change, not the state before it. Was this tested for the "add formatting on top of existing tracked formatting" scenario?

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
```diff
@@ -0,0 +1,75 @@
...
+
+  return attrsExactlyMatch(mark.attrs || {}, snapshot.attrs || {});
+};
+
+export const findMarkInRangeBySnapshot = ({ doc, from, to, snapshot }) => {
+  let exactMatch = null;
+  let typeOnlyMatch = null;
+  const shouldFallbackToTypeOnly = !snapshot?.attrs || Object.keys(snapshot.attrs).length === 0;
+
+  doc.nodesBetween(from, to, (node) => {
```
caio-pizzol, line 71: `return false` in `nodesBetween` only stops descent, not iteration. After `exactMatch` is found, sibling nodes are still visited needlessly.

Should we add `if (exactMatch)` return at the top of the callback?

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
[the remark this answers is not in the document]
palmer-cl: Yes we can loosen restrictions here. Updated.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
palmer-cl: good idea. updated.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/getLiveInlineMarksInRange.js
palmer-cl: Ok makes sense will keep an eye out for this pattern

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.js
palmer-cl: > Was this tested for the "add formatting on top of existing tracked formatting" scenario?

Can you elaborate on this one? do we have a VRT for that scenario that would fail in the test suite?

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
palmer-cl: Yes good idea. Added a short circuit return in the function.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.js
caio-pizzol: My concern about `liveMarks` was wrong - it's read before `newTr.addMark`, so the before snapshot is correct.    
                                                                                                             
No VRT for this. The "rejecting multi-format suggestions" test covers it partially but doesn't assert fontFamily restoration after rejection.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
```diff
@@ -0,0 +1,105 @@
+import { objectIncludes } from '@core/utilities/objectIncludes.js';
+
+const normalizeAttrs = (attrs = {}) => {
+  return Object.fromEntries(Object.entries(attrs).filter(([, value]) => value !== null && value !== undefined));
+};
+
+export const attrsExactlyMatch = (left = {}, right = {}) => {
+  const normalizedLeft = normalizeAttrs(left);
+  const normalizedRight = normalizeAttrs(right);
+  return objectIncludes(normalizedLeft, normalizedRight) && objectIncludes(normalizedRight, normalizedLeft);
```
afn: Could use lodash's `isEqual` instead.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
```diff
@@ -0,0 +1,105 @@
+import { objectIncludes } from '@core/utilities/objectIncludes.js';
+
+const normalizeAttrs = (attrs = {}) => {
+  return Object.fromEntries(Object.entries(attrs).filter(([, value]) => value !== null && value !== undefined));
```
afn, line 4: Could use lodash's `omitBy` to be a bit more succinct:
```suggestion
  return omitBy(attrs, value => value === null || value === undefined);
```

palmer-cl: this statement broke functionality. left what was above.

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/markSnapshotHelpers.js
palmer-cl: added!

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.js
palmer-cl: @caio-pizzol just confirming you want me to add an assertion for font family? Did you want me to add this to the new folders you send me through the loom? Or add to the tests I have as part of this PR?

caio-pizzol: Just as part of this PR through unit-test! But a behavior test would also be beneficial here.

caio-pizzol: Just added!

harbournick:
:tada: This PR is included in **superdoc** v1.14.0

The release is available on [GitHub release](https://github.com/superdoc-dev/superdoc/releases/tag/v1.14.0)

PR #1985, opened 2026-02-09
fix(super-converter): resolve table style conditional shading on cell import (SD-1833)

caio-pizzol:
before:
image

after:
image

word:
image

### DIFF
--- a/devtools/visual-testing/pnpm-lock.yaml
+++ b/devtools/visual-testing/pnpm-lock.yaml
@@ -2062,8 +2062,8 @@ packages:
     resolution: {integrity: sha512-l63NF9y/cLROq/yqKXSLtcMeeyOfnSQlfMSlzFt/K73oIaD8DGaQWd7Z34X9GPiKqP5rbSh84Hl4bOlLcjiSrQ==}
 
   superdoc@file:../../packages/superdoc/superdoc.tgz:
-    resolution: {integrity: sha512-riLJi3/uPgpUeuHBaisa67AfG0u/KtpSqzIx5TuJbMarSCmwDsasUgIWzE8+wGE2Wg/TwjICrK+ZEQwVlDAjXQ==, tarball: file:../../packages/superdoc/superdoc.tgz}
-    version: 1.11.0
+    resolution: {integrity: sha512-uv32JEnNC45h5pivFHQsOOP1rUnUucwHy9fGnWVwr483QNYDJkAXqbkKZVZPizj8+VBxX/ETiLP0tSpG4kXUfw==, tarball: file:../../packages/superdoc/superdoc.tgz}
+    version: 1.0.2
     peerDependencies:
       '@hocuspocus/provider': ^2.13.6
       pdfjs-dist: '>=4.3.136 <=4.6.82'


packages/super-editor/src/core/super-converter/v3/handlers/w/tc/helpers/legacy-handle-table-cell-node.js
```diff
@@ -267,6 +285,67 @@ function normalizeTableCellContent(content, editor) {
...
+  if (isLastColumn && isColStyleEnabled('lastColumn')) {
+    const color = getShading(referencedStyles?.lastCol);
+    if (color) return color;
+  }
+
+  // Banding: odd/even rows and columns
+  const isOddRow = rowIndex % 2 === 0;
+  if (isOddRow) {
+    const color = getShading(referencedStyles?.band1Horz);
+    if (color) return color;
```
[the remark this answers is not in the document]
caio-pizzol: Addressed — removed banding (band1Horz/band2Horz) entirely from this PR.

luccas-harbour:
@caio-pizzol I would advise against resolving the style cascade during import. This causes a property defined in a style to be assigned as an inline one during export. I believe this needs to be handled during rendering inside the layout-engine.

Also, table properties are similar to paragraph/run properties and I believe it would be simpler to implement the style cascade rules for all properties at once instead of handling each property separately. 

We used to have this individual property handling for paragraphs/runs, which was hard to maintain and buggy as not all properties were applying the cascading rules correctly. This was modified so that everything is handled in a unified manner [here](https://github.com/Harbour-Enterprises/SuperDoc/blob/0c4bddab0fd1c47e9530860492480748497ad51d/packages/layout-engine/style-engine/src/ooxml/index.ts).

I think it would make sense to have a similar approach in this case. What do you think?

caio-pizzol:
> @caio-pizzol I would advise against resolving the style cascade during import. This causes a property defined in a style to be assigned as an inline one during export. I believe this needs to be handled during rendering inside the layout-engine.
> 
> Also, table properties are similar to paragraph/run properties and I believe it would be simpler to implement the style cascade rules for all properties at once instead of handling each property separately.
> 
> We used to have this individual property handling for paragraphs/runs, which was hard to maintain and buggy as not all properties were applying the cascading rules correctly. This was modified so that everything is handled in a unified manner [here](https://github.com/Harbour-Enterprises/SuperDoc/blob/0c4bddab0fd1c47e9530860492480748497ad51d/packages/layout-engine/style-engine/src/ooxml/index.ts).
> 
> I think it would make sense to have a similar approach in this case. What do you think?

100% I totally missed this point, let me do a second pass to move it to layout-engine

Thanks for the headsup!

packages/super-editor/src/extensions/table/table-style-shading-plugin.js
luccas-harbour, line 1: I have a feeling that decorations are not being used anymore now with the layout-engine. It might be worth double checking if this code is having any effect on rendering.

packages/layout-engine/pm-adapter/src/converters/table.ts
```diff
@@ -198,6 +208,13 @@ const parseTableCell = (args: ParseTableCellArgs): TableCell | null => {
       }
     }
   }
+  // Fall back to resolved style shading if no inline background
+  if (!cellBackgroundColor && resolvedTcProps?.shading?.fill) {
+    const fill = resolvedTcProps.shading.fill;
+    if (fill !== 'auto') {
+      cellBackgroundColor = fill.startsWith('#') ? fill : `#${fill}`;
```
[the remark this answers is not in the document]
caio-pizzol: Not actionable. Table style conditional formatting uses val="clear" - no pattern blending needed.

packages/super-editor/src/extensions/table/table-style-shading-plugin.js
caio-pizzol: Removed

packages/layout-engine/pm-adapter/src/converters/table.ts
```diff
@@ -185,7 +185,17 @@ const parseTableCell = (args: ParseTableCellArgs): TableCell | null => {
...
+  const tableInfo = tableProperties ? { tableProperties, rowIndex, cellIndex, numCells, numRows } : undefined;
+  const resolvedTcProps = resolveTableCellProperties(
+    inlineTcProps as Parameters<typeof resolveTableCellProperties>[0],
+    tableInfo,
+    context.converterContext?.translatedLinkedStyles,
+  );
+
   // Extract cell background color for auto text color resolution
+  // Priority: inline background attr > resolved style shading
   const cellBackground = cellNode.attrs?.background as { color?: string } | undefined;
```
luccas-harbour, line 199: Perhaps the `cellBackground` should also be derived from `resolvedTcProps`? Otherwise if the background is defined in a style it wouldn't be applied correctly?

packages/layout-engine/style-engine/src/ooxml/index.ts
```diff
@@ -306,7 +312,7 @@ export function getNumberingProperties<T extends PropertyObject>(
 
   if (abstractProps != null) {
     if (levelDefinition?.styleId) {
-      abstractProps.styleId = levelDefinition?.styleId;
+      (abstractProps as Record<string, unknown>).styleId = levelDefinition?.styleId;
```
luccas-harbour: I think the generic `T` in `getNumberingProperties` should be typed as `ParagraphProperties | RunProperties`, which both have a `styleId` property so this cast wouldn't be necessary. The `getNumberingProperties` function will never be called with `TableCellProperties`.

packages/layout-engine/style-engine/src/ooxml/index.ts
```diff
@@ -388,6 +394,41 @@ export function resolveCellStyles<T extends PropertyObject>(
...
+  }
+
+  const cellStyleProps = resolveCellStyles<TableCellProperties>(
+    'tableCellProperties',
+    tableInfo,
+    translatedLinkedStyles,
+  );
+
+  if (cellStyleProps.length === 0) {
+    return (inlineProps ?? {}) as TableCellProperties;
```
luccas-harbour, line 420: I noticed that the `tableProperties` also have a `shading` attribute (accessible from `tableInfo.tableProperties.shading`). I wonder how it affects the cascading rules. 

This if statement will return nothing even if `shading` is defined for the table. It might be the case that that `shading` value doesn't apply at the cell level but maybe worth double checking?

packages/layout-engine/style-engine/src/ooxml/index.ts
caio-pizzol: Good catch. Narrowing the generic is cleaner than the cast.

packages/layout-engine/pm-adapter/src/converters/table.ts
caio-pizzol: Already handled — lines 211-216 fall back to `resolvedTcProps.shading.fill` when no inline background exists.

packages/layout-engine/style-engine/src/ooxml/index.ts
caio-pizzol: `tableProperties.shading` is table-level shading (w:tblPr/w:shd) - a background on the table element itself, not on cells. 

Cell-level shading from the table style comes from `tableStyleProperties.wholeTable.tableCellProperties.shading`, which `resolveCellStyles` already collects via `determineCellStyleTypes()`.

harbournick:

PR #1986, opened 2026-02-10
bug: can't load document with block level permission elements (SD-1840)

luccas-harbour:
- The `w:permStart` and `w:permEnd` tags can show up as block-level nodes but we used to support them only at the inline-level
- New `permStartBlock` and `permEndBlock` nodes have been added for this and the [permissionRanges plugin](https://github.com/Harbour-Enterprises/SuperDoc/blob/f81558dae3bed28fa0297bc68f015a9ee2181653/packages/super-editor/src/extensions/permission-ranges/permission-ranges.js#L232) has been updated to support them.

### DIFF
--- a/tests/behavior/tests/importing/sd-1840-permission-ranges.spec.ts
+++ b/tests/behavior/tests/importing/sd-1840-permission-ranges.spec.ts
@@ -1,0 +1,178 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { test, expect } from '../../fixtures/superdoc.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const DOC_PATH = path.resolve(__dirname, '../../../../test-corpus/permissions/sd-1840-perm-tags.docx');
+
+test.skip(
+  !fs.existsSync(DOC_PATH),
+  'Test document not available — expected test-corpus/permissions/sd-1840-perm-tags.docx',
+);
+
+test('loads block-level permission tags and enforces permission ranges in viewing mode (SD-1840)', async ({
+  superdoc,
+}) => {
+  await superdoc.loadDocument(DOC_PATH);
+  await superdoc.waitForStable();
+  await expect
+    .poll(
+      () =>
+        superdoc.page.evaluate(() => {
+          const editor = (window as any).superdoc?.activeEditor ?? (window as any).editor;
+          const doc = editor?.state?.doc;
+          if (!doc) return false;
+
+          let permStartBlockCount = 0;
+          let permEndBlockCount = 0;
+          doc.descendants((node: any) => {
+            if (node.type?.name === 'permStartBlock') permStartBlockCount += 1;
+            if (node.type?.name === 'permEndBlock') permEndBlockCount += 1;
+            return;
+          });
+
+          const ranges = editor.storage?.permissionRanges?.ranges ?? [];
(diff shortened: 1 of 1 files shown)


packages/super-editor/src/extensions/perm-start/perm-start.js
```diff
@@ -21,22 +39,24 @@ export const PermStart = Node.create({
...
-    };
+    return sharedAttributes();
+  },
+});
+
+export const PermStartBlock = Node.create({
+  name: 'permStartBlock',
+  group: 'block',
+  inline: false,
+  atom: true,
```
caio-pizzol: `PermStartBlock` and `PermEndBlock` are identical except for name and attributes -- could share a factory. also, inline `PermStart` is missing `atom: true` that other marker nodes have.

packages/super-editor/src/extensions/perm-end/perm-end.js
```diff
@@ -11,6 +11,18 @@ import { Node } from '@core/index.js';
  * @sidebarTitle PermEnd
  * @snippetPath /snippets/extensions/perm-end.mdx
  */
+const sharedAttributes = () => ({
```
caio-pizzol, line 15: both files call this `sharedAttributes()` but they return different shapes. rename to `permEndAttributes` / `permStartAttributes` so it's clear they're not the same.

packages/super-editor/src/extensions/permission-ranges/permission-ranges.js
```diff
@@ -255,7 +285,11 @@ export const PermissionRanges = Extension.create({
           apply(tr, value, _oldState, newState) {
             let permissionState = value;
             if (tr.docChanged) {
-              permissionState = buildPermissionState(newState.doc, getAllowedIdentifiers());
+              permissionState = buildPermissionState(
+                newState.doc,
```
caio-pizzol: schema doesn't change after init -- compute this once and reuse instead of calling `getPermissionTypeInfo` on every transaction.

packages/super-editor/src/core/super-converter/v3/handlers/w/perm-start/perm-start-translator.js
```diff
@@ -1,22 +1,25 @@
 // @ts-check
 import { NodeTranslator } from '@translator';
+import { isInlineContext } from '../../../../v2/importer/inlineContext.js';
```
caio-pizzol: v3 handler importing from `v2/importer/` -- move `isInlineContext` to a shared location like `super-converter/utils/`.

packages/super-editor/src/extensions/perm-end/perm-end.js
luccas-harbour: These are local functions defined in separate files so I don't think there's confusion here. Their names indicate they are shared between the inline and the block nodes

packages/super-editor/src/core/super-converter/v3/handlers/w/perm-start/perm-start-translator.js
luccas-harbour: done

packages/super-editor/src/extensions/perm-start/perm-start.js
luccas-harbour: done

packages/super-editor/src/extensions/permission-ranges/permission-ranges.js
luccas-harbour: done

harbournick:

PR #1988, opened 2026-02-10
bug: superdoc webpage misrendered text (SD-1819)

luccas-harbour:
- There was an issue that happened during export if a linked style contained an ampersand. The issue was that if that character was represented as `&amp;`, the ampersand would be double-escaped, turning into `&amp;amp;`
- The `sdpr.docx` test document had been exported from SD and because of that had the incorrect `&amp;amp;` entity. This had to be fixed manually.

### DIFF
--- a/packages/super-editor/src/core/super-converter/exporter.js
+++ b/packages/super-editor/src/core/super-converter/exporter.js
@@ -565,7 +565,7 @@ export class DocxExporter {
   #replaceSpecialCharacters(text) {
     if (text === undefined || text === null) return text;
     return String(text)
-      .replace(/&/g, '&amp;')
+      .replace(/&(?!#\d+;|#x[0-9a-fA-F]+;|(?:amp|lt|gt|quot|apos);)/g, '&amp;')
       .replace(/</g, '&lt;')
       .replace(/>/g, '&gt;')
       .replace(/"/g, '&quot;')
--- a/packages/super-editor/src/tests/export/docxExporter.test.js
+++ b/packages/super-editor/src/tests/export/docxExporter.test.js
@@ -40,6 +40,31 @@ describe('DocxExporter', () => {
     expect(xml).toContain('Format=&lt;&lt;NUM&gt;&gt;_&lt;&lt;VER&gt;&gt;');
   });
 
+  it('does not double-escape ampersands in text nodes', () => {
+    const exporter = new DocxExporter(createConverterStub());
+
+    const data = {
+      name: 'w:document',
+      attributes: {},
+      elements: [
+        {
+          name: 'w:t',
+          elements: [
+            {
+              type: 'text',
+              text: 'Rock & Roll &amp; Jazz',
+            },
+          ],
+        },
+      ],
+    };
+
+    const xml = exporter.schemaToXml(data);
+
+    expect(xml).toContain('Rock &amp; Roll &amp; Jazz');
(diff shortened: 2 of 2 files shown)


packages/super-editor/src/core/Editor.ts
```diff
@@ -2603,8 +2603,9 @@ export class Editor extends EventEmitter<EditorEventMap> {
         'word/_rels/document.xml.rels': String(rels),
         'word/numbering.xml': String(numbering),
 
-        // Replace & with &amp; in styles.xml as DOCX viewers can't handle it
-        'word/styles.xml': String(styles).replace(/&/gi, '&amp;'),
+        // Replace bare & with &amp; in styles.xml as DOCX viewers can't handle it
+        // but avoid double-escaping existing entities like &amp;.
+        'word/styles.xml': String(styles).replace(/&(?!#\d+;|#x[0-9a-fA-F]+;|[a-zA-Z][a-zA-Z0-9]+;)/g, '&amp;'),
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/super-editor/src/core/Editor.ts
```diff
@@ -2603,8 +2603,9 @@ export class Editor extends EventEmitter<EditorEventMap> {
         'word/_rels/document.xml.rels': String(rels),
         'word/numbering.xml': String(numbering),
 
-        // Replace & with &amp; in styles.xml as DOCX viewers can't handle it
-        'word/styles.xml': String(styles).replace(/&/gi, '&amp;'),
+        // Replace bare & with &amp; in styles.xml as DOCX viewers can't handle it,
+        // but avoid double-escaping valid XML entities and numeric references.
+        'word/styles.xml': String(styles).replace(/&(?!#\d+;|#x[0-9a-fA-F]+;|(?:amp|lt|gt|quot|apos);)/g, '&amp;'),
```
caio-pizzol: This regex is the core of the fix - would be good to add a unit test covering the key cases

packages/super-editor/src/core/Editor.ts
caio-pizzol, line 2599: All XML files go through the same `schemaToXml()` → `#replaceSpecialCharacters()` pipeline, which already escapes & → &amp; in all text nodes and attribute values.

Only `styles.xml` gets the extra regex treatment. The deeper question is: if `schemaToXml` already escapes everything, why does `styles.xml` need post-processing at all?

VladaHarbour: Also wondering about the same thing.
We had `.replace(/&/g, '&amp;')` in `#replaceSpecialCharacters()` but than we had another `.replace(/&/gi, '&amp;')` while adding styles file to the exported doc. Maybe this double replace is a root cause?

luccas-harbour: You guys are right. The issue was the double escaping caused by the extra replacement in `Editor.ts`. I removed it and also added an extra guard to the regex that replaces ampersands in `#replaceSpecialCharacters()` so that it will never double-escape an ampersand. The tests were also updated to cover this.

Thank you!

packages/super-editor/src/core/Editor.ts
luccas-harbour: This was actually removed as explained in the other comment.

VladaHarbour:
lgtm

harbournick:

PR #1990, opened 2026-02-10
fix: return null instead of blank num definition when not found

luccas-harbour:
- During the computation of a list item's properties in the numberingPlugin, if the `numId` was not found, a numbering definition with null properties was returned which could cause issues further ahead.
- This was changed so that `null` is returned instead and the definition is ignored.


harbournick:

PR #1991, opened 2026-02-11
fix: handle semicolon font fallback separators

mattConnHarbour:
Some DOCX generators list fonts with a semicolon separator. This PR makes sure those are replaced with commas before turning into CSS rules.

This fixes an issue with incorrect font being rendering in the layout engine, leading to text extending beyond document boundaries.

### DIFF
--- a/shared/font-utils/index.js
+++ b/shared/font-utils/index.js
@@ -69,6 +69,50 @@ const normalizeParts = (value) =>
     .filter(Boolean);
 
 /**
+ * Splits a string by a delimiter, but only when the delimiter is outside of quotes.
+ *
+ * @private
+ * @param {string} str - The string to split
+ * @param {string} delimiter - The delimiter to split on (single character)
+ * @returns {string[]} Array of parts, trimmed and filtered for empty strings
+ */
+const splitOutsideQuotes = (str, delimiter) => {
+  // Fast path: no quotes, just split normally
+  if (!str.includes('"') && !str.includes("'")) {
+    return str
+      .split(delimiter)
+      .map((p) => p.trim())
+      .filter(Boolean);
+  }
+
+  const parts = [];
+  let current = '';
+  let inQuote = false;
+  let quoteChar = null;
+
+  for (const char of str) {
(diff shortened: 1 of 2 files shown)


shared/font-utils/index.js
```diff
@@ -190,7 +190,15 @@ export function mapWordFamilyFallback(wordFamily) {
  */
 export function toCssFontFamily(fontName, options = {}) {
   if (!fontName || typeof fontName !== 'string') return fontName;
-  const trimmed = fontName.trim();
+  let trimmed = fontName.trim();
+  // replace semicolon font fallback separators
+  if (trimmed.includes(';')) {
```
caio-pizzol, line 241: once semicolons become commas, the comma check on line 202 returns early -- no sans-serif fallback gets added. 

works today because the import path already adds one, but if this function gets called with raw "Liberation Sans;Arial" it'll be missing a generic fallback.

caio-pizzol: also.. needs tests for semicolon inputs -- the early return on line 202 skipping fallbacks is easy to miss without one. `toCssFontFamily("Liberation Sans;Arial")` and `toCssFontFamily("Foo;Bar", { wordFamily: 'swiss' })` would cover it.

shared/font-utils/index.js
```diff
@@ -190,7 +190,15 @@ export function mapWordFamilyFallback(wordFamily) {
...
+    trimmed = trimmed
+      .split(';')
+      .map((part) => part.trim())
```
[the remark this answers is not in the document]
caio-pizzol: @mattConnHarbour check and add a response if applicable/or not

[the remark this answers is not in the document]
mattConnHarbour: Done, add quote check.

shared/font-utils/index.js
mattConnHarbour: Added tests and moved early return for strings with commas ahead of semicolon check.

mattConnHarbour:
@caio-pizzol Addressed your remarks, thanks for the review. This should be ready now with added tests.

PR #1994, opened 2026-02-11
fix: preserve text selection highlight on right-click

### LINKED ISSUE #1764: Selected text resets after right clicking. [Only on Mozilla Firefox]
### What's on your mind?

When you select a text and hit right click then the selection gets reset immediately.

### SuperDoc version (if relevant)

1.6.1

### Additional context

_No response_

henriquedevelops:
## Summary
  - Selection highlight no longer disappears when right-clicking to open context menu
  - Fixed by checking SlashMenu open state before clearing selection overlay in PresentationEditor
  - Reproduced in both Chrome and Firefox (pagination mode)

## Before

https://github.com/user-attachments/assets/ff7587dd-1dcc-4b41-919c-a0da6202d4e9

## After

https://github.com/user-attachments/assets/67381ad0-ba1e-46ea-ba56-ccb48ef49297

## Test plan
  - [x] Select text in pagination mode
  - [x] Right-click on selected text
  - [x] Verify blue highlight stays visible while context menu is open

Fixes #1764

### DIFF
--- a/packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
+++ b/packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
@@ -3369,7 +3369,7 @@ export class PresentationEditor extends EventEmitter {
     const activeEditor = this.getActiveEditor();
     const hasFocus = activeEditor?.view?.hasFocus?.() ?? false;
     // Keep selection visible when context menu (SlashMenu) is open
-    const slashMenuOpen = !!SlashMenuPluginKey.getState(activeEditor?.state)?.open;
+    const slashMenuOpen = activeEditor?.state ? !!SlashMenuPluginKey.getState(activeEditor.state)?.open : false;
 
     if (!hasFocus && !slashMenuOpen) {
       try {


packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -3367,8 +3368,10 @@ export class PresentationEditor extends EventEmitter {
 
     const activeEditor = this.getActiveEditor();
     const hasFocus = activeEditor?.view?.hasFocus?.() ?? false;
+    // Keep selection visible when context menu (SlashMenu) is open
+    const slashMenuOpen = !!SlashMenuPluginKey.getState(activeEditor?.state)?.open;
```
caio-pizzol: `SlashMenuPluginKey.getState(undefined)` will throw an error -- ProseMirror's `getState` does `state[this.key]` with no `null` check. 

The `?.` on `activeEditor?.state` protects against `null` `activeEditor`, but if state is `undefined` it still crashes.


```suggestion
    const slashMenuOpen = activeEditor?.state ? !!SlashMenuPluginKey.getState(activeEditor.state)?.open : false;
```

caio-pizzol (commented):
Nice fix, clean and minimal - just added one comment reg Editor state.

We'll also want to add a visual regression test for this (we have an internal interaction testing harness that covers selection scenarios across browsers). We can handle that on our side after merging.

Thanks for the contribution!

henriquedevelops:
> Nice fix, clean and minimal - just added one comment reg Editor state.
> 
> We'll also want to add a visual regression test for this (we have an internal interaction testing harness that covers selection scenarios across browsers). We can handle that on our side after merging.
> 
> Thanks for the contribution!

Thanks for the catch! Applied your suggestion. ✅

harbournick (approved):
LGTM

harbournick:

PR #2001, opened 2026-02-11
feat: enabled telemetry by default and added documentation

andrii-harbour:


packages/super-editor/src/core/super-converter/SuperConverter.js
```diff
@@ -752,44 +778,32 @@ class SuperConverter {
 
   /**
    * Generate identifier hash from documentGuid and dcterms:created
-   * Uses CRC32 of the combined string for a compact identifier
+   * Uses SHA-256 of the combined string for a compact identifier
    * Only call when both documentGuid and timestamp exist
    * @returns {string} Hash identifier in format "HASH-XXXXXXXX"
    */
-  #generateIdentifierHash() {
+  async #generateIdentifierHash() {
```
harbournick: why were these changes necessary? seems a bit risky since the testing days have been using the previous lib and system. Is there a strong reason for removing crc lib and replacing with this approach?

andrii-harbour: @harbournick you're right. I found a solution when my changes are not needed. Rollbacked to the previous tested version!

PR #2027, opened 2026-02-14
feat(slash-menu): notion-like slash menu UX with inline filtering

tupizz:
Resolves SD-1945

## Summary

Rewrites the slash menu to follow **Notion's UX pattern**: typing `/` inserts the character into the document and opens the menu below. Subsequent keystrokes filter items in real-time from the document text. This replaces the previous hidden-input approach with a more intuitive inline editing experience.

Merged with main's context-menu rename (SD-1824) — all naming now uses `contextMenu` conventions (`contextMenu:open`, `contextMenu:navigate`, `ContextMenuPluginKey`, `.context-menu` CSS classes).

### Slash menu opens with `/` visible in document
- Type `/` in a paragraph → the character appears in the document and the menu opens below the cursor
- First item is auto-selected with a visible blue highlight

pr-slash-open

### Real-time filtering from document text
- Continue typing after `/` to filter items (e.g., `/tab` shows only "Insert table")
- A "FILTERED RESULTS" header appears when filtering is active
- If no items match, the menu auto-closes

### Keyboard navigation with wrapping
- **ArrowDown**: moves selection down; from last item wraps to first
- **ArrowUp**: moves selection up; from first item wraps to last
- **Enter**: executes the selected item, removes `/query` text from document
pr-arrowdown

pr-arrowdown-wrap

### Escape cleans up
- Pressing **Escape** deletes the entire `/query` text from the document and closes the menu
- The document returns to its original state before the `/` was typed

slash-menu-enter

### Right-click context menu preserved
- Right-click opens the same menu at the click position with a wider set of options (includes "Insert link")
- Uses a separate `click` trigger flow — no filtering, no document text insertion

pr-contextmenu

### PM Plugin (`context-menu.js`)
- **Editor event emission**: ArrowDown/ArrowUp/Enter now emit `contextMenu:navigate` events via `editor.emit()` instead of relying on DOM event bubbling, which was fragile in presentation mode (PresentationInputBridge creates synthetic KeyboardEvents that don't bubble naturally to document-level listeners)
- **Slash trigger**: `/` is inserted into the document in the same transaction as the menu open — the text is visible and used for filtering
- **Auto-close**: Menu closes when cursor leaves the `/query` range or moves to a different block
- **Removed cooldown**: The previous cooldown mechanism (preventing menu reopen for 300ms after close) has been removed — the menu can now be reopened immediately

### Vue Component (`ContextMenu.vue`)
- **Editor events instead of document keydown**: Listens via `editor.on('contextMenu:navigate')` instead of `document.addEventListener('keydown')` — reliable in both flow and presentation modes
- **Wrapping navigation**: ArrowUp from first item wraps to last, ArrowDown from last wraps to first
- **Document-based filtering**: Search query extracted from document text between `/` anchor and cursor position via `state.doc.textBetween()`
- **Improved highlight visibility**: Selected item background changed from `rgba(55, 53, 47, 0.08)` (8% gray, nearly invisible) to `rgba(35, 131, 226, 0.14)` (blue tint, clearly visible)
- **No hidden input**: Removed the hidden input element — keyboard events flow through ProseMirror plugin → editor events → Vue component

### VerticalNavigation fix (`vertical-navigation.js`)
- **ArrowUp/Down yield to context menu**: When the context menu is open, the VerticalNavigation plugin now yields (`return false`) instead of intercepting arrow keys for cursor movement. This prevents ArrowUp from moving the cursor to the previous line and closing the menu.
- **Root cause**: VerticalNavigation was registered before ContextMenu in the ProseMirror plugin array, so it processed ArrowUp first, dispatching a cursor-move transaction that triggered the menu's auto-close logic.

### Naming alignment with main (SD-1824)
- All event names: `slashMenu:*` → `contextMenu:open`, `contextMenu:close`, `contextMenu:navigate`
- Plugin key: `SlashMenuPluginKey` → `ContextMenuPluginKey` (deprecated alias preserved)
- CSS classes: `.slash-menu-*` → `.context-menu-*`
- Extension name: `slashMenu` → `contextMenu`
- Test mocks and assertions updated throughout

### Test Fixes
- **Mock path fix**: `vi.mock('@extensions/context-menu')` resolved to `index.js` barrel, but the component imports `context-menu.js` directly — different module identifiers meant the mock didn't apply. Fixed to `vi.mock('@extensions/context-menu/context-menu')`
- **Mock state reset**: Added `ContextMenuPluginKey.getState.mockReset()` in top-level `beforeEach` to prevent mock state leaking between tests
- **Mock `tr.delete`**: Added missing `delete` method to mock transaction object
- **Updated assertions**: Event listener setup/cleanup assertions now check for `contextMenu:navigate` instead of document `keydown`

## Test plan

- [x] All 663 test files pass (6161 tests, 10 skipped)
- [x] `/` opens menu with character visible in document
- [x] Typing after `/` filters items in real-time
- [x] ArrowDown/ArrowUp navigate with wrapping
- [x] ArrowUp works in presentation mode with content above cursor (no longer closes menu)
- [x] Enter selects highlighted item and removes `/query` text
- [x] Escape removes `/query` text and closes menu
- [x] Right-click context menu still works
- [x] Menu auto-closes when cursor leaves `/query` range
- [x] Menu auto-closes when cursor moves to a different paragraph
- [x] Works in presentation mode (layout engine on)
- [x] Browser-tested with agent-browser on localhost:9098


packages/super-editor/src/components/context-menu/ContextMenu.vue
```diff
@@ -25,94 +25,52 @@ const props = defineProps({
...
-  if (!searchQuery.value) {
-    return sections.value;
-  }
+  const q = searchQuery.value.toLowerCase().trim();
+  if (!q) return sections.value;
+
+  return sections.value
+    .map((section) => ({
+      ...section,
+      items: section.items.filter((item) => item.label.toLowerCase().includes(q)),
```
caio-pizzol, line 49: the old code had `item.label?.` here. without it, a custom item with no label crashes the menu.

```suggestion
      items: section.items.filter((item) => (item.label ?? '').toLowerCase().includes(q)),
```

packages/super-editor/src/extensions/context-menu/context-menu.js
```diff
@@ -408,44 +439,53 @@ export const ContextMenu = Extension.create({
...
+                  tr.delete(anchorPos, cursorPos);
+                }
+                tr.setMeta(ContextMenuPluginKey, { type: 'close' });
+                view.dispatch(tr);
+                view.focus();
+                return true;
+              }
+              // All other keys pass through to PM (chars go into the document).
+              // apply() auto-closes if cursor leaves the /query range.
+              return false;
```
caio-pizzol, line 478: ArrowRight passes through to ProseMirror when the slash menu is open, so the cursor can move past the typed query into existing text. then `executeCommand` deletes from `anchorPos` to `cursorPos` — which now includes text the user didn't type. easiest fix: cap the delete range to only the `/` + query length.

packages/super-editor/src/components/context-menu/ContextMenu.vue
```diff
@@ -25,94 +25,52 @@ const props = defineProps({
...
   }
 });
 
-watch(flattenedItems, (newItems) => {
-  if (newItems.length > 0) {
-    selectedId.value = newItems[0].id;
+// Auto-close when search yields no results (only after sections have loaded)
+watch(filteredItems, (items) => {
+  if (isOpen.value && searchQuery.value && items.length === 0 && sections.value.length > 0) {
+    closeMenu();
```
caio-pizzol, line 68: Escape deletes the `/query` text, but this auto-close path doesn't. intentional? if yes, a short comment explaining why would help.

packages/super-editor/src/components/context-menu/ContextMenu.vue
```diff
@@ -25,94 +25,52 @@ const props = defineProps({
...
-    nextTick(() => {
-      if (searchInput.value) {
-        // Use preventScroll to avoid scrolling the page when focusing the search input.
-        // Without this, the browser may scroll parent containers to bring the input into view,
-        // which causes unwanted page jumps when opening the context menu.
-        searchInput.value.focus({ preventScroll: true });
-      }
-    });
+// Keep selectedId in sync with filtered results
+watch(filteredItems, (items) => {
```
caio-pizzol, line 59: these two watchers watch the same thing and their conditions don't overlap. merge into one?

packages/super-editor/src/extensions/context-menu/context-menu.js
```diff
@@ -200,6 +195,50 @@ export const ContextMenu = Extension.create({
...
 
           if (!meta) {
+            // Auto-close slash menu when cursor leaves the /query range
+            if (value.open && value.trigger === 'slash' && value.anchorPos !== null) {
+              const mappedAnchor = tr.mapping.map(value.anchorPos);
+              const $cursor = tr.selection?.$cursor ?? null;
+
+              // Close if no cursor or cursor at/before the / character
+              if (!$cursor || $cursor.pos <= mappedAnchor) {
+                editor.emit('contextMenu:close');
```
caio-pizzol, line 205: this emit + return pattern shows up 3 times in 30 lines. a small `closeState()` helper would clean it up.

packages/super-editor/src/extensions/context-menu/context-menu.js
```diff
@@ -200,6 +195,50 @@ export const ContextMenu = Extension.create({
...
+                return ensureStateShape({ open: false, anchorPos: null, trigger: null });
+              }
+
+              // Close if cursor moved to a different block (paragraph/heading)
+              try {
+                const $anchor = tr.doc.resolve(mappedAnchor);
+                // Find the nearest block ancestor for each position (handles run wrappers)
+                let anchorBlock = null;
```
caio-pizzol, line 213: same loop twice, just different variable. pull out a `findBlockAncestor($pos)` helper?

packages/super-editor/src/components/context-menu/ContextMenu.vue
```diff
@@ -330,150 +257,203 @@ const handleRightClick = async (event) => {
...
-    // Set sections and selectedId when menu opens
+
+    // Load context and menu items if not already set (right-click sets them before dispatch)
     if (!currentContext.value) {
       const context = await getEditorContext(props.editor);
-      currentContext.value = context; // Store context for later use
-      sections.value = getItems({ ...context, trigger: 'slash' });
+      currentContext.value = context;
+      const pluginState = ContextMenuPluginKey.getState(props.editor.state);
+      const trigger = pluginState?.trigger === 'click' ? 'click' : 'slash';
```
caio-pizzol, line 407: `pluginState`, `trigger`, and `selectedId` are computed the same way in both branches. hoist above the `if`?

caio-pizzol (changes requested):
@tupizz good stuff — the inline slash menu feels much better than the old hidden-input approach.

two things to fix: a missing `?.` on `item.label` that can crash filtering, and a bug where ArrowRight lets the cursor drift past the query text so `executeCommand` can accidentally delete real content. details inline.

one question: should the menu closing on zero results also delete the `/query` text? right now Escape deletes it but auto-close doesn't.

a few small cleanup ideas inline too (optional). test coverage for the new auto-close logic in `apply()` would be valuable.

left inline comments.

caio-pizzol:
hey @tupizz - checking in on this. the branch is pretty far behind main now and will need a rebase before we can merge.

let me know if you're still working on this or if plans changed.

caio-pizzol:
hey @tupizz - circling back. the idea is still wanted (it's on hold, not dropped, and main still uses the old menu), so this isn't wasted work :)

the catch: every file you touched has since been moved under `editors/v1/`, plus a few related fixes have landed in the same area. a rebase here is really a rewrite.

cleanest path: close this and redo it on current main, picking up the four things i flagged last round. happy to pair if useful.

needs work (redo on current main).

PR #2033, opened 2026-02-15
fix: correctly set color and highlight of pasted text

gm1357:
# Problem

When pasting text with color or highlight from Word, Google Docs, or within SuperDoc, the toolbar shows the correct color but the text renders without it.                                                                                                          
                  
# Cause

Pasted colors were coming as rgb() format, but the rendering pipeline's `normalizeColor` only handles hex, it prepends # to non-hex values, producing invalid strings like #rgb(255, 0, 0).

# Fix
 - Added `cssColorToHex` utility that converts rgb()/rgba() values to hex format
 - Applied it in the Color extension's parseDOM so text colors are stored as hex at the point of entry
 - Applied it in the Highlight extension's parseDOM
  - Added a `{ style: 'background-color' }` parse rule so highlights pasted as `<span style="background-color: ...">` (Word/Google Docs format) are correctly captured,  previously only `<mark>` tags were matched
  
# Comparisons
  
## Before

https://github.com/user-attachments/assets/e753536b-b986-449a-8d12-effe55a95cc3

https://github.com/user-attachments/assets/39df4d3e-b807-423a-a7ff-43758e1f71f7

## After

https://github.com/user-attachments/assets/a6d12a91-1337-4412-9b02-54a68f15ac95

https://github.com/user-attachments/assets/5a6e3eb4-359e-4ed1-84cd-25918ad67ddf

### DIFF
--- a/tests/visual/tests/behavior/formatting/paste-color-and-highlight.spec.ts
+++ b/tests/visual/tests/behavior/formatting/paste-color-and-highlight.spec.ts
@@ -1,0 +1,51 @@
+import { test } from '../../fixtures/superdoc.js';
+
+test('@behavior pasting html with rgb() background-color applies highlight', async ({ superdoc }) => {
+  await superdoc.page.evaluate(() => {
+    (window as any).editor.commands.insertContent(
+      '<span style="background-color: rgb(255, 255, 0)">Yellow highlighted text</span>',
+    );
+  });
+  await superdoc.screenshot('paste-rgb-background-color-highlight');
+});
+
+test('@behavior pasting html with transparent background-color applies no highlight', async ({ superdoc }) => {
+  await superdoc.page.evaluate(() => {
+    (window as any).editor.commands.insertContent(
+      '<span style="background-color: transparent">No highlight text</span>',
+    );
+  });
+  await superdoc.screenshot('paste-transparent-background-no-highlight');
+});
+
+test('@behavior pasting html with hex background-color applies highlight', async ({ superdoc }) => {
+  await superdoc.page.evaluate(() => {
+    (window as any).editor.commands.insertContent(
+      '<span style="background-color: #ffff00">Yellow highlighted text</span>',
+    );
+  });
+  await superdoc.screenshot('paste-hex-background-color-highlight');
+});
+
+test('@behavior pasting html with rgba zero-alpha background applies no highlight', async ({ superdoc }) => {
+  await superdoc.page.evaluate(() => {
+    (window as any).editor.commands.insertContent(
+      '<span style="background-color: rgba(255, 0, 0, 0)">No highlight text</span>',
+    );
+  });
+  await superdoc.screenshot('paste-rgba-zero-alpha-no-highlight');
+});
+
+test('@behavior pasting html with rgb() text color is applied', async ({ superdoc }) => {
+  await superdoc.page.evaluate(() => {
+    (window as any).editor.commands.insertContent('<span style="color: rgb(255, 0, 0)">Red text</span>');
+  });
+  await superdoc.screenshot('paste-rgb-text-color');
+});
+
+test('@behavior pasting html with hex text color is applied', async ({ superdoc }) => {
+  await superdoc.page.evaluate(() => {
+    (window as any).editor.commands.insertContent('<span style="color: #ff0000">Red text</span>');
+  });
+  await superdoc.screenshot('paste-hex-text-color');
+});


packages/super-editor/src/extensions/highlight/highlight.js
```diff
@@ -49,7 +50,16 @@ export const Highlight = Mark.create({
...
 
   parseDOM() {
-    return [{ tag: 'mark' }];
+    return [
+      { tag: 'mark' },
+      {
+        style: 'background-color',
+        getAttrs: (value) => {
+          if (!value || value === 'transparent' || value === 'inherit') return false;
+          return { color: cssColorToHex(value) };
```
caio-pizzol: `cssColorToHex(value)` can return `null` here -- for example when the pasted HTML has `rgba(0, 0, 0, 0)` (a common way browsers encode "no background"). That value passes the transparent/inherit guard on line 58, but `cssColorToHex` correctly returns `null` for zero-alpha.

The problem is this still returns `{ color: null }`, which tells ProseMirror to create a highlight mark with no color -- a phantom mark in the document.
```
  getAttrs: (value) => {
    if (!value || value === 'transparent' || value === 'inherit') return false;
    const color = cssColorToHex(value);
    return color ? { color } : false;
  },
```

packages/super-editor/src/extensions/highlight/highlight.js
```diff
@@ -34,7 +35,7 @@ export const Highlight = Mark.create({
     return {
       color: {
         default: null,
-        parseDOM: (element) => element.getAttribute('data-color') || element.style.backgroundColor,
+        parseDOM: (element) => element.getAttribute('data-color') || cssColorToHex(element.style.backgroundColor),
```
caio-pizzol: `data-color` is used as-is without going through `cssColorToHex`. This matters because SuperDoc sets `data-color` to whatever the stored color attribute is (line 44). If a document was edited before this fix, the stored value could be `rgb(255, 0, 0)`. When that content is copy-pasted within SuperDoc, the pasted `<mark>` has `data-color="rgb(255, 0, 0)"` which gets stored raw.

Downstream, `normalizeColor` in `pm-adapter/src/utilities.ts` prepends # to anything that doesn't start with it, producing `"#rgb(255, 0, 0)"` which breaks rendering in presentation mode.

```
  parseDOM: (element) => cssColorToHex(element.getAttribute('data-color')) || cssColorToHex(element.style.backgroundColor),
```

packages/super-editor/src/extensions/highlight/highlight.js
```diff
@@ -49,7 +50,16 @@ export const Highlight = Mark.create({
   },
 
   parseDOM() {
-    return [{ tag: 'mark' }];
+    return [
+      { tag: 'mark' },
+      {
+        style: 'background-color',
+        getAttrs: (value) => {
```
caio-pizzol, line 57: Minor: the transparent/inherit check here and the `null` handling inside `cssColorToHex` are doing related work in two places. If you add the `null` guard from comment 1, you could simplify this to just:
```
  getAttrs: (value) => {
    if (!value) return false;
    const color = cssColorToHex(value);
    return color ? { color } : false;
  },
```

Since `cssColorToHex('transparent')` would return "transparent" (named color passthrough), you'd want to either keep the explicit check here, or teach `cssColorToHex` to return `null` for CSS keywords like transparent/inherit/initial.

Either way works -- just pick one place for that logic.

packages/super-editor/src/extensions/color/color.js
```diff
@@ -48,7 +49,7 @@ export const Color = Extension.create({
         attributes: {
           color: {
             default: null,
-            parseDOM: (el) => el.style.color?.replace(/['"]+/g, ''),
+            parseDOM: (el) => cssColorToHex(el.style.color?.replace(/['"]+/g, '')),
```
caio-pizzol: The `.replace(/['"]+/g, '')` is left over from the previous code -- `el.style.color` is a browser-computed property that never contains quotes.

Safe to drop it now that you're already editing this line:
```
parseDOM: (el) => cssColorToHex(el.style.color),
```

packages/super-editor/src/core/utilities/cssColorToHex.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Converts a CSS color value to hex format (#RRGGBB).
+ * Handles rgb(), rgba(), hex, and returns null for empty input, transparent rgba, invalid rgb values.
+ * Named colors are returned as-is.
+ *
+ * @param {string|null|undefined} cssColor - A CSS color string
+ * @returns {string|null} Hex color string or null
```
caio-pizzol: `@returns {string|null} Hex color string or null` -- this can also return named colors ("red") and 3-digit hex ("#f00") unchanged, so "Hex color string" isn't quite right.

Something like Normalized color string or null would match the actual contract better.

packages/super-editor/src/extensions/highlight/highlight.js
gm1357: Adressed on [36d4a5b](https://github.com/superdoc-dev/superdoc/pull/2033/commits/36d4a5bc4ec781caef3b7ecbc3737d3da0f7da55)

packages/super-editor/src/extensions/highlight/highlight.js
gm1357: Adressed on [36d4a5b](https://github.com/superdoc-dev/superdoc/pull/2033/commits/36d4a5bc4ec781caef3b7ecbc3737d3da0f7da55)

packages/super-editor/src/extensions/color/color.js
gm1357: Adressed on [36d4a5b](https://github.com/superdoc-dev/superdoc/pull/2033/commits/36d4a5bc4ec781caef3b7ecbc3737d3da0f7da55)

packages/super-editor/src/core/utilities/cssColorToHex.js
gm1357: Addressed on [8611418](https://github.com/superdoc-dev/superdoc/pull/2033/commits/8611418bc4e6bf883570996c1083a04c823436a8)

packages/super-editor/src/extensions/highlight/highlight.js
gm1357: Adressed on [7a02659](https://github.com/superdoc-dev/superdoc/pull/2033/commits/7a026596d1f282e66fa538954142a303e6d5313b).

Chose to keep the logic explictly here so that `cssColorToHex` can be a bit more agnostic on how to handle those values.

gm1357:
Thanks for the reviews @tupizz and @caio-pizzol!

@caio-pizzol I adressed your in-line comments. Regarding the behaviour tests you suggested, I will take a look at it later and add it.

gm1357:
@caio-pizzol I added some behaviour tests for this change, can you do another review when possible?

packages/super-editor/src/extensions/highlight/highlight.js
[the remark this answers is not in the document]
caio-pizzol: Already handled — `cssColorToHex` returns `null` for zero-alpha (line 24), and `getAttrs` returns false for `null` (line 61).

packages/super-editor/src/core/utilities/cssColorToHex.js
```diff
@@ -0,0 +1,28 @@
...
+  if (/^#([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$/.test(trimmed)) {
+    return trimmed;
+  }
+
+  // Parse rgb(r, g, b) or rgba(r, g, b, a)
+  const rgbMatch = trimmed.match(/^rgba?\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)/);
+  if (rgbMatch) {
+    const [, r, g, b] = rgbMatch;
+    return '#' + [r, g, b].map((c) => Number(c).toString(16).padStart(2, '0')).join('');
+  }
```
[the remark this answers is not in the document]
caio-pizzol: Browsers normalize pasted colors to lowercase `rgb()`. Channels >255 return `null` intentionally (line 25)

packages/super-editor/src/extensions/highlight/highlight.js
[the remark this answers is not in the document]
caio-pizzol: Line 38 already wraps both sources in `cssColorToHex()`

packages/super-editor/src/extensions/highlight/highlight.js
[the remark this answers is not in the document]
caio-pizzol: `cssColorToHex` trims internally, and browser style values are always lowercase

packages/super-editor/src/extensions/highlight/highlight.js
```diff
@@ -49,7 +50,16 @@ export const Highlight = Mark.create({
...
+          if (!value || value === 'transparent' || value === 'inherit') return false;
```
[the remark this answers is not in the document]
caio-pizzol: Lines 60-61 already check: `return color ? { color } : false`

packages/super-editor/src/core/utilities/cssColorToHex.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Converts a CSS color value to hex format (#RRGGBB).
+ * Handles rgb(), rgba(), hex, and returns null for empty/invalid input.
+ *
+ * @param {string|null|undefined} cssColor - A CSS color string
+ * @returns {string|null} Hex color string or null
```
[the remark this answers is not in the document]
caio-pizzol: Fair point, but the docstring already mentions named colors are returned as-is

harbournick:

PR #2088, opened 2026-02-18
feat: support rendering of column line separators

### LINKED ISSUE #2067: Feature: support rendering of line between columns
### Summary

Support rendering the "Line between" separator when a document uses multi-column layouts (`w:cols` with `w:sep` or `lineBetween` attribute in `w:sectPr`).

### Problem

When a DOCX uses columns with the "Line between" setting enabled, the vertical separator line between columns does not render in SuperDoc.

### Repro

1. Open attached docx in SuperDoc.
2. The document has two columns with "Line between" enabled.

### Expected

A vertical line renders between the columns, matching Word's behavior.

### Notes

The setting lives in `w:sectPr/w:cols` as either a `w:sep` child element or `w:lineBetween` attribute. This is a rendering-only change — the column layout itself already works.

### Attachments

[Column borders (1).docx](https://uploads.linear.app/92b189d9-e15d-4f28-8a87-d9ca147ba33f/96a61f28-8b5e-4759-8dcb-16372748094a/4a5c9d1a-0903-4479-974e-d0e28596c2b6?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkyYjE4OWQ5LWUxNWQtNGYyOC04YTg3LWQ5Y2ExNDdiYTMzZi85NmE2MWYyOC04YjVlLTQ3NTktOGRjYi0xNjM3Mjc0ODA5NGEvNGE1YzlkMWEtMDkwMy00NDc5LTk3NGUtZDBlMjg1OTZjMmI2IiwiaWF0IjoxNzcxMjYyNTE5LCJleHAiOjE4MDI4MzMwNzl9.NmIPnO4oytoDGETml8sLPyx1WmIYIMr79baR-erLAMc)

---

*Ref:* [SD-1655](https://linear.app/superdocworkspace/issue/SD-1655/feature-support-rendering-of-line-between-columns)

JoaaoVerona:
### Summary

Adds support for rendering "line between" separators between page columns in DOM Painter, extracting their presence by looking up the `w:sep` tag in the layout engine and ProseMirror adapter. Closes #2067.

---

### Changes

- layout-engine, pm-adapter: Extracts `w:sep` tag according to OOXML spec [[1]](http://officeopenxml.com/WPsectionCols.php) [[2]](http://www.datypic.com/sc/ooxml/a-w_sep-1.html)
- painters: Renders one separator for each page column (`w:num`)
- contracts, layout-bridge, super-editor: pass down `withSeparator` property
- New tests

---

### Questions & Additional details

- There were multiple occurrences (within the existing code) of copying the `columns` data (i.e. `y.columns = { count: x.count, gap: x.gap }`) that needed to be updated also to copy `withSeparator`; I could have used spread operator to shallow clone the data, but I went with a more conservative approach to avoid, potentially, copying unnecessary/unintended properties that could be declared in these objects at runtime;
- `withSeparator` will be true when `w:sep="true"`, `w:sep="1"` or `w:sep="on"`, according to the spec;
- Separator color hard-coded as `#b3b3b3`, as I'm not sure if we could infer this somehow;
- I'm not sure about the retrocompatibility of this change; I marked `withSeparator` as optional to make sure we don't break userland, but a second look from someone more experienced with the packaging/release of superdoc would be great.

---

### Showcase

GDocs - Two-column layout ([docx here](https://github.com/user-attachments/files/25378049/GDocs-Two-Columns.docx))             |  SuperDoc - Imported from GDocs
:-------------------------:|:-------------------------:
  |  

GDocs - Three-column layout ([docx here](https://github.com/user-attachments/files/25378048/GDocs-Three-Columns.docx))             |  SuperDoc - Imported from GDocs
:-------------------------:|:-------------------------:
  |  

MSWord - Two-column layout ([docx here](https://github.com/user-attachments/files/25378050/Word-Two-Columns.docx))             |  SuperDoc - Imported from MSWord
:-------------------------:|:-------------------------:
  |


JoaaoVerona:
@caio-pizzol I could not find, in ECMA-376, a definition for `w:sectPr` > `lineBetween` as specified on #2067. Should we add support for that attribute too? Are programs out there exporting docx files with it?

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -1566,6 +1568,37 @@ export class DomPainter {
...
+    const gap = page.columns.gap;
+    const columnWidth = (contentWidth - gap * (columnCount - 1)) / columnCount;
+
+    for (let i = 0; i < columnCount - 1; i++) {
+      const separatorX = leftMargin + (i + 1) * columnWidth + i * gap + gap / 2;
+      const separatorEl = this.doc.createElement('div');
+
+      separatorEl.style.position = 'absolute';
+      separatorEl.style.left = `${separatorX}px`;
+      separatorEl.style.top = `${topMargin}px`;
```
caio-pizzol: the separator currently runs the full height of the content area (top margin to bottom margin). in Word, the line only extends as tall as the actual text in the columns -- so on pages that aren't fully filled (like the last page of a multi-column section), Word's line stops where the text stops.

here, it'll keep going all the way down. totally fine for a first pass -- just flagging it as something we might refine later.

packages/layout-engine/layout-engine/src/index.ts
```diff
@@ -972,6 +978,11 @@ export function layoutDocument(blocks: FlowBlock[], measures: Measure[], options
     if (activeOrientation) {
       page.orientation = activeOrientation;
     }
+
+    if (activeColumns.count > 1) {
```
caio-pizzol, line 1084: heads up on an edge case here: `page.columns` is set from `activeColumns`, which is the raw config from the section properties. but further down in the pipeline, there's a normalization step (`normalizeColumns`) that checks whether columns actually fit on the page -- if the gap is so large that columns would have zero width, it silently falls back to single-column layout. since `page.columns` was already set before that check, the renderer would still draw separators even though the page is actually rendering as one column.

unlikely to hit in real documents, but the fix is straightforward -- either set `page.columns` after normalization, or check the normalized count in the renderer.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -1566,6 +1568,37 @@ export class DomPainter {
...
+      separatorEl.style.height = `${pageHeight - topMargin - bottomMargin}px`;
+      separatorEl.style.width = '1px';
+      separatorEl.style.backgroundColor = '#b3b3b3';
```
caio-pizzol: Word renders this separator as a dark line (close to black). `#b3b3b3` is a light gray, which will look noticeably different when comparing side-by-side. the rest of our renderer also defaults to `#000` for line elements like borders and underlines, so swapping to match would keep things consistent both with Word and with our own codebase.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -1566,6 +1568,37 @@ export class DomPainter {
     }
   }
 
+  private renderColumnSeparators(pageEl: HTMLElement, page: Page, pageWidth: number, pageHeight: number): void {
```
caio-pizzol, line 2315: one thing to be aware of: `page.columns` is a single value per page, set when the page is created. but in OOXML, a "continuous" section break can change the column count mid-page (e.g., a newsletter with a single-column header, then 2-column body, then single-column footer -- all on one page). since `page.columns` doesn't capture that, the separator would draw across the entire page even if only part of it is multi-column. 

this is more of an architectural limitation than something to fix here -- just flagging it so we can track it.

packages/layout-engine/layout-engine/src/paginator.ts
```diff
@@ -4,7 +4,7 @@ export type NormalizedColumns = ColumnLayout & { width: number };
 
 export type ConstraintBoundary = {
   y: number;
-  columns: { count: number; gap: number };
+  columns: { count: number; gap: number; withSeparator?: boolean };
```
caio-pizzol: you'll notice the type `{ count: number; gap: number; withSeparator?: boolean }` is written out inline about 14 times across these files. we actually have a `ColumnLayout` type in `@superdoc/contracts` that's the exact same shape -- and it's already imported in this file. 

using it in the files you touched would mean the next person adding a field (like you did with `withSeparator`) only has to update it in one place.

packages/layout-engine/layout-bridge/src/incrementalLayout.ts
```diff
@@ -130,21 +130,23 @@ const normalizeColumnsForFootnotes = (input: ColumnLayout | undefined, contentWi
...
   }
 
-  return { count, gap, width };
+  return { count, gap, width, withSeparator };
 };
 
 const resolveSectionColumnsByIndex = (options: LayoutOptions, blocks?: FlowBlock[]): Map<number, ColumnLayout> => {
   const result = new Map<number, ColumnLayout>();
-  let activeColumns: ColumnLayout = options.columns ?? { count: 1, gap: 0 };
+  let activeColumns: ColumnLayout = options.columns ?? { count: 1, gap: 0, withSeparator: false };
```
caio-pizzol: this fallback `{ count: 1, gap: 0, withSeparator: false }` shows up 6 times across `incrementalLayout.ts` and `index.ts`. there's already a `SINGLE_COLUMN_DEFAULT` constant in `section-breaks.ts` that represents the same thing, but it's not exported.

if you export it and reuse it here, the next field added won't require hunting down 6 scattered literals -- which is exactly the situation you ran into when adding `withSeparator`.

caio-pizzol (commented):
Hey @JoaaoVerona, this is a great start! The `withSeparator` flag is threaded cleanly through the whole pipeline - extraction, layout, rendering — nice work.

**Tests:** would be great to add a visual rendering test with a sample .docx that has `w:sep="1"` on `<w:cols>`. The test infra auto-discovers .docx files so it's mostly just dropping one in tests/visual/. This is the best way to catch regressions in separator positioning and styling down the road.

On your `lineBetween` question: `lineBetween` isn't part of the ECMA-376 spec - the actual XML attribute for column separators is `w:sep on <w:cols> (§17.6.4)`, which is exactly what this PR implements. The name `lineBetween` probably comes from the Open XML SDK's property naming, not from the XML itself. Word and LibreOffice both use `w:sep`, so no need to support `lineBetween`. 

left some inline comments on a few things worth tweaking before merging.

caio-pizzol:
@JoaaoVerona btw we've created our own OOXML MCP server (https://ooxml.dev/mcp) :)

caio-pizzol:
@JoaaoVerona, let us know if you have any questions on this one

packages/layout-engine/layout-engine/src/index.ts
JoaaoVerona: got it! I could not find a proper way to set `page.columns` after normalization in the layout engine, so instead I'm checking if the column width is higher than 1px (which is the separator width) in the renderer -- https://github.com/superdoc-dev/superdoc/pull/2088/changes/e09b8ca9e89b2b7dff3b10cdaaa40923c545e17d (this commit got linted after). given that normalization checks if column width <= `COLUMN_EPSILON` (0.0001) to fallback to a single-column layout, it seems the check will work in such scenarios...

packages/layout-engine/painters/dom/src/renderer.ts
JoaaoVerona: done (https://github.com/superdoc-dev/superdoc/pull/2088/changes/1e73669c4882cf902dee617c1c96474e0f5cb9ec)

packages/layout-engine/layout-engine/src/paginator.ts
JoaaoVerona: done (https://github.com/superdoc-dev/superdoc/pull/2088/changes/2701ca62b55a744073a04f31e6943e375130e369)

packages/layout-engine/layout-bridge/src/incrementalLayout.ts
JoaaoVerona: done (https://github.com/superdoc-dev/superdoc/pull/2088/changes/2701ca62b55a744073a04f31e6943e375130e369)

JoaaoVerona:
hey @caio-pizzol, thanks for the info!

also, great work with the mcp -- been using it already with CC!

additionally to the inline comments:

- I simply added a test `.docx` directly to `tests/visual/` directory as instructed; I'm guessing there's no way right now for external contributors to run the visual tests (given that the `docs:download` script requires auth)?
- I'm not sure yet why this doc [test.docx](https://github.com/user-attachments/files/25506853/test.docx) has "misleading" `w:cols` entries and the actual columns in each page. it is a two-page doc; each page has a different column layout. SuperDoc itself seems to render the content correctly, but it seems the `w:cols` I'm extracting is misleading (perhaps I should not be looking at them and should infer columns another way? or maybe it could be related to the continuous section break you mentioned?).

Google Docs             |  SuperDoc
:-------------------------:|:-------------------------:
  |  

by looking at [document.xml](https://github.com/user-attachments/files/25506909/document.xml), I can see there are 3 `w:cols` entries (although there are only 2 pages). will need to investigate that further whenever I have additional time to work on this again, but please let me know if you have any insights!

caio-pizzol:
> I simply added a test `.docx` directly to `tests/visual/` directory as instructed; I'm guessing there's no way right now for external contributors to run the visual tests (given that the `docs:download` script requires auth)?

correct — the visual test infra isn't set up for external contributors to run locally yet. we'll upload the baseline on our end. we're looking into better ways to support this for the community going forward.

> I'm not sure yet why this doc test.docx has "misleading" `w:cols` entries [...] by looking at document.xml, I can see there are 3 `w:cols` entries (although there are only 2 pages).

i looked at your document.xml — OOXML sections are **end-tagged**, meaning the `sectPr` on a paragraph defines the section that *ends* at that paragraph, not the one that starts there. your doc has 3 sections:

| Section | `w:cols` | `w:type` | What it controls |
|---------|----------|----------|-----------------|
| 1 (paragraph `sectPr`, line 7) | `num="3"` | *(default = nextPage)* | just "Start line." — a single-paragraph section |
| 2 (paragraph `sectPr`, line 895) | `num="2"` | `continuous` | "Start first column." through "End." — page 1's 2-column layout |
| 3 (body `sectPr`, line 2195) | `num="3"` | `continuous` | "Start line page 2." through "End." — page 2's 3-column layout |

the confusing part is that the **first** `w:cols` says `num="3"`, but it only applies to a throwaway single-paragraph section that Google Docs emits before the actual column content begins. the `num="2"` in section 2 is what drives page 1's layout, and the `num="3"` in the body `sectPr` drives page 2.

so the extraction is correct — the values just read counterintuitively because of end-tagged semantics. nothing to fix here.

packages/layout-engine/pm-adapter/src/sections/extraction.ts
```diff
@@ -6,6 +6,7 @@
 
 import type { PMNode } from '../types.js';
 import type { ParagraphProperties, SectionVerticalAlign } from './types.js';
+import { ColumnLayout } from '@superdoc/contracts';
```
caio-pizzol: this file already uses `import type` on line 8 for other imports, but `ColumnLayout` is also only used as a type. worth making consistent — other files added in this PR already do it correctly.

```suggestion
import type { ColumnLayout } from '@superdoc/contracts';
```

packages/layout-engine/pm-adapter/src/sections/types.ts
```diff
@@ -5,6 +5,8 @@
  * Includes section ranges, signatures, and OOXML structures.
  */
 
+import { ColumnLayout } from '@superdoc/contracts';
```
caio-pizzol: same as above — `ColumnLayout` is only used as a type here.

```suggestion
import type { ColumnLayout } from '@superdoc/contracts';
```

packages/layout-engine/layout-engine/src/section-breaks.d.ts
```diff
@@ -23,10 +23,12 @@ export type SectionState = {
   activeColumns: {
     count: number;
     gap: number;
+    withSeparator?: boolean;
```
caio-pizzol: the `.d.ts` still has the inline object shape instead of referencing `ColumnLayout`, and it's missing the `SINGLE_COLUMN_DEFAULT` export that the `.ts` source now exports. since this file is manually maintained it won't pick up those changes automatically — worth updating it to match?

packages/layout-engine/layout-engine/src/index.ts
```diff
@@ -725,7 +726,7 @@ export function layoutDocument(blocks: FlowBlock[], measures: Measure[], options
 
   // Track active and pending columns
   let activeColumns = options.columns ?? { count: 1, gap: 0 };
```
caio-pizzol: this spot was missed when replacing the other five `{ count: 1, gap: 0 }` literals with `SINGLE_COLUMN_DEFAULT`. not a bug since the values are the same, but it's the odd one out now.

```suggestion
  let activeColumns = options.columns ?? { ...SINGLE_COLUMN_DEFAULT };
```

caio-pizzol (commented):
@JoaaoVerona the round 1 feedback is all addressed — the separator color, the `columnWidth <= 1` guard, the `ColumnLayout` type usage, and the `SINGLE_COLUMN_DEFAULT` constant are all looking good. left a few inline comments on some remaining rough edges before this is ready to merge.

JoaaoVerona:
hey @caio-pizzol, thanks for the context! inline comments should be solved now. rebased it with main too.

packages/layout-engine/layout-engine/src/section-props.ts
```diff
@@ -59,7 +59,7 @@ const _snapshotSectionProps = (block: FlowBlock): SectionProps | null => {
   }
   if (block.columns) {
     hasProps = true;
-    props.columns = { count: block.columns.count, gap: block.columns.gap };
+    props.columns = { count: block.columns.count, gap: block.columns.gap, withSeparator: block.columns.withSeparator };
```
caio-pizzol, line 62: the manual `{ count, gap, withSeparator }` copy here and at line 138 drops `equalWidth` and `widths` — so documents with different-sized columns lose that info in the snapshot. `getColumnConfig()` in `section-breaks.ts` already copies everything correctly. worth reusing it here?

packages/layout-engine/layout-bridge/src/incrementalLayout.ts
```diff
@@ -128,27 +129,32 @@ const COLUMN_EPSILON = 0.01;
...
+const normalizeColumnsForFootnotes = (
+  input: Readonly<ColumnLayout> | undefined,
+  contentWidth: number,
+): NormalizedColumns => {
   const rawCount = Number.isFinite(input?.count) ? Math.floor(input!.count) : 1;
   const count = Math.max(1, rawCount || 1);
   const gap = Math.max(0, input?.gap ?? 0);
   const totalGap = gap * (count - 1);
   const width = (contentWidth - totalGap) / count;
+  const withSeparator = input?.withSeparator ?? false;
```
caio-pizzol: this does the same thing as `normalizeColumns` in `layout-engine/src/index.ts:2633` — both are one-liners wrapping `normalizeColumnLayout`. not blocking, but if one changes and the other doesn't, things will break quietly.

caio-pizzol (approved):
@JoaaoVerona all the feedback from previous rounds is addressed — looking good.

two small things:

1. the column snapshot in `section-props.ts` copies fields by hand and misses `equalWidth` and `widths`. there's already a helper (`getColumnConfig`) that copies everything — reusing it would avoid this.

2. `normalizeColumnsForFootnotes` in `incrementalLayout.ts` is a copy of `normalizeColumns` in `index.ts`. worth consolidating so the next person doesn't update one and forget the other.

on tests: `renderColumnSeparators` doesn't have a unit test for its positioning math and guard clauses. also, `tests/visual/columns-with-line-separator.docx` won't get picked up by the layout regression suite — it needs to go through `pnpm corpus:upload` to land in the R2 corpus where `rendering.spec.ts` auto-discovers test files.

left a couple inline comments.

caio-pizzol:
Hey @JoaaoVerona, this PR has had no activity for 7 days.

PR #2099, opened 2026-02-18
fix(collab): prevent stale view when remote Y.js changes bypass sdBlockRev increment

tupizz:
## Summary

- **Root cause**: `FlowBlockCache` uses `sdBlockRev` for O(1) invalidation, but the collab plugin skips incrementing `sdBlockRev` for Y.js-origin transactions (to prevent an infinite sync loop). This created a window where remote content changes arrive but the cache returns stale blocks because the rev appears unchanged — only reproducible in production with real network latency.
- **Fix**: Add `setHasExternalChanges()` to `FlowBlockCache`. When `PresentationEditor` detects a Y.js-origin transaction, it flags the cache to fall back to JSON comparison for that render cycle instead of trusting the matching rev. The flag auto-clears on `commit()` so normal local editing stays O(1).
- **Also fixed**: File upload in collab mode was destroying and recreating the editor instead of using `replaceFile()`, causing a race condition where the empty Y.js room state could overwrite the new document content.

## Changes

| File | Change |
|------|--------|
| `pm-adapter/src/cache.ts` | Add `setHasExternalChanges(flag)` — forces JSON fallback when Y.js changes bypass rev increment |
| `pm-adapter/src/cache.d.ts` | Update type declaration |
| `pm-adapter/src/cache.test.ts` | 4 new tests including regression test proving the original bug |
| `PresentationEditor.ts` | Detect Y.js-origin transactions via `ySyncPluginKey` meta and signal cache |
| `SuperdocDev.vue` | Use `replaceFile()` on existing editor in collab mode instead of full reinit |

## Test plan

- [ ] All 1543 cache unit tests pass
- [ ] Two-tab collab: edits in Tab A appear in Tab B (typing, suggesting mode, accept/reject tracked changes)
- [ ] File upload in collab mode replaces content without race condition
- [ ] Local editing performance unaffected (O(1) cache path unchanged for non-Y.js transactions)

Closes SD-1852

### DIFF
--- a/packages/layout-engine/pm-adapter/src/cache.test.ts
+++ b/packages/layout-engine/pm-adapter/src/cache.test.ts
@@ -40,6 +40,31 @@ describe('FlowBlockCache', () => {
     expect(result.entry!.blocks).toBe(mockBlocks);
   });
 
+  it('retains serialized node across fast-path hits so external fallback stays incremental', () => {
+    const cache = new FlowBlockCache();
+    const node = makeParagraphNode('hello', 5);
+
+    // Render 1: cache is populated with serialized JSON.
+    cache.begin();
+    cache.set('p1', JSON.stringify(node), 5, mockBlocks, 0);
+    cache.commit();
+
+    // Render 2: local-only fast path hit, caller writes lookup payload into next generation.
+    cache.begin();
+    const fastPathHit = cache.get('p1', node);
+    expect(fastPathHit.entry).not.toBeNull();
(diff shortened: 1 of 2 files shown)


harbournick (approved):
LGTM

harbournick:
:tada: This PR is included in **superdoc** v1.15.0

The release is available on [GitHub release](https://github.com/superdoc-dev/superdoc/releases/tag/v1.15.0)

PR #2109, opened 2026-02-19
fix(tables): expand auto-width tables to fill available page width

tupizz:
## Summary

Closes  SD-1956 https://linear.app/superdocworkspace/issue/SD-1956/bug-auto-layout-tables-with-explicit-tblgrid-tiny-tcw-render-squished

- Tables without explicit `w:tblW` (or with `type="auto"`) were rendering at ~1/3 page width instead of filling the available content area like Microsoft Word does
- Root cause was in the measuring-dom layer: auto-layout tables only scaled **down** (when too wide) but never **up** to fill available space
- Also fixed `resolveTableWidth()` not recognizing `type='dxa'` — the most common explicit width type in OOXML — which caused dxa-width tables to incorrectly fall into the auto-layout path

### Why this was happening

The table measurement pipeline in `measureTableBlock()` has two branches:

1. **Explicit width branch** (`hasExplicitWidth || hasFixedLayout`): scales columns both up and down to match the declared table width
2. **Auto-layout branch** (no explicit width): only scaled **down** when columns exceeded available space, but **never scaled up** — leaving narrow imported column widths as-is

When a DOCX file has no `w:tblW` element (or `type="auto"`), `resolveTableWidth()` returned `undefined`, so `hasExplicitWidth` was `false`. The auto-layout branch would receive narrow column widths from the import (e.g., 100px per column from `w:tcW` cell widths or the fallback grid), and since the total was less than the page width, no scaling occurred.

Additionally, `resolveTableWidth()` only recognized `type='pct'` and `type='px'`/`'pixel'`, but not `type='dxa'` — even though `dxa` values are already converted to pixels during import by the tbl-translator. This meant tables with explicit dxa widths also fell into the auto-layout path.

### What we changed to prevent this

1. **Scale up auto-layout tables** (ECMA-376 §2.4.64): When a table has no explicit width, the layout algorithm (§2.4.49/§2.4.50) uses autofit, which in practice expands the table to fill the available page content width. The auto-layout branch now scales columns proportionally in both directions to match this behavior.

2. **Handle `dxa` type in `resolveTableWidth()`**: Added `type='dxa'` alongside `'px'`/`'pixel'` so that tables with explicit dxa widths are correctly recognized as having an explicit width, preventing them from being affected by the auto-layout scale-up.

## Test plan

- [x] All 224 measuring-dom tests pass (updated 12 existing tests to reflect new auto-layout behavior)
- [x] All 663 super-editor test files pass (6162 tests)
- [x] All 474 layout-engine tests pass
- [x] TypeScript typecheck passes
- [ ] Visual verification: load AI-generated DOCX with auto-width tables → tables fill page width
- [ ] Visual verification: existing DOCX files with explicit table widths render unchanged

### DIFF
--- a/packages/layout-engine/measuring/dom/src/index.test.ts
+++ b/packages/layout-engine/measuring/dom/src/index.test.ts
@@ -2992,10 +2992,9 @@ describe('measureBlock', () => {
 
       expect(measure.kind).toBe('table');
       if (measure.kind !== 'table') throw new Error('expected table measure');
-      // Auto layout scales columns to fill available width (OOXML autofit)
-      // Original ratio 100:150:200 = 2:3:4, scaled to 600px total
-      expect(measure.columnWidths).toEqual([133, 200, 267]);
-      expect(measure.totalWidth).toBe(600);
+      // Auto layout preserves explicit w:tblGrid widths (no scale-up)
+      expect(measure.columnWidths).toEqual([100, 150, 200]);
+      expect(measure.totalWidth).toBe(450);
     });
 
     it('scales column widths proportionally when exceeding available width', async () => {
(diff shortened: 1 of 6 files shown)


harbournick (approved):
LGTM

harbournick:

PR #2140, opened 2026-02-22
fix(layout-engine): text clipping inside table in document with multi-orientation, recursive pagination for deeply nested tables (SD-1962)

tupizz:
## Summary
- Embedded tables inside table cells now paginate correctly at row boundaries across pages, even when nested multiple levels deep (table-in-table-in-table)
- Previously, a deeply nested table (e.g., the second "Collection Time" table inside the first "Collection Time" table inside the outer amendment table) was treated as a single unsplittable segment, causing its content to be clipped or lost on continuation pages
- The fix makes both the layout engine's segment expansion and the renderer's segment counting recursive, so any depth of table nesting paginates correctly

## Test plan
- [x] All 480 layout-engine tests pass
- [x] All 629 painter-dom tests pass
- [x] Browser verification: `text_cut_table.docx` renders all three nesting levels correctly across pages 11-13
- [x] Visual regression tests

🔗 Relates to SD-1962

### DIFF
--- a/packages/layout-engine/painters/dom/src/table/renderTableCell.ts
+++ b/packages/layout-engine/painters/dom/src/table/renderTableCell.ts
@@ -1041,8 +1041,8 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
     // Calculate total segments across all blocks for proper global index mapping.
     // Embedded tables expand recursively (matching the layout engine's getCellLines()
     // which uses getEmbeddedRowLines() for recursive nested table expansion).
-    // Non-paragraph blocks (images, drawings) occupy 1 segment each, except anchored
-    // blocks which are rendered out-of-flow and do not consume segment indices.
+    // Non-paragraph blocks (images, drawings) occupy 1 segment each when height > 0,
+    // including anchored blocks (matching getCellLines() in layout-table.ts).
     const blockLineCounts: number[] = [];
     for (let i = 0; i < Math.min(blockMeasures.length, cellBlocks.length); i++) {
       const bm = blockMeasures[i];
@@ -1053,15 +1053,11 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
         // Embedded tables: recursively count segments (matches getCellLines expansion)
         blockLineCounts.push(getEmbeddedTableSegmentCount(bm as TableMeasure));
       } else {
-        // Anchored blocks are rendered out-of-flow — they don't consume segment slots.
-        // Skip them to stay in sync with the rendering loop which also skips them.
-        const anchor = (blk as ImageBlock | DrawingBlock)?.anchor;
-        if (anchor?.isAnchored) {
-          blockLineCounts.push(0);
-        } else {
-          // Non-anchored non-paragraph blocks (image, drawing) occupy 1 segment
-          blockLineCounts.push(1);
-        }
+        // Non-paragraph/non-table blocks (images, drawings) occupy 1 segment when
+        // their height > 0, matching getCellLines() in layout-table.ts which only
+        // counts non-paragraph blocks with positive height.
+        const blockHeight = 'height' in bm ? (bm as { height: number }).height : 0;
+        blockLineCounts.push(blockHeight > 0 ? 1 : 0);
       }
     }
     const totalLines = blockLineCounts.reduce((a, b) => a + b, 0);
@@ -1109,6 +1105,11 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
       if (blockMeasure.kind === 'image' && block?.kind === 'image') {
         if (block.anchor?.isAnchored) {
           anchoredBlocks.push({ block, measure: blockMeasure as ImageMeasure });
+          // Advance cumulative count only when height > 0 to stay aligned with
+          // getCellLines() which only counts non-paragraph blocks with positive height.
+          if (blockMeasure.height > 0) {
+            cumulativeLineCount += 1;
+          }
           continue;
         }
 
@@ -1154,6 +1155,11 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
       if (blockMeasure.kind === 'drawing' && block?.kind === 'drawing') {
         if (block.anchor?.isAnchored) {
           anchoredBlocks.push({ block, measure: blockMeasure as DrawingMeasure });
+          // Advance cumulative count only when height > 0 to stay aligned with
+          // getCellLines() which only counts non-paragraph blocks with positive height.
+          if (blockMeasure.height > 0) {
+            cumulativeLineCount += 1;
+          }
           continue;
         }
 
--- a/packages/layout-engine/layout-bridge/src/incrementalLayout.ts
+++ b/packages/layout-engine/layout-bridge/src/incrementalLayout.ts
@@ -768,8 +768,14 @@ export async function incrementalLayout(
     hasPreviousMeasures &&
     previousConstraints?.measurementWidth === measurementWidth &&
     previousConstraints?.measurementHeight === measurementHeight;
+  const previousPerSectionConstraints = canReusePreviousMeasures
+    ? computePerSectionConstraints(options, previousBlocks)
+    : null;
   const previousMeasuresById = canReusePreviousMeasures
     ? new Map(previousBlocks.map((block, index) => [block.id, previousMeasures![index]]))
+    : null;
+  const previousConstraintsById = canReusePreviousMeasures
+    ? new Map(previousBlocks.map((block, index) => [block.id, previousPerSectionConstraints![index]]))
     : null;
 
   const measureStart = performance.now();
@@ -794,7 +800,12 @@ export async function incrementalLayout(
 
     if (canReusePreviousMeasures && dirty.stableBlockIds.has(block.id)) {
       const previousMeasure = previousMeasuresById?.get(block.id);
-      if (previousMeasure) {
+      const previousBlockConstraints = previousConstraintsById?.get(block.id);
+      if (
+        previousMeasure &&
+        previousBlockConstraints?.maxWidth === blockMeasureWidth &&
+        previousBlockConstraints?.maxHeight === blockMeasureHeight
+      ) {
         measures.push(previousMeasure);
         reusedMeasures++;
         continue;
--- a/packages/layout-engine/layout-bridge/test/incrementalLayout.previous-measure-reuse.test.ts
+++ b/packages/layout-engine/layout-bridge/test/incrementalLayout.previous-measure-reuse.test.ts
@@ -1,0 +1,76 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { FlowBlock, ParagraphMeasure, SectionBreakBlock } from '@superdoc/contracts';
+import { incrementalLayout, measureCache } from '../src/incrementalLayout';
+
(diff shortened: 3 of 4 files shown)


packages/layout-engine/layout-engine/src/layout-table.ts
caio-pizzol: wrong JSDoc - this block describes `getCellLines` but sits above `getEmbeddedRowLines`. move it to line 377 or drop it

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -764,26 +886,147 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
...
+            for (const cell of row.cells) {
+              const cellTotal = getCellSegmentCount(cell);
+              const cellFrom = Math.min(rowLocalFrom, cellTotal);
+              const cellTo = Math.min(rowLocalTo, cellTotal);
+              fromLineByCell.push(cellFrom);
+              toLineByCell.push(cellTo);
+
+              // Compute visible height for this cell's segment range
+              let cellVisHeight = 0;
+              if (cell.blocks && cell.blocks.length > 0) {
```
caio-pizzol: `getCellSegmentCount` handles the `cell.paragraph` fallback (line 118) but this height loop only checks `cell.blocks`.

if a nested cell uses the legacy path, `partialHeight` stays 0 and the embedded table fragment could disappear.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -764,26 +886,147 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
...
+                      }
+                      segIdx++;
+                    }
+                  } else if (blk.kind === 'table') {
+                    const nestedTable = blk as TableMeasure;
+                    for (const nestedRow of nestedTable.rows) {
+                      const nestedRowSegs = getEmbeddedRowSegmentCount(nestedRow);
+                      for (let s = 0; s < nestedRowSegs; s++) {
+                        if (segIdx >= cellFrom && segIdx < cellTo) {
+                          cellVisHeight += (nestedRow.height || 0) / nestedRowSegs;
```
caio-pizzol: dividing `nestedRow.height` evenly by `nestedRowSegs` is approximate - rows with non-uniform line heights will get the wrong split height.

maybe we could use `getEmbeddedRowLines(nestedRow)` to get actual segment heights like the layout engine does.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -764,26 +886,147 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
...
+                      }
+                      segIdx++;
+                    }
+                  }
+                }
+              }
+              partialHeight = Math.max(partialHeight, cellVisHeight);
+            }
+
+            partialRowInfo = {
```
caio-pizzol: `partialRowInfo` gets overwritten each iteration - if the visible range cuts through two different rows (middle-page fragment of a tall table), only the last one survives. 

`TableFragment` only supports one `partialRow`, so this is a design gap. worth a TODO?

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
caio-pizzol, line 1106: anchored images get counted as 1 segment at line 865 but skip `cumulativeLineCount++` here (the continue at line 1036 fires before the increment at line 1041).

this drifts all subsequent blocks' segment indices. 

we should either exclude anchored blocks from the count at line 865 or add `cumulativeLineCount += 1` before the continue.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -386,17 +456,62 @@ const EMBEDDED_TABLE_VERSION = 'embedded-table';
...
+  let visibleHeight = 0;
+  for (let r = effectiveFromRow; r < effectiveToRow; r++) {
+    if (paramPartialRow && paramPartialRow.rowIndex === r) {
+      visibleHeight += paramPartialRow.partialHeight;
+    } else {
+      visibleHeight += measure.rows[r]?.height || 0;
+    }
+  }
+
+  // Rescale column widths when measurement-scale exceeds render-scale (SD-1962).
```
caio-pizzol, line 552: this rescaling block is the same logic as `rescaleColumnWidths()` in `layout-table.ts:173-194` - same `Math.max(1, Math.round(w * scale))`, same rounding `fixup` on the last column.

maybe import and call the existing function instead of duplicating it?

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -764,26 +886,147 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
 
       if (blockMeasure.kind === 'table' && block?.kind === 'table') {
```
caio-pizzol, line 1082: this entire embedded-table branch (~120 lines to line 1030) does segment mapping, row range detection, partial row computation, height calc, and DOM creation all inline.

extracting it to a function like `renderPartialEmbeddedTable(...)` would make it easier to understand/maintain

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -386,17 +456,62 @@ const EMBEDDED_TABLE_VERSION = 'embedded-table';
...
+    toRow: paramToRow,
+    partialRow: paramPartialRow,
+  } = params;
+
+  const effectiveFromRow = paramFromRow ?? 0;
+  const effectiveToRow = paramToRow ?? table.rows.length;
+
+  // Calculate the height for the visible row range.
+  // For rows with partial rendering (mid-row split), use the partial height.
+  let visibleHeight = 0;
```
caio-pizzol: the "sum row heights, use `partialHeight` for the partial row" loop appears here and again at line 996.

a small helper like `computeVisibleHeight(rows, fromRow, toRow, partialRow)` would deduplicate.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -88,6 +89,67 @@ type WordLayoutInfo = {
...
 type TableRowMeasure = TableMeasure['rows'][number];
+type TableCellMeasure = TableRowMeasure['cells'][number];
+
+/**
+ * Compute the total segment count for a cell's blocks, matching the layout engine's
+ * recursive getCellLines() expansion. Paragraph blocks contribute their line count,
+ * embedded tables contribute the sum of their rows' recursive segment counts,
+ * and other blocks (images, drawings) contribute 1 segment.
+ */
+function getCellSegmentCount(cell: TableCellMeasure): number {
```
caio-pizzol: these three functions (`getCellSegmentCount`, `getEmbeddedRowSegmentCount`, `getEmbeddedTableSegmentCount`) mirror the layout engine's `getCellLines` + `getEmbeddedRowLines`.

any change to one side must be replicated on the other - maybe should consider adding a test asserting `getCellSegmentCount(cell) === getCellLines(cell).length` for a few cell shapes so drift gets caught.

packages/layout-engine/layout-engine/src/layout-table.ts
tupizz: Fixed in d44f370 — removed the orphaned JSDoc block. `getCellLines` at line 368 is self-documenting with inline comments.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
tupizz: Fixed in d44f370 — extracted `computeCellVisibleHeight()` helper (renderTableCell.ts:181) that handles both `cell.blocks` and `cell.paragraph` paths, matching `getCellSegmentCount`. The old inline height loop is replaced with a call to this helper inside the new `renderPartialEmbeddedTable()` function.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
tupizz: Added a TODO in `computeCellVisibleHeight()` (renderTableCell.ts:205). Importing `getEmbeddedRowLines` from layout-engine for this edge case (partially-visible nested rows within nested tables) would add cross-package coupling for a marginal visual improvement — the approximation only affects wrapper div height, not content rendering. Worth revisiting if we see visual artifacts from uneven splits.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
tupizz: Added a TODO in `renderPartialEmbeddedTable()` documenting the design gap. `TableFragment` only supports a single `partialRow`, so fixing this properly requires a contract change — not something to tackle in this PR.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
tupizz: Fixed in d44f370 — the initial segment counting loop (renderTableCell.ts:1062) now checks `anchor?.isAnchored` and pushes `0` for anchored blocks instead of `1`, matching the rendering loop which skips them with `continue` before incrementing `cumulativeLineCount`.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
tupizz: Fixed in d44f370 — exported `rescaleColumnWidths` from `@superdoc/layout-engine` (layout-table.ts:173, index.ts:2664) and imported it in renderTableCell.ts. The ~15-line inline rescaling block in `renderEmbeddedTable` is replaced with a single call (renderTableCell.ts:560). Added `@superdoc/layout-engine` as a dependency in painter-dom's package.json.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
tupizz: Fixed in d44f370 — extracted `renderPartialEmbeddedTable()` (renderTableCell.ts:610). The 143-line inline block is now a named function that handles segment mapping, row range detection, partial row computation, height calculation, and delegates DOM creation to `renderEmbeddedTable`. The call site in the rendering loop is now ~15 lines.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
tupizz: Fixed in d44f370 — extracted `computeVisibleHeight()` helper (renderTableCell.ts:157). Used in both `renderEmbeddedTable` (line 555) and `renderPartialEmbeddedTable` (line 720). Identical loop deduplicated.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
tupizz: Added in d44f370 — 8 sync tests in renderTableCell.test.ts asserting `getCellSegmentCount(cell) === getCellLines(cell).length` for: simple paragraph, legacy single-paragraph, multi-block with image, nested table, double-nested table, triple-nested table (table-in-table-in-table), empty cell, and zero-height image. Exported `getCellSegmentCount` from renderTableCell.ts and `getCellLines` from layout-engine to make them testable. All 637 painter-dom tests pass.

harbournick (approved):
LGTM - such a great improvement! nested tables ftw!

harbournick:

PR #2157, opened 2026-02-24
fix(comments): remove synchronous dispatch from plugin apply() (SD-1940)

tupizz:
## Summary

- Fixes the comment focus loop when programmatically selecting text and immediately calling `addComment()` ([SD-1940](https://linear.app/superdocworkspace/issue/SD-1940))
- Removes 4 lines of redundant code from the comments plugin `apply()` method
- Adds 2 regression tests covering the exact customer scenario

## Root Cause

The comments plugin's `apply()` method (line ~514) was synchronously calling `view.dispatch()` inside `apply()` when the active comment thread changed due to a selection change:

```javascript
// REMOVED — synchronous dispatch inside apply()
const { tr: newTr } = editor.view.state;
const { dispatch } = editor.view;
newTr.setMeta(CommentsPluginKey, { type: 'force' });
dispatch(newTr);
```

This dispatch was **redundant** — its only purpose was to set `shouldUpdate = true` via the `'force'` meta, but `shouldUpdate = true` was already set 3 lines above. The `view.update()` method runs automatically after every transaction and checks `shouldUpdate`, so the extra dispatch accomplished nothing beyond creating a re-entrant transaction cycle.

When a consumer programmatically dispatches a selection change followed immediately by `addComment()`, the re-entrant dispatch from `apply()` competed with the `addComment` transaction, causing a recursive loop that froze the browser.

## Why this approach

We considered two approaches:

1. **`queueMicrotask` wrapper** (ticket's original suggestion) — defer the dispatch to break the synchronous cycle
2. **Remove the dispatch entirely** (chosen) — since `shouldUpdate = true` already triggers the same code path in `view.update()`, the dispatch is dead code. Deferring dead code is worse than removing it.

The `shouldUpdate` flag at line 511 is sufficient because:
- It's checked by `view.update()` which runs after every transaction
- `view.update()` also independently detects `activeThreadChanged` (line 548-551) as a secondary guard
- In presentation mode, `view.update()` returns early anyway — DomPainter handles rendering

## Test plan

- [x] Existing 6,388 super-editor tests pass
- [x] New test: `apply()` does not call `view.dispatch()` when selection moves onto a comment
- [x] New test: programmatic `setTextSelection` + `addComment` completes without dispatch loop
- [ ] Manual: run the customer's code in browser console — no freeze

### DIFF
--- a/packages/superdoc/src/stores/comments-store.js
+++ b/packages/superdoc/src/stores/comments-store.js
@@ -280,15 +280,12 @@ export const useCommentsStore = defineStore('comments', () => {
 
     // If no ID, we clear any focused comments
     if (id === undefined || id === null) {
-      if (activeComment.value === null) return;
       activeComment.value = null;
       activeEditor?.commands?.setActiveComment({ commentId: null });
       return;
     }
 
     const comment = getComment(id);
-    const newCommentId = comment ? comment.commentId : null;
-    if (activeComment.value === newCommentId) return;
     if (comment) activeComment.value = comment.commentId;
     activeEditor?.commands?.setActiveComment({ commentId: activeComment.value });
   };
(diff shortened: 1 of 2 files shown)


harbournick:

PR #2164, opened 2026-02-24
fix(tracked-changes): sync tracked changes store on undo and redo

palmer-cl:
- fix undo and redo logic with tracked changes
- add behavior tests and visual tests

### DIFF
--- a/packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
+++ b/packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
@@ -2519,13 +2519,6 @@ export class PresentationEditor extends EventEmitter {
         if (shouldBypassFastRevision) {
           this.#flowBlockCache?.setHasExternalChanges(true);
         }
-        // History undo/redo can restore prior paragraph content while preserving/reusing
-        // sdBlockRev values, which makes the cache's fast revision check unsafe.
-        // Force JSON comparison for this render cycle to avoid stale paragraph reuse.
-        const isHistoryType = inputType === 'historyUndo' || inputType === 'historyRedo';
-        if (isHistoryType && transaction.docChanged) {
-          this.#flowBlockCache?.setHasExternalChanges(true);
-        }
       }
       if (trackedChangesChanged || transaction?.docChanged) {
         this.#pendingDocChange = true;
--- a/tests/behavior/tests/comments/undo-tracked-insert-removes-sidebar.spec.ts
+++ b/tests/behavior/tests/comments/undo-tracked-insert-removes-sidebar.spec.ts
@@ -1,5 +1,6 @@
 import { test, expect } from '../../fixtures/superdoc.js';
-import { listTrackChanges } from '../../helpers/document-api.js';
+import { getDocumentText, listTrackChanges } from '../../helpers/document-api.js';
+import { activateCommentDialog } from '../../helpers/comments.js';
 
 test.use({ config: { toolbar: 'full', comments: 'panel', trackChanges: true } });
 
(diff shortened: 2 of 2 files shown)


packages/superdoc/src/stores/comments-store.js
```diff
@@ -154,6 +154,115 @@ export const useCommentsStore = defineStore('comments', () => {
...
+        const resolvedDocumentId = comment?.fileId ?? null;
+        if (resolvedDocumentId && resolvedDocumentId !== documentId) return;
+      }
+
+      const positionKey = getCommentPositionKey(comment);
+      if (!positionKey) return;
+      const normalizedPositionKey = String(positionKey);
+      // mirror editorCommentPositions keying: positions are indexed by getCommentPositionKey
+      // (importedId when present, otherwise commentId), so we normalize against that key first.
+      if (currentPositions[normalizedPositionKey] === undefined) return;
```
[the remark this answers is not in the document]
palmer-cl: added more robust handling for edge cases around commentId or importedId

packages/superdoc/src/components/CommentsLayer/FloatingComments.vue
```diff
@@ -156,6 +156,34 @@ watch(activeZoom, () => {
   }
 });
 
+watch(
+  getFloatingComments,
+  (comments) => {
+    const activeIds = new Set(
+      comments
+        .map((comment) => comment?.commentId ?? comment?.importedId)
```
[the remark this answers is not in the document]
palmer-cl: Made a small change to account for both importID and commentId

NOTE: @caio-pizzol might be worth in another ticket to do an audit of comment ids. I think it warrants its own PR, but doing an app wide normalization for comment id ordering might be helpful. currently the vue layer looks for commentId first, but the store looks for importId first. Usually its fine, but this would be a small QOL change and can then remove some extra code.

packages/superdoc/src/stores/comments-store.js
```diff
@@ -154,6 +154,114 @@ export const useCommentsStore = defineStore('comments', () => {
...
+  const syncTrackedChangePositionsWithDocument = ({ documentId, editor } = {}) => {
+    // Keep editor-driven comment anchors in sync with live tracked-change marks
+    if (!editor?.state) return 0;
+    if (!commentsList.value?.length) return 0;
+
+    const currentPositions = editorCommentPositions.value || {};
+    if (!Object.keys(currentPositions).length) return 0;
+
+    const toNormalizedId = (id) => (id === undefined || id === null ? null : String(id));
+    const getCommentAliasIds = (comment) =>
```
caio-pizzol: alias strategies are different in three places - store's `getCommentAliasIds` includes `getCommentPositionKey`, vue's version only uses `[importedId, commentId]`, and `getFloatingComments` uses `commentId || importedId`.

they converge for new tracked changes but will mismatch on imported docs with both IDs.

packages/superdoc/src/stores/comments-store.js
```diff
@@ -823,9 +932,6 @@ export const useCommentsStore = defineStore('comments', () => {
    * @returns {void}
    */
   const handleEditorLocationsUpdate = (allCommentPositions) => {
```
caio-pizzol, line 990: correct for undo, but changes behavior for all callers.

heads up: #2195 is incoming and touches the same files (`FloatingComments.vue`, `comments-store.js`, `SuperDoc.vue`) - it adds stale cleanup in `FloatingComments` that mitigates any flicker from transient empty payloads, so should be safe once both land

caio-pizzol (commented):
solid fix - guard removal + explicit sync on undo/redo addresses the actual data flow gap (PR #2195 will prob conflict here)

packages/superdoc/src/stores/comments-store.js
palmer-cl: @caio-pizzol - reworked comment id normalization. lets do a thorough pass to make sure this is working correctly.

packages/superdoc/src/stores/comments-store.js
palmer-cl: @caio-pizzol can you elaborate on:

> correct for undo, but changes behavior for all callers.

> so should be safe once both land

palmer-cl: I merged your changes in as well. Just make sure thats working as you would expect.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -4168,6 +3955,67 @@ export class PresentationEditor extends EventEmitter {
...
+   * caret element's screen-space position (via getBoundingClientRect) to determine if
+   * scrolling is needed, keeping a small margin for comfortable viewing.
+   *
+   * If the caret element doesn't exist (page may be virtualized / not mounted),
+   * falls back to scrolling the target page into view to trigger virtualization.
+   */
+  #scrollCaretIntoViewIfNeeded(caretLayout: { pageIndex: number }): void {
+    const caretEl = this.#localSelectionLayer?.querySelector(
+      '.presentation-editor__selection-caret',
+    ) as HTMLElement | null;
```
caio-pizzol: the caret element can exist even when the page isn't actually visible (the rendering has a fallback that creates it with estimated coordinates). when that happens, this skips `#scrollPageIntoView` and scrolls to the wrong spot. worth checking if the page is actually mounted before using the caret rect?

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -4124,6 +3923,7 @@ export class PresentationEditor extends EventEmitter {
           console.warn('[PresentationEditor] Failed to render caret overlay:', error);
         }
       }
+      this.#scrollCaretIntoViewIfNeeded(caretLayout);
```
caio-pizzol: this scrolls on every selection change, not just arrow keys — collab edits, undo, find-and-replace all trigger it. the margin check avoids scrolling when the caret is already visible, but are there cases where this causes unexpected jumps?

packages/super-editor/src/extensions/vertical-navigation/vertical-navigation.js
```diff
@@ -118,8 +118,37 @@ export const VerticalNavigation = Extension.create({
...
+          // Check if the hit test result is plausible: if the adjacent line has PM
+          // position data, the hit should land within or very close to that range.
+          // A miss indicates off-screen coordinate mapping failure or fragment
+          // boundary misalignment (adjacent line center Y mapping to wrong fragment).
+          //
+          // Tolerance is kept small (5 positions) to catch cases where the hit
+          // lands on the current line's fragment start instead of the adjacent
+          // line — this causes the cursor to appear stuck since the "new" position
+          // equals the current one.
+          if (adjacent.pmStart != null && adjacent.pmEnd != null) {
```
caio-pizzol: the fallback path (`resolvePositionAtGoalX`) has zero test coverage — no test sets `data-pm-start`/`data-pm-end`, so this block never runs. two cases would cover it: hit lands in range (fallback skipped) and hit lands out of range (fallback runs). the null-skip fix should get a case too.

packages/super-editor/src/core/presentation-editor/tests/scrollCaretIntoView.test.ts
```diff
@@ -0,0 +1,214 @@
...
+ * we extract its logic into a testable helper and verify it against a real DOM
+ * setup with mock scroll containers and caret elements.
+ */
+
+/**
+ * Extracted logic from PresentationEditor.#scrollCaretIntoViewIfNeeded.
+ * This mirrors the implementation to allow direct unit testing without
+ * bootstrapping the full PresentationEditor.
+ */
+function scrollCaretIntoViewIfNeeded(
```
caio-pizzol: re-flagging from last round — the test copies the full method. if the real code changes, these tests still pass on the old version. worth extracting into a shared function both can use?

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -2506,18 +2506,22 @@ export class PresentationEditor extends EventEmitter {
...
+        // cannot be trusted. History undo/redo can also restore tracked-mark-only
+        // changes where visible text stays the same, so use the same JSON fallback.
         const ySyncMeta = transaction.getMeta?.(ySyncPluginKey);
-        if (ySyncMeta?.isChangeOrigin && transaction.docChanged) {
+        const shouldBypassFastRevision =
+          transaction.docChanged &&
+          (ySyncMeta?.isChangeOrigin || inputType === 'historyUndo' || inputType === 'historyRedo');
+        if (shouldBypassFastRevision) {
           this.#flowBlockCache?.setHasExternalChanges(true);
         }
```
caio-pizzol: this now covers undo/redo too, but the `isHistoryType` block below (lines 2525-2528) still does the same thing. it causes the new test to fail (expects 2 calls, gets 4). removing it fixes the test.

palmer-cl: Ok these should be resolved now! added 2 new tests!

harbournick:

PR #2165, opened 2026-02-24
docs(dev): add dev:docs workflow for local SuperDoc development

lucbic:
## The problem
I was trying to fix a bug in the Image extension section of the docs, but I noticed that changes in the `packages/super-editor` were not being reflected in the docs local dev server.

## The idea
Instead of just fetching the superdoc bundle from unpkg, it detects if the user is running the docs in development and fetches the built files from the vite dev server. The file changes are watched to trigger new builds, which run instantaneously thanks to vite.

## The feature
Enable docs contributors to preview interactive widgets with local source changes. `pnpm dev:docs` starts the Vite dev server, UMD watcher, and Mintlify in parallel.

  - Refactor SuperDocEditor widget to load from local dev server on localhost
  - Add serve-dist-for-docs Vite plugin to serve UMD bundle over CORS
  - Add watch:umd script with debounced rebuilds (300ms)
  - Document the workflow in apps/docs/README.md

### DIFF
--- a/apps/docs/snippets/components/superdoc-editor.jsx
+++ b/apps/docs/snippets/components/superdoc-editor.jsx
@@ -44,8 +44,9 @@ export const SuperDocEditor = ({
     if (existingScript) {
       if (window.SuperDocLibrary) return Promise.resolve();
 
-      return new Promise((resolve) => {
+      return new Promise((resolve, reject) => {
         existingScript.addEventListener('load', resolve, { once: true });
+        existingScript.addEventListener('error', reject, { once: true });
       });
     }
 


packages/superdoc/vite.config.umd.js
```diff
@@ -20,6 +20,9 @@ export default defineConfig(({ command }) => {
     },
     build: {
       emptyOutDir: false,
+      watch: {
+        buildDelay: 300,
+      },
```
caio-pizzol: in Vite 7, setting `build.watch` to any non-null value forces watch mode regardless of CLI flags, so `pnpm build` will hang and never finish — remove this block since `watch:umd` already passes `--watch` via CLI.

apps/docs/snippets/components/superdoc-editor.jsx
```diff
@@ -9,33 +9,88 @@ export const SuperDocEditor = ({
...
+          if (onReady) onReady(editorRef.current);
+        },
+      });
+    }, 100);
+  };
+
+  useEffect(() => {
+    let cancelled = false;
+
+    const boot = async () => {
```
caio-pizzol, line 83: wrap the call in a try/catch with `console.error` so a network failure during script load surfaces instead of leaving the editor silently blank.

apps/docs/snippets/components/superdoc-editor.jsx
```diff
@@ -9,33 +9,88 @@ export const SuperDocEditor = ({
...
-          });
-        }
-      }, 100);
+  const loadSuperDocLibrary = (baseUrl) => {
+    if (window.SuperDocLibrary) return Promise.resolve();
+
+    const scriptSrc = `${baseUrl}/superdoc.umd.js`;
+    const existingScript = document.querySelector(`script[src="${scriptSrc}"]`);
+
+    if (existingScript) {
```
caio-pizzol, line 44: if the script already finished loading before this code runs, the `load` event never fires again and this promise hangs forever — add a `window.SuperDocLibrary` check right after finding the existing script to resolve immediately in that case.

packages/superdoc/vite.config.umd.js
lucbic: @caio-pizzol Got it! I have changed the watch config to be conditional according to the --watch flag because we need this buildDelay property though. The IDE was freezing without this debouncing logic.

apps/docs/snippets/components/superdoc-editor.jsx
lucbic: @caio-pizzol good catch! done ✅

apps/docs/snippets/components/superdoc-editor.jsx
lucbic: @caio-pizzol done ✅

apps/docs/snippets/components/superdoc-editor.jsx
```diff
@@ -9,33 +9,94 @@ export const SuperDocEditor = ({
...
+      if (window.SuperDocLibrary) return Promise.resolve();
+
+      return new Promise((resolve) => {
+        existingScript.addEventListener('load', resolve, { once: true });
+      });
```
caio-pizzol: if the script tag exists but already failed (e.g. dev server was down on first load), neither `load` nor `error` will fire again — the promise hangs forever and the editor silently stalls. adding an `error` listener fixes it:

```suggestion
      return new Promise((resolve, reject) => {
        existingScript.addEventListener('load', resolve, { once: true });
        existingScript.addEventListener('error', reject, { once: true });
      });
```

lucbic: @caio-pizzol done!

harbournick:

Issue #2155, opened 2026-02-23
Docs: Images not appearing in the Image and Node Resizer extension pages


lucbic:
### What happened?

There are two issues happening: The external url image set in the example is being blocked by CORS policy, and the local files in the `apps/docs/public` folder are not accepted because the `registerImages` helper is not prepared to handle local files.

<img width="1560" height="1370" alt="Image" src="https://github.com/user-attachments/assets/c71d0f3d-87fe-4899-bb80-db2f92a6c7c4" />

<img width="1190" height="418" alt="Image" src="https://github.com/user-attachments/assets/c6113360-68dc-4de9-99c2-2fe3bf44bf55" />

### Steps to reproduce

1) pnpm run dev:docs
2) http://localhost:3001/extensions/image
3) Images don't appear

### Browser

Chrome

caio-pizzol:
Fixed on https://github.com/superdoc-dev/superdoc/pull/2531

PR #2167, opened 2026-02-24
docs: fix relative path images not appearing

lucbic:
sanitizeHref previously rejected all relative paths, which broke image rendering in docs widgets that use paths like /public/images/photo.png.

Instead of rejecting, relative paths are now resolved against the page origin with security enforcement: control character rejection, same-origin gate, and redirect blocklist support. This also skips the destructive image registration pipeline for relative-path images since the browser resolves them directly.

Refs: #2155

### DIFF
--- a/packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.browser.test.js
+++ b/packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.browser.test.js
@@ -1,4 +1,5 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { flushPromises } from '@vue/test-utils';
 
 // ── Hoisted mock objects (available to vi.mock factories) ─────────────
 const { mockDecoSet, mockPluginKeyInstance } = vi.hoisted(() => {
@@ -69,8 +70,6 @@ import { urlToFile, validateUrlAccessibility } from './handleUrl';
 import { addImageRelationship } from '@extensions/image/imageHelpers/startImageUpload.js';
 
 // ── Helpers ───────────────────────────────────────────────────────────
-const flushPromises = () => new Promise((resolve) => setTimeout(resolve, 0));
-
 const createImageNode = (attrs) => ({
   type: { name: 'image' },
   attrs,
@@ -240,5 +239,52 @@ describe('registerRelativeImages (via handleBrowserPath)', () => {
       originalSrc: 'word/media/photo.jpg',
     });
     expect(view.dispatch).toHaveBeenCalled();
+  });
+
+  it('skips duplicate relative URL when the first is still being registered', async () => {
+    // Create a deferred promise so the first urlToFile call hangs
+    let resolveFirst;
+    const firstCall = new Promise((r) => (resolveFirst = r));
+    urlToFile.mockReturnValueOnce(firstCall);
+
+    const imageA = { node: createImageNode({ src: 'assets/dup.png' }), pos: 0, id: {} };
+    const imageB = { node: createImageNode({ src: 'assets/dup.png' }), pos: 5, id: {} };
+
+    // Both arrive in the same batch
+    handleBrowserPath([imageA, imageB], editor, view, { tr: createTrStub() });
+    await flushPromises();
+
+    // Only one fetch is initiated — the second is blocked by pendingRelativeRegistrations
+    expect(urlToFile).toHaveBeenCalledTimes(1);
+
+    // Complete the first registration
+    resolveFirst(null);
+    await flushPromises();
+
+    // pendingRelativeRegistrations is cleaned up so a future insert could register again
+    expect(editor.storage.image.pendingRelativeRegistrations.size).toBe(0);
+  });
+
+  it('does not register or dispatch when urlToFile returns null (fetch failure)', async () => {
+    urlToFile.mockResolvedValueOnce(null);
+
+    const foundImages = [{ node: createImageNode({ src: 'assets/missing.png' }), pos: 0, id: {} }];
+
+    handleBrowserPath(foundImages, editor, view, { tr: createTrStub() });
+    await flushPromises();
+
+    expect(urlToFile).toHaveBeenCalledWith('assets/missing.png', 'missing.png');
+
+    // No media stored
+    expect(Object.keys(editor.storage.image.media)).toHaveLength(0);
+
+    // No relationship created
+    expect(addImageRelationship).not.toHaveBeenCalled();
+
+    // No transaction dispatched
+    expect(view.dispatch).not.toHaveBeenCalled();
+
+    // Pending set cleaned up so the src can be retried later
+    expect(editor.storage.image.pendingRelativeRegistrations.has('assets/missing.png')).toBe(false);
   });
 });


packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
```diff
@@ -30,6 +30,11 @@ export const needsImageRegistration = (node) => {
   // Data URI with rId means it was converted (e.g., EMF→SVG) but already has export metadata
   if (src.startsWith('data:') && node.attrs?.rId) return false;
 
+  // Relative or absolute path (e.g., /images/photo.png) — the browser resolves these directly
+  if (!src.startsWith('http') && !src.startsWith('data:') && !src.startsWith('blob:')) {
+    return false;
```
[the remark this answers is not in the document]
lucbic: on it

shared/url-validation/index.js
```diff
@@ -458,11 +513,20 @@ export function sanitizeHref(raw, config = {}) {
...
+  if (trimmed.startsWith('//') || /^www\./i.test(trimmed)) {
     return null;
   }
 
+  // Normalize redirect blocklist once and enforce across all URL forms
+  const blocklist = normalizeBlocklist(config.redirectBlocklist);
+
+  // Relative paths have no protocol to validate
+  // resolve against page origin instead
+  if (trimmed.startsWith('/')) {
```
caio-pizzol: dot-relative paths like `./images/chart.jpg` skip this branch, hit the scheme regex, fail to match, and return null — the fix is `if (trimmed.startsWith('/') || trimmed.startsWith('.'))`, which also aligns with what the tests at line 130 (`index.test.js`) and 5789 (`index.test.ts`) expect.

caio-pizzol:
p.s. also rebase with main looks like all tests are failing here

shared/url-validation/index.js
lucbic: @caio-pizzol done ✅

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
```diff
@@ -181,9 +181,16 @@ export const handleNodePath = (foundImages, editor, state) => {
  * @param {import('prosemirror-state').EditorState} state - The current editor state.
  * @returns {import('prosemirror-state').Transaction} - The updated transaction with image nodes replaced by placeholders and registration process initiated.
  */
+const isRelativePath = (src) => !src.startsWith('http') && !src.startsWith('data:') && !src.startsWith('blob:');
+
```
caio-pizzol, line 267: bare paths like `images/photo.png` (no `/` or `./` prefix) get skipped by registration here but also blocked by `sanitizeHref`, so they'd silently disappear. the `/`-prefixed paths this PR targets work fine — just worth a follow-up for bare paths.

lucbic: @caio-pizzol another good catch! Done ✅

shared/url-validation/index.js
```diff
@@ -365,6 +374,52 @@ function isBlockedByRedirectList(normalizedHref, parsed, blocklist) {
...
+  // Determine base origin
+  const origin = typeof window !== 'undefined' ? window.location.origin : null;
+  if (!origin) {
+    return null;
+  }
+
+  // 2. Resolve against the origin (normalises .., //, encoded sequences)
+  let resolved;
+  try {
+    resolved = new URL(pathString, origin);
```
caio-pizzol: this resolves against the site root, not the current page. so `./photo.png` on `/docs/intro` gives `/photo.png` instead of `/docs/photo.png`. using `window.location.href` as the base would fix it.

shared/url-validation/index.js
```diff
@@ -365,6 +374,52 @@ function isBlockedByRedirectList(normalizedHref, parsed, blocklist) {
...
+  } catch {
+    return null;
+  }
+
+  // 3. Same-origin gate — prevents open-redirect style attacks
+  if (resolved.origin !== new URL(origin).origin) {
```
caio-pizzol: `origin` is already `window.location.origin` — no need to parse it again.

```suggestion
  if (resolved.origin !== origin) {
```

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
```diff
@@ -182,8 +183,13 @@ export const handleNodePath = (foundImages, editor, state) => {
  * @returns {import('prosemirror-state').Transaction} - The updated transaction with image nodes replaced by placeholders and registration process initiated.
  */
 const handleBrowserPath = (foundImages, editor, view, state) => {
+  // Relative/absolute paths are resolved by the browser natively — leave them in the doc.
+  const imagesToProcess = foundImages.filter(({ node }) => !isRelativeUrl(node.attrs?.src));
```
caio-pizzol, line 266: relative images are skipped here in the browser, which makes sense for display. but does export need these images to be registered? if so, they'd be missing their metadata.

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
```diff
@@ -251,6 +268,87 @@ export const getImageRegistrationMetaType = (tr) => {
   return null;
 };
 
+// Tracks relative URLs currently being registered to prevent duplicate async work.
+// Between the initial detection (no rId) and the async rId update, appendTransaction
+// may re-detect the same images — this Set causes them to be skipped.
+const pendingRelativeRegistrations = new Set();
```
caio-pizzol: this is shared across all editors on the page. if two editors load the same image path, the second one gets silently skipped. not blocking — but worth moving to per-editor storage in a follow-up.

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
```diff
@@ -251,6 +268,87 @@ export const getImageRegistrationMetaType = (tr) => {
...
+ * The browser displays relative images natively via their src attribute. This function
+ * fetches the binary in the background and stores export metadata (rId, media path) on
+ * the node so that DOCX export can include the image in the zip.
+ *
+ * @param {Array} images - Array of found image nodes with their positions and IDs.
+ * @param {Object} editor - The editor instance.
+ * @param {import('prosemirror-view').EditorView} view - The editor view instance.
+ */
+const registerRelativeImages = async (images, editor, view) => {
+  const mediaStore = editor.storage.image.media ?? {};
```
caio-pizzol: this setup code (lines 288-293) is the same as `handleNodePath` (lines 149-155). small helper would keep them in sync.

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
lucbic: Didn't realize this scope issue! Spot on. I'm changing the approach

lucbic:
@caio-pizzol all done! I have changed the approach for registering the pendingRelativeRegistrations, now it's using the instance storage, so I guess it would be nice if you could take a look again.

I have also rebased from main and solved a few conflicts. A Few tests are failing, but they are unrelated to the changes in this branch

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
```diff
@@ -318,6 +340,78 @@ export const getImageRegistrationMetaType = (tr) => {
...
+ * @param {import('prosemirror-view').EditorView} view - The editor view instance.
+ */
+const registerRelativeImages = async (images, editor, view) => {
+  const { mediaStore, existingFileNames } = getOrInitMediaStore(editor);
+  const { pendingRelativeRegistrations } = editor.storage.image;
+
+  for (const { node } of images) {
+    const src = node.attrs.src;
+
+    if (pendingRelativeRegistrations.has(src)) continue;
```
caio-pizzol, line 361: if the same image gets inserted twice quickly, the second copy can end up without export data — it'll be missing from the DOCX. editing near it fixes it, but that's easy to miss. worth a fix or TODO?

packages/super-editor/src/extensions/image/imageHelpers/imageRegistrationPlugin.js
```diff
@@ -318,6 +340,78 @@ export const getImageRegistrationMetaType = (tr) => {
...
+    pendingRelativeRegistrations.add(src);
+
+    try {
+      const filename = src.split('/').pop()?.split(/[?#]/)[0] || 'image.bin';
```
caio-pizzol, line 365: `derivePreferredFileName` (line 125) already does this.

```suggestion
      const filename = derivePreferredFileName(src);
```

caio-pizzol:
@lucbic thanks for all the work here - the approach is solid.                                                                                      
                                                                                                                                                                                  
picked this up in #2531. your commits are squash-merged with co-author credit, plus a few small fixes on top (reuse existing filename helper, sync image data for collab, mark an internal export).                                                                                                                                      
                                                                                                                                                                                
closing in favor of the new PR. thanks again!

PR #2171, opened 2026-02-25
fix(virtualization): correct scroll mapping and viewport sizing at non-100% zoom

tupizz:
## Summary

Fixes blank pages and excess scroll space when zoomed out (e.g., 75%) in long documents with virtualization enabled.

**Relates to:** IT-579 (cursor accelerates to bottom in long documents), IT-581 (scroll jumps on long documents)

### Root Cause

Three interrelated bugs caused the virtualization system to break at zoom != 100%:

**1. Virtual window scroll calculation ignores CSS transform scale**

`DomPainter.updateVirtualWindow()` uses `getBoundingClientRect().top` to determine scroll position. This returns **screen-space** coordinates (affected by `transform: scale(0.75)`), but `virtualOffsets` are in **layout space** (unscaled). The mismatch grows linearly with scroll position — at scrollTop=15000 with 75% zoom, the error is ~5000px, causing the virtualization to mount pages above the actual viewport (blank pages).

**2. Gap mismatch between PresentationEditor and DomPainter**

PresentationEditor calculated viewport height using 24px page gaps (`DEFAULT_PAGE_GAP`), but DomPainter's virtualization defaulted to 72px (`DEFAULT_VIRTUALIZED_PAGE_GAP`). This made the actual rendered content taller than the viewport expected, causing the spacer math to be inconsistent.

**3. CSS box overflow inflating scroll range**

CSS `transform: scale()` does **not** change an element's CSS box dimensions. At zoom < 1, painterHost's CSS box (full unscaled height, e.g., 74304px) overflowed viewportHost's `minHeight` (scaled, e.g., 53442px), making the scroll container see ~30% extra unusable scrollable space at the bottom.

### Changes

| File | Change |
|------|--------|
| `painters/dom/src/renderer.ts` | Added `zoomFactor` property + `setZoom()` method; fixed `updateVirtualWindow()` to divide `rect.top` by zoom to convert screen-space → layout-space |
| `painters/dom/src/index.ts` | Exposed `setZoom()` through the `createDomPainter` facade |
| `PresentationEditor.ts` | Wire zoom to DomPainter via `setZoom()` in both `setZoom()` and `#ensurePainter()`; normalize virtualization gap to match effective page gap; use explicit `height` + `overflow: hidden` on viewportHost |

### Before / After (75% zoom, scrollTop=15000)

| | Before | After |
|---|--------|-------|
| **Visible content** | Blank page | Paragraphs 172-175 visible |
| **scrollHeight** | 74304px (unscaled CSS box) | 53462px (correct scaled size) |
| **Viewport sizing** | `minHeight: 53442px` (overridden by child CSS box) | `height: 53442px` + `overflow: hidden` |

## Test plan

- [x] painter-dom tests pass (638/638)
- [x] layout-engine tests pass (480/480)
- [x] pm-adapter tests pass (1544/1544)
- [x] TypeScript builds clean
- [x] Verified in browser: 75% zoom — content visible at scroll positions 0, 5000, 10000, 15000, 20000, 30000, and end-of-document
- [x] Verified 100% zoom still works correctly
- [x] Verified 150% zoom works correctly
- [ ] Manual QA: test with various long documents at 50%, 75%, 100%, 125%, 150% zoom
- [ ] Manual QA: verify cursor click-to-position works at zoomed-out levels

### DIFF
--- a/packages/layout-engine/painters/dom/src/virtualization.test.ts
+++ b/packages/layout-engine/painters/dom/src/virtualization.test.ts
@@ -411,18 +411,13 @@ describe('DomPainter virtualization (vertical)', () => {
 
     painter.onScroll!();
 
-    // At layoutScrollY=5000, the anchor page should be around index 8-9
-    // (each page is 500+72=572px, so page 8 starts at 8*572=4576).
+    // At layoutScrollY=5000, the anchor page is index 8 (topOfIndex(8)=4576 <= 5000, topOfIndex(9)=5148 > 5000).
+    // With window=3, overscan=0, the window is centered around the anchor: pages [7, 8, 9].
+    // Without the zoom correction, scrollY would be 3750 (screen-space), giving anchor=6 and pages [5, 6, 7].
     const pages = mount.querySelectorAll('.superdoc-page');
     const indices = Array.from(pages).map((p) => Number((p as HTMLElement).dataset.pageIndex));
 
-    // The anchor page for scroll=5000 should be 8 (topOfIndex(9)=5148 > 5000, topOfIndex(8)=4576 <= 5000).
-    // With window=3, we expect pages around index 7-9 or 8-10.
-    expect(indices.some((i) => i >= 7 && i <= 10)).toBe(true);
-
-    // Verify the window doesn't include page 0 (would indicate uncorrected zoom — scrollY=3750 → page 6).
-    // With correct zoom handling, we shouldn't see early pages.
-    expect(indices.every((i) => i >= 5)).toBe(true);
+    expect(indices).toEqual([7, 8, 9]);
   });
 
   it('renders drawing fragments inside virtualized windows', () => {
--- a/tests/behavior/tests/virtualization/zoom-scroll-visibility.spec.ts
+++ b/tests/behavior/tests/virtualization/zoom-scroll-visibility.spec.ts
@@ -1,0 +1,88 @@
+import { test, expect } from '../../fixtures/superdoc.js';
+
+test.use({ config: { toolbar: 'none' } });
+
+test('content is visible at 75% zoom when scrolled to mid-document', async ({ superdoc }) => {
+  // Generate a long document (~60 paragraphs) to span many pages and trigger virtualization.
+  await superdoc.page.evaluate(() => {
+    const editor = (window as any).editor;
+    const { state } = editor;
+    const { schema } = state;
+
+    const paragraphs: any[] = [];
+    for (let i = 0; i < 60; i++) {
+      const text = schema.text(
+        `Paragraph number ${i + 1}. ` +
+          'This line contains enough text to be clearly visible when rendered ' +
+          'in the paginated layout engine viewport.',
+      );
+      const run = schema.nodes.run.create(null, text);
+      paragraphs.push(schema.nodes.paragraph.create(null, run));
+    }
+
+    const doc = schema.nodes.doc.create(null, paragraphs);
+    const tr = state.tr.replaceWith(0, state.doc.content.size, doc.content);
+    editor.view.dispatch(tr);
+  });
+
+  await superdoc.waitForStable(2000);
+
+  // Verify multiple pages exist before proceeding.
+  const initialPageCount = await superdoc.page.locator('.superdoc-page[data-page-index]').count();
+  expect(initialPageCount).toBeGreaterThanOrEqual(3);
+
+  // Set zoom to 75%.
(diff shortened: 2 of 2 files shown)


packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -4849,7 +4866,9 @@ export class PresentationEditor extends EventEmitter {
 
       this.#viewportHost.style.width = `${scaledWidth}px`;
       this.#viewportHost.style.minWidth = `${scaledWidth}px`;
-      this.#viewportHost.style.minHeight = `${scaledHeight}px`;
+      this.#viewportHost.style.minHeight = '';
+      this.#viewportHost.style.height = `${scaledHeight}px`;
```
caio-pizzol: collab cursor names sit above the caret (top: -1.05em), so overflow: hidden here cuts them off when they're near the top of the page.

page margins usually give enough room, but small-margin docs will hit this.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -610,10 +610,12 @@ export class PresentationEditor extends EventEmitter {
...
 
-      // Register this instance in the static registry
-      if (options.documentId) {
-        PresentationEditor.#instances.set(options.documentId, this);
-      }
+      // Register this instance in the static registry.
+      // Generate a fallback ID when documentId is not provided (e.g., blank documents)
+      // so that setGlobalZoom() can always find and update all instances.
+      const registryKey = options.documentId || `__anonymous_${Date.now()}_${Math.random().toString(36).slice(2)}`;
+      this.#options.documentId = registryKey;
```
caio-pizzol: `this.#options` points to the same object the caller passed in, so this changes their object too. also, line 453 already read `options.documentId` to set the overlay ID - that ran before this line, so the overlay ID and registry key end up different.

separate `#registryKey` field fixes both.

caio-pizzol (changes requested):
good start!

the core fix (zoom × virtualization) isn't covered by tests - each side is tested separately but the interaction isn't.

a unit test in `virtualization.test.ts` that calls s`etZoom(0.75)` + triggers scroll and checks the virtual window would catch regressions here.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
tupizz: @caio-pizzol fixed — added a `#registryKey` private field so we don't mutate the caller's options object. The overlay ID set at line ~453 and the registry key now stay consistent.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
tupizz: @caio-pizzol good catch — replaced `overflow: hidden` with negative `margin-bottom` on painterHost. This compensates for the unscaled CSS box without clipping cursor labels or any other overlay that extends beyond the page bounds. Applied to both horizontal and vertical layout paths.

packages/layout-engine/painters/dom/src/virtualization.test.ts
```diff
@@ -369,6 +369,62 @@ describe('DomPainter virtualization (vertical)', () => {
...
+    const pages = mount.querySelectorAll('.superdoc-page');
+    const indices = Array.from(pages).map((p) => Number((p as HTMLElement).dataset.pageIndex));
+
+    // The anchor page for scroll=5000 should be 8 (topOfIndex(9)=5148 > 5000, topOfIndex(8)=4576 <= 5000).
+    // With window=3, we expect pages around index 7-9 or 8-10.
+    expect(indices.some((i) => i >= 7 && i <= 10)).toBe(true);
+
+    // Verify the window doesn't include page 0 (would indicate uncorrected zoom — scrollY=3750 → page 6).
+    // With correct zoom handling, we shouldn't see early pages.
+    expect(indices.every((i) => i >= 5)).toBe(true);
```
caio-pizzol: these assertions pass even without the zoom fix — scrollY=3750 gives page ~6, which still satisfies `>= 5`. the pages are deterministic here, so `toEqual([7, 8, 9])` or similar would actually catch the regression.

caio-pizzol:
@tupizz a behavior test that loads a long doc at 75% zoom, scrolls to mid-document, and checks content is visible (not blank) would go a long way — this bug is easy to regress when virtualization or zoom code changes.

harbournick:

PR #2178, opened 2026-02-25
fix(paste): detect heading levels from Google Docs styled paragraphs

### LINKED ISSUE #2152: Detect heading levels from Google Docs styled paragraphs
### Summary

Detect and preserve heading levels (H1, H3-H6) when pasting from Google Docs, which converts them to styled `<p>` tags.

### Problem

When pasting content with H1, H3-H6 headings from Google Docs, they lose their heading level and paste as plain paragraphs. Only H2 is reliably preserved as a heading tag.

### Repro

1. Create a Google Doc with headings at all levels (H1 through H6).
2. Copy the content and paste into SuperDoc.
3. Only H2 headings are preserved. H1 and H3-H6 paste as regular paragraphs.

### Expected

All heading levels from pasted content are preserved in SuperDoc.

### Notes

Google Docs converts most heading levels to `<p>` tags with inline `font-size`/`font-weight` styling instead of semantic `<h1>`-`<h6>` tags. The `paragraph.parseDOM` has rules for `h1`-`h6` but they never fire for these styled `<p>` elements.

The fix should detect heading-like paragraphs by font-size/font-weight in the Google Docs paste handler before ProseMirror parsing, and convert them to proper `<h1>`-`<h6>` tags. Google Docs typically uses predictable font sizes for each heading level (e.g., 24pt=H1, 18pt=H2, 14pt=H3, etc.).

**File:** `packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.js`

[copy-paste-test.docx]()

---

**Want to work on this?** Check the [Contributing Guide](<https://github.com/superdoc-dev/superdoc/blob/main/CONTRIBUTING.md>) to get started. Comment on the GitHub issue to let us know you're picking it up.

---

**Want to work on this?** Check the [Contributing Guide](https://github.com/superdoc-dev/superdoc/blob/main/CONTRIBUTING.md) to get started. Comment on this issue to let us know you're picking it up.

*Ref: SD-1981*

ErickPetru:
# Problem

> When pasting content with H1, H3-H6 headings from Google Docs, they lose their heading level and paste as plain paragraphs. Only H2 is reliably preserved as a heading tag.

Closes #2152

SuperDoc with pasted Google Docs headings after the fix is applied

# Cause

Google Docs converts most heading levels to `<p>` tags with inline `font-size`/`font-weight` styling instead of semantic `<h1>`–`<h6>` tags. ProseMirror's `paragraph.parseDOM` has rules for `h1`–`h6` but they never fire for these styled `<p>` elements, so the heading level is silently dropped.

# Fix

- Added `convertStyledHeadings` to the Google Docs paste pipeline, running before ProseMirror parses the pasted DOM
- Detects `<p>` elements carrying bold + a recognised font-size (checked on both the `<p>` itself and its first child `<span>`, covering both Google Docs style placements) and replaces them with the corresponding `<h1>`–`<h5>` element
- Font-size thresholds accommodate the default Google Docs theme and alternate themes (e.g. 20 pt and 24 pt both resolve to `h1`)

### DIFF
--- a/packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.js
+++ b/packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.js
@@ -5,7 +5,8 @@ import { createSingleItemList } from '../html/html-helpers.js';
 import { getLvlTextForGoogleList, googleNumDefMap } from '../../helpers/pasteListHelpers.js';
 import { wrapTextsInRuns } from '../docx-paste/docx-paste.js';
 
-// Ordered largest → smallest; first match wins.
+// Match Google Docs default heading sizes (H1=20pt, H2=18pt, H3=14pt, H4=12pt, H5=11pt).
+// Descending order so oversized fonts (e.g. 24pt) still resolve to closest heading.
 const headingSizeMap = [
   { minPt: 20, tag: 'h1' },
   { minPt: 16, tag: 'h2' },
--- a/packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.test.js
+++ b/packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.test.js
@@ -187,6 +187,22 @@ describe('handleGoogleDocsHtml', () => {
       expect(dom.querySelector('h1')?.textContent?.trim()).toBe('Bold p, size on span');
     });
 
+    it('does not convert when one span is missing a font-size', () => {
+      const html = `
+        <p><span style="font-size:20pt;font-weight:700">A</span><span style="font-weight:700">B</span></p>
+      `;
+      const dom = parseHeadings(html);
+      expect(dom.querySelector('h1,h2,h3,h4,h5')).toBeNull();
+    });
+
+    it('does not convert when spans have inconsistent font sizes', () => {
+      const html = `
+        <p><span style="font-size:20pt;font-weight:700">A</span><span style="font-size:14pt;font-weight:700">B</span></p>
+      `;
+      const dom = parseHeadings(html);
+      expect(dom.querySelector('h1,h2,h3,h4,h5')).toBeNull();
+    });
+
     it('converts when font-size is on <p> but font-weight is only on the child <span>', () => {
       const html = `
         <p style="font-size:20pt"><span style="font-weight:700">Split style heading</span></p>


ErickPetru:
Thanks for the thorough review, @caio-pizzol!

Both points addressed:

1. **Multi-span headings**: replaced the `singleSpan` guard with an all-spans approach which uses `querySelectorAll('span')` then requires them all to share the same font-size and bold status. This detects headings with links/bookmarks while still rejecting body paragraphs with a single bold opening word (since inconsistent bold across spans is not a heading).

2. **`closest('li')`**: swapped the `parentElement` check for `p.closest('li')` as suggested.

packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.js
```diff
@@ -253,3 +266,85 @@ function buildListPath(level, map) {
...
+
+    // if not all spans declare a font-size, not a heading
+    const sizes = spans.map((span) => parsePtValue(span.style.fontSize));
+    if (sizes.some((size) => size === null)) return notHeading;
+
+    // if inconsistent sizes, mixed body text, not a heading
+    const [firstSpanSize] = sizes;
+    if (sizes.some((size) => size !== firstSpanSize)) return notHeading;
+
+    return { fontSize: firstSpanSize, isBold: allSpansBold };
```
caio-pizzol: `fromSpans` checks bold on the child spans, but if bold is set on the `<p>` itself (`<p style="font-weight:700">`) with font-size only on spans, `allSpansBold` will be `false` and the heading gets skipped. worth patching for consistency?

```suggestion
    return { fontSize: firstSpanSize, isBold: allSpansBold || boldWeightRegex.test(el.style.fontWeight || '') };
```

ErickPetru: Good catch, applied your suggestion and also extended it so `elIsBold` (the root element's bold) is checked consistently in both `fromElement` and `fromSpans` paths.

packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.js
```diff
@@ -5,6 +5,17 @@ import { createSingleItemList } from '../html/html-helpers.js';
 import { getLvlTextForGoogleList, googleNumDefMap } from '../../helpers/pasteListHelpers.js';
 import { wrapTextsInRuns } from '../docx-paste/docx-paste.js';
 
+// Ordered largest → smallest; first match wins.
```
caio-pizzol: a comment saying these numbers come from Google Docs default heading sizes (H1=20pt, H2=16pt, etc.) would help — not obvious where they came from otherwise.

packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.js
```diff
@@ -253,3 +266,88 @@ function buildListPath(level, map) {
...
+    // no span children, size is indeterminate
+    if (spans.length === 0) return notHeading;
+
+    // if not all spans declare a font-size, not a heading
+    const sizes = spans.map((span) => parsePtValue(span.style.fontSize));
+    if (sizes.some((size) => size === null)) return notHeading;
+
+    // if inconsistent sizes, mixed body text, not a heading
+    const [firstSpanSize] = sizes;
+    if (sizes.some((size) => size !== firstSpanSize)) return notHeading;
```
caio-pizzol, line 333: these two checks don't have tests yet. if they broke, paragraphs with mixed spans would quietly turn into headings. two quick cases to add:

```html
<!-- one span has no font-size — should stay as <p> -->
<p><span style="font-size:20pt;font-weight:700">A</span><span style="font-weight:700">B</span></p>

<!-- spans have different sizes — should stay as <p> -->
<p><span style="font-size:20pt;font-weight:700">A</span><span style="font-size:14pt;font-weight:700">B</span></p>
```

packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.js
ErickPetru: Good suggestion; what's obvious to one developer might not be to another.

packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.js
ErickPetru: Yes, more coverage makes perfect sense. Thank you for the patient review, I've sent the new tests.

packages/super-editor/src/core/inputRules/google-docs-paste/google-docs-paste.js
```diff
@@ -21,7 +33,9 @@ export const handleGoogleDocsHtml = (html, editor, view) => {
   const tempDiv = document.createElement('div');
   tempDiv.innerHTML = cleanedHtml;
 
-  const htmlWithMergedLists = mergeSeparateLists(tempDiv);
+  const tempDivWithHeadings = convertStyledHeadings(tempDiv);
+
+  const htmlWithMergedLists = mergeSeparateLists(tempDivWithHeadings);
```
caio-pizzol, line 38: Google Docs clipboard already uses `<h1>`–`<h6>` for headings set via the styles dropdown — ProseMirror parses these natively. This conversion only runs on `<p>` elements (normal text), so bold body text at 11pt gets wrongly turned into h5.

```suggestion
  // Google Docs already outputs semantic <h1>–<h6> for headings set via
  // the paragraph style dropdown — ProseMirror handles them natively.
  const htmlWithMergedLists = mergeSeparateLists(tempDiv);
```

ErickPetru: Hey @caio-pizzol, I want to make sure I'm not missing something here, because this seems to contradict the issue https://github.com/superdoc-dev/superdoc/issues/2152 itself.

The issue description states:

> Google Docs converts most heading levels to `<p>` tags with inline `font-size`/`font-weight` styling instead of semantic `<h1>`-`<h6>` tags. The `paragraph.parseDOM` has rules for `h1`-`h6` but they never fire for these styled `<p>` elements.

That's exactly the pattern this PR detects. So could you clarify? Did something change in how Google Docs serializes to clipboard, or was the issue description inaccurate?

If Google Docs already outputs semantic heading tags upfront, then the issue itself would be invalid and need to be closed, but the repro steps suggest otherwise.

caio-pizzol: Hey @ErickPetru, you're right to question it - the issue description is wrong, and that's on us for not verifying the assumption before writing it up.

I just tested both paste flows with debug logging on the clipboard HTML:
- Google Docs into SuperDoc: 6 semantic `<h1>–<h6>` tags, 2 `<p>` elements (body text)
- Word into SuperDoc: 6 semantic `<h1>–<h6>` tags, 5 `<p>` elements (body text)

Both sources already output proper heading tags in the clipboard. ProseMirror handles them natively — no conversion needed.

The `convertStyledHeadings` function only operates on `<p>` elements, which in practice are body text. That's why bold text at 11pt (Google Docs' default body size) gets incorrectly promoted to h5.

Sorry for the confusion on this - but appreciate your patience working through the reviews.

If you're gained, there are other `good first issue` [here](https://github.com/superdoc-dev/superdoc/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22) - feel free to pick any from the list (I will also double check if we might have any there are misinterpreted

PR #2188, opened 2026-02-26
fix(editor): render styles applied inside SDT fields (SD-2011)

tupizz:
## Demo

https://github.com/user-attachments/assets/1294a68d-9e16-406e-bd83-7cd7839ee356

## Summary

Fixes [SD-2011](https://linear.app/superdocworkspace/issue/SD-2011): Styles (bold, italic, font changes) applied to text inside SDT template fields were not visually rendered in the DomPainter, even though ProseMirror state was correct and DOCX export preserved the styles.

## Root Cause

Two issues in the style rendering pipeline:

### 1. `wrapTextInRunsPlugin` skipped SDT children

`wrapTextInRunsPlugin.js` (line 100) explicitly skipped text nodes whose parent was `structuredContent`:

```js
// Before:
if (!node.isText || !parent || parent.type === runType || parent.type?.name === 'structuredContent') return;
```

When `insertStructuredContentInline` creates an SDT field, it inserts bare text (no `run` wrapper). Without a `run` node, `calculateInlineRunPropertiesPlugin` cannot sync PM marks (bold, italic, etc.) into `run.attrs.runProperties`, breaking the entire style pipeline downstream.

### 2. `applyInlineRunProperties` overwrote mark-derived styles with `undefined`

In `common.ts`, the merge used object spread:

```ts
// Before:
const merged = { ...run, ...runAttrs };
```

When `runProperties` is empty (no `run` wrapper → no synced properties), `computeRunAttrs()` returns `{ bold: undefined, italic: undefined, ... }`. The spread overwrites mark-derived values (e.g., `bold: true` from `applyMarksToRun`) with `undefined`.

## Fix

1. **`wrapTextInRunsPlugin.js`** — Removed the `structuredContent` skip so all bare text nodes get wrapped in `run` nodes regardless of parent type. The `run` schema already allows being a child of `structuredContent` (`content: 'inline*'`).

2. **`common.ts`** (`applyInlineRunProperties`) — Changed from spread to an explicit loop that skips `undefined` values, preserving mark-derived properties when `runProperties` doesn't specify them:

```ts
// After:
const merged = { ...run };
for (const key of Object.keys(runAttrs)) {
  if (runAttrs[key] !== undefined) {
    merged[key] = runAttrs[key];
  }
}
```

## Pipeline flow (after fix)

```
User applies Bold to text inside SDT
  → PM adds bold mark to text node
  → wrapTextInRunsPlugin wraps text in run node (was skipped before)
  → calculateInlineRunPropertiesPlugin syncs bold mark → run.attrs.runProperties.bold = true
  → pm-adapter: applyMarksToRun() sets TextRun.bold = true
  → pm-adapter: applyInlineRunProperties() merges without overwriting (was clobbering before)
  → DomPainter renders bold text ✓
```

## Test plan

- [x] 2 new tests in `wrapTextInRunsPlugin.test.js` — verifies text inside SDT gets wrapped in `run` nodes
- [x] 3 new tests in `common.test.ts` — verifies defensive merge preserves mark-derived properties
- [x] All 6,509 super-editor tests pass
- [x] All 1,547 pm-adapter tests pass
- [x] Verified in browser: bold, italic, and font changes render correctly inside SDT fields

### DIFF
--- a/packages/super-editor/src/extensions/run/wrapTextInRunsPlugin.js
+++ b/packages/super-editor/src/extensions/run/wrapTextInRunsPlugin.js
@@ -107,8 +107,10 @@ const buildWrapTransaction = (state, ranges, runType, editor, markDefsFromMeta =
       let textNode = node;
 
       // For the first node in a paragraph, inherit run properties from previous paragraph
-      // and merge marks (this preserves existing marks like italic while adding inherited ones like bold)
-      if (index === 0) {
+      // and merge marks (this preserves existing marks like italic while adding inherited ones like bold).
+      // Only apply when the text is a direct child of the paragraph — not when it is
+      // first inside an inline wrapper like structuredContent (SDT).
+      if (index === 0 && parent.type.name === 'paragraph') {
         ({ runProperties, textNode } = copyRunPropertiesFromPreviousParagraph(state, pos, textNode, runType, editor));
       }
 
--- a/packages/super-editor/src/extensions/run/wrapTextInRunsPlugin.test.js
+++ b/packages/super-editor/src/extensions/run/wrapTextInRunsPlugin.test.js
@@ -566,10 +566,7 @@ describe('wrapTextInRunsPlugin', () => {
       const view = createView(schema, doc);
 
       // Insert SDT with bare text content (simulates template builder insertion)
-      const sdtNode = schema.nodes.structuredContent.create(
-        { id: '123', alias: 'Field' },
-        schema.text('John Doe'),
-      );
+      const sdtNode = schema.nodes.structuredContent.create({ id: '123', alias: 'Field' }, schema.text('John Doe'));
       const tr = view.state.tr.insert(1, sdtNode);
       view.dispatch(tr);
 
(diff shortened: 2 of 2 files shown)


packages/layout-engine/pm-adapter/src/converters/inline-converters/common.ts
```diff
@@ -80,11 +80,13 @@ export const applyInlineRunProperties = (
...
   const runAttrs = computeRunAttrs(runProperties, converterContext);
-  const merged = { ...run, ...runAttrs };
-  // Preserve existing run color when runProperties doesn't specify one.
-  // Object spread with undefined values overwrites the original, so we restore it.
-  if (runAttrs.color === undefined && run.color !== undefined) {
-    merged.color = run.color;
+  // Merge runAttrs onto run, but skip undefined values to avoid overwriting
+  // mark-derived properties (e.g., bold from a mark) with absent runProperties fields.
+  const merged = { ...run };
+  for (const key of Object.keys(runAttrs) as Array<keyof typeof runAttrs>) {
```
caio-pizzol, line 86: underline comes back as `null` from `computeRunAttrs` when it's absent, not `undefined` - so it gets past this check and still overwrites the run's `underline. != null` instead of `!== undefined` would catch both.

not urgent, fine as a follow-up.

harbournick (approved):
LGTM

harbournick:

PR #2189, opened 2026-02-26
fix: text highlight on export

artem-harbour:


packages/super-editor/src/core/super-converter/v3/handlers/utils.js
```diff
@@ -317,7 +317,9 @@ export function decodeProperties(params, translatorsBySdName, properties) {
     if (translator) {
       const result = translator.decode({ ...params, node: { attrs: { [key]: properties[key] } } });
       if (result != null) {
-        result.name = translator.xmlName;
+        if (result.name == null) {
```
artem-harbour, line 320: If `result.name` contains a name, then we should not replace it with the default `xmlName`.

For custom hex colors `w:shd` instead `w:highlight`.

harbournick (approved):
LGTM

harbournick:

PR #2191, opened 2026-02-26
fix(editor): arrow key navigation across page boundaries and auto-scroll (SD-1950)

tupizz:
## Demo

https://github.com/user-attachments/assets/7a3b22ba-ca82-4e0e-b22a-337e57df4888

## Summary

Fixes arrow key `(↑/↓)` AND `SHIFT + (↑/↓)`  navigation in presentation mode that would **jump entire sections** when moving down across page boundaries, and **get stuck in a loop** when moving back up. Also adds auto-scroll to keep the caret visible during keyboard navigation.

## The Problem

Two distinct bugs in the vertical-navigation extension caused broken cursor movement in multi-page documents:

### Bug 1: ArrowDown jumps sections instead of line-by-line

When pressing ArrowDown near a page boundary, the cursor would suddenly jump thousands of characters forward — skipping entire pages. For example, in a 7-page Lorem Ipsum document, pressing ArrowDown at position 3126 would jump to position 6178 (a +3052 character jump), then 9350, then 12378 — each ArrowDown skipping thousands of characters.

### Bug 2: ArrowUp gets stuck / loops on the same page

When navigating back up (ArrowUp), the cursor would get stuck at certain positions and refuse to move further. Pressing ArrowUp repeatedly at these positions had zero effect.

## Root Cause Analysis

Both bugs originate in the same code path: the vertical-navigation extension's `handleKeyDown` → `getAdjacentLineClientTarget` → `getHitFromLayoutCoords` pipeline.

The extension correctly finds the adjacent line DOM element using `findAdjacentLineElement()` (which traverses fragments and pages). However, it then uses `getBoundingClientRect()` on that element to get client-space coordinates, and passes those to `hitTest()` to map to a ProseMirror position.

**Bug 1 root cause:** When the adjacent line is on the next page below the viewport (user hasn't scrolled there yet), `getBoundingClientRect()` returns off-screen coordinates (e.g., `clientY=1137` when the viewport bottom is at `882`). `hitTest()` with these off-screen coordinates produces wildly incorrect PM positions, causing the cursor to jump thousands of characters.

**Bug 2 root cause:** Even when the adjacent line is on-screen (same page, different fragment), the center of its DOM bounding rect can fall in a zone where `hitTest()` maps to the wrong fragment. This was traced by testing `hitTest()` at every Y pixel across the adjacent line:

```
y=215-229: hitTest returns pos 15428 (WRONG — current line's fragment)
y=231-239: hitTest returns pos 15363 (CORRECT — adjacent line)
y=241+:    hitTest returns pos 15428 (WRONG — current line's fragment)
```

The adjacent line's DOM center is at `y ≈ 229.5`, which falls right at the boundary where `hitTest()` returns the wrong fragment. So the "new" position equals the current one, and the cursor doesn't move.

### Approach: Validate hit test results against the adjacent line's known PM range

Instead of blindly trusting `hitTest()` results (which are unreliable when coordinates are off-screen or at fragment boundaries), we now:

1. **Read `data-pm-start`/`data-pm-end`** from the adjacent line's DOM element — these attributes contain the exact PM position range for that line.

2. **Validate the hit test result** against this range with a tight tolerance of 5 positions. If the hit falls outside `[pmStart - 5, pmEnd + 5]`, it's considered unreliable.

3. **Fall back to `resolvePositionAtGoalX()`** — a binary search using `computeCaretLayoutRect` to find the position within the line's PM range whose layout X is closest to the goal X coordinate. This uses the layout engine's position-to-coordinate mapping (which is always accurate regardless of scroll position), requiring only ~7 `computeCaretLayoutRect` calls for a typical 80-character line (O(log N)).

### Why this approach

We considered several alternatives:

- **Scrolling the adjacent line into view first**: Would cause visual jank (scroll → measure → set selection → scroll back) and add latency to every arrow key press.
- **Using only the layout engine (no hit test)**: Would require reimplementing goal-X tracking and character-level position resolution from scratch — the current `hitTest()` works correctly 95%+ of the time.
- **Increasing the hit test tolerance**: Too loose a tolerance (e.g., `lineRange`) fails to catch the "stuck cursor" case where the hit lands just a few positions past the adjacent line's end.

The chosen approach is minimally invasive: the existing hit test pipeline handles the common case correctly; the validation + fallback only activates when the hit test produces an implausible result. The binary search fallback is efficient and uses the same `computeCaretLayoutRect` API that the rest of the caret positioning system relies on.

### Auto-scroll

Added `#scrollCaretIntoViewIfNeeded()` to `PresentationEditor.ts`, called after the caret overlay is rendered. It checks if the caret is outside the scroll container bounds and scrolls minimally (with 20px margin) to keep it visible. Falls back to `#scrollPageIntoView()` for virtualized pages.

## Files Changed

| File | Change |
|------|--------|
| `vertical-navigation.js` | Added PM range validation, `resolvePositionAtGoalX()` binary search fallback, `pmStart`/`pmEnd` extraction from adjacent line |
| `PresentationEditor.ts` | Added `#scrollCaretIntoViewIfNeeded()` for auto-scroll on selection change |
| `scrollCaretIntoView.test.ts` | 9 unit tests for the scroll-into-view logic |

## Test plan

- [x] All 688 test files pass (6646 tests)
- [x] ArrowDown navigates smoothly through all 7 pages of a long Lorem Ipsum document (pos 2 → 42980)
- [x] ArrowUp navigates smoothly back through all pages (pos 42980 → 2) with no looping
- [x] Previously stuck position (15428) now correctly moves past on ArrowUp
- [x] Round-trip test: 100 ArrowDown + 105 ArrowUp returns to position 2
- [ ] Manual testing with various document lengths and structures
- [ ] Verify no regressions in table cell navigation

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,7 +17,7 @@ catalogs:
       version: 9.39.2
     '@floating-ui/dom':
       specifier: ^1.7.0
-      version: 1.7.5
+      version: 1.7.6
     '@hocuspocus/provider':
       specifier: ^2.13.6
       version: 2.15.3
@@ -56,7 +56,7 @@ catalogs:
       version: 22.19.2
     '@types/react':
       specifier: ^19.2.6
-      version: 19.2.11
+      version: 19.2.14
     '@types/react-dom':
       specifier: ^19.2.3
       version: 19.2.3
@@ -194,7 +194,7 @@ catalogs:
       version: 1.11.0
     prosemirror-view:
       specifier: ^1.33.8
-      version: 1.41.5
+      version: 1.41.6
     react:
       specifier: 19.2.4
       version: 19.2.4
@@ -218,7 +218,7 @@ catalogs:
       version: 11.0.0
     rollup:
       specifier: ^4.31.0
-      version: 4.57.1
+      version: 4.59.0
     rollup-plugin-copy:
       specifier: ^3.5.0
       version: 3.5.0
@@ -587,7 +587,7 @@ importers:
         version: 1.4.4
       prosemirror-view:
         specifier: 'catalog:'
-        version: 1.41.5
+        version: 1.41.6
       superdoc:
         specifier: 'workspace:'
         version: link:../superdoc
@@ -599,7 +599,7 @@ importers:
         version: 5.9.3
       vitest:
         specifier: 'catalog:'
-        version: 3.2.4(@types/debug@4.1.12)(@types/node@22.19.2)(esbuild@0.27.2)(happy-dom@20.4.0)(jiti@2.6.1)(jsdom@27.3.0(canvas@3.2.1))(tsx@4.21.0)(yaml@2.8.2)
+        version: 3.2.4(@types/debug@4.1.12)(@types/node@22.19.2)(esbuild@0.27.3)(happy-dom@20.4.0)(jiti@2.6.1)(jsdom@27.3.0(canvas@3.2.1))(tsx@4.21.0)(yaml@2.8.2)
       vue:
         specifier: 3.5.25
         version: 3.5.25(typescript@5.9.3)
@@ -659,16 +659,16 @@ importers:
         version: 6.9.1
       '@testing-library/react':
         specifier: 'catalog:'
-        version: 16.3.2(@testing-library/dom@10.4.1)(@types/react-dom@19.2.3(@types/react@19.2.11))(@types/react@19.2.11)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
+        version: 16.3.2(@testing-library/dom@10.4.1)(@types/react-dom@19.2.3(@types/react@19.2.14))(@types/react@19.2.14)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
       '@testing-library/user-event':
         specifier: 'catalog:'
         version: 14.6.1(@testing-library/dom@10.4.1)
       '@types/react':
         specifier: 'catalog:'
-        version: 19.2.11
+        version: 19.2.14
       '@types/react-dom':
         specifier: 'catalog:'
-        version: 19.2.3(@types/react@19.2.11)
+        version: 19.2.3(@types/react@19.2.14)
       '@vitejs/plugin-react':
         specifier: 'catalog:'
         version: 5.1.3(rolldown-vite@7.3.1(@types/node@25.3.5)(esbuild@0.27.3)(jiti@2.6.1)(tsx@4.21.0)(yaml@2.8.2))
@@ -726,10 +726,10 @@ importers:
     devDependencies:
       '@types/react':
         specifier: 'catalog:'
-        version: 19.2.11
+        version: 19.2.14
       '@types/react-dom':
         specifier: 'catalog:'
-        version: 19.2.3(@types/react@19.2.11)
+        version: 19.2.3(@types/react@19.2.14)
       '@vitejs/plugin-react':
         specifier: 'catalog:'
         version: 5.1.3(rolldown-vite@7.3.1(@types/node@25.3.5)(esbuild@0.27.3)(jiti@2.6.1)(tsx@4.21.0)(yaml@2.8.2))
@@ -957,16 +957,16 @@ importers:
     devDependencies:
       '@testing-library/react':
         specifier: 'catalog:'
-        version: 16.3.2(@testing-library/dom@10.4.1)(@types/react-dom@19.2.3(@types/react@19.2.11))(@types/react@19.2.11)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
+        version: 16.3.2(@testing-library/dom@10.4.1)(@types/react-dom@19.2.3(@types/react@19.2.14))(@types/react@19.2.14)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
       '@types/node':
         specifier: 'catalog:'
         version: 22.19.2
       '@types/react':
         specifier: 'catalog:'
-        version: 19.2.11
+        version: 19.2.14
       '@types/react-dom':
         specifier: 'catalog:'
-        version: 19.2.3(@types/react@19.2.11)
+        version: 19.2.3(@types/react@19.2.14)
       '@typescript-eslint/eslint-plugin':
         specifier: 'catalog:'
         version: 8.54.0(@typescript-eslint/parser@8.54.0(eslint@9.39.2(jiti@2.6.1))(typescript@5.9.3))(eslint@9.39.2(jiti@2.6.1))(typescript@5.9.3)
@@ -1015,7 +1015,7 @@ importers:
         version: 22.19.2
       rollup:
         specifier: 'catalog:'
-        version: 4.57.1
+        version: 4.59.0
       typescript:
         specifier: 'catalog:'
         version: 5.9.3
@@ -1104,7 +1104,7 @@ importers:
         version: 1.11.0
       prosemirror-view:
         specifier: 'catalog:'
-        version: 1.41.5
+        version: 1.41.6
       rehype-parse:
         specifier: 'catalog:'
         version: 9.0.1
@@ -1137,14 +1137,14 @@ importers:
         version: 1.6.11
       y-prosemirror:
         specifier: 'catalog:'
-        version: 1.3.7(prosemirror-model@1.25.4)(prosemirror-state@1.4.4)(prosemirror-view@1.41.5)(y-protocols@1.0.7(yjs@13.6.19))(yjs@13.6.19)
+        version: 1.3.7(prosemirror-model@1.25.4)(prosemirror-state@1.4.4)(prosemirror-view@1.41.6)(y-protocols@1.0.7(yjs@13.6.19))(yjs@13.6.19)
       yjs:
         specifier: 'catalog:'
         version: 13.6.19
     devDependencies:
       '@floating-ui/dom':
         specifier: 'catalog:'
-        version: 1.7.5
+        version: 1.7.6
       '@superdoc/common':
         specifier: workspace:*
         version: link:../../shared/common
@@ -1346,16 +1346,16 @@ importers:
         version: 6.9.1
       '@testing-library/react':
         specifier: 'catalog:'
-        version: 16.3.2(@testing-library/dom@10.4.1)(@types/react-dom@19.2.3(@types/react@19.2.11))(@types/react@19.2.11)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
+        version: 16.3.2(@testing-library/dom@10.4.1)(@types/react-dom@19.2.3(@types/react@19.2.14))(@types/react@19.2.14)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
       '@testing-library/user-event':
         specifier: 'catalog:'
         version: 14.6.1(@testing-library/dom@10.4.1)
       '@types/react':
         specifier: 'catalog:'
-        version: 19.2.11
+        version: 19.2.14
       '@types/react-dom':
         specifier: 'catalog:'
-        version: 19.2.3(@types/react@19.2.11)
+        version: 19.2.3(@types/react@19.2.14)
       '@vitejs/plugin-react':
         specifier: 'catalog:'
         version: 5.1.3(rolldown-vite@7.3.1(@types/node@25.3.5)(esbuild@0.27.3)(jiti@2.6.1)(tsx@4.21.0)(yaml@2.8.2))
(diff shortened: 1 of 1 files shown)


packages/super-editor/src/extensions/vertical-navigation/vertical-navigation.js
```diff
@@ -334,6 +375,56 @@ function findAdjacentLineElement(currentLine, direction) {
...
+  let hi = pmEnd;
+
+  while (lo <= hi) {
+    const mid = Math.floor((lo + hi) / 2);
+    const rect = presentationEditor.computeCaretLayoutRect(mid);
+    if (!rect || !Number.isFinite(rect.x)) {
+      // Can't measure this position — fall back to pmStart
+      break;
+    }
+
```
caio-pizzol: when `computeCaretLayoutRect(mid)` returns null (e.g. inline node boundary), the whole search stops and returns `pmStart` — caret jumps to line start. skipping the position keeps the search going:

```suggestion
    if (!rect || !Number.isFinite(rect.x)) {
      // Can't measure this position — skip it
      lo = mid + 1;
      continue;
    }
```

packages/super-editor/src/extensions/vertical-navigation/vertical-navigation.js
```diff
@@ -334,6 +375,56 @@ function findAdjacentLineElement(currentLine, direction) {
...
+ * @param {number} goalX - Target X coordinate in layout space.
+ * @returns {{ pos: number } | null}
+ */
+function resolvePositionAtGoalX(editor, pmStart, pmEnd, goalX) {
+  const presentationEditor = editor.presentationEditor;
+  let bestPos = pmStart;
+  let bestDist = Infinity;
+
+  // Binary search: characters within a single line have monotonically increasing X
+  let lo = pmStart;
```
caio-pizzol, line 400: this assumes LTR (X increases with PM position). for RTL text X goes the other way, so the search moves in the wrong direction. `bestPos`/`bestDist` tracking limits the damage but it'll be less accurate. worth a `// NOTE: assumes LTR` comment so it's clearly a known limitation?

packages/super-editor/src/extensions/vertical-navigation/vertical-navigation.js
```diff
@@ -334,6 +375,56 @@ function findAdjacentLineElement(currentLine, direction) {
...
+ * visible viewport (e.g., after crossing a page boundary).
+ *
+ * @param {Object} editor
+ * @param {number} pmStart - Start PM position of the target line.
+ * @param {number} pmEnd - End PM position of the target line.
+ * @param {number} goalX - Target X coordinate in layout space.
+ * @returns {{ pos: number } | null}
+ */
+function resolvePositionAtGoalX(editor, pmStart, pmEnd, goalX) {
+  const presentationEditor = editor.presentationEditor;
```
caio-pizzol, line 393: no tests for `resolvePositionAtGoalX` yet — `vertical-navigation.test.js` already mocks `computeCaretLayoutRect`, so adding a few cases (null midpoint, goalX at line edges, exact match) should be quick. worth doing?

packages/super-editor/src/core/presentation-editor/tests/scrollCaretIntoView.test.ts
```diff
@@ -0,0 +1,214 @@
...
+ * we extract its logic into a testable helper and verify it against a real DOM
+ * setup with mock scroll containers and caret elements.
+ */
+
+/**
+ * Extracted logic from PresentationEditor.#scrollCaretIntoViewIfNeeded.
+ * This mirrors the implementation to allow direct unit testing without
+ * bootstrapping the full PresentationEditor.
+ */
+function scrollCaretIntoViewIfNeeded(
```
caio-pizzol, line 16: this copies the production logic inline — if the real method changes, these tests still pass on the old version. extracting the scroll function into a shared module (like `renderCaretOverlay` in `LocalSelectionOverlayRendering.ts`) would keep them in sync. worth considering?

tupizz:
Addressed Caio's feedback in e88919cd9:

**1. Binary search null handling** (applied): Changed `break` to `lo = mid + 1; continue` so the search skips unmeasurable positions (e.g. inline node boundaries) instead of bailing to `pmStart`. This prevents the caret from jumping to line start when `computeCaretLayoutRect` returns null mid-search.

**2. LTR assumption comment** (applied): Added a `// NOTE: assumes LTR` comment documenting the known limitation.

**3. Tests for `resolvePositionAtGoalX`**: Skipped for now — the function is exercised through the integration path (round-trip ArrowDown/ArrowUp tests), and adding dedicated unit tests can be a follow-up.

**4. Test copies production logic**: Skipped — extracting `scrollCaretIntoViewIfNeeded` into a shared module adds architectural overhead for a simple 35-line method. The current pattern (testing a mirror of the logic) is acceptable for now.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -3954,6 +3955,67 @@ export class PresentationEditor extends EventEmitter {
...
+   * caret element's screen-space position (via getBoundingClientRect) to determine if
+   * scrolling is needed, keeping a small margin for comfortable viewing.
+   *
+   * If the caret element doesn't exist (page may be virtualized / not mounted),
+   * falls back to scrolling the target page into view to trigger virtualization.
+   */
+  #scrollCaretIntoViewIfNeeded(caretLayout: { pageIndex: number }): void {
+    const caretEl = this.#localSelectionLayer?.querySelector(
+      '.presentation-editor__selection-caret',
+    ) as HTMLElement | null;
```
caio-pizzol: the caret element can exist even when the page isn't actually visible (the rendering has a fallback that creates it with estimated coordinates). when that happens, this skips `#scrollPageIntoView` and scrolls to the wrong spot. worth checking if the page is actually mounted before using the caret rect?

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -3923,6 +3923,7 @@ export class PresentationEditor extends EventEmitter {
           console.warn('[PresentationEditor] Failed to render caret overlay:', error);
         }
       }
+      this.#scrollCaretIntoViewIfNeeded(caretLayout);
```
caio-pizzol: this scrolls on every selection change, not just arrow keys — collab edits, undo, find-and-replace all trigger it. the margin check avoids scrolling when the caret is already visible, but are there cases where this causes unexpected jumps?

packages/super-editor/src/extensions/vertical-navigation/vertical-navigation.js
```diff
@@ -118,8 +118,37 @@ export const VerticalNavigation = Extension.create({
...
+          // Check if the hit test result is plausible: if the adjacent line has PM
+          // position data, the hit should land within or very close to that range.
+          // A miss indicates off-screen coordinate mapping failure or fragment
+          // boundary misalignment (adjacent line center Y mapping to wrong fragment).
+          //
+          // Tolerance is kept small (5 positions) to catch cases where the hit
+          // lands on the current line's fragment start instead of the adjacent
+          // line — this causes the cursor to appear stuck since the "new" position
+          // equals the current one.
+          if (adjacent.pmStart != null && adjacent.pmEnd != null) {
```
caio-pizzol, line 137: the fallback path (`resolvePositionAtGoalX`) has zero test coverage — no test sets `data-pm-start`/`data-pm-end`, so this block never runs. two cases would cover it: hit lands in range (fallback skipped) and hit lands out of range (fallback runs). the null-skip fix should get a case too.

packages/super-editor/src/core/presentation-editor/tests/scrollCaretIntoView.test.ts
caio-pizzol, line 16: re-flagging from last round — the test copies the full method. if the real code changes, these tests still pass on the old version. worth extracting into a shared function both can use?

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -3954,6 +3955,67 @@ export class PresentationEditor extends EventEmitter {
...
+  #scrollCaretIntoViewIfNeeded(caretLayout: { pageIndex: number }): void {
+    const caretEl = this.#localSelectionLayer?.querySelector(
+      '.presentation-editor__selection-caret',
+    ) as HTMLElement | null;
+
+    if (!caretEl) {
+      // Caret page may not be mounted (virtualized) — scroll page into view
+      // to trigger mount; next selection update will handle precise scroll.
+      this.#scrollPageIntoView(caretLayout.pageIndex);
+      return;
```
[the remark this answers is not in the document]
tupizz: Addressed in 82a37d180. Now checks if the page element is actually mounted in the DOM (`querySelector('[data-page-index="N"]')`) before using the caret rect. Falls back to `#scrollPageIntoView` when the page isn't mounted.

packages/super-editor/src/extensions/vertical-navigation/vertical-navigation.js
tupizz: Already addressed in e88919cd9 — changed `break` to `lo = mid + 1; continue` as suggested. Also added unit tests in 82a37d180 covering the null midpoint case.

packages/super-editor/src/extensions/vertical-navigation/vertical-navigation.js
tupizz: Already addressed in e88919cd9 — added the `// NOTE: assumes LTR` comment.

packages/super-editor/src/extensions/vertical-navigation/vertical-navigation.js
tupizz: Added 7 unit tests in 82a37d180 covering: closest match, exact match, goalX before/after all positions, null midpoints, all-null positions, and single-position range. Exported `resolvePositionAtGoalX` for direct testing.

packages/super-editor/src/core/presentation-editor/tests/scrollCaretIntoView.test.ts
tupizz: Acknowledged. Keeping as-is for now — extracting a 35-line private method into a shared module adds architectural overhead. The test mirrors the logic, which is acceptable for this scope.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
tupizz: No, the unconditional scroll is correct... 

The 20px margin check already prevents scrolling when the caret is visible. When the caret IS off-screen, scrolling to it is the right behavior for every selection source: undo should show the cursor, find-and-replace should show the match, collab edits should show where you landed.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
tupizz: Addressed in 3c5aca1b6 
added a page-mount guard that queries `this.#painterHost` (a stable class field) for the page element. When the page isn't mounted, falls back to `#scrollPageIntoView` instead of using the caret's potentially estimated coordinates.

packages/super-editor/src/extensions/vertical-navigation/vertical-navigation.js
tupizz: Addressed in 84f5f3d85 — added two integration tests that set `data-pm-start`/`data-pm-end` on the adjacent line element:

1. **Hit in range**: `hitTest` returns pos 5 within [3, 8] → used directly, no fallback fired (`computeCaretLayoutRect` called only once for initial goalX)
2. **Hit out of range**: `hitTest` returns pos 100, way outside [3, 8] → `resolvePositionAtGoalX` binary search kicks in, resolves to a position within the line's range

Combined with the 7 unit tests for `resolvePositionAtGoalX` from the earlier commit, the fallback path now has full coverage including the null-skip fix.

packages/super-editor/src/core/presentation-editor/tests/scrollCaretIntoView.test.ts
tupizz: Keeping as-is. The method is private and depends on class internals (`#localSelectionLayer`, `#scrollContainer`, `#scrollPageIntoView`). The test already parameterizes these as arguments --- that IS the extracted function. 

also 
making production code also call it means a new file, a new export, and indirection for a 35-line method that we just proved is better left simpl.

tupizz: if you prefer I can remove this test

harbournick:

PR #2201, opened 2026-02-27
fix: allow backspace to move to next run

palmer-cl:
This is related to: https://github.com/superdoc-dev/superdoc/pull/2175 

- If we try to limit the cursor from moving to another run, we cannot move from runs that contain a single character. 
- In the doc listed https://linear.app/superdocworkspace/issue/SD-1942/bug-tracked-changes-deleting-character-deletes-whole-paragraph-chrome, periods are their own run at some points in the doc, meaning that if we backspace next to a period, without this fix, we can never get to the next character. 
- Chrome handles this correctly without this fix, but Firefox does not.


packages/super-editor/src/core/commands/backspaceNextToRun.js
```diff
@@ -1,7 +1,7 @@
 import { Selection } from 'prosemirror-state';
 
-const findPreviousTextDeleteRange = (doc, cursorPos, minPos) => {
-  for (let pos = cursorPos - 1; pos >= minPos; pos -= 1) {
+const findPreviousTextDeleteRange = (doc, cursorPos) => {
+  for (let pos = cursorPos - 1; pos > 0; pos -= 1) {
```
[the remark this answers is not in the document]
palmer-cl: Not relevant. Sometimes we do need to move across runs.

caio-pizzol: without a lower bound, this scan can walk past paragraph boundaries and delete a character from the wrong paragraph. that only gets caught by `normalizeReplaceStepSingleCharDelete` in track-changes mode — in normal editing it goes through silently.

instead of removing `minPos` entirely, scope it to the current paragraph:

```js
const paragraphStart = $pos.start($pos.depth);
const deleteRange = findPreviousTextDeleteRange(state.doc, $pos.pos, paragraphStart);
```

that still lets you cross sibling runs (fixing the single-char run issue) without leaving the paragraph.

packages/super-editor/src/core/commands/backspaceNextToRun.test.js
```diff
@@ -109,27 +109,4 @@ describe('backspaceNextToRun', () => {
     });
```
caio-pizzol, line 109: this test covers the cross-paragraph case — with a paragraph-scoped bound it should still pass. worth keeping.

packages/super-editor/src/core/commands/backspaceNextToRun.js
palmer-cl: I did try this as a solution:

image


here were the results:

Screen Cast 2026-02-27 at 9 49 35 AM

palmer-cl: @caio-pizzol what are your thoughts?

caio-pizzol: the reason `$pos.start($pos.depth)` didn't work is that `$pos.depth` points to different levels depending on cursor position:

- **between runs** (case 1: `$pos.nodeBefore` is a run) → `$pos.depth` is the paragraph depth → `$pos.start($pos.depth)` = paragraph start ✓
- **at the start of a run** (case 2: `$pos.pos === $pos.start()`) → `$pos.depth` is the **run's** depth → `$pos.start($pos.depth)` = run content start ✗

so in case 2 the bound collapses back to the run, which is why backspace still couldn't cross into the previous run in your video.

we already solve this exact problem in `backspaceAcrossRuns.js:32` — just need to compute `paraDepth` the same way:

```js
const paragraphDepth = $pos.parent.type === runType ? $pos.depth - 1 : $pos.depth;
const paragraphStart = $pos.start(paragraphDepth);
const deleteRange = findPreviousTextDeleteRange(state.doc, $pos.pos, paragraphStart);
```

that gives paragraph-scoped bounds for both cursor positions — crosses sibling runs but not paragraph boundaries.

check: https://github.com/superdoc-dev/superdoc/pull/2225

palmer-cl:
Closing this PR this is no longer an issue in Firefox post merging main, it looks like `findPreviousTextDeleteRange` was updated.

PR #2202, opened 2026-02-27
fix: add currentTotalPages getter and pagination-update event

### LINKED ISSUE #958: some files load failed when pagination:true
### Type

- [x] Bug

### Description

some files ok, some not ok. 

Image

[superdoc-debug.docx](https://github.com/user-attachments/files/22342766/superdoc-debug.docx)

### Current Behavior

some files ok, some not ok. 

### Expected Behavior

all files ok

### Steps to Reproduce

use upload

### SuperDoc Version

0.18.0

### Browser & Version

0.140

### OS

macos14.5

### Additional Context

_No response_

### Checklist

- [x] Tested with latest version
- [x] Checked for duplicates
- [ ] Added relevant labels

asumaran:
## Summary

Fixes #958.

`activeEditor.currentTotalPages` returns `undefined` when accessed in the `onReady` callback. This is because the property was part of the old decoration-based pagination system and was never carried over when the layout engine (`PresentationEditor`) replaced it.

There is also a timing issue: `onReady` fires via `setTimeout(0)` from `Editor.mount()`, but the first layout runs on the next `requestAnimationFrame`. Since `setTimeout(0)` always executes before RAF in the browser event loop, page data is never available inside `onReady`.

### Changes

**A. `currentTotalPages` getter on `Editor`** (`Editor.ts`)

Adds a getter that delegates to `PresentationEditor.getPages()`:

```typescript
get currentTotalPages(): number | undefined {
  if (this.presentationEditor) {
    const pages = this.presentationEditor.getPages();
    return pages.length > 0 ? pages.length : undefined;
  }
  return undefined;
}
```

This makes `activeEditor.currentTotalPages` work for synchronous reads after the first layout completes (e.g., in a button handler or a later callback).

**B. `pagination-update` event on SuperDoc** (`SuperDoc.vue`, `SuperDoc.js`, `types/index.js`)

Bridges `PresentationEditor`'s internal `paginationUpdate` event to a SuperDoc-level `pagination-update` event, with a corresponding `onPaginationUpdate` config callback. This gives consumers a way to know when pages are actually ready:

```javascript
new SuperDoc({
  onPaginationUpdate: ({ totalPages, superdoc }) => {
    console.log('Pages:', totalPages);
  },
});
```

This solves the timing problem — instead of reading `currentTotalPages` inside `onReady` (where it will always be `undefined` because layout hasn't run yet), users can listen for `pagination-update`.

### Why both?

- **A** restores the property consumers expect, so code that reads `currentTotalPages` after layout completes works.
- **B** provides the event to know *when* layout completes, since `onReady` fires too early for pagination data.

## Test plan

- [x] `currentTotalPages` returns `undefined` when no `PresentationEditor` is set
- [x] `currentTotalPages` returns `undefined` when pages array is empty (before first layout)
- [x] `currentTotalPages` returns the page count when pages exist
- [x] `onPaginationUpdate` callback fires when `pagination-update` is emitted
- [x] Default no-op `onPaginationUpdate` does not throw
- [x] All existing tests pass (`pnpm test` in both `super-editor` and `superdoc`)

### DIFF
--- a/apps/docs/core/supereditor/methods.mdx
+++ b/apps/docs/core/supereditor/methods.mdx
@@ -860,4 +860,5 @@ editor.commands.goToSearchResult(results[0]);
 | `isDestroyed` | `boolean` | Whether editor has been destroyed |
 | `isFocused` | `boolean` | Whether editor has focus |
 | `docChanged` | `boolean` | Whether any edits have been made |
+| `currentTotalPages` | `number \| undefined` | Page count after the first layout completes. `undefined` until then. Use the `pagination-update` event to know when it's available. |
 | `sourcePath` | `string \| null` | Source file path (null if opened from Blob) |


asumaran:
@caio-pizzol thanks for the review! You're right, `paginationUpdate` only fires after a successful layout pass, so `layout.pages` is always present. Simplified to `layout.pages.length `in 8a9b48e7b.

Also added docs for both the new event and the `currentTotalPages` property (8483b47f4, 16cc4c6f1).

Let me know if it looks good!

harbournick:

PR #2204, opened 2026-02-27
fix: always call resolveComment after custom TC bubble handlers (SD-2049)

caio-pizzol:
Custom `onTrackedChangeBubbleAccept/Reject` handlers replaced the default behavior entirely, skipping the internal `resolveComment()` call. This left the last tracked change bubble stuck on screen because `resolvedTime` was never set and the editor doesn't emit a position update for the last mark removal.


harbournick:

PR #2205, opened 2026-02-27
fix(editor): selection highlight flickers when dragging across mark boundaries (SD-2024)

tupizz:
## Demo

https://github.com/user-attachments/assets/d012372f-dcda-4ffd-959c-f7b619d24b9b

## Summary

Fixes the selection highlight flickering/disappearing when drag-selecting across inline mark boundaries (e.g. bold → italic, colored → uncolored text).

## Problem

When a user drags to select text that spans two adjacent runs with different marks (e.g., **bold** followed by *italic*), the blue selection overlay flickers — it briefly disappears for one frame, then reappears on the next pointer event. This makes drag selection feel broken on any document with mixed inline formatting.

## How we debugged it

1. **Reproduced in browser** — loaded a complex DOCX (`SuperDoc (5).docx`) with mixed marks (bold, italic, colored text, tracked changes). Drag-selected across formatting boundaries and observed the highlight flash.

2. **Instrumented selection monitoring** — injected a 3ms-interval monitor on `window.editor.state.selection` to track every selection change during drag. The PM selection was stable — the flicker was purely visual, not a state issue.

3. **Traced the rendering path** — followed the selection overlay pipeline:
   - `PresentationEditor.#updateSelectionFromEditor()` reads the PM selection `{from, to}`
   - calls `computeSelectionRectsFromDom()` → `DomPositionIndex.findEntriesInRange(from, to)` to find DOM elements in the selection range
   - builds `DOMRange` objects from those elements → `range.getClientRects()` → draws overlay divs

4. **Found the gap** — when the selection boundary falls exactly on a run boundary, the PM positions land in the 2-position structural gap between adjacent text spans. Example:
   ```
   <span data-pm-start="1" data-pm-end="5">bold</span>   ← run close token at 5, open token at 6
   <span data-pm-start="7" data-pm-end="12">italic</span> ← positions 5-7 are structural tokens
   ```
   A selection at `{from: 5, to: 7}` spans only structural tokens — no text content.

5. **Identified the root cause** — `findEntriesInRange` uses half-open `[start, end)` semantics. For the range `[5, 7)`:
   - Entry `[1,5]`: `pmEnd (5) <= start (5)` → **excluded** (boundary touch)
   - Entry `[7,12]`: `pmStart (7) >= end (7)` → **excluded** (boundary touch)
   - Result: **zero entries** → zero DOM rects → overlay cleared → flicker

## Solution

Three targeted changes:

### 1. `DomPositionIndex.findEntriesInRange` — add `boundaryInclusive` option

Added an optional `{ boundaryInclusive: true }` parameter that switches from half-open `[start, end)` to closed `[start, end]` overlap checks:

```typescript
// Before (half-open): excludes entries touching at boundary
if (entry.pmStart >= end) break;
if (entry.pmEnd <= start) continue;

// After (inclusive): includes entries touching at boundary
if (entry.pmStart > end) break;
if (entry.pmEnd < start) continue;
```

This is opt-in — the default half-open behavior is preserved for all other callers.

### 2. `DomSelectionGeometry.computeSelectionRectsFromDom` — use inclusive boundaries

All calls to `findEntriesInRange` in the selection rect computation now pass `{ boundaryInclusive: true }`. This ensures the two adjacent text spans are both found when a selection crosses a run boundary.

### 3. `PresentationEditor.#updateSelectionFromEditor` — safety net

If a non-empty selection (`from !== to`) produces zero DOM rects (edge case where even inclusive boundaries don't find entries), the method now returns early instead of clearing the overlay. This preserves the last valid highlight.

## Test plan

- [x] 4 new unit tests in `DomPositionIndex.test.ts`:
  - Default half-open semantics exclude boundary entries
  - `boundaryInclusive` includes both adjacent entries at a run gap
  - `boundaryInclusive` includes entry whose `pmEnd` equals query start
  - `boundaryInclusive` includes entry whose `pmStart` equals query end
- [x] All 7083 super-editor tests pass
- [x] All 1220 layout-bridge tests pass
- [x] Manual browser test: drag-selecting across bold/italic/color boundaries no longer flickers

### DIFF
--- a/packages/layout-engine/layout-bridge/test/clickToPosition-dom-page.test.ts
+++ b/packages/layout-engine/layout-bridge/test/clickToPosition-dom-page.test.ts
@@ -1,0 +1,301 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * Tests for DOM-based page-relative coordinate resolution in clickToPosition.
+ *
+ * When the geometry fallback is used (DOM mapping returns null for fragment/line),
+ * clickToPosition must compute page-relative Y from the actual DOM page element
+ * position, not from geometry-based page-top calculations. This is critical on
+ * page 2+ where container-space Y diverges from page-relative Y.
+ *
+ * Scenario (SD-2024): A paragraph sits just above a table on page 2. Clicking on
+ * the paragraph's left margin (where no fragment/line is found by elementsFromPoint)
+ * triggers the geometry fallback. If container-space Y is used as page-relative Y,
+ * the click resolves to a position far below the paragraph — inside or past the
+ * table. The fix computes page-relative Y directly from the page element's
+ * bounding rect.
+ */
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { clickToPosition } from '../src/index.ts';
+import type { Layout, FlowBlock, Measure } from '@superdoc/contracts';
+
+/**
+ * Helper: temporarily mock document.elementsFromPoint for the duration of `run()`.
+ */
+function withMockedElementsFromPoint(elements: Element[], run: () => void): void {
+  const originalElementsFromPoint = document.elementsFromPoint;
+  document.elementsFromPoint = (_x: number, _y: number) => elements;
+
+  try {
+    run();
+  } finally {
+    document.elementsFromPoint = originalElementsFromPoint;
+  }
+}
+
+describe('clickToPosition: DOM-based page-relative Y on page 2+ (SD-2024)', () => {
+  let container: HTMLElement;
+
+  // Layout: page 1 has a filler paragraph, page 2 has a paragraph followed by a table.
+  // The paragraph on page 2 is at y=40 (page-relative), table at y=120.
+  const paraBlock: FlowBlock = {
+    kind: 'paragraph',
+    id: 'page2-para',
+    runs: [{ text: 'In witness to the commitment', fontFamily: 'Arial', fontSize: 11, pmStart: 100, pmEnd: 128 }],
+  };
+
+  const tableBlock: FlowBlock = {
+    kind: 'table',
+    id: 'page2-table',
+    rows: [
+      {
+        id: 'row-0',
+        cells: [
+          {
+            id: 'cell-0',
+            blocks: [
+              {
+                kind: 'paragraph',
+                id: 'cell-para',
+                runs: [{ text: 'Signature', fontFamily: 'Arial', fontSize: 11, pmStart: 130, pmEnd: 139 }],
+              },
+            ],
+            attrs: { padding: { top: 2, bottom: 2, left: 4, right: 4 } },
+          },
+        ],
+      },
+    ],
+  };
+
+  const paraMeasure: Measure = {
+    kind: 'paragraph',
+    lines: [
+      {
+        fromRun: 0,
+        fromChar: 0,
+        toRun: 0,
+        toChar: 28,
+        width: 300,
+        ascent: 10,
+        descent: 4,
+        lineHeight: 16,
+      },
+    ],
+    totalHeight: 16,
+  };
+
+  const tableMeasure: Measure = {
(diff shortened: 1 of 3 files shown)


packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -3946,6 +3946,13 @@ export class PresentationEditor extends EventEmitter {
       return;
     }
 
+    // When a non-empty selection spans only structural tokens (e.g., run open/close
+    // tokens at a mark boundary), the DOM Range produces no visible rects. Preserve
+    // the last overlay to prevent flicker during drag selection across mark boundaries.
+    if (domRects.length === 0 && from !== to) {
```
caio-pizzol: this early return catches all zero-rect non-collapsed selections, not just the mark-boundary case it's guarding against. if DOM nodes become disconnected or virtualized after a scroll, the old overlay stays visible at the wrong position until the next interaction. worth adding a `debugLog('warn', ...)` here like line 3945 does so future zero-rect cases are diagnosable?

```suggestion
    if (domRects.length === 0 && from !== to) {
      debugLog('warn', '[drawSelection] zero rects for non-collapsed selection — preserving last overlay', { from, to });
      return;
    }
```

packages/super-editor/src/core/presentation-editor/dom/DomPositionIndex.ts
```diff
@@ -330,7 +343,7 @@ export class DomPositionIndex {
    * Finds all DOM elements whose position ranges overlap with the given range.
    *
    * @param from - The start of the query range (inclusive)
-   * @param to - The end of the query range (exclusive)
+   * @param to - The end of the query range (inclusive)
```
caio-pizzol: the doc says `to` is now "inclusive" but `findElementsInRange` still calls `findEntriesInRange(from, to)` without `{ boundaryInclusive: true }`, so the range is still half-open. the test at line 904 confirms this too.

```suggestion
   * @param to - The end of the query range (exclusive)
```

tupizz: @caio-pizzol fixed — reverted to "exclusive". findElementsInRange doesn't pass boundaryInclusive so the doc was wrong.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
tupizz: @caio-pizzol added the debugLog — matches the pattern at line 3945. If this fires outside drag-select (virtualized pages, disconnected DOM), we'll see it in the logs now.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -3946,6 +3946,17 @@ export class PresentationEditor extends EventEmitter {
...
+      debugLog('warn', '[drawSelection] zero rects for non-collapsed selection — preserving last overlay', {
+        from,
+        to,
+      });
+      return;
```
caio-pizzol, line 3959: still broad — you have `#isDragSelecting`, worth scoping this to drag-only?

```suggestion
    if (domRects.length === 0 && from !== to && this.#isDragSelecting) return;
```

caio-pizzol (commented):
@tupizz the debugLog and the JSDoc fix look good. one thing still open — left an inline comment on the zero-rect guard.

caio-pizzol:
@tupizz 

one thing worth adding as a follow-up: a behavior test that reproduces the actual user interaction.                                   
                                                                                                                                                     
something like:                                                                                                                                                                                                                                                                               1. load a doc with mixed inline formatting (bold → italic) adjacent to a table
2. drag-select across the mark boundary
3. assert the selection range stays continuous and doesn't collapse

harbournick:

PR #2183, opened 2026-02-26
fix: preserve line spacing and indentation on Google Docs paste

gpardhivvarma:
## Summary

- Add CSS inline style fallback in `parseAttrs.js` so line spacing (`lineHeight`), paragraph spacing (`marginTop`/`marginBottom`), indentation (`marginLeft`), and first-line/hanging indent (`textIndent`) are preserved when pasting from Google Docs or other sources that use inline CSS instead of SuperDoc's `data-*` attributes
- Use a single `parseCssLength` helper backed by a `CSS_LENGTH_TO_PT` conversion map supporting `pt`, `px`, `in`, `cm`, `mm` — unrecognized units (`em`, `rem`, etc.) are safely ignored
- Use exact `72 / 96` ratio for px→pt conversion, matching `PX_PER_PT` in pm-adapter
- Correctly parse CSS `line-height` units: unitless and `%` values are stored as auto multipliers, while absolute lengths are stored as exact line spacing in twips
- Line spacing formula inverts pm-adapter's `normalizeLineValue` (`value * 1.15 / 240`) so values round-trip correctly
- Map positive `text-indent` to `indent.firstLine` and negative `text-indent` to `indent.hanging`
- Preserve explicit zero margins (`margin-top: 0`) as `0` twips to prevent the style-engine from applying non-zero defaults
- Fix pre-existing bug: skip `Number()` coercion for `lineRule` in `data-spacing` parsing to preserve the `'auto'` string value
- Data attributes still take priority, so existing DOCX paste behavior is unchanged

Closes #2151

## Test plan

- [ ] Paste from Google Docs with different line spacings (single, 1.5, double) and verify they are preserved
- [ ] Paste indented paragraphs from Google Docs and verify indentation is preserved
- [ ] Paste HTML with first-line and hanging indentation and verify correct rendering
- [ ] Paste HTML with percentage/px/pt/in/cm/mm values and verify correct conversion
- [ ] Paste HTML with explicit zero margins and verify no unintended spacing appears
- [ ] Paste from Word and verify existing `data-spacing`/`data-indent` behavior is unchanged
- [x] Run `pnpm test` — all parseAttrs tests pass (28/28)

PR #2208, opened 2026-02-27
fix: preserve text-align on paste from Google Docs

gpardhivvarma:
## Summary

- Adds CSS `text-align` inline style fallback in `parseAttrs.js` so that paragraph alignment (center, right, justify) is preserved when pasting from Google Docs
- Maps CSS logical values (`start`/`end`) to physical OOXML justification values (`left`/`right`)
- Follows the same pattern as the existing spacing/indent CSS fallbacks from PR #2183

## Test plan

- [x] Added 8 unit tests covering all alignment values, invalid values, and combination with other CSS fallbacks
- [x] All 717 test files pass (7167 tests)
- [ ] Manual test: paste centered/right-aligned/justified text from Google Docs and verify alignment is preserved

### DIFF
--- a/packages/super-editor/src/extensions/paragraph/helpers/parseAttrs.js
+++ b/packages/super-editor/src/extensions/paragraph/helpers/parseAttrs.js
@@ -1,4 +1,5 @@
 const CSS_LENGTH_TO_PT = { pt: 1, px: 72 / 96, in: 72, cm: 28.3465, mm: 2.83465 };
+const CSS_ALIGN_TO_OOXML = { center: 'center', right: 'right', justify: 'justify', end: 'right' };
 
 /**
  * Parse a CSS length value and return { points, unit }.
@@ -110,14 +111,8 @@ export function parseAttrs(node) {
   let justification;
   if (!justification && node.style) {
     const textAlign = node.style.textAlign;
-    const alignMap = {
-      center: 'center',
-      right: 'right',
-      justify: 'justify',
-      end: 'right',
-    };
-    if (textAlign && alignMap[textAlign]) {
-      justification = alignMap[textAlign];
+    if (textAlign && CSS_ALIGN_TO_OOXML[textAlign]) {
+      justification = CSS_ALIGN_TO_OOXML[textAlign];
     }
   }
 
--- a/tests/behavior/tests/formatting/paste-text-align.spec.ts
+++ b/tests/behavior/tests/formatting/paste-text-align.spec.ts
@@ -39,5 +39,5 @@ test('pasted left-aligned paragraph does not store alignment (default)', async (
   await superdoc.waitForStable();
 
   // left is the default — parseAttrs skips it to avoid baking in direct formatting
-  await superdoc.assertTextAlignment('Left text', null as any);
+  await superdoc.assertTextAlignment('Left text', null);
 });
--- a/tests/behavior/fixtures/superdoc.ts
+++ b/tests/behavior/fixtures/superdoc.ts
@@ -780,7 +780,7 @@ function createFixture(page: Page, editor: Locator, modKey: string) {
       throw new Error(`assertTextMarkAttrs only supports "link" and "textStyle" via document-api; got "${markName}".`);
     },
 
-    async assertTextAlignment(text: string, expectedAlignment: string, occurrence = 0) {
+    async assertTextAlignment(text: string, expectedAlignment: string | null, occurrence = 0) {
       await expect
         .poll(() =>
           page.evaluate(


packages/super-editor/src/extensions/paragraph/helpers/parseAttrs.js
```diff
@@ -104,6 +104,23 @@ export function parseAttrs(node) {
     }
   }
 
+  // CSS inline style fallback for text-align (e.g. Google Docs paste)
+  let justification;
+  if (node.style) {
```
caio-pizzol: spacing and indent both check for a `data-*` attribute first and only fall back to CSS when it's missing (`if (!spacing && node.style)` on line 56, `if (!indent && node.style)` on line 87). this skips that check. no `data-justification` exists today, but adding the guard keeps the pattern consistent.

```suggestion
  let justification;
  if (!justification && node.style) {
```

caio-pizzol (commented):
@gpardhivvarma thanks for picking this up! clean approach that follows the existing pattern well. left two inline comments — nothing blocking.

packages/super-editor/src/extensions/paragraph/helpers/parseAttrs.js
```diff
@@ -104,6 +104,23 @@ export function parseAttrs(node) {
...
+    const textAlign = node.style.textAlign;
+    const alignMap = {
+      left: 'left',
+      center: 'center',
```
caio-pizzol: Google Docs sets `text-align: left` on every paragraph by default, so all pasted paragraphs will get `justification: 'left'` baked in as direct formatting. on export that writes `<w:jc w:val="left"/>` which is valid but unnecessary since left is already the default. worth skipping `left` here so only non-default alignments get stored?

gpardhivvarma:
@caio-pizzol Mind taking a look

gpardhivvarma:
@caio-pizzol Pls take a look

packages/super-editor/src/extensions/paragraph/helpers/parseAttrs.js
```diff
@@ -104,6 +104,23 @@ export function parseAttrs(node) {
     }
   }
 
+  // CSS inline style fallback for text-align (e.g. Google Docs paste)
+  // Skip 'left' — Google Docs sets text-align: left on every paragraph,
+  // and storing it would bake in unnecessary direct formatting on export.
+  let justification;
+  if (!justification && node.style) {
+    const textAlign = node.style.textAlign;
+    const alignMap = {
```
caio-pizzol: `alignMap` is rebuilt on every call. `CSS_LENGTH_TO_PT` at the top of this file is already hoisted for the same reason — worth doing here too?

tests/behavior/tests/formatting/paste-text-align.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  await superdoc.assertTextAlignment('Justified text', 'justify');
+});
+
+test('pasted left-aligned paragraph does not store alignment (default)', async ({ superdoc }) => {
+  await insertHTML(superdoc.page, '<p style="text-align: left">Left text</p>');
+  await superdoc.waitForStable();
+
+  // left is the default — parseAttrs skips it to avoid baking in direct formatting
+  await superdoc.assertTextAlignment('Left text', null as any);
```
caio-pizzol: `null as any` bypasses type checking. `assertTextAlignment` in `tests/behavior/fixtures/superdoc.ts` takes `string` but `null` is a valid expected value here. widening the param to `string | null` would let you drop the cast.

packages/super-editor/src/extensions/paragraph/helpers/parseAttrs.js
gpardhivvarma: Done — hoisted to module-level `CSS_ALIGN_TO_OOXML` constant, same pattern as `CSS_LENGTH_TO_PT`.

tests/behavior/tests/formatting/paste-text-align.spec.ts
gpardhivvarma: Done — widened to `string | null` and dropped the `null as any` cast.

packages/super-editor/src/extensions/paragraph/helpers/parseAttrs.js
gpardhivvarma: Already done — the code has `if (\!justification && node.style)` matching the spacing/indent pattern.

packages/super-editor/src/extensions/paragraph/helpers/parseAttrs.js
gpardhivvarma: Already done — `left` is excluded from the map so Google Docs default `text-align: left` doesn't get stored as direct formatting.

harbournick:

PR #2212, opened 2026-02-27
fix: remove persisting yellow underline after canceling track-format suggestion

gpardhivvarma:
## Summary

- **Fix 1 (addMarkStep.js):** Clamp TrackFormat `addMark` range to per-node boundaries (`Math.max(step.from, pos)` / `Math.min(step.to, pos + node.nodeSize)`) instead of using the full step range, preventing incorrect `before`/`after` arrays from leaking onto adjacent text nodes.
- **Fix 2 (removeMarkStep.js):** When a tracked format addition is fully reversed (`after` becomes empty after filtering), remove the TrackFormat mark entirely instead of creating a new one with `after=[]` and a stale `before` array — which would keep the gold/yellow underline visible for a no-op change.

## Test plan

- [x] All 7203 unit tests pass (`pnpm --filter super-editor test`)
- [ ] Behavior tests: `pnpm --filter behavior test -- reject-format-suggestion` (requires Playwright browsers installed)
- [ ] Manual: enable track changes → select text → bold → bold again (toggle off) → verify yellow underline disappears
- [ ] Manual: enable track changes → select text → bold → reject via suggestion bubble → verify both bold and underline disappear

### DIFF
--- a/packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.test.js
+++ b/packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.test.js
@@ -1,6 +1,6 @@
 import { beforeEach, afterEach, describe, expect, it, vi } from 'vitest';
 import { EditorState, TextSelection } from 'prosemirror-state';
-import { AddMarkStep, RemoveMarkStep } from 'prosemirror-transform';
+import { AddMarkStep } from 'prosemirror-transform';
 import { trackedTransaction } from './index.js';
 import { TrackFormatMarkName } from '../constants.js';
 import { initTestEditor } from '@tests/helpers/helpers.js';
@@ -86,6 +86,9 @@ describe('trackChangesHelpers addMarkStep / removeMarkStep (track format)', () =
 
     const ids = new Set(tfMarks.map((m) => m.id));
     expect(ids.size).toBe(1);
+
+    // Ranges should be clamped to per-node boundaries (no overlap between marks)
+    expect(tfMarks[0].to).toBeLessThanOrEqual(tfMarks[1].from);
 
     // The "after" array should include bold
     for (const tf of tfMarks) {


gpardhivvarma:
@codex review

gpardhivvarma:
@codex review

gpardhivvarma:
@caio-pizzol please review at the earliest

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.js
```diff
@@ -38,7 +39,7 @@ export const addMarkStep = ({ state, step, newTr, doc, user, date }) => {
     const existingChangeMark = liveMarks.find((mark) =>
       [TrackDeleteMarkName, TrackFormatMarkName].includes(mark.type.name),
     );
-    const wid = existingChangeMark ? existingChangeMark.attrs.id : uuidv4();
+    const wid = existingChangeMark ? existingChangeMark.attrs.id : (sharedWid ?? (sharedWid = uuidv4()));
```
caio-pizzol, line 42: the three fixes here (range clamping, shared IDs, empty-mark removal) don't have tests yet. a behavior test with two runs in a paragraph — apply bold across both, check they share one trackFormat ID, then remove bold and check the mark is gone — would cover all three. worth adding?

gpardhivvarma: Yeah makes sense will add and lyk

gpardhivvarma:
@caio-pizzol Please take a look at it

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.test.js
```diff
@@ -0,0 +1,169 @@
...
+    // Use a single AddMarkStep to ensure both nodes are processed in one call
+    const boldMark = schema.marks.bold.create();
+    let tr = state.tr;
+    tr.step(new AddMarkStep(helloPos, worldEnd, boldMark));
+    tr.setMeta('inputType', 'programmatic');
+    const tracked = trackedTransaction({ tr, state, user });
+    state = state.apply(tracked);
+
+    // Both text nodes should have TrackFormat marks with the same ID
+    const tfMarks = getTrackFormatMarks(state.doc);
```
caio-pizzol, line 84: the shared-ID test doesn't assert on the mark ranges, so if the range-clamping fix (using `Math.max`/`Math.min` instead of `step.from`/`step.to`) were reverted, this test would still pass. pinning the boundary would make the test actually guard the fix — worth doing?

```js
expect(tfMarks[0].to).toBe(tfMarks[1].from);
```

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.test.js
```diff
@@ -0,0 +1,169 @@
+import { beforeEach, afterEach, describe, expect, it, vi } from 'vitest';
+import { EditorState, TextSelection } from 'prosemirror-state';
+import { AddMarkStep, RemoveMarkStep } from 'prosemirror-transform';
```
caio-pizzol: `RemoveMarkStep` is imported but never used — the remove tests go through `state.tr.removeMark()`. triggers a `no-unused-vars` warning.

```suggestion
import { AddMarkStep } from 'prosemirror-transform';
```

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.test.js
gpardhivvarma: Made the addition

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/addMarkStep.test.js
gpardhivvarma: Done

gpardhivvarma:
@caio-pizzol

caio-pizzol:
Thanks for the work on this @gpardhivvarma! The range clamping and `sharedWid` fixes are correct.

We found one remaining edge case when canceling a format change on nodes with pre-existing marks (e.g., bold on/off on italic text leaves a ghost "removed italic" mark). We've built on your commits and added the fix in #2233.

Closing in favor of #2233. Appreciate the contribution!

PR #2221, opened 2026-02-28
fix(scroll): wait for virtualized page mount and center text element

mattConnHarbour:


packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -2053,6 +2055,98 @@ export class PresentationEditor extends EventEmitter {
...
+   * Find the DOM element containing a specific document position.
+   * Returns the most specific (smallest range) matching element.
+   */
+  #findElementAtPosition(pageEl: HTMLElement, pos: number): HTMLElement | null {
+    const elements = Array.from(pageEl.querySelectorAll('[data-pm-start][data-pm-end]'));
+    let bestMatch: HTMLElement | null = null;
+    let smallestRange = Infinity;
+
+    for (const el of elements) {
+      const htmlEl = el as HTMLElement;
```
caio-pizzol, line 2070: this scans all `[data-pm-start][data-pm-end]` under the page, but header/footer elements also have those attributes in their own coordinate space. a body position could accidentally match a header element and scroll there instead. two other places in the codebase already filter these out — worth adding the same guard here?

```suggestion
    for (const el of elements) {
      const htmlEl = el as HTMLElement;
      if (htmlEl.closest('.superdoc-page-header, .superdoc-page-footer')) continue;
      const start = Number(htmlEl.dataset.pmStart);
      const end = Number(htmlEl.dataset.pmEnd);
```

packages/super-editor/src/extensions/search/search.js
```diff
@@ -458,9 +458,11 @@ export const Search = Extension.create({
           if (dispatch) dispatch(tr);
 
           const presentationEditor = editor.presentationEditor;
-          if (presentationEditor && typeof presentationEditor.scrollToPosition === 'function') {
-            const didScroll = presentationEditor.scrollToPosition(from, { block: 'center' });
-            if (didScroll) return true;
+          const scrollFn = presentationEditor?.scrollToPositionAsync ?? presentationEditor?.scrollToPosition;
+          if (typeof scrollFn === 'function') {
+            // Fire-and-forget: async version handles virtualized pages, sync is fallback
+            scrollFn.call(presentationEditor, from, { block: 'center' });
```
caio-pizzol: the old code fell through to `domAtPos` when scroll failed. now it returns `true` regardless, so failures are silent. also, if the promise rejects it's unhandled.

```suggestion
            // Fire-and-forget: async version handles virtualized pages, sync is fallback
            Promise.resolve(scrollFn.call(presentationEditor, from, { block: 'center' })).catch(() => {});
```

packages/super-editor/src/core/presentation-editor/tests/PresentationEditor.scrollToPosition.test.ts
```diff
@@ -0,0 +1,611 @@
...
+        element: container,
+        documentId: 'test-doc',
+      });
+
+      await vi.waitFor(() => expect(mockIncrementalLayout).toHaveBeenCalled());
+      await new Promise((resolve) => setTimeout(resolve, 50));
+
+      // Create mock page element
+      const pagesHost = container.querySelector('.presentation-editor__pages') as HTMLElement;
+      if (pagesHost) {
```
caio-pizzol, line 376: if `.presentation-editor__pages` gets renamed, `pagesHost` is null, the `if` block is skipped, and the test passes without checking anything. an `expect(pagesHost).toBeTruthy()` before the block would catch that.

packages/super-editor/src/core/presentation-editor/tests/PresentationEditor.scrollToPosition.test.ts
```diff
@@ -0,0 +1,611 @@
...
+
+      // Page 2 never mounts - should timeout
+      // Note: This test may be slow due to the 2000ms timeout
+      // In real tests, you might want to mock the timeout value
+      const result = await editor.scrollToPositionAsync(150);
+
+      // The page never mounted, so it should fail
+      // However, due to the test setup, the page might already exist
+      // This test verifies the warning path when timeout occurs
+      if (!result) {
```
caio-pizzol: the assertion is inside `if (!result)`, so if the page happens to mount, the test passes without verifying the timeout path. should be unconditional: `expect(result).toBe(false)`.

caio-pizzol (changes requested):
@mattConnHarbour good approach — mirrors the `goToAnchor` pattern nicely. left a few inline comments, two worth fixing before merge (header/footer collision + silent scroll failure), two smaller test things.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -2053,6 +2055,98 @@ export class PresentationEditor extends EventEmitter {
     }
   }
 
+  /**
+   * Find the DOM element containing a specific document position.
+   * Returns the most specific (smallest range) matching element.
+   */
+  #findElementAtPosition(pageEl: HTMLElement, pos: number): HTMLElement | null {
+    const elements = Array.from(pageEl.querySelectorAll('[data-pm-start][data-pm-end]'));
```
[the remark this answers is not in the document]
harbournick: fixed

packages/super-editor/src/extensions/search/search.js
```diff
@@ -458,9 +458,11 @@ export const Search = Extension.create({
...
+            return true;
```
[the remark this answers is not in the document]
harbournick: fixed

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
harbournick: fixed

packages/super-editor/src/core/presentation-editor/tests/PresentationEditor.scrollToPosition.test.ts
harbournick: fixed

packages/super-editor/src/core/presentation-editor/tests/PresentationEditor.scrollToPosition.test.ts
harbournick: fixed

packages/super-editor/src/extensions/search/search.js
harbournick: fixed

mattConnHarbour:
@harbournick Big thanks for addressing Caio's review! @caio-pizzol I think this is ready for merge when you have a chance?

harbournick:

PR #2233, opened 2026-03-02
fix(track-changes): remove ghost TrackFormat on multi-node format cancel

caio-pizzol:
Builds on #2212 by @gpardhivvarma (range clamping + sharedWid fixes) and adds the missing cancel logic for nodes with pre-existing marks.

- When bold is toggled on then off across a multi-node selection (e.g., "Hello " plain + "world" italic), the `removeMarkStep` cancel path only cleaned up when both `before` and `after` were empty
- This missed nodes like "world" where `before=[italic]` (a snapshot of pre-existing marks) — leaving a ghost TrackFormat showing "Format: removed italic" even though italic was never touched
- The fix checks whether all marks in `before` still exist on the node; if they do, the tracked change is a no-op and the TrackFormat is removed


harbournick:

PR #2236, opened 2026-03-02
fix(editor): prevent scroll-to-top when clicking toolbar buttons

tupizz:
## Problem

When SuperDoc is embedded in a page where the **window** is the scroll container (not a div with `overflow: auto`), clicking any toolbar button scrolls the page to the top. This affects the Version Tester app and any external integration where the editor lives in a scrollable page.

Fixes SD-1780 and SD-1785.

## How we found it

The bug was reported through the Version Tester app. It was not reproducible in the SuperDoc dev app because the dev app uses a div with `overflow: auto` as its scroll container, keeping `window.scrollY` at 0 at all times. The Version Tester loads SuperDoc inside a page where the window itself scrolls, which exposed the problem.

The dev app has a built-in scroll test mode (`?scrolltest=1`) that sets `.dev-app__main` to `overflow: visible`, making the window the scroll container. This mode reproduces the exact same behavior.

## Root cause

The hidden ProseMirror editor is positioned at `position: fixed; top: 0; left: -9999px`. When a toolbar button is clicked:

1. The browser transfers focus from the PM editor to the toolbar button (default mousedown behavior)
2. The toolbar calls `view.focus()` to refocus the PM editor
3. ProseMirror's internal `onFocus` handler runs `updateState` then `selectionToDOM`, which manipulates `window.getSelection()`
4. The browser **asynchronously** scrolls to make the DOM selection visible in the hidden editor at `top: 0`
5. This async scroll fires AFTER the `wrapHiddenEditorFocus` wrapper finishes its synchronous scroll restoration

## Solution

Two complementary changes:

### 1. Toolbar mousedown prevention (primary fix)

Added a `@mousedown` handler on the toolbar root div that calls `preventDefault()` for non-input elements. This prevents the browser from transferring focus when clicking toolbar buttons. The PM editor stays focused, no blur/refocus cycle occurs, and no scroll happens.

Input elements (font size input, search bar) are excluded so they still receive native focus and cursor placement.

This is the standard pattern used by ProseMirror's own example editor, Tiptap, and most WYSIWYG editors to keep the editor focused during toolbar interaction.

### 2. RAF backup in wrapHiddenEditorFocus (safety net)

After the synchronous scroll restoration in the focus wrapper, a `requestAnimationFrame` callback now checks and restores scroll positions again. This catches async browser scroll triggered by `selectionToDOM` from any code path, not just toolbar clicks.

## How to reproduce

1. Run `pnpm dev`
2. Open `http://localhost:9096/?scrolltest=1`
3. Scroll down past the spacer to reach the editor
4. Click any toolbar button (Bold, Italic, etc.)
5. **Before fix**: page scrolls to top
6. **After fix**: scroll position stays put

## Why this approach

The mousedown prevention is the correct fix because it eliminates the problem at the source. Preventing focus transfer means ProseMirror never goes through a blur/focus cycle, so `selectionToDOM` never fires in the context of a focus event, and the browser never has a reason to scroll.

The RAF backup is a defense-in-depth measure. Other code paths (programmatic focus calls, browser extensions, accessibility tools) could still trigger the same async scroll behavior. The RAF callback catches these cases without adding complexity to every call site.

## Test plan

- [x] All existing tests pass (733 files, 7411 tests)
- [x] New test verifies RAF scheduling in focus wrapper
- [x] Verified in browser with `?scrolltest=1`: Bold, Italic, Underline, Color buttons all maintain scroll position
- [x] Font size input still receives focus and accepts text input
- [ ] Verify in Version Tester app

### DIFF
--- a/packages/super-editor/src/core/presentation-editor/tests/PresentationEditor.focusWrapping.test.ts
+++ b/packages/super-editor/src/core/presentation-editor/tests/PresentationEditor.focusWrapping.test.ts
@@ -356,20 +356,33 @@ describe('PresentationEditor - Focus Wrapping (#wrapHiddenEditorFocus)', () => {
       }).not.toThrow();
     });
 
-    it('schedules requestAnimationFrame as async scroll safety net', () => {
+    it('schedules requestAnimationFrame that restores scroll on async drift', () => {
       editor = new PresentationEditor({
         element: container,
         documentId: 'test-doc',
         pageSize: { w: 612, h: 792 },
       });
 
-      const rafSpy = vi.spyOn(window, 'requestAnimationFrame');
+      // Capture the RAF callback so we can invoke it manually
+      let rafCallback: FrameRequestCallback | null = null;
+      const rafSpy = vi.spyOn(window, 'requestAnimationFrame').mockImplementation((cb) => {
+        rafCallback = cb;
+        return 1;
+      });
+      const scrollToSpy = vi.spyOn(window, 'scrollTo').mockImplementation(() => {});
 
+      // At focus time, scrollX=0 scrollY=0 → captured as beforeX=0 beforeY=0
       editor.editor.view.focus();
 
-      // RAF should be scheduled to catch async browser scroll after focus
       expect(rafSpy).toHaveBeenCalledTimes(1);
-      expect(rafSpy).toHaveBeenCalledWith(expect.any(Function));
+      expect(rafCallback).not.toBeNull();
+
+      // Simulate the browser async-scrolling to the hidden editor (drift from 0 to 500)
+      Object.defineProperty(window, 'scrollY', { value: 500, configurable: true });
+
+      // Run the RAF callback — it should detect drift and restore to beforeY=0
+      rafCallback!(0);
+      expect(scrollToSpy).toHaveBeenCalledWith(0, 0);
     });
 
     it('cancels focus-scroll RAF when scrollToPosition is called', () => {


packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -743,6 +743,15 @@ export class PresentationEditor extends EventEmitter {
       if (win.scrollX !== beforeX || win.scrollY !== beforeY) {
         win.scrollTo(beforeX, beforeY);
       }
+
+      // Safety net: the browser may asynchronously scroll after ProseMirror's
+      // selectionToDOM() modifies the DOM selection inside the hidden editor.
+      // A single requestAnimationFrame catches this post-layout scroll.
+      win.requestAnimationFrame(() => {
```
caio-pizzol: this saves the scroll position when `view.focus()` runs and puts it back one frame later. but search calls `view.focus()` then scrolls to the match — the RAF fires after and jumps back to where the user was before. breaks find-and-navigate when the window scrolls.

fix idea: cancel this RAF from `scrollToPosition` so intentional scrolls aren't undone.

caio-pizzol (changes requested):
@tupizz toolbar mousedown fix is clean and follows the standard pattern. one issue: the RAF scroll restore at `PresentationEditor.ts:750` fights with search navigation — it undoes the scroll-to-match after `goToSearchResult` calls `scrollToPosition`. left an inline comment with a fix idea.

test coverage is decent, `handleToolbarMousedown` could use a test but not blocking.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -743,6 +744,21 @@ export class PresentationEditor extends EventEmitter {
...
+
+      // Safety net: the browser may asynchronously scroll after ProseMirror's
+      // selectionToDOM() modifies the DOM selection inside the hidden editor.
+      // A single requestAnimationFrame catches this post-layout scroll.
+      // The RAF ID is stored so scrollToPosition() can cancel it — otherwise
+      // intentional scrolls (e.g. search navigation) would be undone.
+      if (this.#focusScrollRafId != null) {
+        win.cancelAnimationFrame(this.#focusScrollRafId);
+      }
+      this.#focusScrollRafId = win.requestAnimationFrame(() => {
```
caio-pizzol, line 788: `#focusScrollRafId` isn't cleaned up in `destroy()`, but the other two (`#rafHandle`, `#decorationSyncRafHandle`) are. if the editor is torn down while this is still pending, it fires after and scrolls the window to a wrong position. worth adding the same cleanup block in `destroy()` after line 2372?

```js
if (this.#focusScrollRafId != null) {
  safeCleanup(() => {
    const win = this.#visibleHost?.ownerDocument?.defaultView ?? window;
    win.cancelAnimationFrame(this.#focusScrollRafId!);
    this.#focusScrollRafId = null;
  }, 'Focus scroll RAF');
}
```

packages/super-editor/src/core/presentation-editor/tests/PresentationEditor.focusWrapping.test.ts
```diff
@@ -355,6 +355,39 @@ describe('PresentationEditor - Focus Wrapping (#wrapHiddenEditorFocus)', () => {
...
     });
+
+    it('schedules requestAnimationFrame as async scroll safety net', () => {
+      editor = new PresentationEditor({
+        element: container,
+        documentId: 'test-doc',
+        pageSize: { w: 612, h: 792 },
+      });
+
+      const rafSpy = vi.spyOn(window, 'requestAnimationFrame');
```
caio-pizzol: the spy watches `requestAnimationFrame` but never runs the callback, so the scroll-restore logic inside it is never actually tested. not blocking, but something to keep in mind.

harbournick:
@tupizz mind resolving the conflict and resolving Caio's comments so we can merge this one? Thank you!

harbournick:

PR #2238, opened 2026-03-02
fix(editor): prevent focus loss when typing in header/footer editors (SD-1993)

tupizz:
## Summary

Fixes [SD-1993](https://linear.app/superdocworkspace/issue/SD-1993/bug-headersfooters-do-not-support-typing): headers and footers lose focus after the first keystroke, making them effectively uneditable.

### Root cause

Every keystroke in a header/footer PM editor triggers a full body rerender cycle:

```
User types in H/F editor
  → contentChanged event → requestRerender()
  → RAF → #rerender() → painter.paint()
  → renderVirtualized() → updateVirtualWindow()
  → appendChild(pageElement)   ← browser fires blur on focused H/F editor
```

`DomPainter.updateVirtualWindow()` has an ordering loop that re-appends **all** mounted page elements via `appendChild`. Calling `appendChild` on an element already in the DOM is a move-to-end operation. The browser fires a `blur` event on any focused element inside the moved subtree. This kills focus on the header/footer PM editor after every keystroke.

All three reported symptoms stem from this single focus loss:
- **Header only accepts first keystroke**: first char triggers rerender → blur → second char has no target
- **Space scrolls to bottom**: focus loss + spacer height recalculation → browser adjusts scroll
- **Footer text appears in body**: after focus loss, input bridge may route keystrokes to the body editor

### Fix

**1. Cursor-based DOM reconciliation in `updateVirtualWindow`** (`renderer.ts`)

Replace the unconditional `appendChild` ordering loop with cursor-based reconciliation. Tracks a `cursor` pointer walking through the existing children. For each mounted page, if it's already at the cursor position, skip the DOM mutation and advance. Only call `insertBefore` when elements are actually out of order.

In the common case (H/F typing with no scroll or window change), pages are already in the correct order after `clearGapSpacers()` removes old gaps. Zero moves = zero blur = focus preserved.

**2. Focus save/restore safety net** (`PresentationEditor.ts`)

Captures the active H/F editor's focus state before `#rerender()`. In the `finally` block, if focus was lost (e.g. from edge cases like scroll during editing, viewport resize, page count changes), restores it via `activeHfEditor.view.focus()`.

## Test plan

- [x] All 645 painter-dom tests pass
- [x] All 7411 super-editor tests pass
- [x] Pre-commit hooks pass (format, lint, typecheck, commitlint)
- [ ] Manual: load document with header/footer, double-click header, type multiple characters rapidly. All characters appear, no scroll jump.
- [ ] Manual: double-click footer, type text. Text appears in footer only, not in body.
- [ ] Manual: press Escape to exit, verify changes persist in static render.
- [ ] Manual: verify normal body editing and page virtualization (scrolling) still work correctly.


caio-pizzol (commented):
@tupizz solid fix, the reconciliation approach is clean. one small thing — the focus restore should check the session is still active before refocusing. left an inline comment with a one-line suggestion.

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -3165,13 +3165,34 @@ export class PresentationEditor extends EventEmitter {
...
     try {
       await this.#rerender();
     } finally {
       this.#isRerendering = false;
       if (this.#pendingDocChange) {
         this.#scheduleRerender();
       }
+
+      // Restore focus if the H/F editor lost it during rerender.
+      if (hadHfFocus && activeHfEditor?.view) {
```
caio-pizzol: `activeHfEditor` is captured before the `await`. if the user exits H/F mode while rerender is running, this will steal focus back to the old editor. not triggerable today, but worth guarding against.

```suggestion
      if (hadHfFocus && activeHfEditor?.view && this.#headerFooterSession?.activeEditor === activeHfEditor) {
```

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
```diff
@@ -3165,13 +3165,34 @@ export class PresentationEditor extends EventEmitter {
...
+        const doc = this.#visibleHost.ownerDocument;
+        const editorDom = activeHfEditor.view.dom;
+        if (doc && !editorDom.contains(doc.activeElement)) {
```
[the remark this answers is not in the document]
caio-pizzol: the inline suggestion on my review covers this - same issue

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -1721,10 +1726,16 @@ export class DomPainter {
           this.topOfIndex(idx) - this.topOfIndex(prevIndex) - this.virtualHeights[prevIndex] - this.virtualGap * 2;
         gap.style.height = `${Math.max(0, Math.floor(gapHeight))}px`;
         this.virtualGapSpacers.push(gap);
-        this.virtualPagesEl.appendChild(gap);
+        this.virtualPagesEl.insertBefore(gap, cursor);
```
caio-pizzol, line 1848: `cursor` not moving here is intentional but easy to misread as a bug — a short comment would help.

caio-pizzol (approved):
@tupizz round 1 feedback addressed, looks good. no issues found, tested locally and it works. one small readability comment inline.

no test covers typing in a header and keeping focus — worth adding to `double-click-edit-header.spec.ts` but not blocking.

feel free to merge after addressing relevant ones

packages/layout-engine/painters/dom/src/renderer.ts
tupizz: Addressed in 68d0e2e57 — added a clarifying comment explaining why `cursor` is intentionally not advanced when inserting a gap spacer (inserted *before* cursor, so cursor still points at the next page element to reconcile).

packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
tupizz: Already addressed in 796c23a6e — added the `this.#headerFooterSession?.activeEditor === activeHfEditor` guard as suggested.

harbournick:

PR #2239, opened 2026-03-02
fix: prefer full decoration range

VladaHarbour:
Long highlight decoration range may be partially restored when mark applied inside or this range.

### DIFF
--- a/packages/super-editor/src/core/presentation-editor/dom/DecorationBridge.ts
+++ b/packages/super-editor/src/core/presentation-editor/dom/DecorationBridge.ts
@@ -579,10 +579,7 @@ export class DecorationBridge {
         this.#prevDecorationSets.set(plugin, decorationSet);
         continue;
       }
-      if (previousPluginRanges?.length) {
-        const restoredRanges = restoreRangesFromPrevious(state.doc, docSize, previousPluginRanges);
-        this.#applyRangesToDesired(desired, domIndex, restoredRanges);
-      }
+      this.#setPreviousRanges(plugin, []);
       this.#prevDecorationSets.set(plugin, decorationSet);
     }
 


mattConnHarbour (approved):
LGTM

packages/super-editor/src/core/presentation-editor/dom/DecorationBridge.ts
```diff
@@ -215,6 +215,44 @@ function restoreRangesFromPrevious(
...
+): PreviousRange[] {
+  if (current.length === 0 || !previousRanges?.length) return current;
+  const restored = restoreRangesFromPrevious(doc, docSize, previousRanges);
+  if (restored.length === 0) return current;
+
+  const currentSpan = rangeUnion(current);
+  const restoredSpan = rangeUnion(restored);
+  if (!currentSpan || !restoredSpan) return current;
+  // Prefer restored only when it strictly contains current (plugin returned partial).
+  const contained =
```
caiopizzol, line 258: right now there's no way to tell if a plugin's range got smaller because mapping broke it, or because the plugin actually wanted it smaller. if both happen in the same doc-changing transaction, the bridge always picks the bigger range. it fixes itself on the next non-doc-change transaction, so the window is small — but a comment here explaining that assumption would save future readers some time.

packages/super-editor/src/core/presentation-editor/dom/DecorationBridge.ts
```diff
@@ -439,13 +485,26 @@ export class DecorationBridge {
...
+        this.#lastTransactionWasDocChange && mayRestoreEmpty && previousPluginRanges?.length
+          ? preferFullRestoredWhenPartial(pluginRanges, previousPluginRanges, state.doc, docSize)
+          : pluginRanges;
+
+      // Store ranges for next comparison.
+      // - If plugin reported current ranges and we expanded them due to a doc change, store the expanded ones
+      //   so the highlight stays stable across subsequent syncs in the same update cycle.
+      // - Otherwise, store the plugin-reported current ranges so intentional narrowing is respected.
+      // - If plugin reported nothing, store the restored (if any).
+      const storeExpandedOnDocChange = this.#lastTransactionWasDocChange && effectiveRanges !== pluginRanges;
```
caiopizzol: this block (compute effective ranges → decide what to store → store it) is copy-pasted at line 677 in `sync` with just the variable name changed. if one gets updated without the other they'll quietly disagree. extract a small shared method?

packages/super-editor/src/core/presentation-editor/dom/DecorationBridge.test.ts
```diff
@@ -873,6 +873,49 @@ describe('DecorationBridge', () => {
       expect(ranges[0].from).toBe(13);
       expect(ranges[0].to).toBe(19);
     });
+
+    it('prefers full restored span when plugin returns partial (e.g. after applying mark in long selection)', () => {
```
caiopizzol, line 877: both tests use `docChanged: true` — there's no test showing that `docChanged: false` keeps the expansion off. add a case where the plugin narrows its range after a meta-only transaction and check that the smaller range comes back as-is.

packages/super-editor/src/core/presentation-editor/dom/DecorationBridge.ts
VladaHarbour: So the reason for this is the following: the transaction’s mapping is built from the document steps. In SuperDoc, when marks are applied, calculateInlineRunPropertiesPlugin runs and can append a transaction that splits runs (one run → several runs with different inline properties). That adds more steps and a more complex mapping. When DecorationSet.map(tr.mapping, tr.doc) runs with that mapping, positions and node boundaries can shift or collapse in a way that turns one decoration into a smaller or broken set.
I will left a comment to make it more clear

packages/super-editor/src/core/presentation-editor/dom/DecorationBridge.ts
VladaHarbour: added helpers for collecting and restoring ranges in both places

packages/super-editor/src/core/presentation-editor/dom/DecorationBridge.test.ts
VladaHarbour: added negative test case

harbournick:

PR #2240, opened 2026-03-02
feat: remove naive ui

artem-harbour:
Linear: SD-1190

- Removed naive ui from the codebase.
- Created necessary components for replacing naive ui components.
- The toolbar and comment dropdowns separated for easier migration, as they have different contracts.
- The required contract for toolbar dropdown (naive ui contract) has been preserved to avoid major refactoring.
- Similar styles and animations (as in naive ui) for toolbar dropdowns and tooltips have been preserved.

### DIFF
--- a/packages/super-editor/src/components/SuperEditor.vue
+++ b/packages/super-editor/src/components/SuperEditor.vue
@@ -1,5 +1,4 @@
 <script setup lang="ts">
-import { NSkeleton, useMessage } from 'naive-ui';
 import 'tippy.js/dist/tippy.css';
 import { ref, onMounted, onBeforeUnmount, shallowRef, reactive, markRaw, computed, watch, nextTick } from 'vue';
 import { Editor } from '@superdoc/super-editor';
@@ -9,6 +8,7 @@ import ContextMenu from './context-menu/ContextMenu.vue';
 import { onMarginClickCursorChange } from './cursor-helpers.js';
 import Ruler from './rulers/Ruler.vue';
 import GenericPopover from './popovers/GenericPopover.vue';
+import EditorSkeleton from './EditorSkeleton.vue';
 import LinkInput from './toolbar/LinkInput.vue';
 import TableResizeOverlay from './TableResizeOverlay.vue';
 import ImageResizeOverlay from './ImageResizeOverlay.vue';
@@ -23,6 +23,8 @@ import { DOM_CLASS_NAMES, buildImagePmSelector, buildInlineImagePmSelector } fro
 const emit = defineEmits(['editor-ready', 'editor-click', 'editor-keydown', 'comments-loaded', 'selection-update']);
 
 const DOCX = 'application/vnd.openxmlformats-officedocument.wordprocessingml.document';
+const FILE_LOAD_ERROR_MESSAGE = 'Unable to load the file. Please verify the .docx is valid and not password protected.';
+
 const props = defineProps({
   documentId: {
     type: String,
@@ -264,8 +266,6 @@ const setupRulerObservers = () => {
     rulerContainerResizeObserver.observe(rulerHost);
   }
 };
-
-const message = useMessage();
 
 const editorWrapper = ref(null);
 const editorElem = ref(null);
@@ -780,12 +780,17 @@ const waitForCollaborativeFragmentSettling = async (ydoc, maxWaitMs = 200) => {
   return fragment;
 };
 
+const notifyFileLoadError = () => {
+  console.warn(FILE_LOAD_ERROR_MESSAGE);
+};
+
 const initializeData = async () => {
   // If we have the file, initialize immediately from file
   if (props.fileSource) {
     let fileData = await loadNewFileData();
     if (!fileData) {
-      message.error('Unable to load the file. Please verify the .docx is valid and not password protected.');
+      // TODO: show a visible error to the user (toast removed with naive-ui)
+      notifyFileLoadError();
       await setDefaultBlankFile();
       fileData = await loadNewFileData();
     }
@@ -1196,23 +1201,7 @@ onBeforeUnmount(() => {
       />
     </div>
 
-    <div class="placeholder-editor" v-if="!editorReady">
-      <div class="placeholder-title">
-        <n-skeleton text style="width: 60%" />
-      </div>
-
-      <n-skeleton text :repeat="6" />
-      <n-skeleton text style="width: 60%" />
-
-      <n-skeleton text :repeat="6" style="width: 30%; display: block; margin: 20px" />
-      <n-skeleton text style="width: 60%" />
-      <n-skeleton text :repeat="5" />
-      <n-skeleton text style="width: 30%" />
-
-      <n-skeleton text style="margin-top: 50px" />
-      <n-skeleton text :repeat="6" />
-      <n-skeleton text style="width: 70%" />
-    </div>
+    <EditorSkeleton v-if="!editorReady" />
 
     <GenericPopover
       v-if="activeEditor"
@@ -1288,24 +1277,5 @@ onBeforeUnmount(() => {
   color: initial;
   overflow: hidden;
   position: relative;
-}
-
-.placeholder-editor {
-  position: absolute;
-  top: 0;
-  left: 0;
-  width: 100%;
-  height: 100%;
-  border-radius: 8px;
-  padding: 1in;
-  z-index: 5;
-  background-color: white;
-  box-sizing: border-box;
-}
-
-.placeholder-title {
-  display: flex;
-  justify-content: center;
-  margin-bottom: 40px;
 }
 </style>
--- a/packages/super-editor/src/components/toolbar/ButtonGroup.vue
+++ b/packages/super-editor/src/components/toolbar/ButtonGroup.vue
@@ -1,12 +1,14 @@
 <script setup>
-import { computed, ref, h, watch, onBeforeUnmount } from 'vue';
+import { computed, getCurrentInstance, ref, watch, onBeforeUnmount } from 'vue';
 import ToolbarButton from './ToolbarButton.vue';
 import ToolbarSeparator from './ToolbarSeparator.vue';
 import OverflowMenu from './OverflowMenu.vue';
-import { NDropdown, NTooltip } from 'naive-ui';
+import ToolbarDropdown from './ToolbarDropdown.vue';
+import SdTooltip from './SdTooltip.vue';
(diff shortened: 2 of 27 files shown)


packages/super-editor/src/components/toolbar/ButtonGroup.vue
```diff
@@ -82,8 +117,16 @@ const handleToolbarButtonTextSubmit = (item, argument) => {
...
 const closeDropdowns = () => {
-  if (!currentItem.value) return;
-  currentItem.value.expand = false;
+  const toolbarItems = proxy?.$toolbar?.toolbarItems || [];
+  const overflowItems = proxy?.$toolbar?.overflowItems || [];
+  const allItems = [...toolbarItems, ...overflowItems];
+
+  const itemsToClose = allItems.length ? allItems : props.toolbarItems;
+  itemsToClose.forEach((toolbarItem) => {
+    if (isDropdown(toolbarItem)) {
```
caio-pizzol: `closeDropdowns()` skips overflow items because it only checks for `type === 'dropdown'`. the "..." menu gets stuck open — outside clicks and Escape don't close it. the old code closed everything.

packages/superdoc/src/components/CommentsLayer/CommentsDropdown.vue
```diff
@@ -0,0 +1,223 @@
...
+const menuStyle = computed(() => ({
+  ...props.contentStyle,
+  position: 'fixed',
+  top: menuPosition.value.top,
+  left: menuPosition.value.left,
+  zIndex: 1200,
+}));
+
+const setOpen = (value) => {
+  if (props.disabled) return;
```
caio-pizzol: this guard blocks closing too — if `disabled` becomes true while the menu is open, it gets stuck. only block opening:

```suggestion
  if (props.disabled && value) return;
```

packages/super-editor/src/components/toolbar/ButtonGroup.vue
```diff
@@ -97,11 +140,12 @@ const handleSelect = (item, option) => {
 const dropdownOptions = (item) => {
   if (!item.nestedOptions?.value?.length) return [];
   return item.nestedOptions.value.map((option) => {
+    const isSelected = option?.type !== 'render' && item.selectedValue.value === option.key;
```
caio-pizzol, line 139: the `type !== 'render'` guard here isn't tested. a quick unit test with a render-type option matching `selectedValue` would lock this in. worth adding?

packages/super-editor/src/components/SuperEditor.vue
```diff
@@ -782,12 +782,16 @@ const loadNewFileData = async () => {
...
+  console.warn(FILE_LOAD_ERROR_MESSAGE);
+};
+
 const initializeData = async () => {
   // If we have the file, initialize immediately from file
   if (props.fileSource) {
     let fileData = await loadNewFileData();
     if (!fileData) {
-      message.error('Unable to load the file. Please verify the .docx is valid and not password protected.');
+      notifyFileLoadError();
```
caio-pizzol, line 793: users who load a bad file now silently get a blank document — the old toast is gone and `console.warn` isn't visible. worth a TODO so this doesn't get forgotten:

```suggestion
      // TODO: show a visible error to the user (toast removed with naive-ui)
      notifyFileLoadError();
```

packages/super-editor/src/components/toolbar/ToolbarDropdown.vue
```diff
@@ -0,0 +1,346 @@
...
+  setup(componentProps) {
+    return () => renderIcon(componentProps.option);
+  },
+});
+
+const handlePointerDown = (event) => {
+  if (!isOpen.value) return;
+
+  const target = event.target;
+  if (!(target instanceof HTMLElement)) return;
```
caio-pizzol: SVG elements don't count as `HTMLElement`, so clicking an SVG outside the menu won't close it. same fix needed in `CommentsDropdown.vue:101` and `ButtonGroup.vue:278`:

```suggestion
  if (!(target instanceof Element)) return;
```

packages/superdoc/src/components/HrbrFieldsLayer/CheckboxField.vue
```diff
@@ -43,7 +42,8 @@ const getPreviewStyle = computed(() => {
 
 <template>
   <div class="checkbox-container">
-    <n-checkbox v-if="props.isEditing" :checked="getValue" :disabled="!props.isEditing"></n-checkbox>
+    <!-- TODO: check when migrating CLM -->
+    <div v-if="props.isEditing" class="checkbox-editing-placeholder"></div>
```
caio-pizzol, line 46: this shows an empty box when editing — no checkbox visible, no way to click it. a plain `<input type="checkbox">` would work without any dependencies. is the editing mode still used, or safe to defer?

packages/superdoc/src/components/CommentsLayer/CommentsDropdown.vue
```diff
@@ -0,0 +1,223 @@
...
+}
+
+.comments-dropdown__trigger {
+  display: inline-flex;
+}
+
+.comments-dropdown__menu {
+  min-width: 120px;
+  border-radius: 8px;
+  border: 1px solid #dbdbdb;
```
caio-pizzol: `#dbdbdb` already exists as a design token:

```suggestion
  border: 1px solid var(--sd-border-default);
```

packages/super-editor/src/components/toolbar/ToolbarDropdown.vue
```diff
@@ -0,0 +1,346 @@
...
+.toolbar-dropdown-option.render {
+  padding: 0;
+  cursor: default;
+}
+
+.toolbar-dropdown-option.render:hover {
+  background: transparent;
+}
+
+.fade-in-scale-up-transition-enter-active,
```
caio-pizzol, line 448: the transition CSS (28 lines) is the same in both `ToolbarDropdown` and `SdTooltip`, only `transform-origin` differs. fine for now, but worth pulling into a shared file later.

harbournick (changes requested):
Nice work @artem-harbour! Really clean. The one thing I noticed is that we used to be able to navigate dropdowns with the keyboard up/down arrows (seems to work in linked styles, but not in the others) and also select something just with 'enter'. This is important navigation functionality to retain. Otherwise looks great to me!

packages/superdoc/src/components/HrbrFieldsLayer/CheckboxField.vue
artem-harbour: `CheckboxField.vue` is used only in PDF previews for CLM. As far as I remember, the edit mode and this if branch were never used there (need to double check when we get to that point).

packages/super-editor/src/components/toolbar/ButtonGroup.vue
```diff
@@ -270,18 +326,17 @@ onBeforeUnmount(() => {
...
-        :trigger="item.disabled.value ? null : 'click'"
-        :show="item.expand.value"
+        :disabled="item.disabled.value"
+        :show="getExpanded(item)"
         :content-style="{ fontFamily: props.uiFontFamily }"
-        size="medium"
         placement="bottom-start"
-        class="toolbar-button toolbar-dropdown sd-editor-toolbar-dropdown"
+        class="toolbar-button sd-editor-toolbar-dropdown"
         :class="{ 'high-contrast': isHighContrastMode }"
```
caio-pizzol: the `high-contrast` class is still applied but the new dropdown has no styles for it. the old black-background-on-hover rules were removed with naive-ui and not replaced. high-contrast users now see the same light gray as everyone else. worth adding matching styles for the new dropdown?

packages/super-editor/src/components/toolbar/ToolbarDropdown.vue
```diff
@@ -0,0 +1,465 @@
+<script setup>
```
caio-pizzol, line 1: `ToolbarDropdown` and `CommentsDropdown` copy-paste the same open/close/position/click-outside logic (~60 lines). worth pulling into a shared helper so bug fixes only happen once?

packages/super-editor/src/components/toolbar/ButtonGroup.vue
```diff
@@ -67,11 +69,49 @@ const isButton = (item) => item.type === 'button';
 const isDropdown = (item) => item.type === 'dropdown';
 const isSeparator = (item) => item.type === 'separator';
 const isOverflow = (item) => item.type === 'overflow';
+
+const getExpanded = (item) => {
```
caio-pizzol, line 73: `item.expand` is always a ref — the plain-value branch here never runs. safe to just use `item.expand.value` directly.

packages/super-editor/src/components/toolbar/ToolbarDropdown.vue
artem-harbour: I prefer to keep this logic separate within dropdown components rather than putting it into shared helpers (at least for now).

artem-harbour:
@harbournick @caio-pizzol - all comments were considered, please review and merge to unblock further work.

packages/superdoc/src/components/CommentsLayer/CommentsDropdown.vue
```diff
@@ -0,0 +1,223 @@
...
+          class="comments-dropdown__option"
+          :class="{ disabled: option.disabled }"
+          @click="onOptionClick(option)"
+        >
+          <span v-if="hasIcon(option)" class="comments-dropdown__option-icon">
+            <span v-if="option.iconString" v-html="option.iconString"></span>
+            <OptionIcon v-else :option="option" />
+          </span>
+          <!-- Keep legacy class for click-outside checks in CommentDialog/SuperDoc -->
+          <span class="comments-dropdown__option-label n-dropdown-option-body__label">{{ option.label }}</span>
```
caio-pizzol: `n-dropdown-option-body__label` is a naive-ui class name — now that naive-ui is gone, rename it to something like `comments-dropdown__option-label`. also update the references in `SuperDoc.vue:213`, `CommentDialog.vue:304`, and `edit-comment-text.spec.ts:24`. this PR already touches all those files so it's the right time.

artem-harbour:
@caio-pizzol, all comments are addressed but merging is blocked (needs one more approve).

fyi @harbournick @VladaHarbour @luccas-harbour @tupizz

caio-pizzol:
merged!

harbournick:

PR #2241, opened 2026-03-02
fix(comments): prevent comment mark from extending to adjacent typed text

tupizz:
## Summary

- Set `inclusive: false` on `CommentsMark` so newly typed text at the boundary of a comment range does not inherit the comment mark
- Added behavior tests verifying the mark boundary is respected on both sides (before and after) across all three browsers

## Context

[SD-2029](https://linear.app/superdocworkspace/issue/SD-2029): When typing immediately after text with a comment anchor, the new text was incorrectly associated with the comment. The comment highlight "leaked" to cover newly typed characters.

**Root cause**: ProseMirror marks default to `inclusive: true`, which extends the mark to adjacent typed text. The `CommentsMark` extension was missing `inclusive: false`. The test schemas already used `inclusive: false`, so unit tests passed while the runtime had the wrong behavior.

**Fix**: One-line addition matching the pattern already used by `Link`, `TrackInsert`, `TrackDelete`, and `TrackFormat` marks.

## Test plan

- [x] All 7079 unit tests pass (`pnpm --filter super-editor test`)
- [x] Behavior test passes on Chromium, Firefox, and WebKit (headed and headless)
- [x] Smoke tests pass (page load, text input, undo/redo, bold, scroll)
- [x] Browser reproduction confirmed: comment highlight stays on original text only


caio-pizzol (approved):
@tupizz clean fix, lgtm. tests cover both sides. ship it.

harbournick:

PR #2242, opened 2026-03-02
fix(comments): improve multiline comment input styling

tupizz:
## Summary
- Replace pill shape (`border-radius: 9999px`) with rounded rect (`12px`) on comment and reply input wrappers to avoid the distorted capsule look on multiline text
- Add `max-height: 150px` with `overflow-y: auto` so long replies scroll instead of stretching the card unboundedly

## Test plan
- [ ] Open a comment, type multiline text in the new comment input. Verify it scrolls at ~150px and corners look clean
- [ ] Reply to an existing comment with multiline text. Same behavior
- [ ] Single-line input still looks correct with 12px rounded corners


caiopizzol (approved):
@tupizz clean fix, looks good. verified the scroll behavior works correctly and doesn't clip the mention dropdown.

one optional follow-up: the two input wrapper classes share almost all their CSS — could consolidate into a shared selector to reduce duplication, but not blocking.

harbournick:

PR #2253, opened 2026-03-03
fix(track-changes): correct format change description for already-formatted text (SD-2077)

tupizz:
## Summary

Fixes the tracked change comment showing incorrect "removed" descriptions when adding formatting to text that already has different formatting in suggesting mode.

**Example:** Text has highlight applied. User switches to suggesting mode and bolds it. The tracked change comment said *"Format: bold, removed highlight"* when it should say *"Format: bold"*. The highlight was never removed.

## Root cause

In `addMarkStep.js`, when a format mark is added in suggesting mode, the track changes system records a `before`/`after` snapshot to describe what changed. The display logic diffs these arrays: marks in `before` but not `after` show as "removed", marks in `after` but not `before` show as "added".

The bug was in the `else` branch (no existing `trackFormat` mark on the text): the code dumped **all** existing non-tracking marks into `before`, regardless of whether they were related to the change. So when adding bold to highlighted text, highlight ended up in `before` and bold in `after`, making it look like highlight was removed and bold was added.

```js
// Before (bug): captures ALL marks as "before" state
before = liveMarks
  .filter((mark) => ![TrackDeleteMarkName, TrackFormatMarkName].includes(mark.type.name))
  .map((mark) => ({ type: mark.type.name, attrs: { ...mark.attrs } }));
```

## Fix

Only capture a mark of the **same type** as the one being added into `before`. This correctly handles:

- **Pure additions** (bold on highlighted text): `before = []`, `after = [bold]` → shows "bold"
- **Type replacements** (changing textStyle color): `before = [textStyle{old}]`, `after = [textStyle{new}]` → shows "color"
- **No false removals**: unrelated marks (highlight, italic, etc.) are never included in `before`

```js
// After (fix): only captures same-type mark being replaced
const existingMarkOfSameType = liveMarks.find(
  (mark) =>
    mark.type.name === step.mark.type.name &&
    ![TrackDeleteMarkName, TrackFormatMarkName].includes(mark.type.name),
);
before = existingMarkOfSameType
  ? [{ type: existingMarkOfSameType.type.name, attrs: { ...existingMarkOfSameType.attrs } }]
  : [];
```

This matches the pattern already used by `removeMarkStep.js`, which correctly scopes `before` to only the specific mark being removed.

## Why this approach

The `before`/`after` arrays represent the **diff**, not the full state. They should only contain marks that are actually changing. The `removeMarkStep` handler already follows this principle (it only puts the removed mark in `before`). This fix aligns `addMarkStep` with the same semantic.

Alternatives considered:
- Fixing at the display layer (`translateFormatChangesToEnglish`): would only mask the symptom. The `before`/`after` data would still be wrong, affecting any future consumer of these arrays (e.g., accept/reject, collaboration sync).
- Filtering unrelated marks in `before` by comparing against `after`: more complex and fragile. The root issue is that unrelated marks should never enter `before` in the first place.

## Test plan

- [x] Unit test: adding bold to highlighted text → `before` is empty, `after` has bold
- [x] Unit test: changing textStyle on text with highlight → `before` has only textStyle, not highlight
- [x] Behavior test (Chromium/Firefox/WebKit): bold on highlighted text shows "Format: bold", no "removed"
- [x] Behavior test: italic on bold text shows "Format: italic", no "removed bold"
- [x] Behavior test: color change on highlighted text shows "Format: color", no "removed highlight"
- [x] All 736 existing unit test files pass
- [x] All 41 existing comment/track-change behavior tests pass

### DIFF
--- a/tests/behavior/tests/comments/sd-2077-format-change-display.spec.ts
+++ b/tests/behavior/tests/comments/sd-2077-format-change-display.spec.ts
@@ -1,6 +1,6 @@
 import type { Page } from '@playwright/test';
 import { test, expect } from '../../fixtures/superdoc.js';
-import { assertDocumentApiReady, listTrackChanges } from '../../helpers/document-api.js';
+import { assertDocumentApiReady } from '../../helpers/document-api.js';
 
 test.use({ config: { toolbar: 'full', comments: 'on', trackChanges: true } });
 
@@ -10,6 +10,14 @@ async function runCommands(page: Page, commands: EditorCommand[]): Promise<void>
   for (const [name, ...args] of commands) {
     await page.evaluate(({ name, args }) => (window as any).editor.commands[name](...args), { name, args });
   }
+}
+
+async function expectTrackedFormatDialog(page: Page) {
+  const dialog = page.locator('.comment-placeholder .comments-dialog', {
+    has: page.locator('.tracked-change-text'),
+  });
+  await expect(dialog).toBeVisible({ timeout: 10_000 });
+  return dialog;
 }
 
 test.describe('SD-2077 tracked format change displays correct description', () => {
@@ -36,10 +44,7 @@ test.describe('SD-2077 tracked format change displays correct description', () =
     await superdoc.assertTrackedChangeExists('format');
 
     // Wait for the tracked change comment dialog to appear
-    const dialog = superdoc.page.locator('.comment-placeholder .comments-dialog', {
-      has: superdoc.page.locator('.tracked-change-text'),
-    });
-    await expect(dialog).toBeVisible({ timeout: 10_000 });
+    const dialog = await expectTrackedFormatDialog(superdoc.page);
 
     // The format description should mention bold but NOT mention highlight removal
     const formatText = dialog.locator('.tracked-change-text');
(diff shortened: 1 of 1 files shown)


tests/behavior/tests/comments/sd-2077-format-change-display.spec.ts
```diff
@@ -0,0 +1,135 @@
+import type { Page } from '@playwright/test';
+import { test, expect } from '../../fixtures/superdoc.js';
+import { assertDocumentApiReady, listTrackChanges } from '../../helpers/document-api.js';
```
caio-pizzol: unused import.

```suggestion
import { assertDocumentApiReady } from '../../helpers/document-api.js';
```

tests/behavior/tests/comments/sd-2077-format-change-display.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    // Select text and apply bold
+    await superdoc.selectAll();
+    await superdoc.bold();
+    await superdoc.waitForStable();
+
+    // Verify tracked format change exists
+    await superdoc.assertTrackedChangeExists('format');
+
+    // Wait for the tracked change comment dialog to appear
+    const dialog = superdoc.page.locator('.comment-placeholder .comments-dialog', {
```
caio-pizzol: this dialog lookup + visibility check shows up 3 times — could pull it into a small helper at the top of the file. worth it?

harbournick: fixed in https://github.com/superdoc-dev/superdoc/pull/2253/commits/de42e1facfdb7ee63d90298afcd4236caf7a7342

tests/behavior/tests/comments/sd-2077-format-change-display.spec.ts
harbournick: https://github.com/superdoc-dev/superdoc/pull/2253/commits/de42e1facfdb7ee63d90298afcd4236caf7a7342

harbournick:

PR #2254, opened 2026-03-03
fix(super-converter): handle empty pic:spPr in image import (SD-2085)

caio-pizzol:
Guard against undefined `spPr.elements`` when pic:spPr` is a self-closing empty element (e.g. `<pic:spPr/>`). This is valid OOXML per ECMA-376 §20.2.2.6 — all `CT_ShapeProperties` children are optional. 

Tools like docx-templates v4.15.0 generate this pattern, causing a `TypeError` that silently breaks all image imports in the document.


harbournick:

PR #2259, opened 2026-03-03
fix(comments): resolve double-click activation and edit mode issues (SD-2035)

tupizz:
## Demo

https://github.com/user-attachments/assets/c74a7d49-a898-4033-b6e5-2b940a0227cb

## Summary

Fixes multiple comment UI issues reported in [SD-2035](https://linear.app/superdocworkspace/issue/SD-2035):

- **Double-click required to activate comments**: Clicking a comment card in the sidebar required two clicks. Root cause: `setFocus()` dispatched two separate PM transactions (one for `setActiveComment`, one for `setCursorById`). The second transaction triggered position-based comment detection in the plugin's `apply()`, which returned `null` (cursor lands on `commentRangeStart` marker, not inside comment-marked text) and cleared the active thread.
- **Reply input stays open after sending**: After submitting a reply, the input field remained visible because `isReplying` was only reset indirectly via the `isActiveComment` watcher, which no longer fired after the `removePendingComment` fix.
- **"N more replies" double-click**: Expanding collapsed threads also required double-click. Root cause: `view.focus()` on the hidden PM view triggered a DOM selection sync transaction that overwrote the `activeThreadId`.
- **Edit mode UI inconsistencies**: Edit comment form used different styles than the reply form. Overflow menu "Edit" activated the wrong comment ID (child instead of parent thread).
- **`removePendingComment` clearing active thread**: `removePendingComment` unconditionally set `activeComment = null`, deactivating threads when called from reply/edit flows (not just new comment cancellation).

### `comments-plugin.js` (super-editor)
- **`setCursorById` command**: Accepts optional `{ activeCommentId }` parameter. When provided, sets `setActiveComment` meta on the same transaction as the cursor move, preventing the two-transaction race. Skips `view.focus()` for sidebar-initiated activations to avoid focus-induced DOM selection sync.
- **Plugin `apply()` method**: Added `changedActiveThread` suppression flag. After an explicit `setActiveComment` sets this flag to `true`, the next transaction's position-based detection is suppressed, preventing residual focus transactions from clearing the active thread. Flag resets to `false` after one transaction.

### `CommentDialog.vue` (superdoc)
- **`setFocus()`**: Rewritten to combine cursor move and active comment activation into a single PM transaction via `setCursorById(id, { activeCommentId })`.
- **`handleAddComment()`**: Explicitly resets `isReplying = false` after sending a reply.
- **`handleOverflowSelect()`**: Uses `props.comment.commentId` (parent thread) instead of child comment ID when activating during edit.
- **`isEditingAnyComment` computed**: New computed that checks if any comment in the current thread is being edited. Used by `showButtons` and `showInputSection` to hide reply UI during edits.
- **Edit template**: Uses `reply-expanded` CSS classes for consistent styling with the reply form.
- **CSS**: Added `overflow-wrap: break-word` on `.comments-dialog`, removed unused `.comment-footer` and `.comment-editing` styles.

### `FloatingComments.vue` (superdoc)
- **`editingCommentId` watcher**: Re-measures all visible dialog heights when entering/exiting edit mode, since the edit form changes the dialog dimensions. Includes timer cleanup to prevent stale measurements.

### `comments-store.js` (superdoc)
- **`removePendingComment()`**: Only clears `activeComment` when there was an actual pending comment (`hadPending` guard). Replies and edits call this to reset `currentCommentText` but no longer inadvertently deactivate the thread.

### `CommentDialog.test.js` (superdoc)
- Updated assertions to match the combined `setCursorById` call with `{ activeCommentId }` parameter.
- Updated edit overflow test to expect parent thread activation.

## Test plan

- [x] Unit tests pass (7766 total: 7529 super-editor + 237 superdoc)
- [x] Click a comment card in sidebar: activates on single click (no double-click)
- [x] Click "N more replies" to expand thread: activates on single click
- [x] Send a reply: input field closes, comment stays active
- [x] Edit a comment via overflow menu: edit form appears with consistent styling
- [x] Cancel edit: returns to normal view, comment stays active
- [x] Click away from active comment: deactivates normally
- [x] Create new comment, cancel: comment is removed and deactivated
- [ ] Resolved comments: clicking scrolls to location without activating


packages/super-editor/src/extensions/comment/comments-plugin.js
```diff
@@ -492,9 +503,13 @@ export const CommentsPlugin = Extension.create({
...
 
-          // Check for changes in the actively selected comment
+          // Check for changes in the actively selected comment.
+          // Skip position-based detection if the previous transaction explicitly set the
+          // active comment (changedActiveThread flag). This prevents focus-induced DOM
+          // selection sync transactions from overriding a sidebar-initiated activation.
           const trChangedActiveComment = meta?.type === 'setActiveComment';
-          if ((!tr.docChanged && tr.selectionSet) || trChangedActiveComment) {
+          const suppressPositionDetection = pluginState.changedActiveThread && !trChangedActiveComment;
+          if ((!tr.docChanged && tr.selectionSet && !suppressPositionDetection) || trChangedActiveComment) {
```
caio-pizzol: after a sidebar click, `changedActiveThread` stays `true` because decoration updates (lines 486-491) copy it forward without resetting it. the next time the user clicks in the editor, that click's comment detection gets skipped.

you already prevent the focus race by skipping `view.focus()` at line 423, so this flag isn't needed — consider removing it. if you want to keep it as a safety net, reset it in the decoration early-return too.

packages/superdoc/src/components/CommentsLayer/CommentDialog.vue
```diff
@@ -280,20 +275,20 @@ const hasTextContent = computed(() => {
...
+  // plugin clears the activeThreadId before the setActiveComment meta is processed.
   if (editor) {
-    // For resolved comments, use commentId since prepareCommentsForImport rewrites
-    // commentRangeStart/End nodes' w:id to the internal commentId (not importedId)
     const cursorId = props.comment.resolvedTime
       ? props.comment.commentId
       : props.comment.importedId || props.comment.commentId;
-    editor.commands?.setCursorById(cursorId);
+    const activeCommentId = !props.comment.resolvedTime ? props.comment.commentId : null;
+    editor.commands?.setCursorById(cursorId, { activeCommentId });
```
caio-pizzol: if `setCursorById` can't find the comment's position (returns `false`), the active comment never gets set in the editor — but the sidebar already shows it as active (line 280). the old code called `setActive()` separately so the editor always stayed in sync. worth a fallback?

```js
const didScroll = editor.commands?.setCursorById(cursorId, { activeCommentId });
if (!didScroll && activeCommentId) {
  editor.commands?.setActiveComment({ commentId: activeCommentId });
}
```

packages/superdoc/src/components/CommentsLayer/CommentDialog.vue
```diff
@@ -82,12 +82,7 @@ const isPendingNewComment = computed(() => {
 });
 
 const showButtons = computed(() => {
```
caio-pizzol: `showButtons` is never used in the template — `showInputSection` does the same thing. safe to remove?

packages/super-editor/src/extensions/comment/comments-plugin.js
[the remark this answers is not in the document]
tupizz: Good catch. Removed the `changedActiveThread` flag entirely in 9fb331e60. Since `view.focus()` is now skipped for sidebar activations (line 423), the flag is unnecessary. The flag also had a bug: it persisted through the decoration early-return path (lines 486-491) and was never consumed, causing the next real editor click to be silently ignored.

packages/superdoc/src/stores/comments-store.js
```diff
@@ -594,11 +594,18 @@ export const useCommentsStore = defineStore('comments', () => {
...
     pendingComment.value = null;
-    activeComment.value = null;
     superdocStore.selectionPosition = null;
 
+    // Only clear active comment when removing an actual pending comment.
+    // Replies and edits also call this to reset currentCommentText, but
+    // clearing activeComment would deactivate the thread (SD-2035).
+    if (hadPending) {
+      activeComment.value = null;
+    }
```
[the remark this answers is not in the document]
tupizz: Valid. Fixed in 9fb331e60 — added `isReplying.value = false` in `handleCancel()`. The old code relied on `removePendingComment` clearing `activeComment` to trigger the watcher that reset `isReplying`. With the `hadPending` guard, that side effect is gone for reply cancels, so we reset it explicitly.

packages/super-editor/src/extensions/comment/comments-plugin.js
tupizz: Agreed — removed the flag entirely in 9fb331e60. The `view.focus()` skip is the real fix. The flag had the persistence bug you described (decoration early-return copies it forward without resetting), and since there's no synthetic focus transaction to consume it, it just stays `true` and eats the next real click.

packages/superdoc/src/components/CommentsLayer/CommentDialog.vue
tupizz: Good call. Added the fallback in 9fb331e60 — when `setCursorById` returns false (position not found), we now call `setActiveComment` separately to keep the editor in sync with the Vue store. Also updated the test mock to return `true` for the happy path and added `setActiveComment` to the command stub.

packages/superdoc/src/components/CommentsLayer/CommentDialog.vue
tupizz: Confirmed dead code — removed in 9fb331e60. `showInputSection` has identical logic and is the one used in the template.

packages/super-editor/src/extensions/comment/comments-plugin.js
```diff
@@ -513,7 +522,7 @@ export const CommentsPlugin = Extension.create({
             }
           }
 
-          return pluginState;
+          return { ...pluginState };
```
caio-pizzol, line 536: leftover from removing `changedActiveThread` — nothing checks if this object changed by reference, so the copy isn't needed.

```suggestion
          return pluginState;
```

packages/superdoc/src/components/CommentsLayer/CommentDialog.vue
```diff
@@ -455,6 +451,7 @@ const getProcessedDate = (timestamp) => {
 
 const handleCancel = (comment) => {
```
caio-pizzol, line 479: `comment` isn't used in the function body anymore. safe to drop?

```suggestion
const handleCancel = () => {
```

harbournick:

PR #2271, opened 2026-03-04
fix: add type definitions for PresentationEditor

mattConnHarbour:


packages/super-editor/src/index.d.ts
```diff
@@ -315,13 +453,288 @@ export declare class SuperToolbar {
...
+   */
+  getPaintSnapshot(): any | null;
+
+  /**
+   * Get section-aware page styles.
+   */
+  getCurrentSectionPageStyles(): {
+    pageSize: PageSize;
+    margins: PageMargins;
+  };
```
caio-pizzol, line 777: this doesn't match what the method actually returns — `margins` should be `pageMargins`, `w`/`h` should be `width`/`height`, and `sectionIndex` + `orientation` are missing.

```suggestion
  getCurrentSectionPageStyles(): {
    pageSize: { width: number; height: number };
    pageMargins: { left: number; right: number; top: number; bottom: number };
    sectionIndex: number;
    orientation: 'portrait' | 'landscape';
  };
```

packages/super-editor/src/index.d.ts
```diff
@@ -158,6 +158,144 @@ export interface OpenOptions {
...
+  pageStyles?: Record<string, unknown>;
+  debugLabel?: string;
+  layoutMode?: LayoutMode;
+  trackedChanges?: TrackedChangesOverrides;
+  emitCommentPositionsInViewing?: boolean;
+  enableCommentsInViewing?: boolean;
+  presence?: PresenceOptions;
+  ruler?: {
+    enabled?: boolean;
+    showMarginHandles?: boolean;
```
caio-pizzol: `showMarginHandles` doesn't exist — the real property is `interactive`, and there's also an `onMarginChange` callback. check `RulerOptions` in `renderer.ts:301`.

packages/super-editor/src/index.d.ts
```diff
@@ -315,13 +453,288 @@ export declare class SuperToolbar {
...
+
+  /**
+   * Set the layout mode (vertical or horizontal).
+   */
+  setLayoutMode(mode: LayoutMode): void;
+
+  /**
+   * Hit test at client coordinates to find document position.
+   */
+  hitTest(clientX: number, clientY: number): { pos: number; inside: number } | null;
```
caio-pizzol: looks like this was copied from `posAtCoords` — `hitTest` returns `{ pos, layoutEpoch, blockId, pageIndex, column, lineIndex }`, no `inside` field.

packages/super-editor/src/index.d.ts
```diff
@@ -315,13 +453,288 @@ export declare class SuperToolbar {
...
+
+  /**
+   * Convert a document range into layout-based bounding rects.
+   */
+  getRangeRects(from: number, to: number, relativeTo?: HTMLElement): RangeRect[];
+
+  /**
+   * Get the bounding rect for the current selection.
+   */
+  getSelectionBounds(relativeTo?: HTMLElement): {
```
caio-pizzol: the `{ top, left, bottom, right, width, height }` shape shows up in both `getSelectionBounds` and `getCommentBounds` — worth extracting a shared type?

packages/super-editor/src/index.d.ts
```diff
@@ -315,13 +453,288 @@ export declare class SuperToolbar {
...
+
+  /**
+   * Subscribe to layout error events. Returns an unsubscribe function.
+   */
+  onLayoutError(handler: (error: LayoutError) => void): () => void;
+
+  /**
+   * Get the rendered pages.
+   */
+  getPages(): any[];
```
caio-pizzol: `getPages()`, `getLayoutSnapshot()`, `getPaintSnapshot()`, and `onLayoutUpdated()` all have real types in the source — using them instead of `any` would give consumers better autocomplete.

packages/super-editor/src/index.d.ts
```diff
@@ -315,13 +564,289 @@ export declare class SuperToolbar {
...
+    x: number;
+    y: number;
+    pageIndex?: number;
+    pageLocalY?: number;
+  } | null;
+
+  /**
+   * Get viewport coordinates at a document position.
+   */
+  coordsAtPos(pos: number, side?: number): { left: number; right: number; top: number; bottom: number } | undefined;
```
caio-pizzol: three things off here: `width` and `height` are missing, it returns `null` not `undefined`, and `side` param doesn't exist.

```suggestion
  coordsAtPos(pos: number): { left: number; right: number; top: number; bottom: number; width: number; height: number } | null;
```

packages/super-editor/src/index.d.ts
```diff
@@ -315,13 +564,289 @@ export declare class SuperToolbar {
...
+
+  /**
+   * Scroll to a specific page number.
+   */
+  scrollToPage(pageNumber: number, scrollBehavior?: ScrollBehavior): Promise<boolean>;
 
+  /**
+   * Get document position at viewport coordinates.
+   */
+  posAtCoords(coords: { left: number; top: number }): { pos: number; inside: number } | null;
```
caio-pizzol: the real method also accepts `clientX`/`clientY` (all optional) — the test on line 1734 already uses that form.

```suggestion
  posAtCoords(coords: { left?: number; top?: number; clientX?: number; clientY?: number }): { pos: number; inside: number } | null;
```

harbournick:

PR #2287, opened 2026-03-04
fix: only include inline pPr on export

VladaHarbour:
Exclude inherited props from document.xml on export

### DIFF
--- a/.github/workflows/ci-superdoc.yml
+++ b/.github/workflows/ci-superdoc.yml
@@ -27,7 +27,7 @@ concurrency:
 
 jobs:
   validate:
-    runs-on: superdoc-ci-4cpu
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
 


packages/super-editor/src/core/super-converter/export-helpers/run-properties-export.js
```diff
@@ -0,0 +1,88 @@
...
+    if (rPr?.elements?.length) chain.push(rPr);
+    const basedOn = styleTag.elements?.find((el) => el.name === 'w:basedOn');
+    currentId = basedOn?.attributes?.['w:val'];
+  }
+
+  if (chain.length === 0) return {};
+
+  // Merge rPr elements: base first, then derived (later overrides by element name)
+  const byName = {};
+  chain.forEach((rPr) => {
```
caio-pizzol: the merge loop processes derived styles first, then base styles — so base wins. it should be the other way around. add `chain.reverse()` before the loop.

packages/super-editor/src/core/super-converter/v3/handlers/w/tc/helpers/translate-table-cell.js
```diff
@@ -121,6 +123,18 @@ export function generateTableCellProperties(node) {
     }
   }
 
+  // Only output tcPr keys that were present in the cell's w:tcPr (exclude inherited from table style). When inlineKeys was not set (e.g. new cell), do not filter.
+  if (Array.isArray(inlineKeys)) {
+    if (inlineKeys.length === 0) {
+      tableCellProperties = {};
+    } else {
+      const allowed = new Set(inlineKeys);
```
caio-pizzol: this drops any cell property that wasn't in the original docx. but if a user adds a background color or changes alignment after import, those keys aren't in the allow-list and get silently removed on export. the border command handles this correctly (`table.js:1253`) — other editing commands need the same treatment.

packages/super-editor/src/extensions/run/calculateInlineRunPropertiesPlugin.js
```diff
@@ -117,14 +117,35 @@ export const calculateInlineRunPropertiesPlugin = (editor) =>
...
         }
 
+        const existingInlineKeys = runNode.attrs?.runPropertiesInlineKeys || [];
+        const keysFromMarks = (segment) => {
+          const textNode = segment.content?.find((n) => n.isText);
+          return Object.keys(decodeRPrFromMarks(textNode?.marks || []));
+        };
+
         if (segments.length === 1) {
           if (JSON.stringify(runProperties) === JSON.stringify(runNode.attrs.runProperties)) return;
```
caio-pizzol: when run properties haven't changed, this returns early without setting `runPropertiesInlineKeys`. if any existing collab docs were saved before this PR, they'd have null inline keys — and export would drop all their formatting. is that a real scenario?

packages/super-editor/src/core/super-converter/v3/handlers/w/r/r-translator.js
```diff
@@ -127,9 +147,25 @@ const encode = (params, encodedAttrs = {}) => {
 
   const filtered = contentWithRunMarks.filter(Boolean);
 
+  // Keys from the run's style (styleId) in styles.xml — don't export these (already in styles.xml)
+  let runPropertiesStyleKeys = null;
+  if (runProperties?.styleId && params?.docx) {
+    const styleRPr = getParagraphStyleRunPropertiesFromStylesXml(params.docx, runProperties.styleId, params);
+    if (styleRPr && Object.keys(styleRPr).length > 0) {
+      runPropertiesStyleKeys = Object.keys(styleRPr);
```
caio-pizzol, line 232: `runPropertiesStyleKeys` is saved during import but never used during export — export recomputes it from scratch. either use the saved value or remove it.

packages/layout-engine/style-engine/src/ooxml/index.ts
```diff
@@ -153,6 +153,30 @@ export function resolveRunProperties(
   return finalProps;
 }
 
+/**
+ * Same as resolveRunProperties but also returns which keys came from the inline rPr (so export can output only those).
+ * Used by the converter to mark inline properties at the combine step and export only inline.
+ */
+export function resolveRunPropertiesWithInlineFlag(
```
caio-pizzol: this wrapper just adds an `Object.keys()` call on top of `resolveRunProperties`. only one place uses it. worth inlining at the call site instead?

packages/super-editor/src/core/super-converter/v3/handlers/w/tc/helpers/translate-table-cell.js
VladaHarbour: The issue was related to tcMar only to be honest and we already treat it above. Removed this check

packages/super-editor/src/core/super-converter/export-helpers/run-properties-export.js
VladaHarbour: added

packages/super-editor/src/core/super-converter/v3/handlers/w/r/r-translator.js
VladaHarbour: fixed

packages/layout-engine/style-engine/src/ooxml/index.ts
VladaHarbour: made it inline

packages/super-editor/src/extensions/run/calculateInlineRunPropertiesPlugin.js
VladaHarbour: added condition to support existing collab docs

packages/super-editor/src/core/super-converter/v3/handlers/w/r/r-translator.js
```diff
@@ -216,6 +274,21 @@ const decode = (params, decodedAttrs = {}) => {
 
   const runAttrs = runNodeForExport.attrs || {};
   const runProperties = runAttrs.runProperties || {};
+  const inlineKeys = runAttrs.runPropertiesInlineKeys;
+  const styleKeys = runAttrs.runPropertiesStyleKeys;
+  const overrideKeys = runAttrs.runPropertiesOverrideKeys;
+
+  // Export run properties that were inline or that override the style (so user overrides are preserved).
+  // Exclude keys that are style-only (in styleKeys but not in overrideKeys).
+  const candidateKeys = [...new Set([...(inlineKeys || []), ...(overrideKeys || [])])];
```
caio-pizzol: the collab guard you added helps when users edit text, but if someone opens an old doc and saves without editing anything, `inlineKeys` is still null here — so nothing gets exported. all formatting gone.

```suggestion
    const candidateKeys = inlineKeys != null
      ? [...new Set([...(inlineKeys || []), ...(overrideKeys || [])])]
      : Object.keys(runProperties);
```

packages/super-editor/src/core/super-converter/v3/handlers/w/p/helpers/generate-paragraph-properties.js
```diff
@@ -12,6 +12,25 @@ export function generateParagraphProperties(params) {
   const { attrs = {} } = node;
 
   const paragraphProperties = carbonCopy(attrs.paragraphProperties || {});
+
+  // Only include w:rPr in pPr when the paragraph had inline rPr on import; filter to inline keys and drop if empty.
+  const inlineKeys = paragraphProperties.runPropertiesInlineKeys;
+  delete paragraphProperties.runPropertiesInlineKeys;
+  if (!Array.isArray(inlineKeys) || inlineKeys.length === 0) {
```
caio-pizzol: same thing for paragraphs — old docs hit this and lose their default formatting on save.

packages/super-editor/src/extensions/run/calculateInlineRunPropertiesPlugin.js
```diff
@@ -94,14 +94,42 @@ export const calculateInlineRunPropertiesPlugin = (editor) =>
...
+          return Object.keys(decodeRPrFromMarks(textNode?.marks || []));
+        };
+        const overrideKeysFromInlineProps = (inlineProps) => styleKeys.filter((k) => inlineProps && k in inlineProps);
         if (segments.length === 1) {
-          if (JSON.stringify(runProperties) === JSON.stringify(runNode.attrs.runProperties)) return;
-          tr.setNodeMarkup(mappedPos, runNode.type, { ...runNode.attrs, runProperties }, runNode.marks);
+          const hadInlineKeys =
+            Array.isArray(runNode.attrs?.runPropertiesInlineKeys) && runNode.attrs.runPropertiesInlineKeys.length > 0;
+          if (JSON.stringify(runProperties) === JSON.stringify(runNode.attrs.runProperties) && hadInlineKeys) return;
+          const newInlineKeys = [...new Set([...existingInlineKeys, ...keysFromMarks(segments[0])])];
```
caio-pizzol, line 118: the guard works for the early-return case, but `keysFromMarks` only finds mark-based properties. things like character styles, language, and text direction are in `runProperties` but never get added to the allow-list — so they're dropped on export.

VladaHarbour: I intentionally avoided unioning all runProperties keys in the plugin because it duplicated inherited props and blew up XML. All inline runProperties will get to the allow-list(allow-list) on import

packages/super-editor/src/core/super-converter/v3/handlers/w/r/r-translator.js
```diff
@@ -96,12 +97,29 @@ const ensureReferenceRunFormatting = (runNode, referenceXmlName) => {
...
+  runPropertiesOverrideKeys = null,
+) => {
   const node = {
     type: SD_KEY_NAME,
     content,
-    attrs: { ...encodedAttrs, runProperties },
+    attrs: {
+      ...encodedAttrs,
+      runProperties: resolvedRunProperties,
+      runPropertiesInlineKeys: inlineKeysFromCombine?.length ? inlineKeysFromCombine : null,
```
caio-pizzol: i tested this — runs without `w:rPr` get `[]` here, which becomes `null`. the plugin fixes it before export, so no inflation in practice. but `null` now means two things: "empty" and "unknown." keeping `[]` would separate them:

```suggestion
      runPropertiesInlineKeys: inlineKeysFromCombine,
```

packages/super-editor/src/document-api-adapters/plan-engine/executor.ts
```diff
@@ -529,9 +529,19 @@ function applyRunAttributePatch(
     const currentRunProperties = isRecord(runNode.attrs?.runProperties)
       ? { ...(runNode.attrs.runProperties as Record<string, unknown>) }
       : {};
+    const currentInlineKeys = Array.isArray(runNode.attrs?.runPropertiesInlineKeys)
```
caio-pizzol: this doesn't go through the plugin, so when an old doc has `null` here it becomes `[]` and only the changed key survives — everything else falls off. the `rtl` in your test (line 2228) shows this. seeding from what's already there would keep them:

```js
const currentInlineKeys = Array.isArray(runNode.attrs?.runPropertiesInlineKeys)
  ? runNode.attrs.runPropertiesInlineKeys
  : Object.keys(currentRunProperties);
```

packages/super-editor/src/extensions/run/calculateInlineRunPropertiesPlugin.test.js
```diff
@@ -166,6 +169,40 @@ describe('calculateInlineRunPropertiesPlugin', () => {
     getResolvedParagraphPropertiesMock.mockImplementation(() => null);
```
caio-pizzol, line 169: your explanation on not adding all keys makes sense. this test covers docs where keys are already set. worth adding one for old docs (`null` keys, run has `lang` + `bold`, user toggles italic) — just to lock down what happens there.

packages/super-editor/src/core/super-converter/v3/handlers/w/r/r-translator.js
```diff
@@ -257,6 +316,25 @@ const decode = (params, decodedAttrs = {}) => {
...
+  const overrideKeys = runAttrs.runPropertiesOverrideKeys;
+
+  // Export run properties that were inline or that override the style (so user overrides are preserved).
+  // Exclude keys that are style-only (in styleKeys but not in overrideKeys).
+  // Old collaboration payloads often omit runPropertiesInlineKeys — fall back to all keys on runProperties
+  // so those documents still round-trip formatting (accepts larger document.xml vs strict allow-list only).
+  const candidateKeys =
+    inlineKeys != null ? [...new Set([...(inlineKeys || []), ...(overrideKeys || [])])] : Object.keys(runProperties);
+
+  const exportKeys = candidateKeys.filter(
```
caio-pizzol: this filter decides what goes into the exported file and it's doing a lot in one line. a named function would make it easier to follow:

```js
const shouldExport = (k) =>
  k in (runProperties || {}) &&
  (!(Array.isArray(styleKeys) && styleKeys.includes(k)) ||
    (Array.isArray(overrideKeys) && overrideKeys.includes(k)));

const exportKeys = candidateKeys.filter(shouldExport);
```

worth doing?

packages/super-editor/src/extensions/run/calculateInlineRunPropertiesPlugin.test.js
VladaHarbour: added test case

packages/super-editor/src/document-api-adapters/plan-engine/executor.ts
VladaHarbour: Since runProperties can also contain style-resolved keys seeding from all keys would incorrectly promote style-only properties into inline-owned overrides. I added more extended fallback for old docs here to not miss properties

caio-pizzol:
@VladaHarbour pushed Word-native regression tests for the export filtering.

3 fixtures created by real Microsoft Word — headings, body text, bold/color/font overrides, and a table with style margins. The test imports each one, exports it, and checks:

- plain runs don't get inherited font/size properties
- bold and color overrides are kept
- table cell margins stay on the table, not copied to each cell

round-trip sizes stayed flat (~8-9% growth, no inflation).

PR #2291, opened 2026-03-04
fix(painter-dom): prevent scroll acceleration feedback loop in virtualization [SD-2005]

tupizz:
fixes https://linear.app/superdocworkspace/issue/SD-2005/bug-virtualization-scroll-accelerates-to-bottom-in-long-documents

## Before

https://github.com/user-attachments/assets/9a8c7a63-dd00-4a51-b419-5d2c9cbc034c

## After

https://github.com/user-attachments/assets/0f7e2adb-e0bb-4989-a109-4ec94687b9ea

## Summary

Fixes the scroll acceleration bug (SD-2005) where scrolling through long documents with `modules: { comments: false }` causes uncontrollable scroll acceleration around page 5, rocketing to the bottom of the document.

### Root Cause

The browser picked `virtualPagesEl` (the flex container for pages) as the scroll anchor. When the top spacer grew during virtual window shifts, `virtualPagesEl` moved down, and scroll anchoring adjusted `scrollContainer.scrollTop` to compensate. This fired a new scroll event with a higher `scrollY`, triggering another window shift. Each shift added ~1 page height to scrollTop, cascading to the end of the document in a single frame burst.

The existing `overflow-anchor: none` on spacer elements was insufficient. It prevented spacers from being anchors but left `virtualPagesEl` as the next candidate.

### Fix

- Set `overflow-anchor: none` on `virtualPagesEl` so the browser anchors on **page elements** (children) instead. Individual page elements stay at stable positions regardless of spacer changes (`topSpacer + pageOffsetInFlex = topOfIndex(pageIndex)` is invariant), so scroll anchoring produces zero adjustment.
- Stabilize the scrollY calculation for external scroll containers by caching the mount-to-container offset and using `scrollTop` instead of `getBoundingClientRect()`, which is susceptible to layout shifts during DOM mutations.

## Test plan

- [x] `pnpm --filter @superdoc/painter-dom test` passes (653/653)
- [ ] Load a 66+ page DOCX with `modules: { comments: false }` and scroll slowly past page 5 at non-100% zoom. Verify no scroll acceleration.
- [ ] Test at 100% zoom as well
- [ ] Test with nested scroll wrapper layout (external scroll container)

### DIFF
--- a/packages/layout-engine/painters/dom/src/renderer.ts
+++ b/packages/layout-engine/painters/dom/src/renderer.ts
@@ -1455,7 +1455,6 @@ export class DomPainter {
 
     if (this.mount && this.mount !== mount) {
       this.resetState();
-      this.scrollContainerMountOffset = null; // Invalidate on mount change
     }
     this.layoutVersion += 1;
     this.layoutEpoch = layout.layoutEpoch ?? 0;


packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -1447,6 +1455,7 @@ export class DomPainter {
 
     if (this.mount && this.mount !== mount) {
       this.resetState();
+      this.scrollContainerMountOffset = null; // Invalidate on mount change
```
caio-pizzol: `resetState()` already clears this field (line 2241) — drop the duplicate.

```suggestion
      this.resetState();
```

caio-pizzol (commented):
@tupizz good second pass on the scroll acceleration bug — the first fix (PR #2171) handled the zoom/gap/CSS issues, and this one tackles the remaining feedback loop from `getBoundingClientRect` reacting to spacer changes. caching the offset and disabling `overflow-anchor` on spacers is a clean approach.

no blocking issues. one duplicate line to clean up (inline comment). the cached offset can go stale if siblings above the mount resize, but that's unlikely in our setup — just flagging for awareness.

tests cover the happy path but not the cache reset on zoom/resize/mount change — a couple of small unit tests would catch regressions if those lines get dropped.

left inline comments.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -1700,13 +1723,18 @@ export class DomPainter {
...
-      const mountRect = this.mount.getBoundingClientRect();
-      const containerRect = this.scrollContainer.getBoundingClientRect();
-      scrollY = Math.max(0, (containerRect.top - mountRect.top) / zoom - paddingTop);
+      // Use scrollContainer.scrollTop with a cached mount offset instead of
+      // getBoundingClientRect(). Rects are affected by spacer DOM mutations
+      // which can cause cascading scroll events and runaway scrolling.
+      //
+      // mountOffset = distance from scroll container's content top to mount's top.
+      // Computed once and cached; invalidated on mount/container/zoom change.
+      if (this.scrollContainerMountOffset == null) {
```
caio-pizzol, line 1731: the cached offset won't update if something above the mount changes height inside the scroll container (e.g., a collapsible panel). probably fine for how we use it today, but there's no way for consumers to force a refresh. worth a follow-up?

packages/layout-engine/painters/dom/src/renderer.ts
tupizz: https://github.com/superdoc-dev/superdoc/pull/2291/commits/cc57069326166693f71a0f85e9822da3eed8308f

packages/layout-engine/painters/dom/src/renderer.ts
tupizz: i prefer focusing in the issue here

tupizz:
fixes https://linear.app/superdocworkspace/issue/SD-2005/bug-virtualization-scroll-accelerates-to-bottom-in-long-documents

harbournick:

PR #2292, opened 2026-03-04
feat: llm tools (alpha version)

andrii-harbour:
- Added 'skipAsATool' and 'essential' properties to CLI-only operation definitions to control tool catalog generation and mark essential tools.
- Updated the 'buildSdkContract' function to include these new properties in the SDK contract output.
- Adjusted operation categories to better reflect their usage, consolidating lifecycle operations under the session category.
- Refactored parameter schemas and validation logic to accommodate new operation definitions.

### DIFF
--- a/apps/docs/docs.json
+++ b/apps/docs/docs.json
@@ -111,7 +111,14 @@
               },
               "document-engine/sdks",
               "document-engine/cli",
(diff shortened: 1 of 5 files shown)


harbournick (approved):
LGTM

PR #2312, opened 2026-03-05
fix: update list marker font before adding list item

VladaHarbour:


packages/layout-engine/pm-adapter/src/attributes/paragraph.ts
```diff
@@ -308,20 +310,31 @@ export const computeParagraphAttrs = (
...
   defaultFontFamily = 'Times New Roman',
 ): ResolvedRunProperties => {
   let fontFamily;
-  if (converterContext) {
+  let fontSize;
+
+  if (previousParagraphFont && previousParagraphFont.fontFamily) {
+    fontFamily = previousParagraphFont.fontFamily;
+  } else if (converterContext) {
     fontFamily =
```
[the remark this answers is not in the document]
VladaHarbour: considered explicit paragraph fonts when setting data from previousParagraphFont

packages/layout-engine/pm-adapter/src/converters/paragraph.ts
```diff
@@ -286,7 +291,18 @@ export function paragraphToFlowBlocks({
       : undefined;
   const hasSectPr = Boolean(rawParagraphProps?.sectPr);
   const isSectPrMarker = hasSectPr || paraAttrs.pageBreakSource === 'sectPr';
```
caio-pizzol, line 548: this checks the merged/resolved run properties, which includes everything from styles and defaults — not just what the user typed. if any style in the chain sets font properties, this is non-empty and the fallback never fires. should this check the raw props from the node attrs instead? also same check is duplicated in `attributes/paragraph.ts:298`.

packages/layout-engine/pm-adapter/src/attributes/paragraph.ts
```diff
@@ -296,10 +298,19 @@ export const computeParagraphAttrs = (
       true,
       Boolean(paragraphProperties.numberingProperties),
     );
+
```
caio-pizzol, line 327: if a numbering definition sets a bullet font (Symbol, Wingdings), `previousParagraphFont` replaces it here. worth checking if the marker already has a font before falling back?

packages/layout-engine/pm-adapter/src/attributes/paragraph.ts
```diff
@@ -309,11 +320,15 @@ export const computeParagraphAttrs = (
 export const computeRunAttrs = (
```
caio-pizzol, line 365: only the marker-run caller uses this param, but it changes the signature for everyone — drop-cap has to pass `undefined` now. could apply the override after the call instead?

packages/layout-engine/pm-adapter/src/converters/paragraph.ts
```diff
@@ -692,6 +734,7 @@ export function handleParagraphNode(node: PMNode, context: NodeHandlerContext):
     }
```
caio-pizzol, line 1043: cache hit returns old blocks without rechecking the previous paragraph's font. if user changes font on paragraph A, empty list item B keeps the old font until something clears the cache. narrow window, but worth a comment?

packages/layout-engine/pm-adapter/src/converters/index.ts
```diff
@@ -11,7 +11,7 @@
  */
 
 // Paragraphs (converter + handler)
-export { paragraphToFlowBlocks, mergeAdjacentRuns, handleParagraphNode } from './paragraph.js';
+export { paragraphToFlowBlocks, mergeAdjacentRuns, handleParagraphNode, getLastParagraphFont } from './paragraph.js';
```
caio-pizzol, line 14: is there a reason to export `getLastParagraphFont`? only used inside `paragraph.ts`.

packages/layout-engine/pm-adapter/src/types.ts
```diff
@@ -412,3 +414,8 @@ export interface OoxmlBorder {
  * Underline style type derived from TextRun contract
  */
 export type UnderlineStyle = NonNullable<import('@superdoc/contracts').TextRun['underline']>['style'];
+
+export type ParagraphFont = {
```
caio-pizzol, line 422: `getLastParagraphFont` always returns both fields or undefined. making them required removes the null checks everywhere.

packages/layout-engine/pm-adapter/src/converters/index.ts
VladaHarbour: getLastParagraphFont import is used in tests

packages/layout-engine/pm-adapter/src/converters/paragraph.ts
VladaHarbour: checking raw props here now(moved to a function to share with attributes/paragraph.ts)

packages/layout-engine/pm-adapter/src/attributes/paragraph.ts
VladaHarbour: Added an additional check here but to be honest I don't think that it should cause issues since if there is no font change previousParagraphFont will be the same with the one from numbering definition

packages/layout-engine/pm-adapter/src/attributes/paragraph.ts
VladaHarbour: removed it from params list

harbournick:

PR #2313, opened 2026-03-05
fix(image): sync headless image media to Y.Doc for collab persistence (SD-2150)

caio-pizzol:
handleNodePath (headless/CLI image registration) stored media only in the local editor.storage.image.media object but never wrote to the Y.Doc media map. When a second collab session connected, image nodes synced via XmlFragment but the media binary data was missing.

Write directly to ydoc.getMap('media') instead of using editor.commands.addImageToCollaboration, which doesn't dispatch properly inside ProseMirror's appendTransaction callback.


harbournick:

PR #2314, opened 2026-03-06
fix(collaboration): sync numbering, styles, and headers in real-time (SD-2062, SD-2040)

tupizz:
## Demo

https://github.com/user-attachments/assets/273e7ea8-ca6a-4c07-81fe-19a4d8928230

## Problem

In collaboration mode, three types of document metadata only sync via the 30-second debounced DOCX export. This means other clients see stale or missing data for up to 30 seconds:

1. **Numbering definitions** (SD-2062) — list indent/spacing comes from numbering level definitions stored in `editor.converter.numbering`. Without them, the receiving client renders lists with wrong marker spacing (34.6px vs 10.6px).

2. **Style definitions** (SD-2040) — document default styles live in `editor.converter.translatedLinkedStyles`. The `styles.apply` API was blocked entirely during collaboration because there was no sync mechanism.

3. **Header/footer content** (SD-2040) — header content pushes to the `headerFooterJson` Y.js map on edit, but `headerIds` (the mapping that tells the layout pipeline *which* header to render for each section type) was never synced. Joining clients had header content but no way to resolve it.

## How collaboration sync works now

SuperDoc uses two real-time sync channels:

| Data | Y.js shared type | When pushed |
|------|------------------|-------------|
| Document body | `XmlFragment('supereditor')` via y-prosemirror | Every PM transaction (automatic) |
| Media files | `Map('media')` | On image upload |
| Header/footer content | `Map('headerFooterJson')` | Every header editor keystroke |
| Numbering + Styles + headerFooterIds | `Map('converterMeta')` | On `list-definitions-change`, `stylesDefaultsChanged` events |

Everything else (full DOCX XML) syncs via the debounced 30s export to `Map('meta').docx`.

### 1. Unified converter metadata sync (`converterMeta` Y.js map)

One map, one observer, one anti-ping-pong flag for all converter metadata. Adding a new type is: add a key to `CONVERTER_META_KEYS`, add push/apply logic, add a trigger.

**Files**: `collaboration-helpers.js`, `collaboration.js`

### 2. Header/footer atomicity fix

`headerIds`/`footerIds` are now bundled into every `headerFooterJson` entry alongside the content. Before, they lived in a separate Y.js map (`converterMeta`) and could arrive in a different network message than the content. The receiver would have header content but `headerIds.default === undefined`, so the layout pipeline couldn't resolve which header to render.

**Files**: `collaboration-helpers.js` (pushHeaderFooterToYjs, applyRemoteHeaderFooterChanges)

### 3. Initial state sync on join

Y.js map observers only fire for *future* changes. Data that arrived via Y.js sync *before* the observer was attached is missed. Fixed by reading existing map entries after attaching each observer.

Also push all initial metadata during `initializeMetaMap` (first client) so joining clients get headers, numbering, and styles immediately.

**Files**: `collaboration.js`

### 4. `numberingPlugin` collaboration hardening

The numbering plugin's `appendTransaction` clears `listRendering` when definitions are missing. In collaboration, definitions may arrive slightly after the document change (different Y.js types). The plugin now preserves synced `listRendering` when `ydoc` is present and the value already exists.

**Files**: `numberingPlugin.js`

### 5. Removed `styles.apply` collaboration gate

`styles.apply` was blocked during collaboration with "Stylesheet mutations cannot be synced via Yjs." Now that styles sync via the `converterMeta` map, the gate is removed.

**Files**: `styles-adapter.ts`, `capabilities-adapter.ts`

## Test plan

- [x] Unit tests: 753 files pass, 8146 tests pass (32 new tests added)
- [x] Smoke test: editor loads and edits correctly
- [ ] Manual: open two tabs with `?collab=1&room=<name>`, create a numbered list on tab A, verify correct spacing on tab B
- [ ] Manual: type in header on tab A, verify header appears on tab B
- [ ] Manual: apply style change on tab A, verify tab B re-renders with new styles

### DIFF
--- a/packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
+++ b/packages/super-editor/src/core/presentation-editor/PresentationEditor.ts
@@ -2804,6 +2804,22 @@ export class PresentationEditor extends EventEmitter {
       handler: handleRemoteHeaderFooterChanged as (...args: unknown[]) => void,
     });
 
+    // Remote converter metadata arrived via Y.js (numbering, styles, headerFooterIds, etc.)
+    // Re-render so the style engine and layout pipeline pick up the changes.
+    // For headerFooterIds: refresh the H/F manager so it discovers new sections.
+    const handleRemoteConverterMetaChanged = (payload: { key: string }) => {
+      if (payload?.key === 'headerFooterIds') {
+        this.#headerFooterSession?.manager?.refresh();
+      }
+      this.#pendingDocChange = true;
+      this.#scheduleRerender();
+    };
+    this.#editor.on('remoteConverterMetaChanged', handleRemoteConverterMetaChanged);
+    this.#editorListeners.push({
+      event: 'remoteConverterMetaChanged',
+      handler: handleRemoteConverterMetaChanged as (...args: unknown[]) => void,
+    });
+
     // Listen for comment selection changes to update Layout Engine highlighting
     const handleCommentsUpdate = (payload: { activeCommentId?: string | null }) => {
       if (this.#domPainter?.setActiveComment) {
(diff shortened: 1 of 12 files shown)


caio-pizzol:
@tupizz closing this

#2325 (parts sync) landed on March 8 and covers everything here. That PR replaced per-type metadata syncing with a system that syncs all document parts at once, including numbering, styles, and headers. It also removed the collab block on `styles.apply`.

PR #2317, opened 2026-03-06
feat: make images uploaded into table cell adjust to width of cell

johanneswilm:
* Added check within `getMaxContentSize()` whether cursor is positioned within table cell or table header cell, and if so, attempts to return the widths of these with a fallback to the page width.
* Added relevant tests.

### DIFF
--- a/tests/behavior/tests/tables/image-resize-in-cell.spec.ts
+++ b/tests/behavior/tests/tables/image-resize-in-cell.spec.ts
@@ -1,0 +1,214 @@
+import { test, expect, type SuperDocFixture } from '../../fixtures/superdoc.js';
+
+/**
+ * Behavior test: images inserted into a table cell must be
+ * constrained to the cell's content width.
+ *
+ * Flow:
+ *  1. Insert a 1-column table with an explicit, narrow column width (200 px).
+ *  2. Click into the only cell so the cursor lands inside it.
+ *  3. Simulate a drag-drop of an image that is much wider than the cell.
+ *  4. Assert the image node's stored width is ≤ the cell's colwidth.
+ *
+ * The constraint is applied by Editor.getMaxContentSize(), which, when the
+ * selection is inside a tableCell / tableHeader, returns the cell's colwidth
+ * minus cell margins instead of the full page content width.
+ */
+
+test.use({ config: { toolbar: 'full', showSelection: true } });
+
+// ─── helpers ────────────────────────────────────────────────────────────────
+
+type PlacementSnapshot = {
+  imageCount: number;
+  imageWidth: number | null;
+};
+
+async function getImageSnapshot(superdoc: SuperDocFixture): Promise<PlacementSnapshot> {
+  return superdoc.page.evaluate(() => {
+    const doc = (window as any).editor?.state?.doc;
+    if (!doc) throw new Error('Editor document is unavailable.');
+
+    let imageCount = 0;
+    let imageWidth: number | null = null;
+
+    doc.descendants((node: any) => {
+      if (node.type?.name === 'image') {
+        imageCount += 1;
+        if (imageWidth === null) {
+          imageWidth = node.attrs?.size?.width ?? null;
+        }
+      }
+    });
+
+    return { imageCount, imageWidth };
+  });
+}
+
+/**
+ * Move the editor cursor into the first table cell by using the ProseMirror
+ * command API, so the selection is deterministic regardless of layout.
+ */
+async function placeCursorInFirstTableCell(superdoc: SuperDocFixture): Promise<void> {
+  await superdoc.page.evaluate(() => {
+    const editor = (window as any).editor;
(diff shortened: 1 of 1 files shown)


johanneswilm:
Tables and table cells are an extension to the main editor, and this adds code that references table cells right in the main Editor. However, I see that table cellNodes are referenced numerous other places within the `core` folder, and the way it way this extra reference works is that it will still work even if the editor is run with the table extension deactivated as it checks by node name.

johanneswilm:
@harbournick @caio-pizzol This is ready for review.

packages/super-editor/src/core/Editor.ts
```diff
@@ -71,6 +71,19 @@ const PIXELS_PER_INCH = 96;
 const MAX_HEIGHT_BUFFER_PX = 50;
 const MAX_WIDTH_BUFFER_PX = 20;
 
+/**
```
caio-pizzol, line 97: this same math already lives in `decode-image-node-helpers.js:74` with a different default margin. if one changes the other won't — share a helper?

packages/super-editor/src/core/Editor.ts
```diff
@@ -2059,6 +2077,21 @@ export class Editor extends EventEmitter<EditorEventMap> {
     // All sizes are in inches so we multiply by PIXELS_PER_INCH to get pixels
     const maxHeight = height * PIXELS_PER_INCH - topPx - bottomPx - MAX_HEIGHT_BUFFER_PX;
     const maxWidth = width * PIXELS_PER_INCH - leftPx - rightPx - MAX_WIDTH_BUFFER_PX;
+
+    // When the cursor is inside a table cell, constrain width to the cell's content
+    // width so images inserted into a cell are never wider than that cell.
```
caio-pizzol, line 2301: `cellWrapping()` in `tableHelpers/cellWrapping.js` already does this same walk. worth reusing?

johanneswilm: These are standard prosemirror functions, so you'll often find similar code in prosemirror based editors.

The function in cellWrapping is to go up the document tree and find a cell node, if any, and then to return that node. 

The function below is instead to return the width of such a cell. So we could take the function from the tableHelpers file and check if it returns a node, and if so, then call the `getCellContentWidthPx()` function on. That would replace 3 lines with 1 and make the code a little less readable. On the plus side, it would ensure that we check for table cell/header cell in the same way both places.

The main reason I didn't do it previously is that the tableHelpers file is part of the table extension whereas this code is in the main code base. I don't know if the idea with extensions is to be able to disable some of them, because that will be harder. On the other hand, I see that we are importing from multiple extensions directly into the core, so maybe that is not a priority.

Conclusion:  Sure, if you don't mind us relying on an extension in the core code, we can do that.

packages/super-editor/src/core/Editor.ts
johanneswilm: Do you mean this code: https://github.com/superdoc-dev/superdoc/blob/9c7fc2e61144365fc33f64fefd324ec894429a51/packages/super-editor/src/core/super-converter/v3/handlers/wp/helpers/decode-image-node-helpers.js#L209-L218

Ok, so that is inside the converter, whereas this is in the editor. 

It could work, but notice how we are using px here and emu in the converter. I also noticed the different default margins there (8 vs 0), but I wonder if that could have to do with the difference of the editor internal prosemirror json and the docx format.


I'll investigate, but I am somewhat afraid that it will turn into spaghetti code if we just use make functions that go across editor and converter that look a bit similar but are used in very different scenarios and with different output formats in mind.

caio-pizzol: fair point — the different defaults and units make a shared helper more coupling than it's worth for a 3-line formula. leaving as-is is fine.

packages/super-editor/src/core/Editor.ts
caio-pizzol: makes sense. the savings are small and the extra import isn't worth it for 3 lines. fine as-is.

packages/super-editor/src/core/Editor.webLayout.test.ts
```diff
@@ -177,4 +177,151 @@ describe('Editor Web Layout Mode', () => {
       });
```
caio-pizzol, line 177: would be good to add a behavior test for the full flow — insert a large image into a table cell and check it fits. `drag-drop-image-insertion.spec.ts` already has the image insert pattern and `tables/` has table helpers, so combining them should be quick.

johanneswilm: Ok, agreed. That makes sense. I'll add it.

caio-pizzol:
@johanneswilm behavior test looks great, thanks for adding it.

the webkit failure in scroll-to-comment.spec.ts is a pre-existing flake, unrelated to this PR

harbournick:

PR #2342, opened 2026-03-09
fix(document-api): return NodeAddress from find and getNode instead of SDAddress (SD-2168)

caio-pizzol:
find and getNode now return NodeAddress directly, making their results compatible with mutation targets like create.paragraph. Removes the toSDAddress conversion layer that stripped nodeType and changed kind from 'block' to 'content', which made find results unusable for positional inserts. Also adds docs for the query.match → create.paragraph workflow.


harbournick:

PR #2343, opened 2026-03-09
fix: remove syncing of runProperties with paragraph (SD-2143)

luccas-harbour:
- Removes logic used for syncing paragraph's default run properties with its first child run
- Word performs this only if the paragraph marker is selected, which is not a concept we have currently. Keeping this logic caused weird issues when dealing with lists.
- With this change, users won't be able to style list markers anymore (which was a side-effect).

### DIFF
--- a/packages/super-editor/src/extensions/run/calculateInlineRunPropertiesPlugin.test.js
+++ b/packages/super-editor/src/extensions/run/calculateInlineRunPropertiesPlugin.test.js
@@ -331,70 +331,40 @@ describe('calculateInlineRunPropertiesPlugin', () => {
     expect(paragraph.attrs.paragraphProperties).toBeNull();
   });
 
-  it('does not sync paragraph runProperties for first runs inside inline wrappers', () => {
+  it("does not update a paragraph's runProperties using the run's properties", () => {
     const schema = makeSchema();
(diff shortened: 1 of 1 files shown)


luccas-harbour:
> @luccas-harbour nice and clean — just removing the sync is the right approach. checked whether old documents could still have the bug through a different path, and they can't — the style engine already handles that.
> 
> a couple small things:
> 
> * `carbonCopy` import on line 9 of `calculateInlineRunPropertiesPlugin.js` isn't used anymore after this removal — can be deleted.
> * the test "does not update paragraph runProperties when a non-first run changes" — now that no run syncs to the paragraph at all, the "non-first" distinction doesn't mean anything. worth renaming or dropping?
> 
> tests cover the change well. only thing missing is a behavior test that bolds the first word of a list item and checks the bullet doesn't turn bold — would protect against future regressions.

@caio-pizzol I removed the unused import and adjusted the unit tests, removing the ones that became redundant. 

As for adding a behavior test, I don't really think it's needed, since this is not a UI-related issue but more of a ProseMirror-related issue. I added unit tests that cover this change: https://github.com/Harbour-Enterprises/SuperDoc/blob/917ecac32a0ed540826ca2660f13a497df2b86d1/packages/super-editor/src/extensions/run/calculateInlineRunPropertiesPlugin.test.js#L321-L387

harbournick:

PR #2352, opened 2026-03-10
Added fixes to properly align paragraphs when loading RTL documents

claudiu-ior:
### Summary

- Propagate the `rightToLeft` (bidi) paragraph property through `pm-adapter` into `ParagraphAttrs` so the layout engine and DomPainter know a paragraph is RTL
- Make `normalizeAlignment` direction-aware: `start`/`end` now resolve to physical `left`/`right` based on paragraph direction instead of always mapping to LTR defaults
- Add support for Arabic kashida justify variants (`lowKashida`, `mediumKashida`, `highKashida`) which were previously silently dropped
- Update DomPainter to set `dir="rtl"` on RTL paragraph and line elements, default their `text-align` to `right`, and position segments from the right edge in segment-based rendering (lines with tabs and explicit X coordinates)

### Problem

Opening an RTL Arabic document rendered all text left-aligned. Three independent issues contributed:

1. **RTL flag not propagated** -- `computeParagraphAttrs` in `pm-adapter` read `resolvedParagraphProperties.rightToLeft` but never set `direction` or `rtl` on the output `ParagraphAttrs`, so DomPainter had no way to know a paragraph was RTL.

2. **Logical-to-physical alignment was not direction-aware** -- `normalizeAlignment` always mapped `start` to `left` and `end` to `right`, which is wrong for RTL where `start` means `right`. Additionally, Arabic-specific OOXML justify values (`lowKashida`, `mediumKashida`, `highKashida`) were unrecognized and fell through to `undefined`.

3. **DomPainter assumed LTR** -- No `dir="rtl"` attribute was set on elements. Justified text always used `text-align: left` as the base (incorrect for RTL, where the last line should right-align). Segment-based rendering (triggered by tabs or explicit positioning) always used `style.left` for X coordinates instead of `style.right` for RTL paragraphs.

### Changes

| File | What changed |
|------|-------------|
| `pm-adapter/src/attributes/paragraph.ts` | Derive `isRtl` from `resolvedParagraphProperties.rightToLeft`; pass it to `normalizeAlignment`; spread `direction: 'rtl'` and `rtl: true` onto output attrs |
| `pm-adapter/src/attributes/spacing-indent.ts` | Add `isRtl` parameter to `normalizeAlignment`; flip `start`/`end` mapping for RTL; add `lowKashida`/`mediumKashida`/`highKashida` as `justify` aliases |
| `painters/dom/src/renderer.ts` | Set `dir="rtl"` on paragraph and line elements; default RTL text-align to `right`; use `right` for justify base in RTL; introduce `setHorizontalPos` helper for segment-based rendering to use `style.right` instead of `style.left` for RTL |
| `pm-adapter/src/attributes/spacing-indent.test.ts` | Tests for RTL `start`/`end` mapping and kashida variants |
| `pm-adapter/src/index.test.ts` | Tests for RTL attr propagation, default alignment, explicit overrides, and `start`/`end` flipping |

### Test plan

- [x] Unit tests pass (`pnpm test` -- `spacing-indent.test.ts` and `index.test.ts` cover all new logic)
- [x] Load an RTL Arabic DOCX (e.g. `intalio_template_17529.docx`) -- text should be right-aligned with correct justification
- [x] Load an LTR document -- no regression, text still left-aligned as before
- [x] Verify mixed LTR/RTL documents render each paragraph with the correct direction
- [x] Check paragraphs with tabs in RTL documents -- segments should be positioned from the right edge

### DIFF
--- a/packages/layout-engine/painters/dom/src/renderer.ts
+++ b/packages/layout-engine/painters/dom/src/renderer.ts
@@ -108,11 +108,7 @@ import {
   stampBetweenBorderDataset,
   type BetweenBorderInfo,
 } from './features/paragraph-borders/index.js';
-import {
-  isRtlParagraph,
-  applyRtlStyles,
-  shouldUseSegmentPositioning,
-} from './features/rtl-paragraph/index.js';
+import { applyRtlStyles, shouldUseSegmentPositioning } from './features/rtl-paragraph/index.js';
 
 /**
  * Minimal type for WordParagraphLayoutOutput marker data used in rendering.


packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -5468,6 +5483,14 @@ export class DomPainter {
       const listIndentOffset = isFirstLineOfPara ? (rawTextStartPx ?? indentLeft) : indentLeft;
       const indentOffset = isListParagraph ? listIndentOffset : indentLeft + firstLineOffsetForCumX;
       let cumulativeX = 0; // Start at 0, we'll add indentOffset when positioning
+
+      const setHorizontalPos = (elem: HTMLElement, xPx: number) => {
+        if (useRightPositioning) {
+          elem.style.right = `${xPx}px`;
+        } else {
```
caio-pizzol: `style.right = 200px` means "200px from the right edge", but the X values here are measured from the left edge. so everything lands on the wrong side. only hits RTL paragraphs with tab stops — regular RTL text is fine.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -5180,14 +5180,27 @@ export class DomPainter {
     if (styleId) {
       el.setAttribute('styleid', styleId);
     }
-    const alignment = (block.attrs as ParagraphAttrs | undefined)?.alignment;
+    const pAttrs = block.attrs as ParagraphAttrs | undefined;
+    const alignment = pAttrs?.alignment;
+    const isRtl = pAttrs?.direction === 'rtl' || pAttrs?.rtl;
```
caio-pizzol: this same `direction === 'rtl' || rtl` check shows up in three places now. small helper to keep it in one spot?

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -5180,16 +5180,14 @@ export class DomPainter {
...
```
caio-pizzol: this RTL check is repeated in `applyParagraphBlockStyles` below. consider moving the RTL logic into a feature module (`features/rtl-paragraph/`) like we did for paragraph borders in #2324 — keeps it all in one place as more RTL work comes in.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -5443,10 +5441,11 @@ export class DomPainter {
       el.style.wordSpacing = `${spacingPerSpace}px`;
     }
 
-    if (hasExplicitPositioning && line.segments) {
-      // Use segment-based rendering with absolute positioning for tab-aligned text
-      // When rendering segments, we need to track cumulative X position
-      // for segments that don't have explicit X coordinates.
+    if (hasExplicitPositioning && line.segments && !isRtl) {
```
caio-pizzol: per the spec (§17.3.1.37), tab positions in RTL paragraphs are measured from the right edge. skipping them here avoids the old bug but means tabs won't land where the document says. fine for now — worth a comment or follow-up issue.

packages/layout-engine/painters/dom/src/renderer.ts
claudiu-ior: Take a look please, I've addressed this in the last commit

claudiu-ior:
@caio-pizzol gentle reminder to review this PR

claudiu-ior:
@codex review

packages/layout-engine/painters/dom/src/features/rtl-paragraph/rtl-styles.ts
```diff
@@ -0,0 +1,65 @@
...
+export const resolveTextAlign = (alignment: ParagraphAttrs['alignment'], isRtl: boolean): string => {
+  switch (alignment) {
+    case 'center':
+    case 'right':
+    case 'left':
+      return alignment;
+    case 'justify':
+      return isRtl ? 'right' : 'left';
+    default:
+      return isRtl ? 'right' : 'left';
```
caiopizzol, line 37: these two cases do the same thing — combine them?

```suggestion
    case 'justify':
    default:
      return isRtl ? 'right' : 'left';
```

packages/layout-engine/painters/dom/src/features/rtl-paragraph/index.ts
```diff
@@ -0,0 +1,15 @@
...
+ * - Applying dir="rtl" and the correct text-align to an element
+ * - Resolving text-align for RTL vs LTR (justify → right/left)
+ * - Deciding whether segment-based (absolute) positioning is safe
+ *
+ * @ooxml w:pPr/w:bidi — paragraph bidirectional flag
+ * @ooxml w:rPr/w:rtl  — run-level right-to-left flag
+ * @spec  ECMA-376 §17.3.1.1 (bidi), §17.3.2.30 (rtl)
+ */
+
+export { isRtlParagraph, resolveTextAlign, applyRtlStyles, shouldUseSegmentPositioning } from './rtl-styles.js';
```
caiopizzol: `resolveTextAlign` and `isRtlParagraph` aren't used outside this module — drop them from the export?

```suggestion
export { applyRtlStyles, shouldUseSegmentPositioning } from './rtl-styles.js';
```

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -108,6 +108,11 @@ import {
   stampBetweenBorderDataset,
   type BetweenBorderInfo,
 } from './features/paragraph-borders/index.js';
+import {
+  isRtlParagraph,
```
caiopizzol: `isRtlParagraph` is imported but never used here — safe to remove.

packages/layout-engine/painters/dom/src/features/rtl-paragraph/rtl-styles.ts
caio-pizzol, line 37: these two cases do the same thing — combine them?

```suggestion
    case 'justify':
    default:
      return isRtl ? 'right' : 'left';
```

packages/layout-engine/painters/dom/src/features/rtl-paragraph/index.ts
caio-pizzol: `resolveTextAlign` and `isRtlParagraph` aren't used outside this module — drop them from the export?

```suggestion
export { applyRtlStyles, shouldUseSegmentPositioning } from './rtl-styles.js';
```

packages/layout-engine/painters/dom/src/renderer.ts
caio-pizzol: `isRtlParagraph` is imported but never used here — safe to remove.

PR #2353, opened 2026-03-10
fix: watermark shading mismatch (SD-2147)

luccas-harbour:
## Summary

  Fixes SD-2147, where watermark shading changed after page 1 when the document mixed VML
  and DrawingML header watermarks.

  The root cause was twofold:

  - DrawingML a:lum effects were imported/exported but not rendered end to end.
  - VML gain / blacklevel values were being approximated incorrectly in the DOM painter.

  This PR makes both watermark paths render from the correct source values so first-page
  and later-page watermarks stay visually consistent.

  ## Changes

  ### 1. Add DrawingML luminance plumbing

  - Parse a:lum/@bright and a:lum/@contrast from w:drawing images.
  - Store them on PM image attrs as lum.
  - Pass lum through pm-adapter into layout-engine image contracts.
  - Export lum back to DOCX as <a:lum .../>.

  ### 2. Render DrawingML luminance correctly

  - Render a:lum using its percentage semantics:
      - bright -> brightness(1 + bright / 100000)
      - contrast -> contrast(1 + contrast / 100000)
  - Add DOM painter coverage for DrawingML image luminance rendering.

  ### 3. Normalize VML watermark filters

  - Parse VML gain / blacklevel values ending in f as fixed-fraction values.
  - Render VML watermark filters using conversions that match Word’s watermark UI
    behavior:
      - gain -> CSS contrast
      - blacklevel -> CSS brightness
  - Add DOM painter coverage for the VML watermark values from the repro doc.

  ## Why

  The repro document uses:

  - page 1 watermark in VML (w:pict / v:imagedata)
  - later-page watermarks in DrawingML (w:drawing / a:lum)

  Word shows these as equivalent watermark adjustments, but SuperDoc was applying
  different rendering logic to each path, which made later pages appear darker.

### DIFF
--- a/packages/layout-engine/painters/dom/src/index.test.ts
+++ b/packages/layout-engine/painters/dom/src/index.test.ts
@@ -4701,6 +4701,76 @@ describe('DomPainter', () => {
       expect(filter.brightness).toBeCloseTo(1.7, 4);
     });
 
+    it('preserves zero-valued DrawingML luminance filters', () => {
+      const dataUrl =
+        'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg==';
+      const imageBlocks: FlowBlock[] = [
+        {
+          kind: 'paragraph',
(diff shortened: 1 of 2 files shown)


harbournick:

PR #2362, opened 2026-03-11
fix(tracked-changes): preserve format on export and import

palmer-cl:
- Preserve formatting changes on export 
- Add tests for multiple types of format preservation

### DIFF
--- a/packages/super-editor/src/core/super-converter/v3/handlers/helpers.js
+++ b/packages/super-editor/src/core/super-converter/v3/handlers/helpers.js
@@ -27,6 +27,7 @@ export const createTrackStyleMark = (marks) => {
   const existingNode = marks.find((mark) => mark?.name === 'w:rPrChange');
   if (existingNode) {
     // Import path already produced a valid OOXML change node; re-use it verbatim.
+    // these xml nodes come from ins-translator and del-translator
     return existingNode;
   }
 


packages/super-editor/src/core/super-converter/v3/handlers/helpers.js
```diff
@@ -1,12 +1,36 @@
...
+const normalizeMark = (mark) => {
+  if (!mark) return null;
+  const type = typeof mark.type === 'string' ? mark.type : typeof mark.type?.name === 'string' ? mark.type.name : null;
+  if (!type) return null;
+  return { type, attrs: mark?.attrs || {} };
+};
+
+const toRunPropertyElements = (marks = []) =>
+  marks
+    .map((mark) => normalizeMark(mark))
```
caio-pizzol: this builds XML from marks using the old `processOutputMarks` path. `translate-text-node.js:21-24` already does the same thing through the newer pipeline (`decodeRPrFromMarks` + `wRPrNodeTranslator.decode`) which covers way more properties. worth switching?

packages/super-editor/src/core/super-converter/v3/handlers/helpers.js
```diff
@@ -1,12 +1,36 @@
...
+    .filter((element) => element && Object.keys(element).length);
 /**
  * Creates export element for trackFormat mark
  * @param {Array} marks SD node marks.
  * @returns {Object|undefined} Properties element for trackFormat change or undefined.
  */
 export const createTrackStyleMark = (marks) => {
-  const trackStyleMark = marks.find((mark) => mark.type === 'trackFormat');
+  const existingNode = marks.find((mark) => mark?.name === 'w:rPrChange');
+  if (existingNode) {
```
caio-pizzol, line 28: the comment says what this does but not why marks can contain raw XML. it happens because `ins-translator` and `del-translator` push XML nodes into the marks array. mentioning that would save future readers some digging.

packages/super-editor/src/core/super-converter/v3/handlers/w/r/r-translator.js
```diff
@@ -237,15 +237,50 @@ const decode = (params, decodedAttrs = {}) => {
...
+    const nextRunProperties = runPropsTemplate ? cloneXmlNode(runPropsTemplate) : null;
+    const runPropertiesNode =
+      nextRunProperties || preservedPropertyChanges.length
+        ? nextRunProperties || { name: 'w:rPr', elements: [] }
+        : null;
+
+    if (runPropertiesNode) {
+      if (!Array.isArray(runPropertiesNode.elements)) runPropertiesNode.elements = [];
+
+      const existingChangeKeys = new Set(
```
caio-pizzol, line 270: the dedup logic here can't actually trigger — the template never contains `w:rPrChange` nodes, so the Set is always empty. could just append directly without the key check.

packages/super-editor/src/core/super-converter/v3/handlers/helpers.js
palmer-cl: Yes! we can do that, just had to make a few tweaks. Updated in daf45661d569e1ea31f6a891a3dec23f37135a19

packages/super-editor/src/core/super-converter/v3/handlers/helpers.js
palmer-cl: expanded on comment

packages/super-editor/src/core/super-converter/v3/handlers/w/r/r-translator.js
palmer-cl: Would the preserved changes ever have duplicates we need to account for? or safe to just put them all in?

caio-pizzol, line 270: to answer your question: no, duplicates can't happen. the template never contains `w:rPrChange` nodes, and the spec only allows one per `w:rPr`. safe to just push directly without the Set.

packages/super-editor/src/core/super-converter/v3/handlers/helpers.js
```diff
@@ -1,12 +1,40 @@
...
+// decodeRPrFromMarks + w:rPr translator pipeline.
+const normalizeMark = (mark) => {
+  if (!mark) return null;
+  const type = typeof mark.type === 'string' ? mark.type : typeof mark.type?.name === 'string' ? mark.type.name : null;
+  if (!type) return null;
+  // Some snapshots only store `{ type }`; normalize to empty attrs so decodeRPrFromMarks stays safe.
+  return { type, attrs: mark?.attrs || {} };
+};
+
+const toRunPropertyElements = (marks = []) =>
```
caio-pizzol, line 14: the IIFE wrapper doesn't do anything here — a regular block body works the same.

```suggestion
const toRunPropertyElements = (marks = []) => {
  const normalizedMarks = marks.map((mark) => normalizeMark(mark)).filter(Boolean);
  const runProperties = decodeRPrFromMarks(normalizedMarks);
  const rPrNode = wRPrNodeTranslator.decode({ node: { attrs: { runProperties } } });
  return Array.isArray(rPrNode?.elements) ? rPrNode.elements : [];
};
```

packages/super-editor/src/core/super-converter/v3/handlers/w/t/helpers/translate-text-node.js
```diff
@@ -8,17 +8,33 @@
...
 
+  // Attach tracked formatting change metadata to the run so export keeps it open (not pre-accepted).
+  const trackedStyleChange = createTrackStyleMark(marks || []);
   const textRunProperties = decodeRPrFromMarks(marks || []);
   const parentRunProperties = params.extraParams?.runProperties || {};
   const combinedRunProperties = combineRunProperties([parentRunProperties, textRunProperties]);
   const rPrNode = wRPrNodeTranslator.decode({ node: { attrs: { runProperties: combinedRunProperties } } });
+  const hasRunProperties = Array.isArray(rPrNode?.elements) && rPrNode.elements.length > 0;
+  const hasStyleChange = Boolean(trackedStyleChange);
+  const resolvedRunProperties =
```
caio-pizzol, line 27: the editor side of tracked formatting has good test coverage (display, reject, revert), but the export/import side — which is what this PR fixes — is thin. the r-translator test always pairs `trackFormat` with `bold`, so the branch where tracked formatting is the *only* thing on the run never runs. also `createTrackStyleMark` has no direct tests (both paths: passthrough of existing XML nodes and building from marks). worth adding at least the standalone `trackFormat` case and a direct test for `createTrackStyleMark`?

caio-pizzol:
Hey @palmer-cl, this PR has had no activity for 8 days.

palmer-cl:
> Hey @palmer-cl, this PR has had no activity for 8 days.

@caio-pizzol Im seeing some changes I think fixed this: https://github.com/superdoc-dev/superdoc/pull/2395

caio-pizzol:
Closing this - #2395 landed on March 15 and covers the same fix (tracked formatting preserved through export/import)

caio-pizzol:
Closing this — #2395 landed on March 15 and covers the same fix. The one missing piece (null guards in `decodeRPrFromMarks` for sparse marks) is now in #2638.

PR #1271, opened 2025-11-16
fix: enhance drawing and inline node handling in translation [SD-824]

palmer-cl:
- Updated `drawing-translator` to only wrap valid `wp:inline` and `wp:anchor` children in `w:drawing`.
- Improved `translateInlineNode` to guard against invalid drawing structures and ensure proper handling of image nodes.
- Improved image and signature placeholder exports, ensuring non-zero IDs are generated when missing.
- Added tests

PR #2363, opened 2026-03-11
fix: guard drawing export against invalid structures and zero IDs (SD-824)

caio-pizzol:
Supersedes #1271 — rebased onto current main and incorporates review feedback.

- **Drawing translator guard**: skip `w:drawing` wrap when child is not `wp:inline` or `wp:anchor`, preventing corrupt OOXML when `translateImageNode` falls back to a text run (e.g., invalid signature data)
- **Inline + anchor node guards**: bail out of `wp:inline` / `wp:anchor` construction when `translateImageNode` returns a non-drawing result (missing `wp:extent`)
- **Non-zero IDs**: generate positive random IDs for `wp:docPr` and `pic:cNvPr` instead of defaulting to `0`, which violates OOXML uniqueness requirements (ECMA-376 §20.4.2.5)

### OOXML compliance

| Element | Spec section | Finding |
|---------|-------------|---------|
| `wp:docPr/@id` | §20.4.2.5 | `id=0` passes XSD but violates uniqueness when multiple objects default to it |
| `w:drawing` children | §17.3.3.9 | Only `wp:inline` and `wp:anchor` are standard children |
| `wp:inline` required children | §20.4.2.8 | `wp:extent` is mandatory in `CT_Inline` sequence |

### Changes from #1271

- Rebased onto current main (code was refactored significantly since Nov 2025)
- Simplified guard: plain `!==` check instead of `Set` (review feedback)
- Simplified inline guard: single `wp:extent` check covers all cases (review feedback)
- Added `Math.max(1, ...)` to prevent theoretical `id=0` from `parseInt` (review feedback)
- Computed `drawingId` once, shared by both `buildDocPrElement` and `buildNvPicPrElement`
- Added anchor path guard (was missing in original PR)
- Updated existing anchor tests to include `wp:extent` in mock data

## Test plan

- [x] `drawing-translator.test.js` — 12 tests pass (guard + existing)
- [x] `drawing-guards.test.js` — 7 new tests (inline guard, anchor guard, ID generation)
- [x] `translate-anchor-node.test.js` — 30 tests pass (mock updated for guard)

### DIFF
--- a/packages/super-editor/src/core/super-converter/v3/handlers/wp/helpers/merge-drawing-children.js
+++ b/packages/super-editor/src/core/super-converter/v3/handlers/wp/helpers/merge-drawing-children.js
@@ -15,8 +15,13 @@ import { carbonCopy } from '@core/utilities/carbonCopy.js';
 export function mergeDrawingChildren({ order, generated, original }) {
   const genQueues = groupByName(generated);
   const originalsByIndex = groupByIndex(original);
+  const merged = mergeWithOrder(order, genQueues, originalsByIndex);
 
-  return mergeWithOrder(order, genQueues, originalsByIndex);
+  // Originals may carry invalid IDs (e.g. id="0") that Word rejects.
+  // Patch them using the valid ID from the generated wp:docPr.
+  fixZeroDrawingIds(merged, generated);
+
+  return merged;
 }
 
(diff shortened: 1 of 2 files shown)


harbournick:

PR #2368, opened 2026-03-11
fix(superdoc): expose header/footer edits in update callbacks

luccas-harbour:
## Summary

  This PR makes header and footer edits observable through the same public callback  surface as body edits.

  Before this change, typing in header/footer child editors updated internal state, but those edits were not propagated through the public `onEditorUpdate` and `onTransaction` hooks with the same immediacy as body edits. This caused integrations that rely on those callbacks for dirty-state tracking, autosave, and other live reactions to miss header/footer changes until blur.

  ## What Changed

  - Forward live header/footer child-editor `update` and `transaction` events through the header/footer session flow.
  - Re-emit those events from `PresentationEditor` as header/footer-specific presentation events.
  - Bridge those presentation events into the existing SuperDoc public callback surface.
  - Normalize callback payloads so callers can tell where the change originated via:
    - `sourceEditor`
    - `surface` (`body`, `header`, or `footer`)
    - `headerId`
    - `sectionType`
  - Update the published SuperDoc and React wrapper event types to match the new payload shape.

  ## Testing

  - Added a `PresentationEditor` test that verifies live header/footer child-editor updates and transactions are re-emitted immediately.
  - Added a `SuperDoc.vue` test that verifies header/footer presentation events are forwarded into `editor-update` and `onTransaction`.

### DIFF
--- a/packages/super-editor/src/core/presentation-editor/tests/PresentationEditor.test.ts
+++ b/packages/super-editor/src/core/presentation-editor/tests/PresentationEditor.test.ts
@@ -2156,6 +2156,132 @@ describe('PresentationEditor', () => {
       );
     });
 
+    it('stops re-emitting header/footer child editor events after exiting edit mode', async () => {
+      mockIncrementalLayout.mockResolvedValueOnce(buildLayoutResult());
+
+      editor = new PresentationEditor({
+        element: container,
+        documentId: 'test-doc',
+      });
+
+      await vi.waitFor(() => expect(mockIncrementalLayout).toHaveBeenCalled());
+      await new Promise((resolve) => setTimeout(resolve, 100));
+
+      const pagesHost = container.querySelector('.presentation-editor__pages') as HTMLElement;
+      const mockPage = document.createElement('div');
+      mockPage.setAttribute('data-page-index', '0');
+      pagesHost.appendChild(mockPage);
+
+      const viewport = container.querySelector('.presentation-editor__viewport') as HTMLElement;
+      vi.spyOn(viewport, 'getBoundingClientRect').mockReturnValue({
+        left: 0,
+        top: 0,
+        width: 800,
+        height: 1000,
+        right: 800,
+        bottom: 1000,
+        x: 0,
+        y: 0,
+        toJSON: () => ({}),
+      } as DOMRect);
+
+      const updateSpy = vi.fn();
+      const transactionSpy = vi.fn();
+      editor.on('headerFooterUpdate', updateSpy);
+      editor.on('headerFooterTransaction', transactionSpy);
+
+      viewport.dispatchEvent(new MouseEvent('dblclick', { bubbles: true, clientX: 120, clientY: 50, button: 0 }));
+
+      await vi.waitFor(() => expect(createdSectionEditors.length).toBeGreaterThan(0));
+      await vi.waitFor(() => expect(editor.getActiveEditor()).toBe(createdSectionEditors.at(-1)?.editor));
+
+      const sourceEditor = editor.getActiveEditor();
+      const transaction = { docChanged: true };
+
+      sourceEditor?.emit('update', { editor: sourceEditor });
+      sourceEditor?.emit('transaction', { editor: sourceEditor, transaction, duration: 9 });
+
+      expect(updateSpy).toHaveBeenCalledTimes(1);
+      expect(transactionSpy).toHaveBeenCalledTimes(1);
+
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
```diff
@@ -1537,6 +1610,8 @@ export class HeaderFooterSessionManager {
    * Clean up all resources.
    */
   destroy(): void {
+    this.#teardownActiveEditorEventBridge();
```
caio-pizzol, line 1606: the `resetSession` callback in `initialize()` sets `#activeEditor = null` without calling `#teardownActiveEditorEventBridge()` first. every other place that clears `#activeEditor` does the teardown. not a bug since the editor is already destroyed, but the leftover cleanup ref could log a warning next time around. worth adding for consistency?

packages/super-editor/src/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
```diff
@@ -928,6 +949,58 @@ export class HeaderFooterSessionManager {
...
+      });
+    };
+
+    editor.on('update', emitSurfaceUpdate);
+    editor.on('transaction', emitSurfaceTransaction);
+
+    this.#activeEditorEventCleanup = () => {
+      try {
+        editor.off?.('update', emitSurfaceUpdate);
+      } catch (error) {
```
caio-pizzol: the individual try/catch blocks around each `editor.off()` are heavier than anywhere else in the codebase — other spots just call `editor.off()` directly. the outer try/catch in `#teardownActiveEditorEventBridge` already covers this. ok to simplify?

packages/superdoc/src/SuperDoc.vue
```diff
@@ -299,8 +307,52 @@ const onEditorDocumentLocked = ({ editor, isLocked, lockedBy }) => {
   proxy.$superdoc.lockSuperdoc(isLocked, lockedBy);
 };
 
-const onEditorUpdate = ({ editor }) => {
-  proxy.$superdoc.emit('editor-update', { editor });
+const buildEditorUpdatePayload = ({
```
caio-pizzol: these two builder functions (`buildEditorUpdatePayload` and `buildEditorTransactionPayload`) do the same thing for the first 5 fields — only `transaction` and `duration` differ. could pull the shared part into one helper to keep them in sync.

packages/super-editor/src/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
luccas-harbour: sounds good, done

packages/super-editor/src/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
luccas-harbour: done, I simplified this

packages/superdoc/src/SuperDoc.vue
luccas-harbour: done, I refactored this to use a common helper

packages/superdoc/src/SuperDoc.vue
```diff
@@ -299,8 +307,43 @@ const onEditorDocumentLocked = ({ editor, isLocked, lockedBy }) => {
...
+  return {
+    editor: effectiveEditor,
+    sourceEditor: sourceEditor ?? effectiveEditor,
+    surface,
+    headerId,
+    sectionType,
+  };
+};
+
+const buildEditorUpdatePayload = (payload = {}) => {
```
caio-pizzol, line 327: `buildEditorUpdatePayload` just passes through to `buildEditorPayloadBase` without adding anything — could skip the wrapper and call the base directly. worth it?

packages/super-editor/src/core/presentation-editor/tests/PresentationEditor.test.ts
```diff
@@ -2088,6 +2088,200 @@ describe('PresentationEditor', () => {
       boundingSpy.mockRestore();
     });
 
+    it('re-emits live header/footer child editor updates and transactions', async () => {
```
caio-pizzol, line 2091: the same ~25 lines of setup (create editor, wait for layout, mock viewport, add spies) are copied across all three new tests. a small helper would make these easier to read.

harbournick:

PR #2369, opened 2026-03-11
fix: improve multi-column rendering

VladaHarbour:
Initial issue:
SuperDoc does not match Word layout for early pages in a document that combines:
  1. Continuous section breaks.
  2. A manual column break before the first custom two-column section.
  3. An asymmetric first two-column setup.
  
Screen Shot 2026-02-25 at 1 30 05 PM


packages/layout-engine/layout-engine/src/column-balancing.test.ts
```diff
@@ -290,9 +290,9 @@ describe('shouldSkipBalancing', () => {
   });
 
   it('should NOT skip when content can be meaningfully distributed', () => {
-    // 100px total / 2 columns = 50px per column, above minColumnHeight (20px)
+    // 100px total exceeds the available single-column height (80px), so balancing is needed.
```
caio-pizzol: this comment is wrong — `shouldSkipBalancing` never checks `availableHeight`. the test passes because both values are above `minColumnHeight`.

```suggestion
    // 100px total / 2 columns = 50px per column, above minColumnHeight (20px) → balancing proceeds
```

packages/layout-engine/layout-bridge/src/incrementalLayout.ts
```diff
@@ -128,12 +129,40 @@ const COLUMN_EPSILON = 0.01;
 type NormalizedColumns = ColumnLayout & { width: number };
 type PageColumns = NormalizedColumns & { left: number; contentWidth: number };
```
caio-pizzol, line 132: this function is almost copy-pasted from `normalizeColumns` in `index.ts:2567`. could they share one helper?

packages/layout-engine/layout-bridge/src/incrementalLayout.ts
```diff
@@ -128,12 +129,40 @@ const COLUMN_EPSILON = 0.01;
...
 
+const resolveMaxColumnWidth = (contentWidth: number, columns?: ColumnLayout): number => {
```
caio-pizzol, line 135: footnotes are always measured using the widest column's width. if a footnote lands in a narrower column, its wrapping might be slightly off. maybe add a `// TODO` so it doesn't catch someone off guard later?

caio-pizzol (commented):
@VladaHarbour looks good overall. left a few small inline comments.

one test comment at `column-balancing.test.ts:293` describes a check that doesn't exist — easy fix.

`normalizeColumnsForFootnotes` and `normalizeColumns` are mostly the same ~30 lines — a shared helper would keep them in sync.

on tests: `widthsEqual` and `cloneColumnLayout` have a few branches but no unit tests. also nothing covers the case where two sections share count/gap but differ only in `widths` or `equalWidth`.

packages/layout-engine/layout-bridge/src/incrementalLayout.ts
VladaHarbour: created shared helper

harbournick:

PR #2370, opened 2026-03-11
fix(rendering): show comment highlight on text with Word highlight formatting (SD-2188)

caio-pizzol:
The `!textRun.highlight` guard in the renderer prevented comment highlights from being applied when text also had Word highlight formatting (`<w:highlight>`). This caused comments on highlighted text to be invisible. Removed the guard so comment highlights always take precedence, matching Word's behavior.


harbournick:

PR #2371, opened 2026-03-11
fix: faulty TOC import/export (SD-2183)

luccas-harbour:
## Summary

Fixes the DOCX export/import issues behind SD-2183 for table of contents content controls.

This PR addresses two related problems:

- exported TOC content controls could emit invalid/empty w:id values
- TOC field instructions stored inside a content control were not reliably round-tripped, which caused the exported document to lose the w:fldChar / w:instrText structure Word expects

### 1. Prevent empty SDT IDs during export

- stop emitting w:id for document part SDTs when the value is empty
- sanitize passthrough sdtPr data so empty w:id nodes are not re-exported

### 2. Handle complex fields stored in a single run

- update preProcessNodesForFldChar to correctly process fields when begin, instrText, separate, and end are all stored in the same w:r
- preserve unknown-field fallback behavior for those compressed single-run cases
- add regression coverage for:
  - TOC fields in a single run
  - unknown fields in a single run
  - drawing/pict content inside active field collection
  - nested/recursive field boundaries

### 3. Preserve TOC structure inside document part objects

- update TOC docPartObject import to hoist sd:tableOfContents out of wrapper paragraphs
- keep the imported PM structure aligned with what the layout/export pipeline expects:
  - heading paragraph
  - nested tableOfContents block
- avoid creating empty wrapper paragraphs when a paragraph only contains pPr plus the TOC block
- relax the tableOfContents node content model from paragraph+ to paragraph* so empty imported TOCs remain valid

## Why

Word expects the TOC field to round-trip as a real complex field sequence. Without that, exported files can lose the TOC field markers and fail to behave correctly when reopened in Word.

### DIFF
--- a/packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/handle-doc-part-obj.test.js
+++ b/packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/handle-doc-part-obj.test.js
@@ -268,6 +268,53 @@ describe('tableOfContentsHandler', () => {
     expect(handler).toHaveBeenCalledTimes(1);
   });
 
+  it('preserves paragraph content when a wrapper paragraph contains text and sd:tableOfContents', () => {
+    const sdtPr = {
+      name: 'w:sdtPr',
+      elements: [
+        { name: 'w:id', attributes: { 'w:val': '456' } },
+        {
+          name: 'w:docPartObj',
+          elements: [{ name: 'w:docPartGallery', attributes: { 'w:val': 'Table of Contents' } }],
+        },
+      ],
+    };
+    const contentNode = {
+      name: 'w:sdtContent',
+      elements: [
+        {
+          name: 'w:p',
+          elements: [
+            { name: 'w:pPr', elements: [] },
+            { name: 'w:r', elements: [{ name: 'w:t', elements: [{ type: 'text', text: 'Intro text' }] }] },
+            { name: 'sd:tableOfContents', attributes: { instruction: 'TOC \\o "1-1" \\h \\z \\u' }, elements: [] },
+          ],
+        },
+      ],
+    };
+    const handler = vi.fn(({ nodes }) => {
+      const node = nodes[0];
+      if (node.name === 'sd:tableOfContents') {
+        return [{ type: 'tableOfContents', attrs: { instruction: node.attributes.instruction }, content: [] }];
+      }
+      return [{ type: 'paragraph', content: [{ type: 'text', text: 'Intro text' }] }];
+    });
+    const params = {
+      nodes: [contentNode],
+      nodeListHandler: { handler },
+      extraParams: { sdtPr },
+      path: [],
+    };
+
+    const result = tableOfContentsHandler(params);
+
+    expect(result.content).toEqual([
+      { type: 'paragraph', content: [{ type: 'text', text: 'Intro text' }] },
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/core/super-converter/field-references/preProcessNodesForFldChar.js
```diff
@@ -161,27 +181,38 @@ export const preProcessNodesForFldChar = (nodes = [], docx) => {
...
       collectedNodesStack[collectedNodesStack.length - 1].push(node);
-      rawCollectedNodesStack[rawCollectedNodesStack.length - 1].push(rawNode);
+      captureRawNodeForCurrentField(rawNode, capturedRawStacks);
     } else {
       processedNodes.push(node);
     }
+  };
+
+  for (const node of nodes) {
+    const rawNode = carbonCopy(node);
```
caio-pizzol, line 227: when a field is unhandled, `rawCollectedNodes` contains the full original run as fallback. if a single `w:r` held two field sequences (e.g., unknown + TOC), the raw fallback for the first would include the second field's markers, then the second field also emits its processed result — duplicating content. very unlikely in real files but worth a note or a guard.

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-document-part-obj.js
```diff
@@ -46,30 +53,39 @@ function generateSdtPrForDocPartObj(attrs) {
...
+  const id = sanitizeId(attrs.id ?? attrs.sdtPr?.elements?.find((el) => el.name === 'w:id')?.attributes?.['w:val']);
   // Per OOXML spec: presence of w:docPartUnique element = true, absence = false
   const docPartUnique =
     attrs.docPartUnique ?? existingDocPartObj?.elements?.some((el) => el.name === 'w:docPartUnique') ?? false;
 
   // If we do not know the gallery type, prefer full passthrough to avoid emitting invalid XML
   if (docPartGallery === null) {
     if (attrs.sdtPr) {
+      // Filter out any w:id elements with empty values to avoid invalid XML, but preserve all other passthrough elements
+      if (attrs.sdtPr.elements && Array.isArray(attrs.sdtPr.elements)) {
```
caio-pizzol: every other translator (`translate-document-section.js:82`, `translate-field-annotation.js:84`) creates a filtered copy of sdtPr elements. this branch mutates in place. even though `sdtPr` isn't currently a live PM attribute here, using a copy keeps the pattern consistent.

```suggestion
      const filteredSdtPr = {
        ...attrs.sdtPr,
        elements: attrs.sdtPr.elements.filter(
          (el) => !(el.name === 'w:id' && el.attributes?.['w:val']?.trim() === ''),
        ),
      };
      return filteredSdtPr;
```

packages/super-editor/src/core/super-converter/field-references/preProcessNodesForFldChar.js
luccas-harbour: good call, I adjusted the logic to account for this.

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-document-part-obj.js
luccas-harbour: this was fixed

harbournick:

PR #2378, opened 2026-03-12
fix: handle null declaration in DocxExporter to prevent export crash

aldrinjenson:
## Summary

- Fixes `TypeError: Cannot read properties of null (reading 'attributes')` crash when exporting a SuperDoc to DOCX
- The crash occurs in `DocxExporter.#generate_xml_as_list` when `converter.declaration` is null — which happens for documents created/edited via Yjs collaboration that were never initialized from raw XML
- Adds optional chaining and a standard XML declaration fallback (`version="1.0" encoding="UTF-8" standalone="yes"`)

## Root Cause

`exporter.js` line 580 accessed `this.converter.declaration.attributes` without a null check. `SuperConverter.declaration` is initialized as `null` in the constructor and only populated in `parseFromXml()`, which isn't called for Yjs-only documents.

## Test plan

- [x] Added unit tests for null and undefined declaration cases
- [ ] Verify exporting the specific failing superdoc to DOCX produces a valid file that opens in Word
- [ ] Confirm the fallback encoding (UTF-8 uppercase) is compatible with downstream consumers

### DIFF
--- a/packages/super-editor/src/core/super-converter/exporter.js
+++ b/packages/super-editor/src/core/super-converter/exporter.js
@@ -577,7 +577,11 @@ export class DocxExporter {
 
   #generate_xml_as_list(data, debug = false) {
     const json = JSON.parse(JSON.stringify(data));
-    const declaration = this.converter.declaration.attributes;
+    const declaration = this.converter.declaration?.attributes ?? {
+      version: '1.0',
+      encoding: 'UTF-8',
+      standalone: 'yes',
+    };
     const xmlTag = `<?xml${Object.entries(declaration)
       .map(([key, value]) => ` ${key}="${value}"`)
       .join('')}?>`;
--- a/packages/super-editor/src/tests/export/docxExporter.test.js
+++ b/packages/super-editor/src/tests/export/docxExporter.test.js
@@ -41,6 +41,48 @@ describe('DocxExporter', () => {
     expect(xml).toContain('Format=&lt;&lt;NUM&gt;&gt;_&lt;&lt;VER&gt;&gt;');
   });
 
+  describe('null declaration handling', () => {
+    it('uses default XML declaration when converter.declaration is null', () => {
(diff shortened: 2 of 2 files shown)


packages/super-editor/src/core/super-converter/exporter.js
```diff
@@ -577,7 +577,11 @@ export class DocxExporter {
 
   #generate_xml_as_list(data, debug = false) {
     const json = JSON.parse(JSON.stringify(data));
-    const declaration = this.converter.declaration.attributes;
+    const declaration = this.converter.declaration?.attributes ?? {
```
caio-pizzol, line 580: this same fallback object is copy-pasted in 11+ files across the converter. there's already a `constants.js` next door — move it there and import it?

```js
import { DEFAULT_XML_DECLARATION } from './constants.js';
// ...
const declaration = this.converter.declaration?.attributes ?? DEFAULT_XML_DECLARATION.attributes;
```

packages/super-editor/src/tests/export/docxExporter.test.js
```diff
@@ -41,6 +41,48 @@ describe('DocxExporter', () => {
     expect(xml).toContain('Format=&lt;&lt;NUM&gt;&gt;_&lt;&lt;VER&gt;&gt;');
   });
 
+  describe('null declaration handling', () => {
```
caio-pizzol, line 44: these two tests are the same except for `null` vs `undefined`. could be one test with `it.each` — less to read.

caio-pizzol:
@aldrinjenson thanks for catching this\! your fix was correct — we built on it in #2578 with a small cleanup (reused an existing constant instead of copying it again). your commit is preserved there so you keep credit. closing this one in favor of that. appreciate the help\!

aldrinjenson:
Oh wow. Thank You @caio-pizzol !

PR #2379, opened 2026-03-12
fix(super-editor): guard against style definition nodes without elements

luccas-harbour:
## Summary

Prevents getDefaultStyleDefinition() from crashing when a matching style node exists in word/styles.xml but does not contain an elements array.

## What changed

- Added an early return in packages/super-editor/src/core/super-converter/docx-helpers/get-default-style-definition.js when the matched w:style node has no elements
- Added a regression test in packages/super-editor/src/core/super-converter/docx-helpers/get-default-style-definition.test.js covering that malformed/incomplete style-
definition shape
- Verified the helper returns the minimal fallback object and does not attempt to parse marks in this case

## Why

Some DOCX inputs can include a style definition node with the expected w:styleId but without child elements. Before this change, the helper assumed firstMatch.elements always existed and could throw while traversing the style definition. This makes the converter more defensive and preserves the existing fallback behavior for incomplete style data.


harbournick:

PR #2380, opened 2026-03-12
fix: import of w:sdt nested in paragraph (SD-2190)

luccas-harbour:
## Summary

Fix DOCX import failures caused by `w:sdt` `w:docPartObj` nodes being parsed as block content inside `paragraph` nodes.

This change makes paragraph import resilient when a block-level docPart SDT appears inside `w:p`, especially for cases like Table of Figures / Table of Contents placeholders at the start of a document.

## Problem

Some documents contain XML like:

```xml
<w:p>
  <w:sdt>
    <w:sdtPr>
      <w:docPartObj>
        <w:docPartGallery w:val="Table of Figures"/>
      </w:docPartObj>
    </w:sdtPr>
    <w:sdtContent>...</w:sdtContent>
  </w:sdt>
</w:p>
```

Our importer translated that `w:sdt` into a `documentPartObject` block node, but still attempted to keep it inside `paragraph.content`.

Since `paragraph` only allows inline content, import could fail with:

```text
Invalid content for node type paragraph
```

## What changed

- Added a shared `isInlineNode()` helper for schema-aware inline/block classification.
- Strengthened the fallback inline detection used when schema metadata is unavailable.
- Updated paragraph import normalization so block-producing children are hoisted out of `w:p` instead of being inserted into `paragraph.content`.
- Split mixed paragraph content into sibling nodes when needed:
  - `paragraph` fragments for inline runs
  - block nodes for hoisted content like `documentPartObject`
- Adjusted paragraph attribute distribution for split results:
  - `paraId` / `textId` stay on only one paragraph fragment
  - `rsid*` attrs are still copied to paragraph fragments
- Preserved wrapper paragraph metadata on block-only hoisted `documentPartObject` nodes so paragraph formatting can round-trip through export.
- Re-export wrapped `w:p > w:sdt` when wrapper paragraph metadata is present.
- Preserved `sectPr` placement for hoisted trailing/block-only docPart cases.

## Tests

Added and updated regressions for:

- block-only `docPartObj` inside `w:p`
- mixed inline text around `docPartObj`
- no-schema import paths
- split paragraph identity handling
- block-only wrapper paragraph formatting
- docPart export with preserved paragraph wrapper metadata

## Limitations

The following limitations are known and accepted for now:

- Wrapper paragraph metadata preserved on `documentPartObject` is primarily used for round-trip/export and is not fully reflected in rendering behavior.
- Some paragraph-level attrs may still be duplicated across split `paragraph` fragments in edge cases.
- The no-schema inline fallback is still heuristic-based and may not cover every inline media type.

### DIFF
--- a/packages/super-editor/src/extensions/structured-content/document-part-object.js
+++ b/packages/super-editor/src/extensions/structured-content/document-part-object.js
@@ -44,7 +44,7 @@ export const DocumentPartObject = Node.create({
       id: {},
       docPartGallery: {},
       docPartUnique: {
-        default: true,
+        default: false,
       },
       wrapperParagraph: {
         default: null,


luccas-harbour:
About the OOXML Spec Review failure above, the docx file that originated this PR has exactly the structure that the report is saying isn't valid but was created with Word:

```xml
<w:p w14:paraId="41964671" w14:textId="04598795" w:rsidR="00233D7B" w:rsidRPr="003104CE" w:rsidRDefault="00233D7B" w:rsidP="003104CE">
  <w:sdt>
    <w:sdtPr>
      <w:id w:val="123456789"/>
      <w:docPartObj>
        <w:docPartGallery w:val="Table of Figures"/>
        <w:docPartUnique/>
      </w:docPartObj>
    </w:sdtPr>
    <w:sdtContent>
      <w:p w14:paraId="11111111" w14:textId="11111111">
        <w:r>
          <w:t>Table of Figures</w:t>
        </w:r>
      </w:p>
      <w:p w14:paraId="22222222" w14:textId="22222222">
        <w:r>
          <w:t>Figure 1</w:t>
        </w:r>
        <w:r>
          <w:tab/>
        </w:r>
        <w:r>
          <w:t>1</w:t>
        </w:r>
      </w:p>
    </w:sdtContent>
  </w:sdt>
</w:p>
```

Therefore I don't think the analysis result is valid.

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-document-part-obj.js
```diff
@@ -27,7 +28,37 @@ export function translateDocumentPartObj(params) {
...
+
+  if (pPr) {
+    elements.push(pPr);
+  }
+  elements.push(sdtNode);
+
+  const attributes = {};
+  if (wrapperParagraphAttrs?.rsidRDefault) {
+    attributes['w:rsidRDefault'] = wrapperParagraphAttrs.rsidRDefault;
+  }
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-document-part-obj.js
```diff
@@ -27,7 +29,56 @@ export function translateDocumentPartObj(params) {
...
+    elements,
+    ...(Object.keys(attributes).length ? { attributes } : {}),
+  };
+}
+
+function extractRawParagraphXmlAttributes(attrs = {}) {
+  return Object.fromEntries(Object.entries(attrs).filter(([key]) => key.includes(':')));
+}
+
+function decodeParagraphXmlAttributes(attrs = {}) {
```
caio-pizzol: `decodeParagraphXmlAttributes` does the same thing as `translator.decodeAttributes()` from p-translator — same loop, same attribute configs. you could replace this with `translator.decodeAttributes({ node: { attrs: wrapperParagraphAttrs } })` and drop the extra function + import. worth doing?

caio-pizzol:
> About the OOXML Spec Review failure above, the docx file that originated this PR has exactly the structure that the report is saying isn't valid but was created with Word:
> 
> ```
> <w:p w14:paraId="41964671" w14:textId="04598795" w:rsidR="00233D7B" w:rsidRPr="003104CE" w:rsidRDefault="00233D7B" w:rsidP="003104CE">
>   <w:sdt>
>     <w:sdtPr>
>       <w:id w:val="123456789"/>
>       <w:docPartObj>
>         <w:docPartGallery w:val="Table of Figures"/>
>         <w:docPartUnique/>
>       </w:docPartObj>
>     </w:sdtPr>
>     <w:sdtContent>
>       <w:p w14:paraId="11111111" w14:textId="11111111">
>         <w:r>
>           <w:t>Table of Figures</w:t>
>         </w:r>
>       </w:p>
>       <w:p w14:paraId="22222222" w14:textId="22222222">
>         <w:r>
>           <w:t>Figure 1</w:t>
>         </w:r>
>         <w:r>
>           <w:tab/>
>         </w:r>
>         <w:r>
>           <w:t>1</w:t>
>         </w:r>
>       </w:p>
>     </w:sdtContent>
>   </w:sdt>
> </w:p>
> ```
> 
> Therefore I don't think the analysis result is valid.

agreed - Word creates this structure so we should preserve it on export. changing it would risk breaking things when the file is reopened.

one thing worth noting though: the `docPartUnique` defaulting to true is a real issue. it's not from this PR, but since more nodes now go through this export path it matters a bit more. maybe worth a follow-up ticket?

packages/super-editor/src/core/super-converter/v3/handlers/w/sdt/helpers/translate-document-part-obj.js
luccas-harbour: Sounds good, I refactored this. I also changed the default for `docPartUnique` to fix the other issue

PR #2381, opened 2026-03-12
feat: add toolbar checkboxes example

mattConnHarbour:


caio-pizzol:
@mattConnHarbour really good examples, clear samples that cover the full surface well.

I want to hold off on merging this into `examples/features/` though. that folder is for core SuperDoc features that many customers use. checkboxes is one use case of the content controls API, not a standalone feature. let's chat about where this should live - I'll send you a message.

PR #2383, opened 2026-03-12
fix(painter-dom): skip non-scrollable scroll container in virtualization (SD-2199)

tupizz:
## Demo

https://github.com/user-attachments/assets/9188746f-3e05-42df-95da-a1685d1d9c61

## Summary

Fixes [SD-2199](https://linear.app/superdocworkspace/issue/SD-2199): only the first 7 pages rendered on documents with 8+ pages when SuperDoc was embedded in certain layouts (version tester, manual testing site).

## Root Cause

In v1.18.0 we added scroll container detection (`#findScrollableAncestor` in PresentationEditor, `setScrollContainer` on DomPainter) to fix virtualization offsets when SuperDoc is mounted inside a scrollable wrapper div.

The detection walks up the DOM tree and picks the first ancestor with `overflow: auto` or `overflow: scroll` CSS. `updateVirtualWindow()` then reads that element's `scrollTop` to compute which pages to render.

The problem: **having `overflow: auto` in CSS does not mean the element actually scrolls.** In flex layouts where the parent only has `min-height` (not `height`), the child grows to fit its content instead of constraining it. The element never scrolls, `scrollTop` stays at 0, and the virtual window is permanently stuck at pages 0-6 (window=5, overscan=1).

This is the exact layout in both affected environments:

```
.editorDock       display: flex; min-height: calc(100vh - 120px)  <-- NO height
  .vt-editor-root flex: 1; overflow: auto                        <-- grows to fit, never scrolls
    #superdoc
```

The local dev app works because `.dev-app__layout` has `height: 100vh` (fixed), so `.dev-app__main` is properly height-constrained and actually scrolls.

## The Fix

Added a runtime check in `updateVirtualWindow()` that verifies the scroll container is **actually scrollable** (`scrollHeight > clientHeight`) before using its `scrollTop`. When the container has `overflow: auto` but isn't actually overflowing, the code falls through to the viewport-based `getBoundingClientRect` calculation, which works correctly with window-level scrolling.

```typescript
// Before: trusted the scroll container blindly
} else if (this.scrollContainer) {

// After: verify content actually overflows the container
const scrollCont = this.scrollContainer;
const isScrollContainerActive = scrollCont != null
  && scrollCont.scrollHeight > scrollCont.clientHeight + 1;
} else if (isScrollContainerActive) {
```

### Why this approach

**Alternative considered: fix `#findScrollableAncestor` to check if the element is height-constrained.** This is fragile because at setup time content may not be laid out yet, and determining whether a flex child will be constrained requires walking the entire ancestor chain analyzing flex/grid properties, explicit heights, and viewport units. The detection would need to be repeated on every layout change.

**Why the runtime check is better:** `scrollHeight > clientHeight` is the ground truth. It directly answers "is this element scrollable right now?" on every scroll event, with zero overhead (two property reads). It self-heals if the layout changes dynamically (e.g., window resize causes the container to become constrained). No assumptions about CSS layout models needed.

## Test Plan

- [x] Added regression test: "falls through to viewport-based calculation when scroll container is not actually scrollable (SD-2199)"
- [x] All 784 painter-dom tests pass (including existing scroll container tests)
- [x] Manually verified on the version tester (labs.superdoc.dev layout) with a local build: all pages render correctly on 8+ page documents
- [ ] Verify on the GCS manual testing site after publish
- [ ] Verify on labs.superdoc.dev version tester after publish

### DIFF
--- a/packages/layout-engine/painters/dom/src/renderer.ts
+++ b/packages/layout-engine/painters/dom/src/renderer.ts
@@ -1799,9 +1799,16 @@ export class DomPainter {
     const zoom = this.zoomFactor;
     let scrollY: number;
     const isContainerScrollable = this.mount.scrollHeight > this.mount.clientHeight + 1;
+    // Check if the external scroll container is actually scrollable (content overflows its
+    // visible area). An element can have overflow:auto but still not scroll if it's in an
+    // unconstrained flex layout where the parent has only min-height (no height). In that
+    // case the element grows to fit content and scrollTop stays 0 — fall through to the
+    // viewport-based calculation instead.
+    const scrollCont = this.scrollContainer;
+    const isScrollContainerActive = scrollCont != null && scrollCont.scrollHeight > scrollCont.clientHeight + 1;
     if (isContainerScrollable) {
       scrollY = Math.max(0, this.mount.scrollTop - paddingTop);
-    } else if (this.scrollContainer) {
+    } else if (isScrollContainerActive) {
       // Intermediate scroll ancestor (e.g., a wrapper div with overflow-y: auto).
       // Use scrollContainer.scrollTop with a cached mount offset instead of
       // getBoundingClientRect(). Rects are affected by spacer DOM mutations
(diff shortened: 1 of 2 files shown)


harbournick (approved):
LGTM

harbournick:

PR #2384, opened 2026-03-13
Fix: Don't iterate over every attributeto find w:sectPr

johanneswilm:
@harbournick I noticed you wrote this code a while back. I am not entirely sure why it was written like that - it creates this problem and so I scaled the walking mechanism back. That should save time as well. It may though be that you are trying to catch more than what I do with this. So please have a look and fix it in another way if needed. What it was doing was comparing with all values of all text fields and all other attributes, which surely isn't what we wanted.

Second reviewer @caio-pizzol

### DIFF
--- a/packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
+++ b/packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
@@ -761,12 +761,11 @@ const importHeadersFooters = (docx, converter, mainEditor, numbering, translated
 };
 
 const findSectPr = (obj, result = []) => {
-  for (const key in obj) {
-    if (obj[key] === 'w:sectPr') {
-      result.push(obj);
-    } else if (typeof obj[key] === 'object') {
-      findSectPr(obj[key], result);
-    }
+  if (obj && obj.name === 'w:sectPr') {
+    result.push(obj);
+  }
+  if (obj.elements) {
+    obj.elements.forEach((el) => findSectPr(el, result));
   }
   return result;
 };


johanneswilm:
@caio-pizzol: Not sure why this was assigned to me, as unfortunately I don't have the access rights to merge it. But unless you can see something wrong with it, please go ahead and merge it.

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
```diff
@@ -761,12 +761,11 @@ const importHeadersFooters = (docx, converter, mainEditor, numbering, translated
...
-    if (obj[key] === 'w:sectPr') {
-      result.push(obj);
-    } else if (typeof obj[key] === 'object') {
-      findSectPr(obj[key], result);
-    }
+  if (obj && obj.name === 'w:sectPr') {
+    result.push(obj);
+  }
+  if (obj.elements) {
+    obj.elements.forEach((el) => findSectPr(el, result));
```
caio-pizzol, line 768: line 764 checks if `obj` exists, but this line doesn't — crashes if `obj` is null. the old code handled that fine.

```suggestion
  if (obj?.elements) {
```

johanneswilm: @caio-pizzol under what scenarios would `obj` not exist? As far as I can tell, that shouldn't ever happen.

johanneswilm:
@caio-pizzol 

1. I answered that inline. It's possible that I overlooked something, but I cannot think of a situation where obj doesn't exist. Hence I removed the check for that.

2. The description of the issue in **SD-1855** in tries to go into the possible causes of this based on the error message. However, it turns out that it was something entirely different in the case they found. Do we know for certain whether this can still be triggered once this fix is in place? If so, let's file another issue with that.

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
johanneswilm: the only places `findSectPr()` is called is within itself on all the children - in which case obj will exist. And externally on `docx["word/document.xml"]` and if that object doesn't exist, then I'd assume we would crash much sooner then here when we are scanning for the sections.

caio-pizzol:
1. agreed — not a blocker.

2. that makes sense — if the import was failing because of the old traversal, fixing it would set `this.numbering` correctly and the export crash goes away. I'll file a separate issue for guarding the export path in case import fails for other reasons.

packages/super-editor/src/core/super-converter/v2/importer/docxImporter.js
caio-pizzol: fair enough — the old `for...in` just happened to skip null silently, and `obj.elements` wouldn't, but it's not a realistic scenario. not worth blocking on.

harbournick:

PR #2385, opened 2026-03-13
fix: clear selection on undo/redo

VladaHarbour:


packages/super-editor/src/extensions/history/history.js
```diff
@@ -55,7 +88,8 @@ export const History = Extension.create({
           return yUndo(state);
```
caio-pizzol: collab undo (`yUndo`) returns before this cleanup runs, so the ghost highlight bug would still happen in collab sessions. does it need to be handled there too?

packages/super-editor/src/extensions/history/history.js
```diff
@@ -1,7 +1,40 @@
...
+  return (historyTr) => {
+    let cleared = historyTr.setMeta(CustomSelectionPluginKey, {
+      focused: false,
+      preservedSelection: null,
+      showVisualSelection: false,
+      skipFocusReset: false,
+    });
+
+    const sel = cleared.selection;
+    if (sel && sel instanceof TextSelection && !sel.empty) {
```
caio-pizzol: this removes the text selection after every undo/redo. most editors keep the restored text selected so users see what changed. is that ok here, or could we skip this part and just clear the highlight state (steps 1 and 3)?

packages/super-editor/src/extensions/history/history.js
```diff
@@ -1,7 +1,40 @@
...
 import { history, redo as originalRedo, undo as originalUndo } from 'prosemirror-history';
 import { undo as yUndo, redo as yRedo, yUndoPlugin } from 'y-prosemirror';
 import { Extension } from '@core/Extension.js';
+import { CustomSelectionPluginKey } from '../custom-selection/custom-selection.js';
+
+function createHistoryDispatch(editor, dispatch) {
+  if (!dispatch) return dispatch;
+
+  return (historyTr) => {
+    let cleared = historyTr.setMeta(CustomSelectionPluginKey, {
```
caio-pizzol: this object is the same as `DEFAULT_SELECTION_STATE` in `custom-selection.js`. if one changes and the other doesn't, they'll get out of sync. worth importing it instead?

packages/super-editor/src/extensions/history/history.js
```diff
@@ -1,7 +1,40 @@
...
+      focused: false,
+      preservedSelection: null,
+      showVisualSelection: false,
+      skipFocusReset: false,
+    });
+
+    const sel = cleared.selection;
+    if (sel && sel instanceof TextSelection && !sel.empty) {
+      const headPos = typeof sel.head === 'number' ? sel.head : sel.to;
```
caio-pizzol: after the `instanceof TextSelection` check above, `sel.head` is always a number, so the `typeof` check and `sel.to` fallback never run. can simplify to just `TextSelection.create(cleared.doc, sel.head)`.

packages/super-editor/src/extensions/custom-selection/custom-selection.js
```diff
@@ -378,6 +378,14 @@ export const CustomSelection = Extension.create({
                   skipFocusReset: false,
                 }),
               );
+
+              // Also clear editor-level preserved selection snapshots so that
+              // subsequent commands (linked styles, mark commands, etc.) don't
+              // resurrect an old selection after history undo/redo.
+              this.editor.setOptions({
```
caio-pizzol, line 385: this new clearing code doesn't have a test. the existing blur tests don't hit this branch.

caio-pizzol (commented):
@VladaHarbour good approach — wrapping undo/redo to clear the ghost highlight makes sense.

main question: the fix also collapses the cursor after every undo/redo, not just the ghost highlight case. could it work without that part? also, the collab undo path skips the fix entirely. left inline comments with details.

on DX: one object literal duplicates `DEFAULT_SELECTION_STATE` (drift risk if the shape changes), and a dead `typeof` guard can be simplified — both minor.

test coverage: redo path uses the same wrapper but has no test, and the new `setOptions` call in the non-toolbar blur branch isn't covered.

packages/super-editor/src/extensions/history/history.js
VladaHarbour: There are some limitations of selection in Superdoc. If we want the visual highlight to disappear after undo/redo, we must either:
1. Keep the collapse logic in applySelectionCleanup for history transactions, accepting that SuperDoc behaves a bit differently from some editors (selection turns into a caret after undo/redo).
2. Introduce a new, explicit flag (e.g. selectionHidden on editor/presentation state) and make #updateSelection() skip drawing selection rects when that flag is set, while leaving the PM selection range intact. That’s a deeper change touching PresentationEditor and selection rendering logic.

I won't make any bigger changes for now. Waiting for your reply

packages/super-editor/src/extensions/history/history.js
VladaHarbour: fixed

packages/super-editor/src/extensions/history/history.js
VladaHarbour: fixed

packages/super-editor/src/extensions/history/history.js
VladaHarbour: fixed

packages/super-editor/src/extensions/custom-selection/custom-selection.js
VladaHarbour: added another test case

harbournick:

PR #2416, opened 2026-03-17
fix: seed base docx package for collaboration exports

luccas-harbour:
This fixes DOCX export for collaboration-backed editors that have document state but no original DOCX package to rebuild from.

Previously, export could fail in two steps:
- `DocxExporter` could crash when `converter.declaration` was missing
- after guarding that, `DocxZipper` could still fail with `docx is not iterable` because collaborative export assumed `options.content` contained original DOCX entries

This change makes export resilient when the editor was hydrated from JSON or collaboration state without a backing package. The editor now falls back to the blank DOCX template to synthesize base package entries before zipping, and the zipper emits a clear error if that contract is violated again.

Tests:
- added a regression test for the zipper missing-base-package path
- added an editor-level export regression covering missing base package entries before export


harbournick:

PR #2420, opened 2026-03-17
fix(evals): resolve SDK from local workspace with automatic prepare pipeline

harbournick:
## Summary
When running evals, we should be using the currently active repo / codebase to run the evals, not the SDK from npm. This PR enforces using the locally-built version of the SDK before running evals


harbournick:
@tupizz fyi

tupizz (approved):

PR #2437, opened 2026-03-18
fix(tracked-changes): undo/redo applies to both document and comment bubbles

palmer-cl:
- Added undo/redo tracked-change resync in SuperDoc.vue so history replay restores both tracked-change anchors and the comment bubble

- Updated tracked-change comment reconciliation in comments-store.js and bubble actions in CommentDialog.vue so accepted/rejected tracked changes reopen correctly on undo/redo and keyboard history works immediately

### DIFF
--- a/packages/superdoc/src/stores/comments-store.js
+++ b/packages/superdoc/src/stores/comments-store.js
@@ -1096,6 +1096,7 @@ export const useCommentsStore = defineStore('comments', () => {
     if (!(liveTrackedChangeIds instanceof Set) || !activeDocumentId) return;
 
     const removedIds = new Set();
+    const restoredComments = [];
     const previousComments = [...commentsList.value];
 
     commentsList.value = commentsList.value.filter((comment) => {
@@ -1115,12 +1116,23 @@ export const useCommentsStore = defineStore('comments', () => {
         comment.resolvedTime = resolutionSnapshot.resolvedTime ?? Date.now();
         comment.resolvedByEmail = resolutionSnapshot.resolvedByEmail ?? null;
         comment.resolvedByName = resolutionSnapshot.resolvedByName ?? null;
+        restoredComments.push(comment);
         return true;
       }
 
       if (commentId) removedIds.add(commentId);
       if (importedId) removedIds.add(importedId);
       return false;
+    });
+
+    restoredComments.forEach((comment) => {
+      const payload = getCommentEventPayload(comment);
+      const event = {
+        type: COMMENT_EVENTS.UPDATE,
+        comment: payload,
+      };
+      syncCommentsToClients(superdoc, event);
+      superdoc?.emit?.('comments-update', event);
     });
 
     if (!removedIds.size) return;
--- a/packages/superdoc/src/stores/comments-store.test.js
+++ b/packages/superdoc/src/stores/comments-store.test.js
@@ -1066,6 +1066,22 @@ describe('comments-store', () => {
     expect(existingComment.resolvedTime).toBe(555);
     expect(existingComment.resolvedByEmail).toBe('reviewer@example.com');
     expect(existingComment.resolvedByName).toBe('Reviewer');
+
+    // Should emit UPDATE so collaborators see the re-resolved state
+    expect(syncCommentsToClientsMock).toHaveBeenCalledWith(
+      superdoc,
+      expect.objectContaining({
+        type: comments_module_events.UPDATE,
+        comment: expect.objectContaining({ commentId: 'tc-snapshot-restore' }),
+      }),
+    );
+    expect(superdoc.emit).toHaveBeenCalledWith(
+      'comments-update',
+      expect.objectContaining({
+        type: comments_module_events.UPDATE,
+        comment: expect.objectContaining({ commentId: 'tc-snapshot-restore' }),
+      }),
+    );
   });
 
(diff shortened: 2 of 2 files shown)


packages/superdoc/src/SuperDoc.vue
```diff
@@ -990,7 +990,10 @@ const onEditorTransaction = (payload = {}) => {
   // This could be extended to other listeners in the future
   if (inputType === 'historyUndo' || inputType === 'historyRedo') {
     const documentId = editor?.options?.documentId;
+    // separation is useful because functions have different inputs
+    // different side effects, and different reuse points.
     syncTrackedChangePositionsWithDocument({ documentId, editor });
+    syncTrackedChangeComments({ superdoc: proxy.$superdoc, editor });
```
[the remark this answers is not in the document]
palmer-cl: added event on prune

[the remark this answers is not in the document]
palmer-cl: emit a close event when redo prunes

packages/superdoc/src/stores/comments-store.js
```diff
@@ -966,15 +971,18 @@ export const useCommentsStore = defineStore('comments', () => {
...
+    // reopen them when the tracked marks reappear.
+    const existingUnresolvedIds = new Set();
     commentsList.value.forEach((comment) => {
       if (!comment?.trackedChange) return;
       if (!belongsToDocument(comment, activeDocumentId)) return;
-      if (comment.commentId != null) existingIds.add(String(comment.commentId));
-      if (comment.importedId != null) existingIds.add(String(comment.importedId));
+      if (comment.resolvedTime) return;
+      if (comment.commentId != null) existingUnresolvedIds.add(String(comment.commentId));
+      if (comment.importedId != null) existingUnresolvedIds.add(String(comment.importedId));
```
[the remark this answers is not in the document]
palmer-cl: fix path for imported tcs that are resolved

packages/superdoc/src/stores/comments-store.js
```diff
@@ -1110,6 +1128,24 @@ export const useCommentsStore = defineStore('comments', () => {
...
     }
 
+    const removedComments = previousComments.filter((comment) => {
+      if (!belongsToDocument(comment, activeDocumentId)) return false;
+      const commentId = comment.commentId != null ? String(comment.commentId) : null;
+      const importedId = comment.importedId != null ? String(comment.importedId) : null;
+      return (commentId && removedIds.has(commentId)) || (importedId && removedIds.has(importedId));
+    });
+
+    removedComments.forEach((comment) => {
```
caio-pizzol, line 1171: accept → undo → redo → undo breaks the thread:

1. undo reopens the thread (clears the "accepted" state)
2. redo sees a thread with no mark and no "accepted" flag → deletes it entirely
3. next undo can't find the old thread → creates a new one, old replies are lost

the original accept just marks the thread as done — redo should do the same, not delete it.

(verified at runtime — the thread ID disappears from commentsList after redo, and a fresh one gets created on the next undo)

packages/superdoc/src/stores/comments-store.js
```diff
@@ -960,21 +968,31 @@ export const useCommentsStore = defineStore('comments', () => {
...
 
-    // Build a Set of existing tracked-change IDs for O(1) lookup.
+    // Build a Set of existing unresolved tracked-change IDs for O(1) lookup.
     // Include both runtime and imported IDs to avoid duplicate threads when
     // replay/import flows remap commentId but marks still reference importedId.
-    const existingIds = new Set();
+    // History replay can opt in to excluding resolved tracked-change threads so
+    // undo/redo reopens them when their marks reappear. Initial import rebuilds
+    // keep resolved IDs in the set so resolved DOCX threads do not reopen on load.
+    const existingUnresolvedIds = new Set();
```
caio-pizzol: this name is a bit misleading — resolved IDs also get added to this set (lines 989-990). something like `skipIds` would be clearer since the set means "don't create a thread for these."

```suggestion
    const skipIds = new Set();
```

(and rename below to match)

packages/superdoc/src/SuperDoc.vue
```diff
@@ -990,7 +990,10 @@ const onEditorTransaction = (payload = {}) => {
   // This could be extended to other listeners in the future
   if (inputType === 'historyUndo' || inputType === 'historyRedo') {
     const documentId = editor?.options?.documentId;
+    // separation is useful because functions have different inputs
```
caio-pizzol: the two function calls already show they're different — this comment doesn't add much. ok to drop?

```suggestion
```

packages/superdoc/src/stores/comments-store.js
palmer-cl: Good catch! This is fixed now.

packages/superdoc/src/stores/comments-store.js
palmer-cl: Updated in latest.

packages/superdoc/src/SuperDoc.vue
palmer-cl: Updated in latest.

caio-pizzol (approved):
@palmer-cl all three comments from last round are fixed. I pushed two commits — extra unit tests and a small fix so other users in a collab session see the correct state after redo.

approving

harbournick:

PR #2439, opened 2026-03-18
fix: allow for diacritical symbols in tcs

palmer-cl:
- Update tracked changes to allow for diacritical symbols in tracked changes
- add behavior tests


packages/super-editor/src/extensions/track-changes/trackChangesHelpers/trackedTransaction.js
```diff
@@ -16,7 +16,7 @@ import { CommentsPluginKey } from '../../comment/comments-plugin.js';
  * @returns {import('prosemirror-state').Transaction} Modified transaction.
  */
 export const trackedTransaction = ({ tr, state, user }) => {
-  const onlyInputTypeMeta = ['inputType', 'uiEvent', 'paste', 'pointer'];
+  const onlyInputTypeMeta = ['inputType', 'uiEvent', 'paste', 'pointer', 'composition'];
```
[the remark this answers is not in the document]
palmer-cl: added composition meta grouping

tests/visual/tests/fixtures/superdoc.ts
```diff
@@ -135,6 +137,40 @@ export const test = base.extend<{ superdoc: SuperDocFixture } & SuperDocOptions>
...
+            data: string,
+          ) => visibleHost.dispatchEvent(new CompositionEvent(type, { data, bubbles: true, cancelable: true }));
+
+          dispatchComposition('compositionstart', '');
+          dispatchComposition('compositionupdate', value);
+
+          hiddenEditor.focus();
+          const inserted = document.execCommand('insertText', false, value);
+          if (!inserted) {
+            throw new Error('document.execCommand("insertText") failed during composition simulation.');
```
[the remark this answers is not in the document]
palmer-cl: added more robust handling in the fixture

packages/super-editor/src/extensions/track-changes/trackChangesHelpers/trackedTransaction.js
caio-pizzol, line 316: typing an accented character (e.g. é on macOS with Option+E then E) inserts a leftover `'` character before the accent — so you get `'é` instead of `é`. this only happens after a space.

also, after typing the accent, pressing Enter adds spaces instead of creating a new line. the editor gets stuck and every keypress keeps behaving wrong until you click away.

confirmed by testing in the dev app.

palmer-cl: @caio-pizzol do you have any other repro steps on this? I wasnt able to reproduce it. 

Screen Cast 2026-03-18 at 8 54 11 PM

caio-pizzol: https://www.loom.com/share/640c778a64e8492b9befb720d7e1d836

palmer-cl: @caio-pizzol good find this was tricky. Only was happening on tracked changes and also only at beginning of a sequence.

Had to harden a few paths. 

- added a recovery path that force-ends ProseMirror’s composition if the browser sends a non-composing beforeinput

- made dead-key placeholder folding work even if the composition meta has already been cleared, so suggesting mode still replaces the tracked accent placeholder with the final tracked character 

😵‍💫

packages/super-editor/src/core/extensions/editable.js
```diff
@@ -86,13 +100,20 @@ export const Editable = Extension.create({
               return true;
             }
 
+            if (shouldForceEndStaleComposition(view, event)) {
+              __endComposition(view);
```
caio-pizzol, line 104: in suggesting mode, typing an accent at the start of a paragraph produces `'e` instead of `é`. the `shouldForceEndStaleComposition` guard seems to be ending the composition too early — the final character arrives without the accent.

from what I can tell, the tracked changes transaction rebuild makes the composition look stuck when it's still active. `pendingDeadKeyPlaceholder` might be a good signal to distinguish real stale compositions from this case. also worth checking if `˜` (U+02DC) needs to be added to the placeholder map — some keyboards use it instead of `~`.

tested in the dev app with debug logging on `shouldForceEndStaleComposition` and `normalizeCompositionInsertStep`.

palmer-cl: @caio-pizzol can you elaborate on which keys you are pressing? Or other specific repro steps?

Accent characters are working fine for me even at the beginning of the first paragraph. See attached. 

Screen Cast 2026-03-22 at 9 16 01 PM

caio-pizzol: I will get back to you tomorrow on this!

caio-pizzol (approved):
@palmer-cl tested again — the fix is working. the issue I saw was because of my keyboard layout (Brazilian Portuguese), where `'` works as a dead key. with a US keyboard it behaves the same as Word. approving.

packages/super-editor/src/core/extensions/editable.js
caio-pizzol: @palmer-cl issue was because of my keyboard layour (PT-BR) all good

PR #2441, opened 2026-03-18
fix: add backward compat aliases, fix hardcoded values, add theming docs

caio-pizzol:


apps/docs/getting-started/theming.mdx
```diff
@@ -0,0 +1,119 @@
...
+<html class="my-theme">
+  <div id="editor"></div>
+</html>
+```
+
+Every `--sd-*` CSS custom property can be overridden on any parent element. The changes cascade down to all child components.
+
+## Preset themes
+
+Three preset themes ship out of the box. Add the class to `<html>` or any wrapper element.
```
artem-harbour: Only to `<html>` element. We have elements that are appended to the body.

caio-pizzol: good call — fixed. theme class goes on <html> only.

packages/superdoc/src/assets/styles/helpers/compat.css
```diff
@@ -0,0 +1,81 @@
...
+  --sd-font-family: var(--sd-ui-font-family);
+
+  /* ─── Radius ─── */
+  --sd-radius-sm: var(--sd-radius-50);
+  --sd-radius-md: var(--sd-radius-200);
+  --sd-radius-lg: var(--sd-radius-300);
+
+  /* ─── Comment dialog ─── */
+  --sd-comment-bg: var(--sd-ui-comments-panel-card-background);
+  --sd-comment-bg-hover: var(--sd-ui-comments-panel-card-hover-background);
```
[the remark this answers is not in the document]
caio-pizzol: good catch — fixed. new tokens now fall back to old names via var(--sd-comment-bg, #f3f6fd) so existing overrides propagate automatically.

PR #2472, opened 2026-03-19
fix(track-changes): allow backspace on empty list items in suggesting mode (SD-2187)

tupizz:
## Demo
https://github.com/user-attachments/assets/6e2cd6bb-a706-40bd-bac4-74fa92bdb631

## Summary

- Fixes inability to delete an empty list bullet in suggesting mode after all text has been tracked-deleted
- Root cause: `replaceAroundStep` handler blocked structural changes (e.g. lift out of list) when `findPreviousLiveCharPos` returned null, leaving the user stuck on an undeletable empty bullet
- Fix: when no live characters remain, allow the structural `ReplaceAroundStep` to apply directly instead of blocking it

## Root Cause

The `replaceAroundStep` handler (introduced in PR #2225 / SD-2061) converts backspace `ReplaceAroundStep` operations into tracked single-character deletions. When all text in a list item is already marked as `trackDelete`, `findPreviousLiveCharPos()` returns `null` and the handler returned early, blocking the structural change entirely.

The fix allows the structural step to pass through when the content is effectively empty. Structural changes aren't tracked yet (as noted in the ticket), but blocking them is worse than applying them untracked.

## Test plan

- [x] All 847 super-editor test files pass (9214 tests, 0 failures)
- [x] Updated test case to verify structural step is attempted (not blocked) when no live characters exist
- [x] Verified in browser: delete all text in numbered list item in suggesting mode, then backspace removes the bullet
- [x] Manual: upload TC bulleted test.docx, switch to suggesting mode, delete list item text, backspace to remove bullet
- [x] Manual: verify same fix works for indented list items

### DIFF
--- a/packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceAroundStep.js
+++ b/packages/super-editor/src/extensions/track-changes/trackChangesHelpers/replaceAroundStep.js
@@ -5,6 +5,53 @@ import { TrackDeleteMarkName } from '../constants.js';
 import { TrackChangesBasePluginKey } from '../plugins/index.js';
 
 /**
+ * Check whether the enclosing structural scope (listItem, or paragraph
+ * when outside a list) contains any live (non-tracked-deleted) inline
+ * leaf node (text, tab, line break, image, etc.).
+ *
+ * @param {import('prosemirror-model').Node} doc
+ * @param {number} cursorPos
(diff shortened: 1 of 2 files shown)


harbournick:
:tada: This PR is included in **superdoc** v1.23.0

The release is available on [GitHub release](https://github.com/superdoc-dev/superdoc/releases/tag/v1.23.0)

PR #2501, opened 2026-03-20
fix: inherit run formatting when inserting inline structured content (SD-1421)

luccas-harbour:
## Summary

- When inserting an inline SDT field in the middle of formatted text, the inserted text now inherits the surrounding font/formatting instead of appearing unstyled
- Fixes a structural issue where `insertStructuredContentInline` produced `run > structuredContent > run > text` nesting — the SDT is now placed at paragraph level between split sibling runs
- Handles ranged selections correctly by removing the selected text when splitting the parent run

## Problem

`insertStructuredContentInline` used a bare `tr.replaceWith(from, to, sdtNode)` to insert the SDT. When the cursor was inside a run, ProseMirror placed the SDT as a child of the run (since `run` has `content: 'inline*'`). Then `wrapTextInRunsPlugin` normalized this into `run{null} > SDT > run > text`, creating unwanted nesting with an empty outer run.

Additionally, the SDT's text content was created without any marks or run wrapping, so it appeared unformatted regardless of the surrounding text style.

## Solution

Two changes to `insertStructuredContentInline`:

1. **Formatting copy** — resolves marks and `runProperties` from the cursor position via `getFormattingStateAtPos` and wraps the text in a properly formatted run inside the SDT
2. **Run splitting** — when the cursor is inside a run, manually splits it into left/right halves and inserts `[leftRun, sdtNode, rightRun]` at the paragraph level, producing the correct flat structure


packages/super-editor/src/extensions/structured-content/structured-content-commands.js
```diff
@@ -157,7 +179,37 @@ export const StructuredContentCommands = Extension.create({
...
+              const runDepth = $from.depth;
+              const runStart = $from.before(runDepth);
+              const runEnd = $from.after(runDepth);
+              const parentRun = $from.parent;
+              const startOffset = $from.parentOffset;
+
+              // When the user has a ranged selection, cut the right half from
+              // the end of the selection so the selected text is removed.
+              const $to = state.doc.resolve(to);
+              const endOffset = $to.parent === parentRun ? $to.parentOffset : startOffset;
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/super-editor/src/extensions/structured-content/structured-content-commands.js
```diff
@@ -134,6 +135,27 @@ export const StructuredContentCommands = Extension.create({
               content = schema.text(' ');
             }
 
+            // When content was not provided as structured JSON, wrap the text
+            // in a formatted run inside the SDT so it visually matches the
+            // surrounding text. The run-split logic below prevents an outer run
+            // from wrapping the SDT itself.
+            const runType = schema.nodes.run;
+            if (runType && !options.json && content.isText) {
+              const formattingState = getFormattingStateAtPos(state, from, editor);
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/super-editor/src/extensions/structured-content/structured-content-commands.js
```diff
@@ -134,6 +135,29 @@ export const StructuredContentCommands = Extension.create({
               content = schema.text(' ');
             }
 
+            // When content was not provided as structured JSON, wrap the text
+            // in a formatted run inside the SDT so it visually matches the
+            // surrounding text. The run-split logic below prevents an outer run
+            // from wrapping the SDT itself.
+            const runType = schema.nodes.run;
+            if (runType && !options.json && content.isText) {
```
caiopizzol, line 146: when `options.json` is used, this whole block is skipped — the inserted field won't pick up the surrounding font. that's probably intentional, but a comment here would save the next person from wondering if it's a bug.

packages/super-editor/src/extensions/structured-content/structured-content-commands.js
```diff
@@ -157,7 +181,36 @@ export const StructuredContentCommands = Extension.create({
...
+              const fragments = [];
+              if (leftContent.size > 0) {
+                fragments.push(runType.create(parentRun.attrs, leftContent, parentRun.marks));
+              }
+              fragments.push(node);
+              if (rightContent.size > 0) {
+                fragments.push(runType.create(parentRun.attrs, rightContent, parentRun.marks));
+              }
+
+              tr.replaceWith(runStart, runEnd, fragments);
```
caiopizzol, line 213: after this replace, the cursor ends up at the end of the text instead of right after the new field. users won't notice if it's triggered by a button, but if they keep typing after inserting a field in the template builder, text goes to the wrong spot. worth fixing with a `tr.setSelection(...)` after the replace?

packages/super-editor/src/extensions/structured-content/structured-content-commands.test.js
```diff
@@ -556,6 +556,122 @@ describe('updateStructuredContentByGroup', () => {
   });
```
caiopizzol, line 556: missing a test for when the cursor is at the very start of a run — that's the case where there's no text on the left side of the split. worth adding to make sure no empty run sneaks into the output.

packages/super-editor/src/extensions/structured-content/structured-content-commands.js
luccas-harbour: done. I added a comment explaining this.

packages/super-editor/src/extensions/structured-content/structured-content-commands.js
luccas-harbour: yes, good idea. fixed.

packages/super-editor/src/extensions/structured-content/structured-content-commands.test.js
luccas-harbour: I added a new test case covering this.

PR #2511, opened 2026-03-21
fix(document-api): improve AI tool definitions, add block refs, and fix formatting extraction

tupizz:
## Summary

Fixes AI tool call quality by adding ref handles to API responses, fixing formatting extraction bugs, correcting tool descriptions, and rewriting the system prompt.

### Problems fixed

- **Extra search steps**: Creating content required create > search > format (3 calls). Editing a paragraph required search first even when blocks data was already available.
- **Wrong target format**: Codegen fallback description said `{kind:'text', blockId, range:{start,end}}` but the correct format is `{kind:'selection', start:{kind:'text', blockId, offset}, end:{kind:'text', blockId, offset}}`.
- **Partial replacements**: Search refs only cover matched text. "Rewrite this paragraph" replaced part of the text and left the rest because the ref didn't cover the full block.
- **Confusing flat params**: `blockId`, `start`, `end` in edit tool schemas confused models into building targets manually instead of using refs.
- **Heading numbering**: Templates with outline numbering linked to heading styles showed "1" prefix on programmatically created headings.
- **Wrong color in blocks data**: `extractBlockFormatting` read `color` from all marks including `highlight` marks, reporting `#FFFFFF` as text color when paragraphs had white highlight backgrounds.
- **List creation made separate lists**: Prompt said "for each paragraph" so the LLM called `superdoc_list` once per item, creating N independent lists instead of one.
- **Missing list conversion**: `set_type` action for converting bullet/numbered wasn't mentioned in prompt or tool description.
- **LLMs pass limit=0 instead of omitting**: "Do NOT add limit" was interpreted as "set to zero." API now normalizes `limit: 0` and `nodeTypes: []` to mean "all."

### Changes

**Block refs in `blocks.list`** — Each non-empty block now returns a `ref` handle covering its full text. Agents pass this directly to `superdoc_edit` or `superdoc_format` without searching first.

**Ref from `superdoc_create`** — Create responses include a `ref` for the new block, enabling immediate formatting without a search step (create > format instead of create > search > format).

**Fixed `extractBlockFormatting`** — Only reads `fontFamily`/`fontSize`/`color` from `textStyle` marks and `bold` from `bold` marks. No longer picks up highlight background colors.

**Fixed target descriptions** — Corrected the fallback `target` description in codegen from wrong `{kind:'text'}` format to correct `{kind:'selection'}` format.

**Tool description improvements:**
- `superdoc_edit`: guides toward `ref` usage, explains that search refs are partial while block refs cover full text
- `superdoc_create`: documents the ref in the response, required format-after-create workflow with examples
- `superdoc_get_content`: mentions that blocks returns ref handles and formatting data
- `superdoc_list`: expanded description with key actions (create, set_type, insert, indent, outdent, detach)

**List creation fix** — Prompt now instructs creating all paragraphs first, then one `superdoc_list` call with a `BlockRange` target to produce a single contiguous list. Added `set_type` guidance for bullet/numbered conversion.

**Heading numbering fix** — `insertHeadingAt` sets `numberingProperties: { numId: '0', ilvl: '0' }` to suppress outline numbering from template heading styles.

**Tolerant input normalization** — `blocks.list` now normalizes `limit: 0` to "all blocks" and `nodeTypes: []` to "no filter" before validation.

**Hidden confusing params** — `blockId`, `start`, `end`, `offset` flat params hidden from agent tool schemas (kept for CLI).

**System prompt rewrite** — Blocks-first workflow, concrete formatting examples for paragraphs and headings, correct list creation with `BlockRange` targets, list type conversion, sequential create ordering via nodeId.

**Schema validator** — Added `description`, `minimum`, `maximum` to supported keywords for contract conformance tests.

### Files changed

| Area | Files | What |
|------|-------|------|
| Contract | `operation-definitions.ts`, `schemas.ts` | Tool descriptions, block ref + color schema |
| Types | `blocks.types.ts`, `create.types.ts` | `ref` and `color` fields |
| Codegen | `generate-intent-tools.mjs` | Fixed target fallback description |
| CLI params | `operation-params.ts` | Hidden flat params from agent tools |
| Validation | `blocks/blocks.ts` | Normalize limit=0 and nodeTypes=[] |
| System prompt | `system-prompt.md` | Rewritten workflow, list range targets, set_type |
| Editor commands | `insertHeadingAt.js` | Suppress outline numbering |
| Adapters | `blocks-wrappers.ts` | Block refs, color field, fixed mark reading |
| Adapters | `create-wrappers.ts` | Ref minting on create |
| Tests | `schema-validator.ts` | Support `description`/`minimum`/`maximum` |
| Evals | `execution.yaml` | List creation and conversion tests |
| Docs | `apps/docs/document-api/reference/*` | Generated from contract changes |

## Test plan

- [ ] `pnpm --filter super-editor test` passes
- [ ] Run eval suite, verify pass rate >= 96%
- [ ] Eval: 3-item bullet list creates one contiguous list
- [ ] Eval: bullet to numbered conversion uses set_type
- [ ] Manual: create heading via document API, verify no "1" prefix
- [ ] Manual: get blocks, use block ref to replace entire paragraph

### DIFF
--- a/packages/super-editor/src/editors/v1/core/helpers/importHtml.js
+++ b/packages/super-editor/src/editors/v1/core/helpers/importHtml.js
@@ -47,12 +47,14 @@ export function createDocFromHTML(content, editor, options = {}) {
   // Detect unsupported content when opted in (requires an Element for DOM scanning)
   const domElement =
     parsedContent?.ownerDocument?.defaultView?.Element ?? domDocument?.defaultView?.Element ?? globalThis.Element;
-  if (
+  const parsedElement =
     (options.onUnsupportedContent || options.warnOnUnsupportedContent) &&
     domElement &&
     parsedContent instanceof domElement
-  ) {
-    const unsupported = detectUnsupportedContent(parsedContent, editor.schema);
+      ? /** @type {Element} */ (/** @type {unknown} */ (parsedContent))
+      : null;
+  if (parsedElement) {
+    const unsupported = detectUnsupportedContent(parsedElement, editor.schema);
     if (unsupported.length > 0) {
       if (options.onUnsupportedContent) {
         options.onUnsupportedContent(unsupported);


evals/providers/utils.mjs
```diff
@@ -74,6 +74,49 @@ export function cleanArgs(args) {
   return rest;
 }
 
+// --- Ref revision bump (handles REVISION_MISMATCH after create→format) ---
+
+const REF_PREFIX = 'text:v4:';
+const MAX_REF_RETRIES = 3;
```
andrii-harbour, line 83: please add a comment where text:v4 is coming from and when it should be updated

evals/tests/execution.yaml
```diff
@@ -350,6 +350,50 @@
...
+  vars:
+    fixture: document.docx
+    keepFile: true
+    task: 'Convert the existing bullet list in the document to a numbered list.'
+  assert:
+    - type: javascript
+      value: |
+        const d = JSON.parse(output);
+        if (!d.documentText) return { pass: false, score: 0, reason: 'No content' };
+        return { pass: true, score: 1, reason: 'Conversion attempted' };
```
andrii-harbour: the assertion ('Conversion attempted') is very weak. It passes even if the conversion fails as long as text exists.

examples/collaboration/ai-node-sdk/client/src/components/landing/create-room-form.tsx
```diff
@@ -0,0 +1,259 @@
...
+            <>
+              <Upload className="h-8 w-8 text-muted-foreground/60 mb-2" />
+              <p className="text-sm text-muted-foreground">
+                Drop a <span className="font-medium">.docx</span> file here, or click to browse
+              </p>
+            </>
+          )}
+        </div>
+
+        {/* Quick actions */}
```
andrii-harbour, line 194: suggestion: right now when clicking on any of the quick actions it creates the room immediately without letting us finish the settings flow. What if click on those buttons would just change the UI to something like this

Image

examples/collaboration/ai-node-sdk/client/src/components/shared/sidebar-toggle.tsx
```diff
@@ -0,0 +1,34 @@
...
+} from 'lucide-react';
+import { Button } from '@/components/ui/button';
+
+interface SidebarToggleProps {
+  collapsed: boolean;
+  onToggle: () => void;
+  side: 'left' | 'right';
+}
+
+export function SidebarToggle({ collapsed, onToggle, side }: SidebarToggleProps) {
```
andrii-harbour, line 15: all the traces are duplicated 

Image

examples/collaboration/ai-node-sdk/Makefile
```diff
@@ -0,0 +1,144 @@
+.PHONY: install dev dev-local dev-server dev-client link-local-sdk rebuild-local-sdk kill clean help
+
+# ─── Config ───────────────────────────────────────────────────────────────────
```
andrii-harbour, line 3: Nice!

packages/document-api/src/blocks/blocks.ts
```diff
@@ -39,6 +39,26 @@ const VALID_BLOCK_NODE_TYPES = new Set<string>(BLOCK_NODE_TYPES);
 // blocks.list validation
 // ---------------------------------------------------------------------------
 
+function normalizeBlocksListInput(input?: BlocksListInput): BlocksListInput | undefined {
```
andrii-harbour, line 42: let's add a unit tests for this too

packages/super-editor/src/editors/v1/document-api-adapters/__conformance__/schema-validator.ts
```diff
@@ -24,6 +24,9 @@ const SUPPORTED_SCHEMA_KEYWORDS = new Set([
   'anyOf',
   '$ref',
   '$defs',
+  'description',
+  'minimum',
+  'maximum',
```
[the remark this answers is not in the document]
andrii-harbour: +1

packages/super-editor/src/document-api-adapters/plan-engine/blocks-wrappers.ts
```diff
@@ -83,23 +86,35 @@ function extractBlockFormatting(node: ProseMirrorNode): {
...
       }
-      if (attrs.bold === true) bold = true;
+      if (markName === 'bold' && attrs.value === true) bold = true;
+      if (markName === 'underline') underline = true;
     }
     return false;
   });
 
+  // Default to black when no explicit color is set
+  if (!color) color = '#000000';
```
andrii-harbour: This defaults every block without an explicit color mark to black. If a document's theme defines a non-black default text color (e.g., dark gray), the agent will propagate #000000 to new content. Should this default come from the style-engine cascade instead, or at minimum be documented as an approximation?

packages/sdk/langs/browser/src/system-prompt.ts
```diff
@@ -9,9 +9,9 @@ export const SYSTEM_PROMPT = `You are a document editing assistant. You have a D
 | Tool | Purpose |
```
andrii-harbour, line 9: packages/sdk/langs/browser/src/system-prompt.ts and packages/sdk/tools/system-prompt.md contain nearly identical content. Changes to one must be manually synced to the other. Can use a symlink here?

evals/providers/utils.mjs
tupizz: Done — added a comment referencing `story-ref-codec.ts` and when to update it.

evals/tests/execution.yaml
tupizz: Agreed. Updated the assertion to verify the tool call used either `set_type` or `create` action on `superdoc_list`, not just that text exists.

examples/collaboration/ai-node-sdk/client/src/components/landing/create-room-form.tsx
tupizz: Good idea. Will address in a follow-up — the quick actions should populate the settings form instead of creating immediately.

packages/document-api/src/blocks/blocks.ts
tupizz: Done — added 8 test cases covering all normalization paths (limit=0, nodeTypes=[], both combined, passthrough for valid values). Tests exercise it indirectly through `executeBlocksList`.

packages/super-editor/src/document-api-adapters/plan-engine/blocks-wrappers.ts
tupizz: Good catch. Removed the #000000 default — `color` is now absent when no explicit color mark exists, consistent with how fontFamily/fontSize already behave. Also filtering out the OOXML `"auto"` sentinel since it is a rendering directive, not an actual color value.

packages/sdk/langs/browser/src/system-prompt.ts
tupizz: This file is auto-generated by codegen from `system-prompt.md`. See `generate-intent-tools.mjs:584-592` — it reads the markdown, wraps it in a TS export, and writes it to the browser SDK. The generated file has an `// Auto-generated from packages/sdk/tools/system-prompt.md` header. A symlink wouldn't work (TS can't import .md as a typed constant), but the codegen keeps them in sync automatically via `pnpm run generate:all`.

PR #2515, opened 2026-03-22
fix: update dropdown trigger label on select

Branc0:
### Summary
When a user selects an option from a toolbar dropdown, `handleSelect` updates `selectedValue` and emits the command, but does not update the reactive `item.label` used by `ToolbarButton` in the dropdown trigger. The visible label could stay on the initial value even though the selection changed.

https://github.com/user-attachments/assets/c7a9a49b-8865-4f04-8d91-6dab1af10c37

### Steps to reproduce
- [ ] Open the editor with a mobile viewport (under 768px width) and open the overflow menu.
- [ ] Click on the font size dropdown and change the value.
- [ ] The document font-size changes, but the select remains displaying the same value

### Solution
This change sets `item.label` from `option.label` when there's a change in the selection, keeping the trigger text in sync with the selection.


packages/super-editor/src/components/toolbar/ButtonGroup.vue
```diff
@@ -136,6 +136,10 @@ const handleSelect = (item, option) => {
   const value = item.dropdownValueKey.value ? option[item.dropdownValueKey.value] : option.label;
   emit('command', { item, argument: value, option });
   item.selectedValue.value = option.key;
+
+  if (option?.label && !item.hideLabel?.value) {
+    item.label.value = option.label;
+  }
```
caio-pizzol: `handleSelect` now sets the label, but `onActivate` already does this too for some dropdowns (font family, zoom) — so the label gets written twice. nothing breaks, but two places updating the same thing could get confusing later. worth keeping label updates in one place?

Branc0: Gotcha! Yeah, it makes total sense; that passed under my radar.

I will take a look at this sometime between today and tomorrow, thanks for the feedback

Branc0: So after some debugging, I realized that I had a couple of misconceptions.

First of all, I thought that the ButtonGroup.vue was only being used in the context of the OverflowMenu.vue, but it's also the same component used when the select is being shown directly on the Toolbar.vue. So yeah, definitely not the correct path for the fix.

What is actually happening is that when the ButtonGroup.vue is used from the OverflowMenu.vue context, and we bubble up to super-toolbar.js, the condition on line 1139 is fulfilled, triggering `onActivate` with empty arguments.

I guess that `wasFocused` variable is false because we have to click on the three-dot menu to then display the OverflowMenu, and that loses the focus on the editor.

So I performed this fix, not sure if it's accurate, let me know if it's good, and then I will write some tests for this

packages/super-editor/src/components/toolbar/defaultItems.js
```diff
@@ -1023,6 +1023,7 @@ export const makeDefaultItems = ({
     icon: toolbarIcons.lineHeight,
     hasCaret: false,
     hasInlineTextInput: false,
+    hideLabel: true,
```
Branc0, line 1026: I guess it's good to keep this, right?

packages/super-editor/src/components/toolbar/ButtonGroup.vue
```diff
@@ -136,6 +136,10 @@ const handleSelect = (item, option) => {
   const value = item.dropdownValueKey.value ? option[item.dropdownValueKey.value] : option.label;
   emit('command', { item, argument: value, option });
   item.selectedValue.value = option.key;
+
+  if (option?.label) {
+    item.label.value = option.label;
```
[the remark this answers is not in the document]
caio-pizzol: @Branc0 this comment is about the old approach — the `handleSelect` label write was removed in your latest push, so this no longer applies. no action needed.

caio-pizzol (commented):
@Branc0 great follow-up, verified it locally and it works. one small inline suggestion.

since you mentioned writing tests — two would help:
1. a unit test checking that `activate` gets the right value when the editor isn't focused
2. a behavior test (check `tests/behavior/` for examples): small viewport → open overflow → change font size → check label updated

happy to help with either!

packages/super-editor/src/components/toolbar/super-toolbar.js
```diff
@@ -1138,7 +1138,15 @@ export class SuperToolbar extends EventEmitter {
     // until the next selection update (after the user clicks into the editor).
     if (!wasFocused && isMarkToggle) {
       this.pendingMarkCommands.push({ command, argument, item });
-      item?.activate?.();
+      const shouldPassActivationAttrs = item?.command === 'setFontSize' || item?.command === 'setFontFamily';
+      if (shouldPassActivationAttrs) {
+        const itemName = item?.name?.value;
```
caio-pizzol, line 1143: nice find on the root cause! the items already have a `labelAttr` field with the same info — using that instead of command names would keep this flexible for new dropdowns.

PR #2528, opened 2026-03-23
feat: initial support for multi-column; lists and media fixes(Copy/paste project) 

VladaHarbour:
This PR introduces fixes to support copy/paste experience within Superdoc:
1. Preserves multi-column layout
2. Preserves lists (including nested)
3. Preserves images and headings

(We have hyperlinks and table support in current stable version)

### DIFF
--- a/packages/super-editor/src/editors/v1/core/helpers/superdocClipboardSlice.test.js
+++ b/packages/super-editor/src/editors/v1/core/helpers/superdocClipboardSlice.test.js
@@ -1,7 +1,13 @@
+// @vitest-environment happy-dom
 import { describe, expect, it, vi } from 'vitest';
 import {
   collectReferencedImageMediaForClipboard,
   applySuperdocClipboardMedia,
+  embedSliceInHtml,
+  extractSliceFromHtml,
+  stripSliceFromHtml,
+  extractBodySectPrFromHtml,
+  bodySectPrShouldEmbed,
   SUPERDOC_MEDIA_MIME,
 } from './superdocClipboardSlice.js';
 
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/editors/v1/core/InputRule.js
```diff
@@ -531,3 +668,283 @@ export function handleClipboardPaste({ editor, view }, html, plainText) {
...
 }
+
+/** Cut: put slice + annotated HTML on clipboard, then delete selection (copy uses ProseMirrorRenderer). */
+function handleCutEvent(view, event, editor) {
+  const clipboardData = event.clipboardData;
+  if (!clipboardData) return false;
+
+  const { from, to } = view.state.selection;
+  if (from === to) return false;
+
```
caio-pizzol, line 688: if anything inside the try block fails, `preventDefault` already blocked the browser's cut — so the user presses Cmd+X and nothing happens (no clipboard data, no deletion). moving `preventDefault` to after the clipboard writes means the browser's own cut kicks in as a fallback.

```suggestion
  try {
    const slice = view.state.doc.slice(from, to);
    const fragment = slice.content;
    const sliceJson = JSON.stringify(slice.toJSON());

    clipboardData.setData(SUPERDOC_SLICE_MIME, sliceJson);
    const mediaJson = collectReferencedImageMediaForClipboard(sliceJson, editor);
    if (mediaJson) {
      clipboardData.setData(SUPERDOC_MEDIA_MIME, mediaJson);
    }

    const div = document.createElement('div');
    const serializer = PmDOMSerializer.fromSchema(view.state.schema);
    div.appendChild(serializer.serializeFragment(fragment));

    annotateFragmentDomWithClipboardData(div, fragment, editor);

    const html = unflattenListsInHtml(div.innerHTML);
    const bodySectPr = view.state.doc.attrs?.bodySectPr;
    const bodySectPrJson = bodySectPr && bodySectPrShouldEmbed(bodySectPr) ? JSON.stringify(bodySectPr) : '';
    clipboardData.setData('text/html', embedSliceInHtml(html, sliceJson, bodySectPrJson));
    clipboardData.setData('text/plain', fragment.textBetween(0, fragment.size, '\n\n'));

    event.preventDefault();
    view.dispatch(view.state.tr.deleteSelection().scrollIntoView());
  } catch (error) {
    console.warn('Failed to handle cut:', error);
  }
```

packages/super-editor/src/core/InputRule.js
```diff
@@ -461,6 +565,39 @@ export function sanitizeHtml(html, forbiddenTags = ['meta', 'svg', 'script', 'st
...
+  // Strip Word conditional list-marker spans so paste does not duplicate markers.
+  const stripWordListConditionalPrefixes = (root) => {
+    const stripFromNode = (node) => {
+      if (!node?.childNodes) return;
+
+      for (let i = 0; i < node.childNodes.length; i += 1) {
+        const current = node.childNodes[i];
+        if (current?.nodeType === Node.COMMENT_NODE && current.nodeValue?.includes('[if !supportLists]')) {
+          let j = i + 1;
+          while (j < node.childNodes.length) {
```
caio-pizzol: if the closing `[endif]` comment is missing (broken Word HTML), this loop removes every node after it until there's nothing left. a warning would make this easier to debug:

```suggestion
          while (j < node.childNodes.length) {
            const next = node.childNodes[j];
            if (next?.nodeType === Node.COMMENT_NODE && next.nodeValue?.includes('[endif]')) {
              node.removeChild(next);
              break;
            }
            node.removeChild(next);
          }
          if (j >= node.childNodes.length) {
            console.warn('[SuperDoc] stripWordListConditionalPrefixes: missing [endif], some content may have been removed');
          }
```

packages/super-editor/src/core/InputRule.js
```diff
@@ -1,20 +1,96 @@
...
+    /data-num-id\s*=/i.test(html) &&
+    (/data-level\s*=/i.test(html) || /data-list-numbering-type\s*=/i.test(html) || /data-list-level\s*=/i.test(html))
+  ) {
+    return true;
+  }
+  return false;
+}
+
+/** Apply pasted body `sectPr` when target has single-column layout. */
+function tryApplyEmbeddedBodySectPr(editor, view, bodySectPr) {
```
caio-pizzol: `tryApplyEmbeddedBodySectPr` and `applyEmbeddedBodySectPrToTransaction` do the same checks and the same work — only difference is who owns the transaction. could be one shared function:

```js
function applyBodySectPrToTr(editor, tr, bodySectPr, currentDoc) {
  if (!bodySectPr || typeof bodySectPr !== 'object') return false;
  const incomingCols = getSectPrColumns(bodySectPr);
  if (!incomingCols?.count || incomingCols.count <= 1) return false;
  const currentCols = getSectPrColumns(currentDoc.attrs?.bodySectPr);
  if (currentCols?.count > 1) return false;
  const clone = JSON.parse(JSON.stringify(bodySectPr));
  tr.setDocAttribute('bodySectPr', clone);
  if (editor?.converter) editor.converter.bodySectPr = clone;
  return true;
}
```

worth doing?

packages/super-editor/src/editors/v1/core/InputRule.js
```diff
@@ -531,3 +668,283 @@ export function handleClipboardPaste({ editor, view }, html, plainText) {
...
+      node.content.forEach((child) => collect(child));
+    }
+  };
+
+  fragment.forEach((node) => collect(node));
+
+  if (paragraphMeta.length === 0) {
+    return fragment;
+  }
+
```
caio-pizzol, line 779: `createListIdAllocator` is identical to the one in `html-helpers.js:53` — moving it to the shared numbering helpers would keep them from drifting apart.

packages/super-editor/src/core/inputRules/html/html-helpers.js
```diff
@@ -49,6 +50,83 @@ export function flattenListsInHtml(html, editor, domDocument) {
   return doc.body.innerHTML;
 }
 
+function createListIdAllocator(editor) {
```
caio-pizzol: same `createListIdAllocator` as `InputRule.js:790` — see comment there.

packages/super-editor/src/editors/v1/core/InputRule.js
VladaHarbour: moved preventDefault inside try/catch

packages/super-editor/src/core/InputRule.js
VladaHarbour: created more advanced handling for this corner case + test

PR #2541, opened 2026-03-24
fix(layout-bridge): correct click-to-position mapping for RTL text

claudiu-ior:
## Summary
- Fix text selection and cursor placement in RTL (Arabic, Hebrew, etc.) documents when the layout engine is enabled. Previously, clicking or dragging to select RTL text produced inverted or incorrect positions because the DOM mapping assumed LTR visual order.
- Add RTL-aware boundary snapping, empty-element snapping, and character-level position mapping using the browser's native `caretPositionFromPoint` / `caretRangeFromPoint` APIs.
## What changed
All changes are in `packages/layout-engine/layout-bridge/src/dom-mapping.ts`:
**RTL detection** — Added `isRtlLine()` helper that reads `getComputedStyle(lineEl).direction` to detect RTL lines. This leverages the `dir="rtl"` attribute already set by `DomPainter`'s `rtl-paragraph` feature.
**Boundary snapping** — In both `processFragment` and `processLineElement`, replaced the old assumption that the first DOM span is visually leftmost and the last is visually rightmost. Now computes `visualLeft` / `visualRight` across all spans and swaps `lineStart` / `lineEnd` for RTL lines (clicking on the visual left edge of an RTL line means "end of line" in document order, and vice versa).
**Empty-element snapping** — When a click lands on a non-text element, the "closer to left/right edge" logic now accounts for RTL direction so the cursor snaps to the correct document position.
**Character-level mapping (`findCharIndexAtX`)** — Rewrote to use browser caret APIs (`caretPositionFromPoint` / `caretRangeFromPoint`) as the primary strategy, which natively handles bidi text, Arabic ligatures, and contextual shaping. Falls back to a binary search with per-character `Range` rects, using `r.left` for RTL and `r.right` for LTR as the boundary edge.
## Test plan
- [ ] Open an RTL (Arabic) document with the layout engine enabled
- [ ] Verify single-click places the cursor at the correct character position
- [ ] Verify click-and-drag selects text in the visually correct direction
- [ ] Verify mixed LTR/RTL content (e.g., English words within Arabic text) still maps correctly
- [ ] Verify LTR documents are unaffected (no regression)


claudiu-ior:
@caio-pizzol could I get a review on this PR please?

caio-pizzol:
> @caio-pizzol could I get a review on this PR please?

On it!

packages/layout-engine/layout-bridge/src/dom-mapping.ts
```diff
@@ -642,7 +671,6 @@ function findSpanAtX(spanEls: HTMLElement[], viewX: number): HTMLElement | null
       });
```
caio-pizzol, line 621: when a click lands between two spans (gap), the fallback picks the last span to the left of the click. that works for LTR but in RTL it can pick the wrong neighbor. a simpler approach: just pick the closest span by distance regardless of direction — replace the `viewX > rect.right` accumulator with a min-distance check.

packages/layout-engine/layout-bridge/src/dom-mapping.ts
```diff
@@ -370,17 +378,31 @@ function processFragment(fragmentEl: HTMLElement, viewX: number, viewY: number):
     return lineStart;
```
caio-pizzol: the RTL bounds + edge snapping + empty-element snap is duplicated between `processFragment` and `processLineElement` (~35 lines). the only real difference is the final position calculation, and `mapCharIndexToPm` already handles the simple case. worth extracting a shared helper so future RTL fixes don't need to be applied in two places?

packages/layout-engine/layout-bridge/src/dom-mapping.ts
```diff
@@ -660,61 +688,68 @@ function findSpanAtX(spanEls: HTMLElement[], viewX: number): HTMLElement | null
 /**
```
caio-pizzol, line 640: no RTL test coverage in `dom-mapping.test.ts` — all existing tests use LTR. the most valuable additions would be: clicking left/right of RTL spans returns the reversed positions, and RTL empty-element snap goes the opposite way.

PR #2552, opened 2026-03-25
fix: support hyperlinks on DrawingML images (a:hlinkClick)

### LINKED ISSUE #2065: Feature: Support hyperlink click on DrawingML images
### Summary

Support image hyperlinks on DrawingML images (a:hlinkClick) in DOCX.

### Problem

Image renders, but hyperlink click does not work in SuperDoc.

### Repro

1. Open attached docx in SuperDoc.
2. Click the footer image (it has a hyperlink).

### Expected

Clicking the image opens the hyperlink.

### Notes

This docx isolates the image hyperlink case using a JPEG image.

### Attachments

[footer-jpeg-hlink-min.docx](https://uploads.linear.app/92b189d9-e15d-4f28-8a87-d9ca147ba33f/8537b070-73bf-42f5-a236-4c93cc1ccf61/8af02c8b-1dd2-49c3-bbc2-f7c79c71f9ff?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkyYjE4OWQ5LWUxNWQtNGYyOC04YTg3LWQ5Y2ExNDdiYTMzZi84NTM3YjA3MC03M2JmLTQyZjUtYTIzNi00YzkzY2MxY2NmNjEvOGFmMDJjOGItMWRkMi00OWMzLWJiYzItZjdjNzljNzFmOWZmIiwiaWF0IjoxNzcyMTIyMDA5LCJleHAiOjE4MDM2OTI1Njl9.wip6XxYHCWbiDC6YMhpkm44VsXfrv6k39LS6Wp9Yazg)

---

*Ref:* [SD-1871](https://linear.app/superdocworkspace/issue/SD-1871/feature-support-hyperlink-click-on-drawingml-images)

iguit0:
## Summary

DOCX images with `a:hlinkClick` hyperlinks rendered correctly in Word but clicking them in SuperDoc did nothing. The importer already parsed the relationship and stored `attrs.hyperlink` on the ProseMirror image node — but the hyperlink was silently dropped further down the pipeline, so the renderer never had a URL to render.

Root cause: a three-step gap between import and rendering:
1. `ImageBlock` / `ImageRun` contract types had no `hyperlink` field
2. `imageNodeToBlock` and `imageNodeToRun` did not forward the attr
3. `DomPainter.renderImageFragment` / `renderImageRun` never created an anchor

## What Changed

- Added `hyperlink?: { url: string; tooltip?: string }` to both `ImageBlock` and `ImageRun` in the shared contracts package
- Forwarded the attr in `imageNodeToBlock` and `imageNodeToRun`, with the same validation (non-empty string URL) used for other optional fields
- Added `DomPainter.buildImageHyperlinkAnchor()` — a private helper that wraps an image element in `<a class="superdoc-link">` when a valid hyperlink is present. Uses `sanitizeHref` (same guard as text links) and applies `target`/`rel` for external URLs with accessibility attributes mirroring `applyLinkAttributes`
- Called the helper in `renderImageFragment` (block images) and at all three return points of `renderImageRun` (inline images with/without clip-path wrapper)

The existing `EditorInputManager` click-delegation already detects `a.superdoc-link` on pointer-down and dispatches `superdoc-link-click`. `LinkClickHandler` then opens the URL in viewing mode — no changes needed in either component. Unlinked images are unaffected: `buildImageHyperlinkAnchor` is a no-op when `hyperlink` is undefined, and image resize/selection behavior is preserved because the fragment `<div>` (which carries `data-image-metadata`) remains the outermost element.

## Test Plan

- [ ] Added 5 unit tests to `pm-adapter/src/converters/image.test.ts` — verifies `hyperlink` is forwarded with tooltip, without tooltip, and is omitted for null/empty cases
- [ ] Added 4 integration tests to `painters/dom/src/index.test.ts` — verifies linked images render as `<a class="superdoc-link">` with correct `href`, tooltip becomes `title`, unlinked images have no anchor, and unsafe URLs (e.g. `javascript:`) are blocked
- [ ] All existing tests pass: pm-adapter (1,669), painter-dom (824), super-editor (9,697)
- [ ] Manual: open a DOCX with a hyperlinked image → image renders → clicking in viewing mode opens the correct URL

Fixes #2065


packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -3449,6 +3452,62 @@ export class DomPainter {
...
+    anchor.href = sanitized.href;
+    anchor.classList.add('superdoc-link');
+
+    if (sanitized.protocol === 'http' || sanitized.protocol === 'https') {
+      anchor.target = '_blank';
+      anchor.rel = 'noopener noreferrer';
+    }
+    if (hyperlink.tooltip) {
+      anchor.title = hyperlink.tooltip;
+    }
```
caio-pizzol, line 3727: text links run the tooltip through `encodeTooltip()` which trims whitespace and caps it at 500 chars. this skips that — a crafted docx with a huge tooltip would render it in full. worth matching the text link behavior?

```suggestion
    if (hyperlink.tooltip) {
      const tooltipResult = encodeTooltip(hyperlink.tooltip);
      if (tooltipResult?.text) {
        anchor.title = tooltipResult.text;
      }
    }
```

packages/layout-engine/pm-adapter/src/converters/image.ts
```diff
@@ -313,6 +313,18 @@ export function imageNodeToBlock(
     ...(rotation !== undefined && { rotation }),
     ...(flipH !== undefined && { flipH }),
     ...(flipV !== undefined && { flipV }),
+    // Image hyperlink from OOXML a:hlinkClick
+    ...(() => {
```
caio-pizzol: same parsing logic exists in `inline-converters/image.ts:170`. a small shared helper in `utilities.ts` would keep them in sync and also get rid of the IIFE here.

packages/layout-engine/painters/dom/src/index.test.ts
```diff
@@ -6442,6 +6442,135 @@ describe('ImageFragment (block-level images)', () => {
...
+        src: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg==',
+        width: 100,
+        height: 50,
+        ...(hyperlink ? { hyperlink } : {}),
+      };
+      const measure: Measure = { kind: 'image', width: 100, height: 50 };
+      const fragment = {
+        kind: 'image' as const,
+        blockId: 'linked-img',
+        x: 20,
```
caio-pizzol: `layout` is created here but never used — the tests build their own. can be removed.

caio-pizzol (commented):
@iguit0 nice contribution, thanks for tackling this! the approach looks good — checked the spec and the dual placement, relationship handling, and tooltip are all done right.

one small thing on tooltip encoding (left a suggestion). one optional cleanup on shared code.

tests: block-level is well covered. the inline paths (`imageNodeToRun` and the 3 renderer return paths) don't have hyperlink tests yet — would be great to add those.

left a few inline comments.

PeterHollens:
Nice work on this. I independently implemented the same fix on my fork while digging through #2065, and I converged on the same root cause.

One thing I found useful was reusing the existing shared link pipeline ( /  / ) rather than introducing an image-specific hyperlink path, since it automatically inherits the current blocked-link handling, dataset propagation, tooltip accessibility, and other shared link behavior. I also added explicit inline-image coverage in both the PM adapter and DOM painter tests, since the inline paths were the main gap I noticed while reviewing.

If any of that is helpful, happy to share or compare against my branch here: https://github.com/PeterHollens/superdoc/tree/feature/drawingml-image-hyperlinks

PeterHollens:
Nice work on this. I independently implemented the same fix on my fork while
digging through #2065, and I converged on the same root cause.

One thing I found useful was reusing the existing shared link pipeline
(`FlowRunLink` / `buildFlowRunLink` / `wrapElementWithLink`) rather than
introducing an image-specific hyperlink path, since it automatically inherits
the current blocked-link handling, dataset propagation, tooltip accessibility,
and other shared link behavior. I also added explicit inline-image coverage in
both the PM adapter and DOM painter tests, since the inline paths were the main
gap I noticed while reviewing.

If any of that is helpful, happy to share or compare against my branch here:
https://github.com/PeterHollens/superdoc/tree/feature/drawingml-image-hyperlinks

iguit0:
> Nice work on this. I independently implemented the same fix on my fork while digging through #2065, and I converged on the same root cause.
> 
> One thing I found useful was reusing the existing shared link pipeline (`FlowRunLink` / `buildFlowRunLink` / `wrapElementWithLink`) rather than introducing an image-specific hyperlink path, since it automatically inherits the current blocked-link handling, dataset propagation, tooltip accessibility, and other shared link behavior. I also added explicit inline-image coverage in both the PM adapter and DOM painter tests, since the inline paths were the main gap I noticed while reviewing.
> 
> If any of that is helpful, happy to share or compare against my branch here: https://github.com/PeterHollens/superdoc/tree/feature/drawingml-image-hyperlinks

I diffed your branch against my implementation and we converged on the same root cause, but your version is cleaner.

Reusing the shared link pipeline (`FlowRunLink` / `buildFlowRunLink` / the existing DOM wrapping) is the better call than keeping a separate `image.hyperlink` path around. It means image links inherit the normal blocked-link behavior, `target` / `rel` handling, dataset propagation, tooltip accessibility, and whatever else we add to link behavior later. My version currently reimplements part of that in an image-specific helper. It works, but it's the worse abstraction.

You were also right about the inline gap. That was the missing case. The fix gets a lot more solid once you add explicit PM adapter and DOM painter coverage for the inline-image path.

This is my first issue, so I'm not totally sure what's the best approach here 🤔

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -3449,6 +3452,62 @@ export class DomPainter {
...
+    display: 'block' | 'inline-block',
+  ): HTMLElement {
+    if (!hyperlink?.url || !this.doc) return imageEl;
+
+    const sanitized = sanitizeHref(hyperlink.url);
+    if (!sanitized?.href) return imageEl;
+
+    const anchor = this.doc.createElement('a');
+    anchor.href = sanitized.href;
+    anchor.classList.add('superdoc-link');
```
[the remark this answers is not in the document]
caio-pizzol: @iguit0 this is a good catch — clicking a linked image in edit mode opens the link instead of selecting the image. the link handler runs first and skips the image selection path entirely.

caio-pizzol:
Hey @iguit0, this PR has had no activity for 8 days.

iguit0:
> Hey @iguit0, this PR has had no activity for 8 days.

Hey! Apologies for the delay. I have now resolved the merge conflicts and incorporated your suggestions. Please let me know if anything else needs adjustment.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -3677,6 +3680,65 @@ export class DomPainter {
...
+   * When `hyperlink` is present and its URL passes sanitization, returns an
+   * `<a class="superdoc-link">` wrapping `imageEl`. The existing EditorInputManager
+   * click-delegation on `a.superdoc-link` handles both viewing-mode navigation and
+   * editing-mode event dispatch automatically, with no extra wiring needed here.
+   *
+   * When `hyperlink` is absent or the URL fails sanitization the original element
+   * is returned unchanged.
+   *
+   * @param imageEl   - The image element (img or span wrapper) to potentially wrap.
+   * @param hyperlink - Hyperlink metadata from the ImageBlock/ImageRun, or undefined.
```
caio-pizzol, line 3695: **editing mode**: clicking a linked image selects it (resize handles show up) but doesn't open the link. ctrl+click doesn't work either (in Word ctrl+click or just click will open img URL straight away - no resize selection appears)
image

**tooltip**: shows "Image" instead of the actual tooltip URL (e.g. "https://superdoc.dev").
image

caio-pizzol (commented):
@iguit0 prior feedback is all addressed, nice.

one thing: linked images work in viewing mode but not in editing mode — you can select them but can't follow the link. left an inline comment with details.

rendering looks good, tested with a Word file with different hyperlink setups and everything renders correctly.

packages/layout-engine/painters/dom/src/renderer.ts
iguit0: The editing mode issue was happening because linked images were still going through the text-link popover flow, which only knows how to resolve ProseMirror `link` marks. Image hyperlinks are stored on the image node attrs instead, so the click ended up selecting the image rather than following the link.

I changed that path so linked images open/navigate directly in editing mode, same as viewing mode.

The tooltip issue was happening because the wrapped image was still carrying its own `title` (often `"Image"`), which could win over the link tooltip on hover. I changed linked image rendering to prefer the hyperlink tooltip, fall back to the hyperlink URL when no tooltip is set, and clear the nested image title so the correct text is shown.

Added tests for:
- linked images opening directly in editing mode
- linked image anchor navigation in editing mode
- inline linked images using hyperlink tooltip/URL instead of the image title

Also verified manually with the repro doc: click now follows the image link in editing mode, no resize selection appears on that click, and the hover tooltip is correct.

SCR-20260404-pnds

packages/layout-engine/contracts/src/index.ts
```diff
@@ -361,6 +361,8 @@ export type ImageRun = {
   // OOXML image effects
   grayscale?: boolean; // Apply grayscale filter to image
   lum?: ImageLuminanceAdjustment; // DrawingML luminance adjustment from a:lum
+  /** Image hyperlink from OOXML a:hlinkClick. When set, clicking the image opens the URL. */
+  hyperlink?: { url: string; tooltip?: string };
```
caio-pizzol, line 365: `{ url: string; tooltip?: string }` is copy-pasted in ~5 places. worth pulling into a shared type so they don't drift apart? not blocking.

caio-pizzol (approved):
@iguit0 everything from the previous rounds is fixed — tested manually and it all works. one small cleanup idea inline, nothing blocking. lgtm.

PR #2557, opened 2026-03-25
fix: disable calling splitRunToParagraph when unavailable

VladaHarbour:


packages/super-editor/src/core/extensions/keymap.js
```diff
@@ -3,16 +3,22 @@ import { Extension } from '../Extension.js';
 import { isIOS } from '../utilities/isIOS.js';
 import { isMacOS } from '../utilities/isMacOS.js';
 
+const dispatchHistoryBoundary = (view) => {
+  const tr = view?.state?.tr;
+  if (typeof tr?.setMeta !== 'function') return;
```
caio-pizzol: tested this in the browser — `setMeta` is always there in both the main editor and comment editor. this check only kicks in for the test mock, never for real users. a simpler null check would do the same job.

```suggestion
  if (!tr) return;
```

packages/super-editor/src/core/extensions/keymap.test.js
```diff
@@ -25,6 +26,37 @@ const setupKeymap = async ({ isMacOS, isIOS }) => {
...
+
+      for (const command of chain) {
+        if (command()) return true;
+      }
+
+      return false;
+    });
+
+    const editor = {
+      view: {
```
caio-pizzol, line 49: if you simplify the guard above, give this mock a `setMeta` so the test actually runs through the real code path instead of bailing out early.

```suggestion
        state: { tr: { setMeta: vi.fn(() => ({})) } },
```

PR #2559, opened 2026-03-25
fix: newline style inheritance after clearing inline formatting (SD-2228)

luccas-harbour:
**Summary**
This PR fixes newline formatting inheritance around explicit formatting removal at a collapsed caret.

The main change is to treat explicit cursor formatting state as authoritative when splitting paragraphs. When the user removes bold/italic at the end of a formatted line and presses Enter, the new paragraph should no longer inherit the previous run’s inline formatting. This now works for both split paths: standard paragraph splits and run-to-paragraph splits.

**What Changed**
- Added explicit handling for `storedMarks = []` in formatting-state resolution so cleared formatting is preserved at the caret instead of being rehydrated from inherited paragraph/run formatting.
- Preserved collapsed-selection `storedMarks` through run-wrapping, run cleanup, and inline run-property recalculation so plugins do not accidentally lose the user’s explicit formatting state.
- Extracted shared split inheritance logic into [`splitParagraphRunProperties.js`](/Users/luccascorrea/Dev/superdoc-wt/luccas/sd-2228-new-lines-style-inheritance/packages/super-editor/src/core/helpers/splitParagraphRunProperties.js) and applied it to both [`splitBlock.js`](/Users/luccascorrea/Dev/superdoc-wt/luccas/sd-2228-new-lines-style-inheritance/packages/super-editor/src/core/commands/splitBlock.js) and [`split-run.js`](/Users/luccascorrea/Dev/superdoc-wt/luccas/sd-2228-new-lines-style-inheritance/packages/super-editor/src/extensions/run/commands/split-run.js).
- Added unit coverage for:
  - explicit empty `storedMarks`
  - empty paragraph split inheritance
  - stored-mark precedence over previous run formatting
  - run split behavior preserving explicit marks
- Added behavior coverage for:
  - SD-2228 newline style inheritance after removing bold+italic from line end
  - toggling bold off, pressing Enter, and ensuring the next line is not bold

**Why**
Before this change, pressing Enter after clearing inline formatting at the end of a formatted line could still carry formatting into the next paragraph because the split logic and follow-up run plugins fell back to the previous run’s formatting instead of respecting explicit caret state.

### DIFF
--- a/packages/super-editor/src/core/helpers/splitParagraphRunProperties.js
+++ b/packages/super-editor/src/core/helpers/splitParagraphRunProperties.js
@@ -1,13 +1,5 @@
 import { decodeRPrFromMarks } from '@converter/styles.js';
-
-/**
- * @param {Record<string, unknown> | null | undefined} runProperties
- * @returns {Record<string, unknown> | null}
- */
-function normalizeRunProperties(runProperties) {
-  if (!runProperties || typeof runProperties !== 'object') return null;
-  return Object.keys(runProperties).length > 0 ? runProperties : null;
-}
+import { normalizeRunProperties } from './normalizeRunProperties.js';
 
 /**
  * Extracts runProperties from the cursor context.
--- a/packages/super-editor/src/core/helpers/getMarksFromSelection.js
+++ b/packages/super-editor/src/core/helpers/getMarksFromSelection.js
@@ -3,6 +3,7 @@ import { calculateResolvedParagraphProperties } from '@extensions/paragraph/reso
 import { decodeRPrFromMarks, encodeMarksFromRPr } from '@converter/styles.js';
 
 import { resolveRunProperties } from '@superdoc/style-engine/ooxml';
+import { normalizeRunProperties } from './normalizeRunProperties.js';
 
 export function getMarksFromSelection(state, editor) {
   return getSelectionFormattingState(state, editor).resolvedMarks;
@@ -250,11 +251,6 @@ function getParagraphRunContext($pos, editor) {
     tableInfo,
     numberingDefinedInline: Boolean(paragraphAttrs.paragraphProperties?.numberingProperties),
   };
-}
-
-function normalizeRunProperties(runProperties) {
-  if (!runProperties || typeof runProperties !== 'object') return null;
-  return Object.keys(runProperties).length > 0 ? runProperties : null;
 }
 
 function getSafeResolutionContext(editor, node, $pos, paragraphAttrs) {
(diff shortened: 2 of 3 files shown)


packages/super-editor/src/core/helpers/splitParagraphRunProperties.js
```diff
@@ -0,0 +1,72 @@
...
+ * looks at the node just before the cursor. For empty paragraphs, it falls
+ * back to `paragraphProperties.runProperties`.
+ *
+ * @param {import('prosemirror-model').ResolvedPos} $from
+ * @returns {Record<string, unknown> | null}
+ */
+export function getRunPropertiesAtCursor($from) {
+  const runNode = $from.nodeBefore;
+  if (runNode?.type.name === 'run' && runNode.attrs.runProperties) {
+    return { ...runNode.attrs.runProperties };
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/super-editor/src/core/helpers/splitParagraphRunProperties.js
```diff
@@ -0,0 +1,76 @@
+import { decodeRPrFromMarks } from '@converter/styles.js';
+
+/**
+ * @param {Record<string, unknown> | null | undefined} runProperties
+ * @returns {Record<string, unknown> | null}
+ */
+function normalizeRunProperties(runProperties) {
```
caio-pizzol: this `normalizeRunProperties` function is a copy of the one in `getMarksFromSelection.js` — same code, two places. worth pulling into a shared file so they stay in sync?

tests/behavior/tests/formatting/sd-2228-new-line-style-inheritance.spec.ts
```diff
@@ -0,0 +1,44 @@
+import { test, expect } from '../../fixtures/superdoc.js';
```
caio-pizzol, line 1: the behavior tests cover typing from scratch, but there's no test for opening a Word file that already has bold text, toggling bold off, and pressing Enter. that's a slightly different path since the formatting comes from the imported file instead of the user. could be worth adding with a small fixture doc.

luccas-harbour: The issue here is more of an interaction problem, not really related to document import. The ticket itself exemplifies the problem from a blank editor and the behavior test reproduces it as described. To be honest I don't think adding a fixture doc would test anything else related to this problem.

packages/super-editor/src/core/helpers/splitParagraphRunProperties.js
luccas-harbour: good catch. I refactored it to use a shared helper

PR #2562, opened 2026-03-25
fix: handle newline characters in structured content table cell values (SD-2339)

luccas-harbour:
## Summary
- `appendRowsToStructuredContentTable` was dropping text after `\n` in cell values because `buildFormattedCellBlock` passed the raw string into a single ProseMirror text node
- Now splits newline-containing strings into separate paragraph nodes (one per line) via `Fragment`
- Added behavior tests confirming multi-line cell values render all content across Chromium, Firefox, and WebKit


packages/super-editor/src/extensions/table/tableHelpers/appendRows.js
```diff
@@ -164,6 +164,16 @@ export function buildFormattedCellBlock(schema, value, { blockType, blockAttrs,
     return type.createAndFill(blockAttrs || null, content);
   }
 
+  // Split on newline characters so each line becomes its own paragraph
+  const lines = text.split(/\r?\n/);
+  if (lines.length > 1) {
+    const paragraphs = lines.map((line) => {
+      const content = line ? schema.text(line, marks) : null;
+      return type.createAndFill(blockAttrs || null, content);
+    });
```
caio-pizzol: if any line fails to create a node here, `Fragment.from` will crash. adding `.filter(Boolean)` before it would skip bad entries instead of blowing up.

```suggestion
    }).filter(Boolean);
```

luccas-harbour: nice one. I implemented your suggestion

caio-pizzol (approved):
@luccas-harbour `.filter(Boolean)` is in, nice. no correctness issues — pushed a few unit tests for the multiline path (`\n`, `\r\n`, trailing newline) to round out coverage. approving.

PR #2576, opened 2026-03-26
fix: hide image resize handles in view mode

VladaHarbour:


packages/super-editor/src/components/SuperEditor.vue
```diff
@@ -1112,6 +1123,7 @@ const handleSuperEditorClick = (event) => {
   // Update table resize overlay on click
   if (isViewingMode()) {
     hideTableResizeOverlay();
+    clearSelectedImage();
```
caio-pizzol: we tested this — if resize handles are showing when you switch to viewing mode, they stick around until the mouse moves. clicking doesn’t clear them because `hideImageResizeOverlay()` is missing here. one-line fix:

```suggestion
    hideTableResizeOverlay();
    hideImageResizeOverlay();
    clearSelectedImage();
```

VladaHarbour: pushed some updates to clear selection on mode change

PR #2651, opened 2026-03-30
chore: remove custom mark and node examples

harbournick:


caio-pizzol:
@harbournick remove extensions from docs here: https://github.com/superdoc-dev/superdoc/pull/2674

PR #2658, opened 2026-03-31
SD-2277 - fix: tracked change doesn't undo correctly

chittolinag:
### Issue

Tracked-change bubbles weren’t updated when the history replay rebuilt suggestions, so undoing part of an insertion (typing “add text” → undoing “ text”) removed those characters from the document but left the sidebar bubble showing “add text.” 

### Fix

This fix makes `syncTrackedChangeComments()` re-run `createOrUpdateTrackedChangeComment()` in “update” mode for existing suggestions, making sure that replay re-calculates the tracked-change text/deletion payload from the current state before emitting the bubble update.

The fix keeps full metadata per tracked-change ID (runtime ID + imported ID) and adds a `refreshExisting` mode. When replay runs, the helper now:

1. Collects existing unresolved tracked-change comments and memoizes them by ID.
2. When we find a grouped change whose ID already exists, we let it through and call `createOrUpdateTrackedChangeComment()` with `event: 'update'`.
3. The update path re-extracts insertion/deletion text directly from the doc and emits the updated payload via `handleTrackedChangeUpdate()`.


packages/superdoc/src/stores/comments-store.js
```diff
@@ -1028,9 +1037,17 @@ export const useCommentsStore = defineStore('comments', () => {
...
       const id = insertedMark?.mark.attrs.id || deletionMark?.mark.attrs.id || formatMark?.mark.attrs.id;
-      if (!id || skipIds.has(id)) return;
+      if (id == null) return;
+      const normalizedId = String(id);
+      if (processedIds.has(normalizedId)) return;
+      processedIds.add(normalizedId);
+
+      if (!refreshExisting && skipIds.has(normalizedId)) return;
+      const existingTrackedChange = existingTrackedChangeById.get(normalizedId);
+      if (!refreshExisting && existingTrackedChange) return;
```
caio-pizzol: this check can never trigger — the `skipIds.has()` check above already covers it. the lookup on 1049 should stay though, you need it later on 1061.

```suggestion
      if (!refreshExisting && skipIds.has(normalizedId)) return;
      const existingTrackedChange = existingTrackedChangeById.get(normalizedId);
```

caio-pizzol (commented):
@chittolinag looks good! tested locally — undo/redo updates the sidebar bubble correctly now. one small cleanup inline.

packages/superdoc/src/stores/comments-store.js
chittolinag: nice catch! it turns out that check can be completely removed. thx @caio-pizzol

caio-pizzol (approved):
@chittolinag lgtm! added a test for the bubble text after undo/redo, all green.

PR #2659, opened 2026-03-31
fix: large header is missing and cursor is buggy in doc with footnotes

VladaHarbour:


packages/super-editor/src/editors/v1/dom-observer/index.ts
```diff
@@ -8,7 +8,7 @@
  * @module dom-observer
  */
 
-export { DomPositionIndex, type DomPositionIndexEntry } from './DomPositionIndex.js';
+export { DomPositionIndex, type DomPositionIndexEntry, isFootnotePaintedBlockHost } from './DomPositionIndex.js';
```
caio-pizzol: `isFootnotePaintedBlockHost` is exported from the barrel but nobody imports it from here — `PresentationEditor.ts` imports directly from `DomPositionIndex.ts`. remove?

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
```diff
@@ -2433,6 +2434,9 @@ export class PresentationEditor extends EventEmitter {
       for (let idx = 0; idx < layout.pages.length; idx++) {
         const page = layout.pages[idx];
         for (const fragment of page.fragments) {
+          if (isFootnoteLayoutBlockId((fragment as { blockId?: string }).blockId)) {
```
caio-pizzol: this same page-scan loop (skip footnotes → match position → get page index) is copy-pasted at ~2434, ~2636, and ~6566. small helper like `#findPageIndexForPosition(layout, pos)` would clean it up — worth doing?

caio-pizzol (commented):
@VladaHarbour looks good, no blockers.

both root causes are fixed correctly:
1. first-page headers now get cached (not just the default one), so they render on load
2. header heights survive the footnote relayout pass, so body text doesn't overlap tall headers

the footnote position guards are also correct — footnotes use a separate position space that overlaps with the body, so skipping them prevents wrong cursor/scroll jumps.

### spec check

tested with 6 Word-native documents against ECMA-376 rules:

| test | what it checks | result |
|------|---------------|--------|
| titlePg + first header + footnotes | core bug fix | pass |
| multi-section, sec2 inherits first header | variant inheritance | pass |
| multi-section, sec2 titlePg=false | first header suppressed | pass |
| first header + 3 footnotes | both fixes together | pass |
| evenAndOddHeaders | even page header selection | fail pre-existing gap, tracked as [SD-2450](https://linear.app/superdocworkspace/issue/SD-2450) |
| empty first header | suppresses inheritance | pass |

5 passing test docs uploaded to the rendering corpus for layout + visual regression coverage.

### even/odd gap (pre-existing, tracked as [SD-2450](https://linear.app/superdocworkspace/issue/SD-2450))

SuperDoc always shows the default header — it doesn't switch to the even header on even pages. Word does. this isn't caused by this PR and the cache already includes even refs, so it'll be ready when even/odd rendering is added.

[test5-even-odd-headers.docx](https://github.com/user-attachments/files/26448392/test5-even-odd-headers.docx)

SuperDoc:
image

Word:
image

left 2 small inline comments, neither blocking.

caio-pizzol (approved):
@VladaHarbour both items from last round are fixed, nice.

tested the even/odd headers with the same test file from round 1 — SuperDoc now matches Word. odd pages show the odd/default header, even pages show the even header (correctly inherited across sections).

pushed a commit adding tests for the even/odd behavior.

VladaHarbour:
Hi @caio-pizzol! Resolved conflicts here and checked that PR is still relevant

VladaHarbour:
Hi @harbournick! This PR still fixes odd/even header issue that Caio reported here https://github.com/superdoc-dev/superdoc/pull/2659#pullrequestreview-4053058399 and also improves footnotes rendering

packages/layout-engine/layout-bridge/src/headerFooterUtils.ts
```diff
@@ -357,17 +372,15 @@ export function getHeaderFooterTypeForSection(
...
-    if (pageNumber % 2 === 0 && (hasEven || hasDefault)) {
-      return hasEven ? 'even' : 'default';
-    }
-    if (pageNumber % 2 === 1 && (hasOdd || hasDefault)) {
-      return hasOdd ? 'odd' : 'default';
-    }
+    // Keep parity-based variant selection even when this section doesn't
+    // explicitly define that variant. Resolution/inheritance happens later.
+    if (!hasAny) return null;
+    return pageNumber % 2 === 0 ? 'even' : 'odd';
```
caio-pizzol, line 384: before, this returned `'default'` when no explicit odd/even existed. now it always returns `'odd'`/`'even'`, and the lookups below hit a null slot and bail. minimal repro: section with `alternateHeaders: true` and only `headerRefs.default`, page 1 returns `null` instead of the default ref.

want to fall back to `'default'` here when there's no parity ref to inherit, or add the fallback in the lookups so the inheritance behavior stays intact?

caio-pizzol (commented):
hey @VladaHarbour, one thing to flag.

i ran a quick repro: for a section with `alternateHeaders` on but only a `default` ref, the bridge resolvers (lines 428-447, 505-513) look up the parity slot directly and return `null`. the main render path has this fallback in `HeaderFooterSessionManager`, so the bridge helpers are the odd ones out. left it inline.

packages/layout-engine/layout-bridge/src/headerFooterUtils.ts
VladaHarbour: Hi @caio-pizzol! Pushed a fix to return default header

packages/layout-engine/layout-bridge/src/headerFooterUtils.ts
```diff
@@ -412,21 +425,27 @@ export function getHeaderFooterIdForPage(
   });
   if (!variantType) return null;
 
+  const resolveVariantId = (ids: Partial<SectionHeaderFooterIds> | undefined): string | null => {
+    if (!ids) return null;
+    const direct = ids[variantType];
+    if (direct) return direct;
+    // Under alternate headers/footers, Word still falls back to section default
```
caio-pizzol: tested with a 4-page doc that has different odd/even headers turned on but only the default header set

[word-fixture-pr-2659-only-default.docx](https://github.com/user-attachments/files/27177604/word-fixture-pr-2659-only-default.docx)

all four pages show the default text, including pages 2 and 4. the spec says even pages should be blank in this case. only matters for docs not made by Word, since Word always adds an empty even header (which we render correctly). flagging for later, not blocking.

caio-pizzol (commented):
hey @VladaHarbour! thanks for addressing last round, fix lands cleanly :)

one small thing inline. nits only.

caio-pizzol:
@VladaHarbour feel free to merge after addressing last comment

packages/layout-engine/layout-bridge/src/headerFooterUtils.ts
VladaHarbour: must be fixed now

PR #2661, opened 2026-03-31
fix: disable sdt hover in view mode for web layout

VladaHarbour:


caio-pizzol (commented):
@VladaHarbour tested with Word docs in web layout viewing mode — hover suppression and inline click blocking work nicely.

two small things: (1) clicking a block SDT in viewing mode still shows a blue selection outline in web layout. (2) when the document only contains a block SDT with nothing around it, the cursor stays inside instead of moving out on mode switch. both are minor since the CSS hides the hover, but worth a look.

pushed two behavior tests for click suppression in viewing mode. left inline comments. feel free to merge after addressing those.

packages/super-editor/src/editors/v1/core/helpers/getViewModeSelectionWithoutStructuredContent.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  return null;
+}
+
+export function getViewModeSelectionWithoutStructuredContent(state) {
+  const { selection, doc } = state;
+
+  if (selection instanceof NodeSelection && STRUCTURED_CONTENT_NODE_TYPES.has(selection.node.type.name)) {
+    return Selection.near(doc.resolve(selection.from), -1);
```
caio-pizzol: when the block SDT is the only thing in the document, there's nowhere else to put the cursor — so it ends up back inside the SDT. no visible issue since the CSS hides the highlight, but a quick guard would make this safer:

```suggestion
    const candidate = Selection.near(doc.resolve(selection.from), -1);
    const candidatePos = findEnclosingStructuredContentPosition(candidate.$from);
    if (candidatePos !== null) return null;
    return candidate;
```

packages/super-editor/src/editors/v1/extensions/structured-content/StructuredContentViewBase.js
```diff
@@ -97,6 +98,12 @@ export class StructuredContentViewBase {
       return false;
     }
 
+    // In viewing mode, suppress node-wrapper clicks so SDTs don't become
+    // selected when the PM fallback is used.
+    if (isViewingMode && isClickEvent && isSelectable) {
```
caio-pizzol, line 103: clicking a block SDT in viewing mode still shows a blue selection outline in web layout. might need a CSS rule to hide it:

```css
.super-editor .ProseMirror.view-mode .sd-structured-content-block.ProseMirror-selectednode {
  outline: none;
}
```

caio-pizzol (approved):
lgtm (after addressing comments)

packages/super-editor/src/editors/v1/extensions/structured-content/StructuredContentViewBase.js
VladaHarbour: Created test document with block sdt filed and can't reproduce this. Do you have some specific document where this issue appears?

caio-pizzol (approved):
@VladaHarbour both fixes look good, tests cover the edge cases. approving.

PR #2662, opened 2026-03-31
fix: minor css vars issues

artem-harbour:
A couple of minor fixes for dark themes and in general.

### DIFF
--- a/packages/superdoc/src/assets/styles/helpers/themes.css
+++ b/packages/superdoc/src/assets/styles/helpers/themes.css
@@ -34,6 +34,7 @@
   --sd-ui-comments-card-shadow: 0 1px 2px rgba(60, 64, 67, 0.2), 0 2px 6px rgba(60, 64, 67, 0.18);
   --sd-ui-comments-separator: #c7cdd6;
   --sd-ui-comments-timestamp-text: #5f6368;
+  --sd-ui-comments-input-bg: #fff;
   --sd-ui-comments-option-hover-bg: #e8eaed;
   --sd-ui-comments-input-border: #80868b;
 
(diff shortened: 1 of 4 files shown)


packages/superdoc/src/assets/styles/helpers/themes.css
```diff
@@ -34,6 +34,7 @@
   --sd-ui-comments-card-shadow: 0 1px 2px rgba(60, 64, 67, 0.2), 0 2px 6px rgba(60, 64, 67, 0.18);
   --sd-ui-comments-separator: #c7cdd6;
   --sd-ui-comments-timestamp-text: #5f6368;
+  --sd-ui-comments-input-bg: #fff;
```
artem-harbour, line 37: This variable was not in the original PR intentionally (but apparently added later), because it does not refer to the background of the input/mini-editor (the input is always white), but to its wrapper.

That's why I redefine it for each theme.

Before:
Image

After:
Image

@caio-pizzol

packages/superdoc/src/dev/themes/neon-night.css
```diff
@@ -50,7 +50,7 @@
   --sd-ui-comments-author-text: #e7f7ff;
   --sd-ui-comments-timestamp-text: #8fc4ef;
   --sd-ui-comments-body-text: #d6ecff;
-  --sd-ui-comments-input-bg: #0e1d33;
+  --sd-ui-comments-input-bg: #fff;
```
artem-harbour, line 53: The same.

caio-pizzol, line 53: ~~this sets the comment input background to white, but neon-night is a dark theme — the text in there is light-colored, so it'd be nearly invisible on white. the old `#0e1d33` looked intentional. should it stay dark?~~

caio-pizzol (approved):
@artem-harbour looks good — the color fixes on toolbar icons and comment header are solid, and the input background additions in the preset themes are needed (they'd inherit a gray tint otherwise). one thing on neon-night — left an inline comment.

packages/superdoc/src/assets/styles/helpers/themes.css
caio-pizzol: nice catch

packages/superdoc/src/dev/themes/neon-night.css
caio-pizzol: makes sense — the wrapper should match the input. my inline comment was off, resolving it.

PR #2663, opened 2026-03-31
feat: bookmark api for cross session document addressing (SD-2358)

luccas-harbour:
This PR adds cross-session, story-aware document addressing for bookmarks and introduces a higher-level navigation API for bookmarks, comments, and tracked changes.

On the document API side, bookmark addresses can now include a `story`, and `bookmarks.list()` can be scoped with `in`. The super-editor adapters now resolve bookmark operations against the correct story runtime, including header/footer stories, and return story-qualified bookmark addresses for non-body stories. At the same time, this branch restores document-wide bookmark name uniqueness so we remain compliant with OOXML even when bookmarks are created outside the body.

On the navigation side, `SuperDoc.navigateTo()` now supports entity-based navigation for:
- bookmarks in the body and header/footer stories
- comments in the body
- tracked changes in the body

For bookmarks, anchor navigation can now fall back to resolving the target from the active ProseMirror document when the layout bookmark map is insufficient. For header/footer bookmark navigation, the editor activates the target region, waits for the live sub-editor, and places the caret at the resolved bookmark position.

This branch also updates the document-api contract/types/schema surface to expose these capabilities, adds conformance and unit coverage for story-aware bookmark handling, and adds behavior tests covering:
- header bookmark CRUD
- entity navigation to bookmarks, comments, and tracked changes

### DIFF
--- a/packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
+++ b/packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
@@ -5929,12 +5929,12 @@ export class PresentationEditor extends EventEmitter {
     }
 
     // Comments and tracked changes are only supported in the document body.
-    // Reject non-body story locators early with a clear error.
     if ('story' in target && target.story && target.story.storyType !== 'body') {
-      throw new Error(
-        `navigateTo does not support non-body stories for ${target.entityType}. ` +
+      console.warn(
+        `[PresentationEditor] navigateTo does not support non-body stories for ${target.entityType}. ` +
           `Only bookmarks support cross-story navigation.`,
       );
+      return false;
     }
 
     if (target.entityType === 'trackedChange') {
--- a/packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
+++ b/packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
@@ -115,6 +115,16 @@ import { ySyncPluginKey } from 'y-prosemirror';
 import type * as Y from 'yjs';
 import type { HeaderFooterDescriptor } from '../header-footer/HeaderFooterRegistry.js';
 import { isHeaderFooterPartId } from '../parts/adapters/header-footer-part-descriptor.js';
+import type {
+  BookmarkAddress,
+  NavigableEntityAddress,
+  StoryLocator,
+  ResolveRangeOutput,
+  DocumentApi,
+} from '@superdoc/document-api';
+import { resolveStoryRuntime } from '../../document-api-adapters/story-runtime/resolve-story-runtime.js';
+import { resolveBookmarkTarget } from '../../document-api-adapters/helpers/bookmark-resolver.js';
+import { resolveTrackedChange } from '../../document-api-adapters/helpers/tracked-change-resolver.js';
 import type { PartChangedEvent } from '../parts/types.js';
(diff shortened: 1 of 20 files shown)


packages/superdoc/src/core/SuperDoc.js
```diff
@@ -65,6 +65,7 @@ const DEFAULT_AWARENESS_PALETTE = Object.freeze([
 /** @typedef {import('./types').UpgradeToCollaborationOptions} UpgradeToCollaborationOptions */
 /** @typedef {import('./types').SurfaceRequest} SurfaceRequest */
 /** @typedef {import('./types').SurfaceHandle} SurfaceHandle */
+/** @typedef {import('@superdoc/document-api').NavigableEntityAddress} NavigableEntityAddress */
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
```diff
@@ -5897,6 +5921,190 @@ export class PresentationEditor extends EventEmitter {
...
+      }
+
+      const runtime = resolveStoryRuntime(this.#editor, story);
+      const resolved = resolveBookmarkTarget(runtime.editor.state.doc, target);
+      if (typeof runtime.editor.commands?.setTextSelection !== 'function') return false;
+      runtime.editor.commands.setTextSelection({ from: resolved.pos, to: resolved.pos });
+      runtime.editor.view?.focus?.();
+      return true;
+    } catch (error) {
+      this.#exitHeaderFooterMode();
```
caio-pizzol: when `#waitForHeaderFooterEditor` throws, `#exitHeaderFooterMode()` runs once in the `finally` and again here in the outer `catch`. the session exit itself is safe (it checks if already in body mode), but the second call still triggers an extra rerender and focus. could either guard `#exitHeaderFooterMode()` to skip if already in body mode, or drop this call since `finally` already handles it.

```suggestion
      console.error('[PresentationEditor] navigateTo bookmark failed:', error);
```

caio-pizzol (commented):
@luccas-harbour solid work — the cross-story bookmark API is well structured and tests are thorough.

one small correctness thing: `#exitHeaderFooterMode()` can get called twice when header/footer navigation fails. left an inline comment with a fix.

one cleanup suggestion on duplicated test helpers.

also opened SD-2454 for showing bookmark bracket indicators in the document (like Word's "Show bookmarks" toggle) — not needed for this PR but a nice follow-up.

left a couple inline comments.

tests/behavior/tests/navigation/header-bookmark-crud.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    const story: StoryLocator = {
+      kind: 'story',
+      storyType: 'headerFooterSlot',
+      section: entry.section,
+      headerFooterKind: 'header',
+      variant: entry.variant ?? 'default',
+    } as any;
+
+    const toRanges = (item: any) => {
```
caio-pizzol, line 42: the `toRanges` helper here and in `navigate-to-entity.spec.ts` do the same thing but slightly differently — the other one also handles a legacy fallback. worth pulling into a shared helper in `tests/behavior/helpers/document-api.ts`?

caio-pizzol:
Closing this — the navigation parts (`navigateTo` for comments, tracked changes, bookmarks) shipped separately in #2772 and #2774. The bookmark-only improvements are still valuable, will open a focused PR for those.

PR #2667, opened 2026-03-31
SD-2343 - Table borders being converted twice

chittolinag:
## Issue

DOCX tables store border thickness in eighths of a point (`w:sz`). The table translator (_processTableBorders in `tbl-translator`) already runs `w:sz` through `eighthPointsToPixels`, so the `size` passed downstream is in pixels. However, `convertBorderSpec()` / `convertTableBorderValue()` still assumed `size` was an OOXML eighth-point value and converted it again. This makes SuperDoc render noticeably thinner borders than Word in some cases.

## Fix

When calculating the final border values, do not convert it to px again since it's already converted.


packages/layout-engine/pm-adapter/src/attributes/borders.ts
```diff
@@ -101,11 +105,15 @@ export function convertBorderSpec(ooxmlBorder: unknown): BorderSpec | undefined
     return { style: 'none' as BorderStyle, width: 0 };
   }
 
-  const width = borderSizeToPx(sizeNumber);
```
chittolinag, line 104: We were converting it again here when it already was in pixels. Adding other verifications just as a fail safe.

packages/layout-engine/pm-adapter/src/converters/table.ts
```diff
@@ -853,26 +865,31 @@ export function tableNodeToBlock(
...
     ) {
+      borderUnit = 'px';
       return node.attrs.borders as Record<string, unknown>;
     }
     if (
       hydratedTableStyle?.borders &&
       typeof hydratedTableStyle.borders === 'object' &&
       hydratedTableStyle.borders !== null
     ) {
+      borderUnit = 'eighthPoints';
```
caio-pizzol: this `'eighthPoints'` branch never runs. the borders coming through here are already converted, so they skip right past it. confirmed by loading 6 different tables at runtime — all took the `'px'` path. safe to remove or add a note explaining why.

packages/layout-engine/pm-adapter/src/converters/table.ts
```diff
@@ -853,26 +865,31 @@ export function tableNodeToBlock(
...
   const tableAttrs: Record<string, unknown> = {};
+  let borderUnit: 'px' | 'eighthPoints' = 'px';
   const getBorderSource = (): Record<string, unknown> | undefined => {
     if (
       node.attrs?.borders &&
       typeof node.attrs.borders === 'object' &&
       node.attrs.borders !== null &&
       Object.keys(node.attrs.borders as Record<string, unknown>).length > 0
     ) {
+      borderUnit = 'px';
```
caio-pizzol: `getBorderSource()` quietly changes `borderUnit` as a side effect. returning `{ borders, unit }` instead would make it obvious what's happening.

packages/layout-engine/pm-adapter/src/attributes/borders.ts
```diff
@@ -43,6 +48,16 @@ const borderSizeToPx = (size?: number): number | undefined => {
   return Math.min(MAX_BORDER_SIZE_PX, Math.max(MIN_BORDER_SIZE_PX, pixelValue));
```
caio-pizzol: three different functions now convert eighth-points to pixels: one here (clamps to [0.5, 100]px), one in normalizeNewTableAttrs.js and one in tables-adapter.ts (neither clamps). borders could render differently depending on which path runs. worth consolidating?

packages/super-editor/src/editors/v1/extensions/table/tableHelpers/normalizeNewTableAttrs.js
```diff
@@ -53,9 +54,11 @@ export function normalizeNewTableAttrs(editor) {
   const resolved = resolvePreferredNewTableStyleIdFromEditor(editor);
 
   if (resolved.source === 'none') {
+    const fallbackPixelBorders = cloneBorders(TABLE_FALLBACK_BORDERS, convertBorderSizeToPx);
```
caio-pizzol: `cloneBorders` here and `cloneBorderMap` in tables-adapter.ts do the same thing. ~25 lines of duplication.

caio-pizzol (commented):
@chittolinag fix is correct — I tested with a Word-generated doc (6 tables, thin to thick borders) and all widths come out right. no double conversion.

one dead code path to clean up and a few suggestions to reduce duplication. left inline comments.

packages/super-editor/src/editors/v1/extensions/table/tableHelpers/normalizeNewTableAttrs.js
chittolinag: Good point. they're pretty similar, with a small difference that `cloneBorders` optionally converts them. I'll consolidate them.

packages/layout-engine/pm-adapter/src/attributes/borders.ts
chittolinag: you're right - I've just consolidated them.

packages/layout-engine/pm-adapter/src/converters/table.ts
chittolinag: Interesting - this is from previous commits added 3 months ago. I wonder if by removing this, we could break something else.

packages/layout-engine/pm-adapter/src/attributes/borders.ts
```diff
@@ -32,15 +36,16 @@ const MAX_BORDER_SIZE_PX = 100; // Reasonable maximum
  *
  * Clamps results to reasonable bounds to prevent edge cases.
  */
-const borderSizeToPx = (size?: number): number | undefined => {
-  if (!isFiniteNumber(size)) return undefined;
-  if (size <= 0) return 0;
+export const borderSizeToPx = (size?: number): number | undefined => eighthPointsToPixels(size, { clamp: true });
```
caio-pizzol, line 39: building pm-adapter on its own breaks here. it picks up an older copy of the helper from `super-editor/dist` that still takes one argument, so `pnpm --filter @superdoc/pm-adapter build` fails with `TS2554: Expected 1 arguments, but got 2`. confirmed the snippet below makes it pass.

```suggestion
export const borderSizeToPx = (size?: number): number | undefined => {
  const pixels = eighthPointsToPixels(size);
  if (!isFiniteNumber(pixels)) return undefined;
  if (pixels <= 0) return 0;
  return clampPixelBorderWidth(pixels);
};
```

caio-pizzol (commented):

two things to fix before merge. one inline. also, the `extractCellBorders` JSDoc example at `packages/layout-engine/pm-adapter/src/attributes/borders.ts:275` is out of date - with the new default unit, size 8 returns width 8 (not 1.33) and size 16 returns width 16. the new `options` argument is also missing from the JSDocs of `convertBorderSpec`, `convertTableBorderValue`, and `extractTableBorders`.

caio-pizzol (approved):
hey @chittolinag! pushed the round 2 fixes myself :)

quick correction: the build-break i flagged last round was wrong on my end - it only happens with leftover build files from main. your branch is fine on its own.

lgtm.

PR #2672, opened 2026-04-01
fix: change default line height for header/footer editor

VladaHarbour:
It seems that lineHeight multiplier is not applied neither to header rendered with layout engine nor to header in Word. Removing multiplier which is applied to ProseMirror editor we still use for header/footer editing seems to be the most efficient and quick fix

### DIFF
--- a/packages/super-editor/src/editors/v1/core/renderers/ProseMirrorRenderer.ts
+++ b/packages/super-editor/src/editors/v1/core/renderers/ProseMirrorRenderer.ts
@@ -617,7 +617,7 @@ export class ProseMirrorRenderer implements EditorRenderer {
     }
 
     // Line height
-    proseMirror.style.lineHeight = String(DEFAULT_LINE_HEIGHT);
+    proseMirror.style.lineHeight = editor.options.isHeaderOrFooter ? String(1) : String(DEFAULT_LINE_HEIGHT);
 
     // Mobile styles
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/editors/v1/core/renderers/ProseMirrorRenderer.ts
```diff
@@ -617,7 +617,7 @@ export class ProseMirrorRenderer implements EditorRenderer {
     }
 
     // Line height
-    proseMirror.style.lineHeight = String(DEFAULT_LINE_HEIGHT);
+    proseMirror.style.lineHeight = editor.options.isHeaderOrFooter ? String(1) : String(DEFAULT_LINE_HEIGHT);
```
caio-pizzol: might be worth adding a `const HEADER_FOOTER_LINE_HEIGHT = 1` next to `DEFAULT_LINE_HEIGHT` on line 44 so it's clear why headers use a different value. Up to you though.

caio-pizzol (commented):
@VladaHarbour looks good — fixes the clipping and the value is correct per the spec. Tested locally, all header lines show up fine. Pushed a commit with 4 behavior tests. One small note inline, nothing blocking.

caio-pizzol (approved):
@VladaHarbour looks good, last round's feedback is addressed. approving.

PR #2676, opened 2026-04-01
fix: prevent text overlap when multi-column sections span pages (SD-1869)

luccas-harbour:
## Summary

- Fix text overlap caused by the CSS "strut" (inherited browser-default 16px font-size) misaligning normal-flow inline children relative to absolutely-positioned tab segments inside line containers. Sets `fontSize: '0'` on line containers and restores explicit font sizes on child elements that need them (field annotations, math wrappers, empty-run spans).
- Fix multi-column section overlap when columns have unequal heights by tracking `maxCursorY` across columns and using it as the starting Y for the next region.
- Fix tab run measurement using actual font size instead of a hardcoded 12pt default.


packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -4904,6 +4905,14 @@ export class DomPainter {
...
     wrapper.style.minWidth = `${run.width}px`;
     wrapper.style.minHeight = `${run.height}px`;
+    // Restore font-size so MathML content and the plain-text fallback render
+    // correctly (the line container sets fontSize: 0 to eliminate the CSS strut).
+    // We use run.height as a rough proxy — it's a layout heuristic from
+    // estimateMathDimensions(), not actual font metrics. The exact value doesn't
+    // matter much: MathML has its own internal scaling, and the plain-text
+    // fallback just needs a non-zero size. Don't rely on this for precise
+    // typographic sizing.
+    wrapper.style.fontSize = `${run.height}px`;
```
caio-pizzol: `run.height` can be 80-100px for tall math expressions (fractions, equation arrays). that gets used as the font size here, which would make the plain-text fallback huge. a fixed value like 24px or the paragraph's font size would be safer — worth changing?

caio-pizzol (commented):
@luccas-harbour the overlap fix works — pushed a regression test for it (`e859c96`).

side finding: the ECMA spec says continuous section breaks should balance content evenly across columns (like Word does — 6 paragraphs in 2 columns = 3+3). SuperDoc fills column 1 first, which is why multi-column docs look different from Word. this is pre-existing, not on this PR — filed [SD-2452](https://linear.app/superdocworkspace/issue/SD-2452/implement-column-balancing-for-continuous-section-breaks-ecma-376) for it.

left one inline comment.

caio-pizzol (approved):
feel free to merge after addressing last comment

packages/layout-engine/painters/dom/src/renderer.ts
luccas-harbour: sounds good. I addressed this.

PR #2680, opened 2026-04-01
fix(super-editor): constrain images to table cell width in web layout (SD-2416)

luccas-harbour:
## Summary

- Images inserted into table cells in web layout mode were not constrained to the cell width, causing them to overflow
- Hoisted the table cell detection loop in `getMaxContentSize()` so it runs for both print and web layout modes
- In web layout + table cell, returns `{ width: cellWidth, height: Infinity }` (no page height to constrain against)


caio-pizzol (approved):
@luccas-harbour looks good — images now get constrained in web layout too. pushed a commit with a couple extra tests. approving.

PR #2704, opened 2026-04-02
fix: apply linked style to selected range only

VladaHarbour:


caio-pizzol (approved):
@VladaHarbour nice work — the partial selection fix is correct and matches how Word handles linked styles. verified by creating a real Word document and comparing the output.

found one bug: applying a character style with just a cursor (no selection) was going through the wrong code path and changing the whole paragraph. fixed it so it only affects what you type next.

also cleaned up the split-run code — pulled the inline style check into its own helper and simplified a redundant condition.

added tests: unit tests for the new code paths, plus 5 browser tests (Chromium/Firefox/WebKit) covering partial apply, full apply, formatting preservation, Enter key, and cross-paragraph. all passing. also uploaded a Word-native test document to the corpus.

PR #2705, opened 2026-04-02
feat: add w:pPrChange translator for paragraph property tracked changes (SD-2417)

luccas-harbour:
## Summary
- Add a new `w:pPrChange` translator that encodes/decodes tracked changes on paragraph properties (author, date, id) along with the nested `w:pPr` containing the previous properties
- Register the translator as a child of `w:pPr` and in the global handler index
- Extract base pPr property translators into a shared module to avoid a circular dependency between `pPr` ↔ `pPrChange`


luccas-harbour:
> The current decoder has an interesting behavior — if a `change` object has only `{ id, author, date }` with no `paragraphProperties`, the decode returns `undefined` (as shown in the test at line ~145). That means a `pPrChange` with no prior paragraph properties (empty `<w:pPr/>`) won't round-trip cleanly. The spec explicitly allows an empty `<w:pPr/>` to mean "no prior properties existed." Whether that matters depends on whether this code path is exercised in practice, but it's worth noting for future work.

This issue was fixed.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/pPrChange/pPrChange-translator.js
```diff
@@ -0,0 +1,99 @@
...
+export const translator = NodeTranslator.from({
+  xmlName: 'w:pPrChange',
+  sdNodeOrKeyName: 'change',
+  type: NodeTranslator.translatorTypes.NODE,
+  attributes: ATTRIBUTE_HANDLERS,
+  encode: (params, encodedAttrs = {}) => {
+    const changeNode = params.nodes[0];
+    const pPrNode = changeNode?.elements?.find((el) => el.name === 'w:pPr');
+
+    let paragraphProperties = pPrNode ? pPrTranslator.encode({ ...params, nodes: [pPrNode] }) : undefined;
```
caio-pizzol: when `<w:pPr/>` exists but is empty, `pPrTranslator.encode()` returns `undefined` here — so the `paragraphProperties` key never makes it into the result. later, when decoding back to XML, the guard at line 81 (`hasParagraphProperties`) can't fire because the key is missing. the empty `<w:pPr/>` is lost.

the spec says `<w:pPr>` is required inside `<w:pPrChange>` — even when empty it means "there were no previous properties." tested with a real Word file and confirmed the element disappears after export.

```suggestion
    let paragraphProperties = pPrNode ? (pPrTranslator.encode({ ...params, nodes: [pPrNode] }) ?? {}) : undefined;
```

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/pPr/pPr-translator.js
```diff
@@ -1,82 +1,11 @@
...
-  wTabsTranslator,
-  wTextAlignmentTranslator,
-  wTextDirectionTranslator,
-  wTextboxTightWrapTranslator,
-  wTopLinePunctTranslator,
-  wWidowControlTranslator,
-  wWordWrapTranslator,
-  wRPrTranslator,
-];
+const propertyTranslators = [...basePropertyTranslators, wPPrChangeTranslator];
```
caio-pizzol, line 8: not blocking, just a heads up: the spec expects `sectPr` before `pPrChange` inside `<w:pPr>`, but the exporter always puts `sectPr` last. so if a paragraph has both a section break and a tracked property change, they'll be in the wrong order. this predates your PR — the pattern was already there, it just didn't matter until `pPrChange` existed. rare in practice. maybe a follow-up ticket?

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/utils.js
```diff
@@ -508,6 +510,7 @@ export function createNestedPropertiesTranslator(
   propertyTranslators,
   defaultEncodedAttrs = {},
   attributeHandlers = [],
+  { emitWhenAttributesOnly = false } = {},
 ) {
   const propertyTranslatorsByXmlName = {};
```
caio-pizzol, line 512: `emitWhenAttributesOnly` isn't used anywhere — the pPrChange translator handles this case with its own logic instead. worth removing so it doesn't sit here unused?

caio-pizzol (commented):
@luccas-harbour clean structure — sharing the base translators between pPr and pPrChange is a nice call.

one issue: when the original `<w:pPr/>` inside `<w:pPrChange>` is empty, it gets lost on round-trip. the decode guard you added works, but the encode side never feeds it — the key gets dropped for empty elements. tested with a Word file — the empty `<w:pPr/>` disappears after export.

[pPrChange-empty-pPr.docx](https://github.com/user-attachments/files/26472520/pPrChange-empty-pPr.docx)

also a minor DX thing and a pre-existing ordering note — details in inline comments.

tests look good. left a few comments.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/pPrChange/pPrChange-translator.js
luccas-harbour: sounds good. this was addressed as suggested.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/pPr/pPr-translator.js
luccas-harbour: ok. I addressed this in this PR.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/utils.js
luccas-harbour: oops, I removed it.

caio-pizzol (approved):
lgtm - round-1 items are all cleanly addressed. pushed 8c07f00c6 with the two test regression guards i flagged (mixed sectPr + other props, and pPrChange routing through wPPrTranslator); revert if you'd prefer your own take.

PR #2706, opened 2026-04-02
SD-2440 - fix: toc not loading on paragraphs

chittolinag:
### Issue

Importing documents where the TOC field sits inline inside a paragraph crashed because the converter always emitted block `tableOfContents` nodes that expect paragraph parents. When that type of document is imported, SuperDoc crashes completely.

### Solution

Detect when the parent disallows block nodes and encode the TOC within a paragraph block.


packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/sd/tableOfContents/tableOfContents-translator.js
```diff
@@ -56,8 +70,14 @@ const encode = (params) => {
...
 const decode = (params) => {
   const { node } = params;
+  const isInlineNode = node.type === 'tableOfContentsInline';
   const tocContent = Array.isArray(node.content) ? node.content : [];
-  const contentNodes = tocContent.map((n) => exportSchemaToJson({ ...params, node: n }));
+  const inlineContentNodes = tocContent.flatMap((n) => {
+    const exported = exportSchemaToJson({ ...params, node: n });
+    if (!exported) return [];
+    return Array.isArray(exported) ? exported : [exported];
+  });
```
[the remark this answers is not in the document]
caio-pizzol: @chittolinag agree with this — block TOC children get decoded twice and the first pass changes shared state, so the second pass can lose hyperlinks.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/sd/tableOfContents/tableOfContents-translator.js
```diff
@@ -56,8 +70,14 @@ const encode = (params) => {
...
+  const blockContentNodes = isInlineNode ? [] : tocContent.map((n) => exportSchemaToJson({ ...params, node: n }));
```
caio-pizzol: two things:

1. block TOCs decode every child twice — once for `inlineContentNodes` (thrown away) and once for `blockContentNodes` (kept). some decoders modify data as they go, so the second pass sees corrupted input. hyperlinks can get dropped. fix:

```suggestion
  const inlineContentNodes = isInlineNode
    ? tocContent.flatMap((n) => {
        const exported = exportSchemaToJson({ ...params, node: n });
        if (!exported) return [];
        return Array.isArray(exported) ? exported : [exported];
      })
    : [];
  const blockContentNodes = isInlineNode ? [] : tocContent.map((n) => exportSchemaToJson({ ...params, node: n }));
```

2. loaded the original `Template_Test_Report.docx` with debug logs — `tableOfContentsInline` was never produced. the paragraph wrapping at lines 42-49 is what prevents the crash. do you have a file that actually triggers the inline path?

caio-pizzol (commented):
@chittolinag tested with the original bug file + 13 other Word files. no crashes.

block TOCs decode children twice and the second pass can lose data — fix in the inline comment.

`tableOfContentsInline` was never triggered during testing, even with the original file. if no file produces it, the wrapping logic alone is the fix and the inline node type can be removed.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/sd/tableOfContents/tableOfContents-translator.js
chittolinag: Great catches. just fixed the 1st issue. I am now checking the 2nd issue you mentioned, it seems like there's also a problem when exporting the file (the opening in Word). I'll fix it.

chittolinag: Ok I've been trying to fix the export issue for a while now but no luck. I am pretty sure the export issue is unrelated though - after removing the TOC & re-exporting the document, the error persists. 

If you agree, I'll file a separate ticket to handle it, but in the mean time I'll make sure to fix the issues you mentioned here in the review.

chittolinag: done ✅

caiopizzol (approved):
@chittolinag fix works, tested with the original file — document loads fine. nice simplification on the latest push.

added a few tests and uploaded the bug file to the test corpus so it's covered going forward.

PR #2707, opened 2026-04-02
fix: table gap clicks selecting wrong paragraph and jumping scroll (SD-2356)

luccas-harbour:
## Summary

- Fix clicks in table cell gaps (between paragraphs) selecting the wrong paragraph and causing a scroll jump. Replaces the flawed "last paragraph" fallback in `hitTestTableFragment` with a nearest-distance algorithm that picks the closest paragraph to the click point.
- Fix non-text clicks (page margins, table padding) resetting the cursor to position 1, which triggered an unwanted scroll-to-top. Now preserves the existing cursor and scroll position.
- Add scoped line-container resolution in `DomPointerMapping` so gap clicks in nested table DOM resolve to the nearest line within the correct cell.


packages/super-editor/src/editors/v1/dom-observer/DomPointerMapping.ts
```diff
@@ -189,6 +189,14 @@ export function clickToPositionDom(domContainer: HTMLElement, clientX: number, c
     return resolveLineAtX(hitChainLine, clientX);
   }
 
+  if (fragmentEl.classList.contains(CLASS.tableFragment)) {
```
caio-pizzol, line 207: when a table splits across pages, the container here can hold lines from all pages of that table. so `findLineAtY` picks the closest line overall, which might be on a different page — and the view scrolls there.

tested with both customer docs, the jump still happens. could you scope this to only lines within the current page fragment?

packages/super-editor/src/editors/v1/core/presentation-editor/pointer-events/EditorInputManager.ts
```diff
@@ -1162,9 +1162,9 @@ export class EditorInputManager {
       }
     }
 
-    // Handle click outside text content
+    // Handle click outside text content — keep cursor and scroll position unchanged.
     if (!rawHit) {
```
caio-pizzol, line 1529: this path never fires during manual testing — the table extends into the margin area, so margin clicks get resolved as table hits instead of reaching here.

packages/super-editor/src/editors/v1/dom-observer/DomPointerMapping.ts
```diff
@@ -278,6 +286,18 @@ function resolveFragment(fragmentEl: HTMLElement, viewX: number, viewY: number):
   return resolveLineAtX(lineEl, viewX);
 }
 
+function resolveLineInContainer(containerEl: HTMLElement, viewX: number, viewY: number): number | null {
```
caio-pizzol: this function is the same as `resolveFragment` above — same query, same two calls. only difference is a debug log that's off by default. could reuse `resolveFragment` and delete this one?

tests/behavior/tests/navigation/click-scroll-jump.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+  expect(page3Box).not.toBeNull();
+
+  await page.mouse.click(page3Box!.x + page3Box!.width / 2, page3Box!.y + 15);
+  await superdoc.waitForStable(1000);
+
+  const scrollAfter = await getScrollTop(page);
+  const scrollDelta = Math.abs(scrollAfter - scrollBefore);
+  expect(
+    scrollDelta,
+    `Scroll jumped by ${scrollDelta}px after clicking page margin — expected no significant scroll change`,
```
caio-pizzol, line 79: `selBefore` is captured but never checked after the click. worth adding:

```ts
const selAfter = await superdoc.getSelection();
expect(selAfter.from).toBe(selBefore.from);
```

caio-pizzol (changes requested):
@luccas-harbour good direction — nearest-paragraph selection is a clear improvement over always picking the last one.

I tested both customer docs manually and the scroll jump still happens. two things going on:

1. when a table splits across pages, the container used to find the nearest line includes lines from ALL pages of that table — so the click can land on a line that's on a different page, and the view scrolls there.

2. clicking on page margins doesn't trigger the "no hit" path. the table is so wide that margin clicks get resolved as table clicks, so the cursor moves and the view jumps.

the `getFirstTextPosition` early-exit fix (the `found` flag) doesn't have a test that exercises it — existing tests use single-paragraph docs that pass either way. a quick test with two paragraphs would cover it.

one small cleanup suggestion and a couple test gaps. left inline comments.

caio-pizzol:
@luccas-harbour still working on this one?

packages/super-editor/src/editors/v1/dom-observer/DomPointerMapping.ts
luccas-harbour: done @caio-pizzol. I made a change to search only within the current page fragment

packages/super-editor/src/editors/v1/core/presentation-editor/pointer-events/EditorInputManager.ts
luccas-harbour: fixed this

packages/super-editor/src/editors/v1/dom-observer/DomPointerMapping.ts
luccas-harbour: sounds good. I refactored it

tests/behavior/tests/navigation/click-scroll-jump.spec.ts
luccas-harbour: done. I added this assertion

luccas-harbour:
@caio-pizzol I addressed your comments

caio-pizzol (approved):
hey @luccas-harbour! thanks for addressing all four from last round. i loaded the customer doc in the dev app and both the margin click and the split-table gap click are clean now (cursor stays put, no scroll jump).

i pushed commit `1877c04ef` with three small tests and two one-line fixes the tests pin down: a click on a tracked-change span no longer gets swallowed, and a footnote session now exits when the click lands in an endnote with the same number.

lgtm.

PR #2708, opened 2026-04-02
fix: allow caret placement inside commented text on click (SD-2442)

luccas-harbour:
## Summary
- Fixes a bug where clicking on text covered by a single comment highlight would short-circuit caret placement, making it impossible to position the cursor within commented text
- Direct clicks on comment highlight spans now fall through to the generic pointer handling, allowing normal caret placement while still supporting nearby/gap clicks for comment activation
- Adds a Playwright behavior test (`sd-2442-commented-text-caret.spec.ts`) and updates existing unit tests to reflect the new behavior

### DIFF
--- a/tests/behavior/helpers/editor-interactions.ts
+++ b/tests/behavior/helpers/editor-interactions.ts
@@ -25,26 +25,10 @@ async function getDocPosCoords(
 
 export async function clickAtDocPos(page: Page, pos: number): Promise<void> {
   const coords = await getDocPosCoords(page, pos);
-  const x = Math.min(Math.max(coords.left + 1, coords.left), Math.max(coords.right - 1, coords.left + 1));
-  const y = (coords.top + coords.bottom) / 2;
-  await page.mouse.click(x, y);
+  await page.mouse.click(coords.left + 1, (coords.top + coords.bottom) / 2);
 }
 
-/**
- * Right-clicks at the screen location corresponding to a document position in the SuperDoc editor.
- *
- * This helper queries the editor's coordinates for the given logical document position, calculates a suitable
- * (x, y) point within the bounding rectangle, and dispatches a mouse right-click at that spot.
- *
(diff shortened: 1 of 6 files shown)


tests/behavior/helpers/editor-interactions.ts
```diff
@@ -29,6 +20,30 @@ export async function rightClickAtDocPos(page: Page, pos: number): Promise<void>
     throw new Error(`Could not resolve coordinates for document position ${pos}`);
   }
 
+  return coords;
```
caio-pizzol, line 23: this math always simplifies to `coords.left + 1` — the min/max wrapping doesn't do anything. same thing in `rightClickAtDocPos` below. worth simplifying?

```suggestion
  const x = coords.left + 1;
```

caio-pizzol (commented):
@luccas-harbour fix looks good, no bugs found. clicking on commented text places the caret correctly and the right comment gets selected — tested all the cases (single comment, overlapping comments, re-clicking same comment).

one side effect: clicking commented text highlights the comment bubble but doesn't open the reply thread like google docs does. filed SD-2460 for that — not a blocker here.

left one small inline comment on the test helpers.

tests/behavior/helpers/editor-interactions.ts
caio-pizzol: addressed in d5ae180f4 — simplified both functions and removed the duplication.

caio-pizzol (approved):
@luccas-harbour fix works well — clicking inside commented text places the cursor correctly now.

found one extra issue while testing: after clicking commented text, the comment bubble would flash active then immediately deselect. the click-outside handler on the comment dialog wasn't recognizing the click landed on a comment highlight (because of how the editor captures mouse events). fixed in 8bf062554.

also cleaned up the test helper math from round 1 (d5ae180f4) and added tests for everything — cursor placement, bubble staying active, and word selection via double-click.

PR #2710, opened 2026-04-02
feat(math): implement m:d delimiter converter (SD-2380)

andrewsrigom:
## Summary
- implement the `m:d` OMML -> MathML converter
- register the converter in the OMML converter registry
- add tests covering default delimiters, custom separators, and empty-expression handling

## Testing
- `pnpm exec vitest run packages/layout-engine/painters/dom/src/features/math/omml-to-mathml.test.ts`
- validated manually in the dev app with the issue test document

Closes #2611


packages/layout-engine/painters/dom/src/features/math/converters/delimiter.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const MATHML_NS = 'http://www.w3.org/1998/Math/MathML';
+
+const DEFAULT_BEGIN_DELIMITER = '(';
+const DEFAULT_END_DELIMITER = ')';
+const DEFAULT_SEPARATOR_DELIMITER = '|';
+
+function getDelimiterValue(properties: OmmlJsonNode | undefined, name: string, fallback: string): string {
+  const property = properties?.elements?.find((element) => element.name === name);
+  return property?.attributes?.['m:val'] ?? fallback;
```
caio-pizzol: the spec treats "element missing" and "element present without m:val" differently. missing → use default char, present without val → no character. right now both return the default because `property.attributes` is undefined in both cases.

```suggestion
  const property = properties?.elements?.find((element) => element.name === name);
  if (!property) return fallback;
  return property.attributes?.['m:val'] ?? '';
```

packages/layout-engine/painters/dom/src/features/math/converters/delimiter.ts
```diff
@@ -0,0 +1,72 @@
+import type { MathObjectConverter, OmmlJsonNode } from '../types.js';
+
+const MATHML_NS = 'http://www.w3.org/1998/Math/MathML';
+
+const DEFAULT_BEGIN_DELIMITER = '(';
+const DEFAULT_END_DELIMITER = ')';
+const DEFAULT_SEPARATOR_DELIMITER = '|';
```
caio-pizzol: spec says the default separator is U+2502, not ASCII `|` (§22.1.2.95). they look similar but are different characters.

```suggestion
const DEFAULT_SEPARATOR_DELIMITER = '\u2502'; // U+2502 BOX DRAWINGS LIGHT VERTICAL (ECMA-376 §22.1.2.95)
```

caio-pizzol (changes requested):
@andrewsrigom clean implementation, follows the existing converter pattern nicely.

two issues to fix — both confirmed by comparing SuperDoc rendering against Word (test file + screenshots below).

**1. element present without `m:val` should mean "no character"** — right now `<m:begChr/>` (no attribute) behaves the same as the element being absent, so it renders `(`. the spec says it should render nothing. same applies to endChr and sepChr.

**2. wrong default separator** — should be `\u2502` (box drawing vertical) not ASCII `|`, per §22.1.2.95.

tests look good. you'll want to add one for the no-val case after fixing #1. left inline comments with fixes.

> 💡 tip: you can look up any ECMA-376 section directly using the [ooxml.dev MCP server](https://ooxml.dev/mcp) — handy for checking defaults and edge cases in the spec.

---

**test file**: attached `.docx` with 21 cases covering every `m:dPr` property. cases 5, 8, 9, 12 show the bugs:
[SD-2380-delimiter-spec-tests.docx](https://github.com/user-attachments/files/26471460/SD-2380-delimiter-spec-tests.docx)

**Case 5** — `<m:begChr/>` should hide the opening delimiter:
SuperDoc:
image
Word:
image

**Case 8** — `<m:endChr/>` should hide the closing delimiter:
SuperDoc:
image
Word:
image

**Case 9** — both present without val, should show no delimiters:
SuperDoc:
image
Word:
image


**Case 12** — `<m:sepChr/>` should hide the separator:
SuperDoc:
image
Word:
image

caio-pizzol:
@andrewsrigom let me know if you have any questions here! thanks for you work ❤️

andrewsrigom:
Hi @caio-pizzol , this should be ready for another review now. Sorry for the delay, I got caught up with some work on my side. I rechecked the delimiter cases and they now match the expected output

I've noticed a timeout a issue in the pipeline, can you check

caio-pizzol (approved):
@andrewsrigom all three items from last round are fixed, looks good.

i pushed two commits — behavior tests with a test document covering all 21 spec cases, and a small refactor inlining `createExpressionGroup` into a single loop. also added the test doc to our rendering corpus for automatic regression checks.

approving.

PR #2729, opened 2026-04-07
Update presentation when typing inside table in sdt

VladaHarbour:


packages/layout-engine/layout-bridge/src/cache.ts
```diff
@@ -141,6 +141,14 @@ const hashRuns = (block: FlowBlock): string => {
         const cellBlocks = cell.blocks ?? (cell.paragraph ? [cell.paragraph] : []);
 
         for (const cellBlock of cellBlocks) {
+          if (cellBlock?.kind === 'table') {
+            // Nested tables inside table cells must contribute to the parent
+            // table cache key, otherwise edits can be missed until a later
+            // broader invalidation.
+            cellHashes.push(`nt:${hashRuns(cellBlock as FlowBlock)}`);
```
caio-pizzol: this works but there's no test for it. `cache.test.ts` already has lots of table tests — adding one for a table-inside-a-table (~30 lines, same pattern) would make sure this path doesn't break silently. worth adding?

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -7335,6 +7335,10 @@ const deriveBlockVersion = (block: FlowBlock): string => {
               hash = hashString(hash, getRunStringProp(run, 'vertAlign'));
               hash = hashNumber(hash, getRunNumberProp(run, 'baselineShift'));
             }
+          } else if (cellBlock?.kind) {
```
caio-pizzol, line 7638: this catches all non-paragraph blocks in cells, while `cache.ts` only catches tables specifically. both approaches work, just different — worth a comment pointing to each other so whoever changes one remembers the other exists.

caio-pizzol (commented):
@VladaHarbour good fix — the three changes work well together: cache updates so edits show up, version tracking so the page re-renders, and click handling so the cursor lands inside the inner table instead of selecting the whole field.

no correctness issues found.

one thing we noticed while testing: opening a Word file that already has this structure (table inside an SDT inside a table cell) doesn't produce the right nodes on import — the inner tables get dropped. that's not something this PR broke, it was already like that, but it means these code paths only get triggered through the template builder (where users build the structure interactively). could be worth a follow-up ticket for import support.

left two inline comments — both about test coverage, not the logic itself.

packages/layout-engine/layout-bridge/src/cache.ts
VladaHarbour: added test

packages/layout-engine/painters/dom/src/renderer.ts
VladaHarbour: comment added

harbournick (approved):
LGTM

PR #2732, opened 2026-04-07
SD-2279 - fix: floating textbox not rendering

chittolinag:
### Issue
Word sometimes emits DrawingML tags with arbitrary prefixes (e.g. `ns6:`). Our importer only looked for literal `a:graphic/a:graphicData`, so it dropped the whole drawing, emitted a `passthroughBlock` inside the run, and `calculateInlineRunPropertiesPlugin` crashed with `RangeError: Invalid content for node type run`.

### Proposed solution
Normalize DrawingML lookups by local name in `handleImageNode` (and helpers) so `graphic/graphicData`, blips, transforms, etc. are found regardless of prefix; do the same in `vector-shape-helpers`.


packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
```diff
@@ -415,10 +415,10 @@ export function handleImageNode(node, params, isAnchor) {
 
   // --- Parse decorative flag from wp:docPr > a:extLst > a:ext > adec:decorative ---
   let decorative = false;
-  const docPrExtLst = docPr?.elements?.find((el) => el.name === 'a:extLst');
+  const docPrExtLst = findChildByLocalName(docPr?.elements, 'extLst');
   if (docPrExtLst) {
     for (const ext of docPrExtLst.elements || []) {
-      if (ext.name !== 'a:ext') continue;
+      if (!hasLocalName(ext, 'ext')) continue;
       const decEl = ext.elements?.find((el) => el.name === 'adec:decorative' || el.name === 'a16:decorative');
```
caio-pizzol: you migrated the outer `extLst` / `ext` lookups but left this one with hardcoded prefixes. intentional? a one-line comment would save the next reader from wondering.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/drawingml-utils.js
```diff
@@ -0,0 +1,58 @@
+/**
```
caio-pizzol, line 1: matching by namespace URI + local name would be stricter than local name alone. fine as-is for this code path - just noting it in case you want to tighten later.

caio-pizzol (approved):
hey @chittolinag! good call on this :)

one out-of-diff thing worth a follow-up: `merge-drawing-children.js:105-106` still hardcodes `a:graphic`. on export, a doc with the `ns6:` prefix won't get the same treatment. not a blocker here.

lgtm - feel free to merge after addressing the inline notes.

caio-pizzol:
hey @chittolinag! quick follow-up.

i pushed the two fixes, added tests so we don't lose this, and uploaded a sample doc so layout and visual tests cover it.

lgtm.

PR #2733, opened 2026-04-07
fix(super-editor): structural text.rewrite for multi-block replacements

tupizz:
## Summary

Fixes SD-2362  
https://linear.app/superdocworkspace/issue/SD-2362/bug-comments-inserted-by-ai-agent-do-not-display-in-document-ui

This PR now preserves multi-paragraph AI rewrites in both direct mode and suggestion mode.

It covers three related cases:
- replacing an entire paragraph with multi-paragraph content
- replacing text in the middle of a paragraph with multi-paragraph content
- agent fallback flows that do `delete` + `text.insert` with leading paragraph breaks

It also keeps the branch’s prompt/tooling improvements and adds better local verification support for the `examples/collaboration/ai-node-sdk` flow.

## Problem

There were two separate structural failures in the AI rewrite pipeline:

1. Full-paragraph rewrites were flattened because `text.rewrite` originally collapsed `\n\n` content into a single text node, and tracked condensation could then re-flatten structural replacements.
2. Middle-of-paragraph rewrites still failed even after the first fix because the selection sits inside inline `run` wrappers in real SuperDoc documents. The structural slice was being built with paragraph-level assumptions, so ProseMirror rejected it and the executor silently fell back to inline text replacement.

A second variant of the same bug showed up when the agent decomposed the rewrite into `delete` plus `text.insert` with a leading newline. That path produced an invalid structural slice shape (`inconsistent open depths`) and also fell back to inline insertion.

### 1. Structural rewrites for full-paragraph and in-paragraph replacements

`executeTextRewrite()` in `packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts` now:
- detects full-inline-range coverage inside a textblock instead of relying on raw paragraph boundaries
- supports structural splicing when the target is only part of a paragraph
- resolves the inline wrapper chain between the paragraph and the text position (for example `run` nodes)
- builds replacement slices with the correct open depth for the real schema, not just simplified paragraph/text schemas
- preserves paragraph attributes while stripping identity attrs such as `paraId` and `sdBlockId`
- falls back to inline replacement only when the surrounding content model truly cannot host the structural result

### 2. Structural inserts for agent fallback flows

`executeTextInsert()` now supports multi-block structural insertion inside paragraphs.

For leading/trailing newline cases, the executor normalizes the payload into edge empty blocks so ProseMirror can split the surrounding paragraph cleanly without invalid open-depth combinations. This is what fixes the agent trace where the model deletes `"hello world"` and then inserts `"\nLorem..."` after the preceding text.

### 3. Track changes preservation

The tracked condensation work in `replaceStep.js` remains part of the fix. Once the executor produces a valid structural replace step, the track-changes helper must still preserve that structure instead of replaying it as a flat inline insertion.

### 4. Debugging and local verification support

To make the example easier to verify and debug:
- `examples/collaboration/ai-node-sdk/Makefile` now rebuilds local artifacts on `make dev-local`
- `packages/sdk/langs/node/src/runtime/host.ts` now forwards host stderr when `SUPERDOC_DEBUG_TEXT_REWRITE=1`
- the executor emits explicit structural rewrite/insert debug logs, including the fallback reason when a slice is rejected

### 5. Prompt and browser tool updates

This branch also includes the prompt/dispatch updates that guide the browser agent toward better block targeting and better formatting decisions (for example, distinguishing titles/headings from body paragraphs).

## Files changed

- `packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts`
- `packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.test.ts`
- `packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/tracked-rewrite.integration.test.ts`
- `packages/super-editor/src/editors/v1/extensions/track-changes/trackChangesHelpers/replaceStep.test.js`
- `packages/sdk/langs/node/src/runtime/host.ts`
- `examples/collaboration/ai-node-sdk/Makefile`
- `packages/sdk/langs/browser/src/intent-dispatch.ts`
- `packages/sdk/langs/browser/src/system-prompt.ts`

## Test plan

- [x] `pnpm vitest packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.test.ts --run`
- [x] `pnpm vitest packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/tracked-rewrite.integration.test.ts --run`
- [x] `pnpm vitest packages/super-editor/src/editors/v1/document-api-adapters/plan-engine --run`
- [x] `pnpm vitest packages/super-editor/src/editors/v1/extensions/track-changes/trackChangesHelpers --run`
- [x] `pnpm --filter @superdoc/super-editor run build`

Focused regression coverage now includes:
- full-paragraph multi-block rewrite in direct mode
- full-paragraph multi-block rewrite in tracked mode
- middle-of-paragraph multi-block rewrite in direct mode
- middle-of-paragraph multi-block rewrite in tracked mode
- tracked `text.insert` with a leading newline inside a paragraph


packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts
```diff
@@ -991,6 +1005,131 @@ function resolveReplacementBlocks(replacement: ReplacementPayload, stepId: strin
...
+  return null;
+}
+
+function findTextRangeWithin(doc: ProseMirrorNode, from: number, to: number): { from: number; to: number } | null {
+  let textFrom: number | null = null;
+  let textTo: number | null = null;
+
+  doc.nodesBetween(from, to, (node, pos) => {
+    if (!node.isText || !node.text) {
+      return;
```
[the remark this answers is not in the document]
andrii-harbour: @tupizz is this a relevant comment?

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts
```diff
@@ -991,6 +1005,131 @@ function resolveReplacementBlocks(replacement: ReplacementPayload, stepId: strin
...
+    if (textFrom === null) {
+      textFrom = nodeFrom;
+    }
+    textTo = nodeTo;
+  });
+
+  return textFrom === null || textTo === null ? null : { from: textFrom, to: textTo };
+}
+
+function sanitizeReplacementParagraphAttrs(
```
andrii-harbour: This function strips paraId, sdBlockId, nodeId, id, blockId, uuid. The existing                                       
  `resolveInheritedParagraphAttrsForReplacement` does the same thing. These should share a single helper to avoid drift if new identity attrs are added later.

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.test.ts
andrii-harbour, line 1: The tests only cover replacement.text with \n\n. There's no test exercising replacement.blocks (the array variant), which also flows through `replacementMayContainStructuralBlocks`

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts
```diff
@@ -746,6 +746,20 @@ export function executeTextRewrite(
...
+
+  if (structuralRewrite) {
+    const nodes = buildReplacementParagraphNodes(
+      editor,
+      structuralRewrite.replacementBlocks,
+      marks,
+      structuralRewrite.paragraphAttrs,
+      step.id,
+    );
+    const slice = new Slice(Fragment.from(nodes), 1, 1);
```
andrii-harbour: Slice open depth (1, 1) assumes paragraph is at depth 1 - This is true for a flat doc > paragraph structure but could be wrong if the textblock is nested deeper (e.g., inside a list item or table cell). The `textblockDepth` is already computed — should the Slice open/close depth be derived from it?

andrii-harbour (changes requested):
just minor questions/suggestions

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts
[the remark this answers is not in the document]
tupizz: Fixed in 763596d. Replaced `findTextRangeWithin` (which only scanned text nodes, missing inline atoms) with ProseMirror's position-based content boundaries: `$from.start(depth)` / `$from.end(depth)`. These account for all inline content including atoms (mentions, footnotes, images, etc.). The `findTextRangeWithin` function has been deleted entirely.

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts
tupizz: Fixed in 763596d. Extracted a shared `stripIdentityAttrs()` helper that both `resolveStructuralRangeRewrite` and `resolveInheritedParagraphAttrsForReplacement` now use. The duplicate delete sequences are gone.

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.test.ts
tupizz: Added in 763596d. New test exercises `replacement.blocks` array variant with a whole-textblock target. The existing tests cover `replacement.text` with \`\\n\\n\` and partial-range preservation.

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts
tupizz: Added explanatory comment in 763596d. The `(1, 1)` open depth is correct at any nesting depth because it's relative to the fragment's own structure, not the document tree. ProseMirror's `tr.replace` fits the slice into the target context regardless of how deeply nested the textblock is. `findSharedTextblockDepth` finds the actual textblock (paragraph), not its container (listItem, tableCell), so the replacement always produces sibling paragraphs at the right level.

harbournick (approved):
LGTM

PR #2735, opened 2026-04-07
SD-2443 - fix: comments being merged by range when they shouldn't

chittolinag:
### Issue

- Imports without `word/commentsExtended.xml` get a `range-based` `commentThreadingProfile`. We passed that straight back into `updateCommentsExtendedXml()`, which returns `null` for range-based profiles, so our export dropped `commentsExtended.xml` entirely.
- Re-importing that DOCX makes `importCommentData()` fall back to `detectThreadingFromNestedRanges()`. In tables the ranges are naturally nested, so we stitched independent comments into one thread.

### Proposed solution

1. Force `commentsExtended.xml` generation for any non–Google Docs export (pass `'word'` to `updateCommentsExtendedXml()` when `exportStrategy !== 'google-docs'`), so we always stamp explicit `w15:commentEx` entries.
2. Add a regression test that exports a `range-based` profile, asserts the file exists, and verifies re-import keeps overlapping table comments separate.
3. (Optional) Cover `detectThreadingFromRanges()` with a unit test to ensure it skips when `commentsExtended.xml` is present.


packages/super-editor/src/editors/v1/core/super-converter/v2/exporter/commentsExporter.js
```diff
@@ -48,6 +47,7 @@ export const getCommentDefinition = (comment, commentId, allComments, editor) =>
     'custom:trackedChangeType': comment.trackedChangeType,
     'custom:trackedChangeDisplayType': comment.trackedChangeDisplayType || null,
     'custom:trackedDeletedText': comment.deletedText || null,
+    'custom:authorEmail': comment.creatorEmail || comment.importedAuthor?.email,
```
palmer-cl: nice good idea here.

packages/super-editor/src/editors/v1/core/super-converter/v2/exporter/commentsExporter.js
```diff
@@ -400,10 +400,14 @@ export const prepareCommentsXmlFilesForExport = ({
   relationships.push(generateRelationship('comments.xml'));
   emittedTargets.add('comments.xml');
 
+  const forceWordThreadingProfile =
+    threadingProfile?.defaultStyle === 'range-based' && exportStrategy !== 'google-docs';
```
caio-pizzol: this guard never fires for the files this ticket is about. anything with comments but no `commentsExtended.xml` gets tagged `origin='google-docs'` on import (`docxImporter.js:87-90`), so the guard skips. the test only passes because the helper defaults comments to `origin='word'`; cleanest fix is to stop calling a missing `commentsExtended.xml` a google-docs signal (word 2010 ships without it).

to repro: feed `tests/data/gdocs-single-comment.docx` through import -> export and check `documentXml['word/commentsExtended.xml']`. it won't be there.

packages/super-editor/src/editors/v1/core/super-converter/v2/exporter/commentsExporter.js
caio-pizzol: main already reads this attribute as `custom:email` (commit `52298526b`), so after rebase the exporter and importer will disagree and emails will drop silently. rename here, at `:143`, and at `documentCommentsImporter.js:38`:

```suggestion
    'custom:email': comment.creatorEmail || comment.importedAuthor?.email,
```

after rebase, the test you added at `documentCommentsImporter.unit.test.js:258` will fail until renamed too.

packages/super-editor/src/editors/v1/core/super-converter/v2/exporter/commentsExporter.test.js
```diff
@@ -381,12 +381,65 @@ describe('prepareCommentsXmlFilesForExport', () => {
...
+
+      const result = prepareCommentsXmlFilesForExport({
+        convertedXml: makeConvertedXml(),
+        defs,
+        commentsWithParaIds,
+        exportType: 'external',
+        threadingProfile,
+      });
+
+      expect(result.documentXml['word/commentsExtended.xml']).toBeDefined();
```
caio-pizzol: this passes even if the exporter writes an empty file, which would bring the bug right back. the importer falls back to guessing threads from overlapping ranges when there are no entries. check the contents: one entry per comment, each with a `w15:paraId`, and none with `w15:paraIdParent` (missing parent ids are the whole point of this fix).

`tests/data/nested-comments-gdocs.docx` is a good fixture for a proper export -> re-import regression test.

caio-pizzol (changes requested):
needs work. three inline. modern Word files already ship `commentsExtended.xml` (no fix needed); legacy Word and Google Docs files get tagged `origin='google-docs'` on import, which the guard then skips. the only path that actually fires is `origin='word'` on a file missing `commentsExtended.xml`, which never happens naturally, only in the test helper. worth a follow-up.

caio-pizzol:
filed the architecture refactor as [SD-2650](https://linear.app/superdocworkspace/issue/SD-2650/separate-comment-origin-metadata-from-comment-export-routing). scope: stop using `origin` for export routing, key off explicit capabilities + target instead. the narrow SD-2443 fix here still needs to land first, then SD-2650 makes the guard unnecessary.

packages/super-editor/src/editors/v1/core/super-converter/v2/exporter/commentsExporter.js
chittolinag: hey @caio-pizzol - I just tested this same file `gdocs-single-comment.docx` and the logic seems to be working as expected. This file contains a comment inside the `Open comment` text. I manually added a smaller comment inside the existing comment (`comm`). After exporting and opening it in Word and Google Docs, the comments seem to be correct. They're still separate and `commentsExtended.xml` is also present on the zip.

See here:

Image


Image

caio-pizzol:
@chittolinag pushed a follow-up commit with the fix + updated tests. happy to back it out if you'd rather own it.

the check that was added (`exportStrategy !== 'google-docs'`) never runs for the files this PR targets. when a file has `comments.xml` but no `commentsExtended.xml`, the importer labels every comment as google-docs. that's the exact shape SD-2443 is about, and the check skips it.

`Testing 123.docx` round-trips fine today only because a different check inside `updateCommentsExtendedXml` catches threaded replies. remove the reply link and `commentsExtended.xml` gets dropped again, and re-import merges the nested comment back into the thread.

i reproduced this end-to-end in the dev app with Testing 123.docx. same fixture, same export path, only difference is the guard:

```
# before the fix (guard never fires for origin=google-docs)
$ unzip -l sd2443-before-fix.docx | grep word/comments
8570  word/comments.xml
```

```
# after the fix (commentsExtended.xml synthesized)
$ unzip -l sd2443-fix-verified.docx | grep word/comments
8570  word/comments.xml
2825  word/commentsExtended.xml
```

[Testing 123.docx](https://github.com/user-attachments/files/27069753/Testing.123.docx)

| case | before | after |
|---|---|---|
| imported as-is (one reply link) | emits | emits |
| no reply links, not resolved | **drops** | emits |
| same with origin forced to word | emits | emits |

the fix keys the check off the file-set the importer recorded, not the origin label:

```js
// before
const forceWordThreadingProfile =
  threadingProfile?.defaultStyle === 'range-based' && exportStrategy !== 'google-docs';
// after
const forceWordThreadingProfile =
  threadingProfile?.defaultStyle === 'range-based' &&
  threadingProfile?.fileSet?.hasCommentsExtended === false;
```

tests:
- flipped the two `commentThreadingProfile.test.js` "Google Docs profile without threading" tests. they were asserting `commentsExtended.xml` should be dropped for `comments.xml`-only imports, which is the regression.
- replaced the `commentsExporter.test.js` "still honors Google Docs export strategy" test with a positive case: origin google-docs, no replies, not resolved -> emits `commentsExtended.xml`.
- added a `commentsExtended` profile test so the override doesn't over-fire.

round-trip verified in the UI: exported `Testing 123.docx`, dropped the export back in, thread and standalone both preserved:

[sd2443-fix-verified.docx](https://github.com/user-attachments/files/27069774/sd2443-fix-verified.docx)

SD-2650 is still the right place to properly decouple origin from export routing later.

caio-pizzol (approved):
lgtm

PR #2736, opened 2026-04-07
feat: support paragraph bar borders

### LINKED ISSUE #2282: Support paragraph bar borders (w:pBdr/w:bar)
## What's happening

The `w:bar` paragraph border — a vertical decorative bar rendered on the left side of a paragraph — is parsed by the converter but never rendered. It's silently dropped in the pm-adapter normalization step, same as the `between` border tracked in .

## Why it's happening

`normalizeParagraphBorders()` in `borders.ts` only processes `['top', 'right', 'bottom', 'left']`. The `bar` value is parsed and stored in `paragraphProperties.borders.bar` but never forwarded to the layout format or DomPainter.

## Suggested fix

Similar pattern to the `between` border fix:

1. **Contracts** (`packages/layout-engine/contracts/src/index.ts`): Add `bar` to the `ParagraphBorder` type on `ParagraphAttrs`.
2. **pm-adapter** (`packages/layout-engine/pm-adapter/src/attributes/borders.ts`): Include `bar` in `normalizeParagraphBorders()` so it flows to the layout format.
3. **DomPainter** (`packages/layout-engine/painters/dom/src/renderer.ts`): Render as a vertical line on the left edge of the paragraph — simpler than `between` since it doesn't require knowledge of adjacent paragraphs.

This is a good candidate for a **good first issue** since the rendering is straightforward (just a vertical CSS border/line on the left side).

See OOXML spec §17.3.1.4 for `w:bar` behavior.

---

*Ref:* [SD-2109](https://linear.app/superdocworkspace/issue/SD-2109/support-paragraph-bar-borders-wpbdrwbar)

iguit0:
## Summary

Adds support for Word paragraph bar borders (`w:pBdr/w:bar`) in the layout/rendering pipeline.

`w:bar` is the vertical decorative line rendered on the left edge of a paragraph. It was already parsed upstream, but it never made it through the layout-engine path, so it could not render in DomPainter.

## Root cause

`w:bar` was being dropped in `normalizeParagraphBorders()` inside the pm-adapter layer.

Because of that:
- `ParagraphBorders` did not expose `bar`
- normalized paragraph attrs never carried `bar`
- paragraph border hashes/equality checks ignored `bar`
- DomPainter never had any `bar` data to render

### Contracts
- added `bar?: ParagraphBorder` to `packages/layout-engine/contracts/src/index.ts`

### PM adapter
- updated `normalizeParagraphBorders()` to preserve `bar`
- kept the existing `between` special-case behavior scoped to `between` only
- `bar` now normalizes like the other paragraph border sides:
  - style mapping
  - width conversion
  - color normalization
  - `space` support

### Hashing / invalidation / grouping
Updated the paragraph-border identity paths so bar-only changes are treated as real visual changes:
- `packages/layout-engine/layout-bridge/src/paragraph-hash-utils.ts`
- `packages/layout-engine/painters/dom/src/paragraph-hash-utils.ts`
- `packages/layout-engine/layout-bridge/src/diff.ts`
- `packages/layout-engine/layout-engine/src/layout-paragraph.ts`

This ensures:
- cache invalidation happens when only `bar` changes
- paragraphs that differ only by `bar` do not incorrectly group for `between` rendering

### DomPainter rendering
- implemented `bar` rendering in `packages/layout-engine/painters/dom/src/features/paragraph-borders/border-layer.ts`
- renders `bar` as a separate absolutely-positioned vertical rule on the left side of the paragraph border box
- keeps `bar` independent from the normal `left` border so both can coexist
- does not change existing `between` behavior

Implementation detail:
- `bar` is rendered as a dedicated child decoration element attached to the existing paragraph border layer
- its horizontal offset is based on the rendered left border width plus normalized `bar.space`
- it uses the normalized `style`, `width`, and `color` values
- pointer events remain disabled

## Test plan

- Import a `.docx` with a paragraph using only `w:bar`; verify a left-side vertical bar renders.
- Import a `.docx` with both `w:left` and `w:bar`; verify both render independently.
- Verify `w:bar` + paragraph shading renders correctly and the bar stays visible.
- Verify consecutive paragraphs with identical borders still group correctly with `between`.
- Verify consecutive paragraphs that differ only by `bar` no longer group.
- Inspect rendered DOM and confirm `bar` is a separate decoration element with pointer events disabled.

Closes #2282


caio-pizzol (commented):
@iguit0 nice work — the bar rendering is the best of the three apps we tested. Word ignores bar borders entirely, Google Docs renders them but loses the color, SuperDoc gets it right.

one thing to look at: both Word and Google Docs ignore bar when deciding whether to group paragraphs together. right now SuperDoc breaks the group when bars differ, but the other two don't. left an inline comment with more detail.

couple of small cleanup suggestions in the other comments. tests look good.

packages/layout-engine/painters/dom/src/paragraph-hash-utils.ts
```diff
@@ -29,6 +29,7 @@ export const hashParagraphBorders = (borders: ParagraphBorders): string => {
   if (borders.right) parts.push(`r:[${hashParagraphBorder(borders.right)}]`);
   if (borders.bottom) parts.push(`b:[${hashParagraphBorder(borders.bottom)}]`);
   if (borders.left) parts.push(`l:[${hashParagraphBorder(borders.left)}]`);
+  if (borders.bar) parts.push(`bar:[${hashParagraphBorder(borders.bar)}]`);
```
caio-pizzol: we tested this file in Word and Google Docs — both still group paragraphs together even when their bar borders are different. including `bar` here breaks that grouping in SuperDoc. the bar should still trigger a re-render when it changes, but it shouldn't affect whether paragraphs are grouped for between borders. worth splitting those two?

packages/layout-engine/painters/dom/src/between-borders.test.ts
```diff
@@ -689,6 +722,25 @@ describe('computeBetweenBorderFlags', () => {
     expect(computeBetweenBorderFlags(fragments, lookup).size).toBe(0);
   });
```
caio-pizzol, line 722: heads up — this test expects bar-only differences to break grouping, but Word and Google Docs both keep them grouped. if the grouping change above goes in, this test would need to flip.

packages/layout-engine/painters/dom/src/features/paragraph-borders/border-layer.ts
```diff
@@ -182,6 +183,93 @@ const computeRenderedBorderWidths = (
...
+  barElement.style.position = 'absolute';
+  barElement.style.pointerEvents = 'none';
+  barElement.style.boxSizing = 'border-box';
+  barElement.style.top = '0px';
+  barElement.style.bottom = '0px';
+  barElement.style.left = `-${renderedLeftBorderWidth + barSpace}px`;
+  barElement.style.width = '0px';
+  barElement.style.borderLeftStyle = resolvedBar.style;
+  barElement.style.borderLeftWidth = `${resolvedBar.width}px`;
+  barElement.style.borderLeftColor = resolvedBar.color;
```
caio-pizzol: these three lines do the same thing as `setBorderSideStyle(barElement, 'left', barBorder)` — worth reusing it?

```suggestion
  setBorderSideStyle(barElement, 'left', barBorder);
```

packages/layout-engine/painters/dom/src/features/paragraph-borders/border-layer.ts
```diff
@@ -182,6 +183,93 @@ const computeRenderedBorderWidths = (
...
+  if (!border) return 0;
+  const resolved = resolveParagraphBorder(border);
+  if (resolved.style === 'none') return 0;
+  return resolved.width;
+};
+
+const getParagraphBarElement = (element: HTMLElement): HTMLElement | undefined => {
+  return Array.from(element.children).find(
+    (child): child is HTMLElement => child instanceof HTMLElement && child.classList.contains(PARAGRAPH_BAR_CLASS),
+  );
```
caio-pizzol: `element.querySelector('.superdoc-paragraph-bar')` does the same thing in one line — worth switching?

iguit0: Yep, much cleaner

packages/layout-engine/painters/dom/src/features/paragraph-borders/border-layer.ts
iguit0: Good catch! The resolved style/width/color logic was duplicated for no reason.

packages/layout-engine/painters/dom/src/paragraph-hash-utils.ts
iguit0: Fixed! Removed `bar` from this hash (grouping) and the layout-engine's local `hashBorders`. The layout-bridge copy keeps `bar` for cache invalidation, so there's now an intentional divergence between the two copies. Added a comment explaining why.

packages/layout-engine/painters/dom/src/between-borders.test.ts
iguit0: Flipped. The test now asserts that bar-only differences keep the group intact and `showBetweenBorder` is `true`

iguit0:
> @iguit0 nice work — the bar rendering is the best of the three apps we tested. Word ignores bar borders entirely, Google Docs renders them but loses the color, SuperDoc gets it right.
> 
> one thing to look at: both Word and Google Docs ignore bar when deciding whether to group paragraphs together. right now SuperDoc breaks the group when bars differ, but the other two don't. left an inline comment with more detail.
> 
> couple of small cleanup suggestions in the other comments. tests look good.

Fixed the grouping behavior: `bar` is now excluded from the grouping hash in both the painter and layout-engine paths, so paragraphs that differ only by bar stay grouped. The layout-bridge hash (cache invalidation) still includes `bar` so bar changes trigger re-render. Updated tests to match.

caio-pizzol:
Hey @iguit0, thanks for the work — the code and tests are solid.

I tested the doc in Word and Word never draws these bars. It keeps them in the file on save but doesn't show them on screen. The spec says apps can skip them, and Word does.

SuperDoc today doesn't draw them either, so main already matches Word. This PR would be the one thing that starts drawing them.

Suggest closing for now. If a real doc needs bars drawn later, we can revisit with that context. If you have a doc in mind, share it and I'll take another look.

Thank you for all the work still! We can add it in the future if relevant

PR #2756, opened 2026-04-08
fix: comment bubble got cut off

VladaHarbour:


packages/superdoc/src/components/CommentsLayer/FloatingComments.vue
```diff
@@ -205,6 +205,13 @@ const totalHeight = computed(() => {
   return max + 50;
 });
 
+// The inner sidebar is translated by sidebarOffsetY. When shifted down, the
+// rendered bottom edge can exceed totalHeight and get clipped by parent scroll
+// containers. Expand wrapper height to include positive translate offset.
+const wrapperMinHeight = computed(() => {
```
caio-pizzol, line 211: this fix works well with a few bubbles, but with 7+ tracked changes the sidebar shifts upward instead of downward — and `Math.max(0, negative)` becomes 0, so no extra height is added. the last bubble still gets cut off. i reproduced this locally on the branch.

google docs handles this by letting the sidebar extend past the page bottom ([test doc](https://docs.google.com/document/d/14fspOexT5XMsKJh4AgdHGxZ4LizfB_kcywakLEZmWdY/edit?usp=sharing)). could we remove `overflow: hidden` from `.floating-comments` in SuperDoc.vue to do the same?

packages/superdoc/src/components/CommentsLayer/FloatingComments.vue
```diff
@@ -287,6 +294,19 @@ const handleResize = (comment) => {
...
+
+    const isActiveThread = key === activeCommentKey.value;
+    const isPending = key === 'pending';
+    const isEditingThread = !!editingCommentId.value && editingCommentId.value === comment?.commentId;
+    if (!isActiveThread && !isPending && !isEditingThread) return;
+
+    // Reflow nearby cards after size changes of the active/pending/editing thread.
+    // Avoid force-snapping to anchor here because it can over-shift the whole lane
+    // near viewport boundaries and make bottom clipping more frequent.
+    remeasureCommentKeys(allPositions.value.map((pos) => pos.id));
```
caio-pizzol, line 306: this clear-timers + set-50ms/350ms pattern is now in three places. fine for this PR, but worth extracting into a small helper as a follow-up.

caio-pizzol (changes requested):
@VladaHarbour fix works well with a few bubbles, but with 7+ tracked changes on a short doc, the last bubble still gets clipped — reproduced it locally. the sidebar shifts up in that case and the height fix doesn't apply. root issue is `overflow: hidden` on `.floating-comments` in SuperDoc.vue. left inline comments.

caio-pizzol (approved):
@VladaHarbour both issues from last round are fixed. verified locally that bubbles show fully even with many tracked changes. nice work.

PR #2757, opened 2026-04-08
feat: toolbar to headless api (state and execute)

artem-harbour:
Linear: SD-2348.

Migrated toolbar to headless API (state and execute path).

Notes:
- The link item is not migrated intentionally, as this is a special case and all the logic lives in the component (`LinkInput.vue`).
- Some legacy logic has been moved to adapters (`getFontFamilyFallbackValue`, `isFontSizeMixedState`) rather than headless api for now.

### DIFF
--- a/packages/super-editor/src/editors/v1/components/toolbar/constants.js
+++ b/packages/super-editor/src/editors/v1/components/toolbar/constants.js
@@ -53,3 +53,59 @@ export const TOOLBAR_FONT_SIZES = [
   { label: '72', key: '72pt', props: { 'data-item': 'btn-fontSize-option' } },
   { label: '96', key: '96pt', props: { 'data-item': 'btn-fontSize-option' } },
 ];
+
+export const HEADLESS_ITEM_MAP = {
+  undo: 'undo',
+  redo: 'redo',
+  bold: 'bold',
+  italic: 'italic',
+  underline: 'underline',
+  strike: 'strikethrough',
+  acceptTrackedChangeBySelection: 'track-changes-accept-selection',
+  rejectTrackedChangeOnSelection: 'track-changes-reject-selection',
+  ruler: 'ruler',
+  zoom: 'zoom',
+  documentMode: 'document-mode',
+  link: 'link',
+  fontFamily: 'font-family',
+  fontSize: 'font-size',
+  list: 'bullet-list',
+  numberedlist: 'numbered-list',
+  table: 'table-insert',
+  image: 'image',
+  color: 'text-color',
+  highlight: 'highlight-color',
+  textAlign: 'text-align',
+  lineHeight: 'line-height',
+  linkedStyles: 'linked-style',
+  indentleft: 'indent-decrease',
+  indentright: 'indent-increase',
+  clearFormatting: 'clear-formatting',
+  copyFormat: 'copy-format',
+};
+
+export const TABLE_ACTION_COMMAND_MAP = {
+  addRowBefore: 'table-add-row-before',
+  addRowAfter: 'table-add-row-after',
+  deleteRow: 'table-delete-row',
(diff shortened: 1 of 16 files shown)


artem-harbour:
@harbournick @caio-pizzol - please review this.

packages/super-editor/src/editors/v1/components/toolbar/super-toolbar.js
```diff
@@ -1158,15 +787,23 @@ export class SuperToolbar extends EventEmitter {
...
-    if (command in this.#interceptedCommands) {
-      const result = this.#interceptedCommands[command]({ item, argument });
-      if (isMarkToggle) this.#syncStickyMarksFromState();
-      return result;
+    if (HEADLESS_EXECUTE_ITEMS.has(item?.name?.value)) {
+      const handledByHeadless = this.#executeHeadlessCommand(item, argument);
+      if (handledByHeadless) {
+        this.updateToolbarState();
+        if (shouldRestoreFocus) this.#scheduleRestoreEditorFocus();
+        return;
```
caio-pizzol, line 807: when you toggle bold/italic/underline on an empty selection via toolbar, the new headless path returns before saving the sticky mark state. so if the user clicks somewhere else after toggling, the formatting is lost. adding the sync call before the return should fix it:

```suggestion
      if (isMarkToggle) this.#syncStickyMarksFromState();
      return;
```

packages/super-editor/src/editors/v1/components/toolbar/super-toolbar.js
```diff
@@ -849,20 +520,157 @@ export class SuperToolbar extends EventEmitter {
     }
```
caio-pizzol, line 520: the old image upload code fired an `'exception'` event when something went wrong, so consumers could handle it via `onException`. the new headless version just logs to console — no event reaches consumers anymore. worth wiring the error back up?

packages/super-editor/src/editors/v1/components/toolbar/constants.js
```diff
@@ -53,3 +53,131 @@ export const TOOLBAR_FONT_SIZES = [
   { label: '72', key: '72pt', props: { 'data-item': 'btn-fontSize-option' } },
   { label: '96', key: '96pt', props: { 'data-item': 'btn-fontSize-option' } },
 ];
+
+export const HEADLESS_TOOLBAR_COMMANDS = [
```
caio-pizzol: `TABLE_ACTION_COMMAND_IDS`, `HEADLESS_TOOLBAR_COMMANDS`, and `HEADLESS_EXECUTE_ITEMS` are all just copies of the data in the two maps above. if someone adds a command to one map but forgets to update these lists, things break silently. can we derive them from the maps instead?

caio-pizzol (commented):
@artem-harbour clean migration, the routing is much easier to follow now. left a few inline comments.

two things to fix: toggling formatting on empty selections loses the toggle when you click away (sticky marks), and image upload errors no longer reach `onException`. also flagged three lists in constants.js that could get out of sync — easy to derive them from the maps.

test-wise: no coverage for destroy/cleanup and no test for hiding tracked-change buttons based on role. not blocking, good follow-ups.

artem-harbour:

caio-pizzol (approved):
@artem-harbour everything from last round is fixed, nice work.

i pushed three small commits:

- `512a46e`: clear pending focus timer on rapid clicks, drop redundant `updateToolbarState()` calls in mode setters
- `f9e22b8`: tests for the headless adapter branches (textAlign, lineHeight, zoom, link, color, highlight, fontSize mixed, tableActions), plus zoom string handling and text-color `"none"`
- `b5e1c41`: fix the CI type error — Editor's `exception` event payload doesn't accept `originalError`, so i merged the original message into the wrapped Error

lgtm now.

PR #2759, opened 2026-04-09
fix(layout): lay out all header/footer variants in per-section path (SD-2500)

caio-pizzol:
`layoutWithPerSectionConstraints` only measured the default header variant per section, ignoring first/even/odd. When `titlePg` was enabled and the decoration provider resolved the first-page rId, it wasn't in `layoutsByRId` — so the cover page header with branding images never rendered.

- Replace `resolveDefaultRIdPerSection` with `buildEffectiveRefsBySection` to collect all variant rIds per section (with OOXML inheritance)
- Iterate and lay out each variant's rId with section-appropriate constraints
- Remove the now-unused `resolveDefaultRIdPerSection`

Fixes SD-2500. Affects Patriot, M3, and Heffernan — any doc where section 1 has `titlePg=true` with a branded first-page header.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/header-footer/HeaderFooterPerRidLayout.ts
+++ b/packages/super-editor/src/editors/v1/core/header-footer/HeaderFooterPerRidLayout.ts
@@ -190,36 +190,6 @@ function collectReferencedRIdsBySection(effectiveRefsBySection: Map<number, Head
 }
 
 /**
- * Resolve the default header/footer rId for each section.
- *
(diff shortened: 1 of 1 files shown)


caio-pizzol:
Closing this. The fix already landed via #2922.

That PR replaced the old single-variant logic with `buildEffectiveHeaderFooterRefsBySection`, which lays out every header/footer variant (default/first/even/odd) per section. Same mechanism this PR proposes, so the titlePg-on-section-1 case is already covered on main.

PR #2761, opened 2026-04-09
fix: improve toolbar responsiveness

artem-harbour:
Linear: SD-2328

Improved responsiveness behavior relative to the container.

### DIFF
--- a/packages/super-editor/src/editors/v1/components/toolbar/Toolbar.vue
+++ b/packages/super-editor/src/editors/v1/components/toolbar/Toolbar.vue
@@ -11,6 +11,7 @@ import {
 } from 'vue';
 import { throttle } from './helpers.js';
 import ButtonGroup from './ButtonGroup.vue';
+import { RESPONSIVE_BREAKPOINTS } from './constants.js';
 
 /**
  * The default font-family to use for toolbar UI surfaces when no custom font is configured.
@@ -23,6 +24,8 @@ const { proxy } = getCurrentInstance();
 const emit = defineEmits(['command', 'toggle', 'select']);
 
 let toolbarKey = ref(1);
+const compactSideGroups = ref(false);
+let containerResizeObserver = null;
 
 /**
  * Computed property that determines the font-family to use for toolbar UI surfaces.
@@ -49,6 +52,9 @@ const getFilteredItems = (position) => {
   return proxy.$toolbar.getToolbarItemByGroup(position).filter((item) => !excludeButtonsList.includes(item.name.value));
 };
 
+const updateCompactSideGroups = () => {
+  compactSideGroups.value = proxy.$toolbar.getAvailableWidth() <= RESPONSIVE_BREAKPOINTS.lg;
+};
 const onKeyDown = async (e) => {
   if (e.metaKey && e.key === 'f') {
     const searchItem = proxy.$toolbar.getToolbarItemByName('search');
@@ -65,6 +71,7 @@ const onKeyDown = async (e) => {
 
 const onWindowResized = async () => {
   await proxy.$toolbar.onToolbarResize();
+  updateCompactSideGroups();
   toolbarKey.value += 1;
 };
 const onResizeThrottled = throttle(onWindowResized, 300);
@@ -72,11 +79,25 @@ const onResizeThrottled = throttle(onWindowResized, 300);
 function teardownWindowListeners() {
   window.removeEventListener('resize', onResizeThrottled);
   window.removeEventListener('keydown', onKeyDown);
+  containerResizeObserver?.disconnect();
+  containerResizeObserver = null;
 }
 
 function setupWindowListeners() {
+  teardownWindowListeners();
   window.addEventListener('resize', onResizeThrottled);
   window.addEventListener('keydown', onKeyDown);
+  if (
+    typeof ResizeObserver !== 'undefined' &&
+    proxy.$toolbar.config?.responsiveToContainer &&
+    proxy.$toolbar.toolbarContainer
+  ) {
+    containerResizeObserver = new ResizeObserver(() => {
+      onResizeThrottled();
+    });
+    containerResizeObserver.observe(proxy.$toolbar.toolbarContainer);
+  }
+  updateCompactSideGroups();
 }
 
 onMounted(setupWindowListeners);
@@ -121,6 +142,7 @@ const handleToolbarMousedown = (e) => {
       tabindex="0"
       v-if="showLeftSide"
       :toolbar-items="getFilteredItems('left')"
+      :compact-side-groups="compactSideGroups"
       :ui-font-family="uiFontFamily"
       position="left"
       @command="handleCommand"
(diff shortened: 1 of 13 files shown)


artem-harbour:
@caio-pizzol @harbournick - please review and merge.

packages/super-editor/src/editors/v1/components/toolbar/Toolbar.vue
```diff
@@ -40,9 +41,14 @@ const getFilteredItems = (position) => {
   return proxy.$toolbar.getToolbarItemByGroup(position).filter((item) => !excludeButtonsList.includes(item.name.value));
 };
 
+const updateCompactSideGroups = () => {
```
caio-pizzol, line 55: this only runs on page load and browser resize. if the container gets narrower without a browser resize (side panel, drawer), the toolbar never updates. I reproduced it — "Editing" button ended up 387px outside the container. a `ResizeObserver` on `toolbarContainer` would fix it.

packages/super-editor/src/editors/v1/components/toolbar/Toolbar.vue
```diff
@@ -40,9 +41,14 @@ const getFilteredItems = (position) => {
...
+  compactSideGroups.value = (proxy.$toolbar.toolbarContainer?.offsetWidth ?? 0) <= 1280;
```
caio-pizzol: this reads the container width directly, but `defaultItems.js` reads a different width depending on the `responsiveToContainer` flag. with the default config they disagree — side groups compact but button labels don't. both should use the same width. also, `1280` should use `RESPONSIVE_BREAKPOINTS.lg` from `defaultItems.js`.

packages/super-editor/src/editors/v1/components/toolbar/defaultItems.js
```diff
@@ -1043,18 +1043,38 @@ export const makeDefaultItems = ({
...
+  // `availableWidth` comes from SuperToolbar and represents either:
+  // - container width when `responsiveToContainer: true`
+  // - viewport/document width when `responsiveToContainer: false`
+  const RESPONSIVE_BREAKPOINTS = {
     sm: 768,
     md: 1024,
     lg: 1280,
     xl: 1410,
   };
+  const XL_OVERFLOW_SAFETY_BUFFER = 20;
```
caio-pizzol, line 1005: no comment on what the 20px is for, and the `<` → `<=` flip means the actual shift is 21px. one-liner explaining the why would help.

packages/super-editor/src/editors/v1/components/toolbar/defaultItems.js
```diff
@@ -1043,18 +1043,38 @@ export const makeDefaultItems = ({
...
   const stickyItemsWidth = 120;
   const toolbarPadding = 32;
 
   const itemsToHideXL = ['linkedStyles', 'clearFormatting', 'copyFormat', 'ruler'];
   const itemsToHideSM = ['zoom', 'fontFamily', 'fontSize', 'redo'];
+  const shouldCompactDocumentMode = availableWidth <= RESPONSIVE_BREAKPOINTS.lg;
+  const shouldCompactLinkedStyles = availableWidth <= RESPONSIVE_BREAKPOINTS.lg;
```
caio-pizzol: `shouldCompactDocumentMode` and `shouldCompactLinkedStyles` are identical — merge or comment why they're separate.

packages/superdoc/src/dev/components/SuperdocDev.vue
```diff
@@ -730,7 +730,7 @@ const init = async () => {
         // },
         // fonts: null,
         // hideButtons: false,
-        // responsiveToContainer: true,
+        responsiveToContainer: true,
```
caio-pizzol: the `ai: {}` comment-out on lines 856-861 is unrelated and breaks the AI button in dev. worth reverting.

caio-pizzol (changes requested):
@artem-harbour good direction on moving from CSS media queries to JS-driven compaction.

one blocking issue: the toolbar only recomputes on browser resize, not container resize — so side panels / layout changes leave it stale. reproduced: "Editing" button 387px outside the container. `ResizeObserver` fixes it.

also: the side-group check and the button-hiding check use different width sources, so they disagree under the default config. and the ticket asks for a regression test at the reproduction width — none added yet. left inline comments.

artem-harbour:
Note: Not sure why the test fails here, but locally it passes.
`pnpm --filter @superdoc-testing/behavior test tests/toolbar/overflow-dropdown-label.spec.ts`

packages/super-editor/src/editors/v1/components/toolbar/Toolbar.vue
```diff
@@ -65,18 +71,29 @@ const onKeyDown = async (e) => {
...
   window.removeEventListener('keydown', onKeyDown);
+  containerResizeObserver?.disconnect();
+  containerResizeObserver = null;
 }
 
 function setupWindowListeners() {
+  teardownWindowListeners();
   window.addEventListener('resize', onResizeThrottled);
   window.addEventListener('keydown', onKeyDown);
+  if (typeof ResizeObserver !== 'undefined' && proxy.$toolbar.toolbarContainer) {
```
caio-pizzol: this watches the container, but with the default config the toolbar gets its width from the window. shrinking the container fires a full refresh that changes nothing, except it closes any open dropdown.

```suggestion
  if (
    proxy.$toolbar.config?.responsiveToContainer &&
    typeof ResizeObserver !== 'undefined' &&
    proxy.$toolbar.toolbarContainer
  ) {
```

packages/super-editor/src/editors/v1/components/toolbar/defaultItems.js
```diff
@@ -996,18 +996,34 @@ export const makeDefaultItems = ({
...
-    xl: 1410,
-  };
+  // Responsive toolbar calculations.
+  // `availableWidth` comes from SuperToolbar and represents either:
+  // - container width when `responsiveToContainer: true`
+  // - viewport/document width when `responsiveToContainer: false`
+
+  // Extra headroom to prevent toolbar jitter at the XL edge;
+  // with `<=` this effectively shifts overflow by 21px.
+  const XL_OVERFLOW_SAFETY_BUFFER = 20;
```
caio-pizzol, line 1005: the comment says `<=` plus `+ 20` actually shifts by 21. pick one: use `<` to keep 20, or set the buffer to 21 so the name matches the math.

packages/superdoc/src/dev/components/SuperdocDev.vue
```diff
@@ -734,7 +734,7 @@ const init = async () => {
...
```
caio-pizzol: this was commented before. prod default is `false`, so turning it on here means everyone running the dev app skips the default path and won't notice bugs in it locally. re-comment, or flip the prod default to `true` if that's the plan.

```suggestion
        // Matches production default; uncomment locally to reproduce SD-2328.
        // responsiveToContainer: true,
```

caio-pizzol (changes requested):
round 1 items all addressed, and i pushed the regression test at `84a3f22d`. three inline comments, one worth fixing before merge.

caio-pizzol (approved):
hey @artem-harbour! thanks for addressing last round.

i pushed a follow-up commit with the missing tests and fixed a selector in the regression spec that was matching nothing.

lgtm once ci is green.

PR #2762, opened 2026-04-09
fix: centered image

artem-harbour:
Linear: SD-2469


VladaHarbour (approved):

artem-harbour:

caio-pizzol (approved):
@artem-harbour fix works for SD-2469. i pushed three small commits on top:

- **indent support** — centering now accounts for paragraph indents, matching Word
- **distribute alignment** — `w:jc="distribute"` now centers groups (regular justify stays left), matching Word
- **7 more unit tests** — guards for wrong drawing types, missing wrap, other alignment values, oversized groups

uploaded a 17-case test file to the corpus covering every alignment value + indents + RTL + edge cases. layout and visual tests pick it up automatically.

note: the Codex bot's suggestion to use `block.wrap?.type` would break the feature — `normalizeWrap` strips `'Inline'` from that field. the current `attrs.wrap` approach is correct.

all 17 cases verified in dev app against Word.

harbournick (approved):
LGTM

PR #2765, opened 2026-04-09
SD-2505 - fix: avoid UI stalls with on Typo.js example

chittolinag:
### Issue
- Typo.js proofing work executed on the main thread, creating typing lag and unresolved imports for the worker bridge (`./typoWorkerMessages`, `./typoWorker`).

### Proposed solution
- Use a worker implementation so the provider compiles and runs its spellcheck logic inside a dedicated Web Worker, keeping the UI responsive.


examples/features/spell-check/typo-js/src/createTypoJsProvider.ts
```diff
@@ -26,42 +65,58 @@ export async function createTypoJsProvider() {
...
+
+        const cleanup = () => {
+          request.signal?.removeEventListener('abort', onAbort);
+        };
+
+        const onAbort = () => {
+          pending.delete(requestId);
+          cleanup();
+          reject(createAbortError());
+        };
```
caio-pizzol, line 88: when a check takes longer than 10s, `ProofingSessionManager` gives up and the provider rejects — but nothing tells the worker to stop. the worker keeps running, eventually replies, and the reply gets thrown away here. next cycle, the same segments come back through and get sent to the worker again. i hit this in the browser against `superdoc@next` — same `segmentId` posted twice in a row, worker still busy from the first one. that's the same "spellcheck never settles" feel the PR is trying to fix, just via timeouts now instead of a blocked main thread. fix is to post a `cancel` message to the worker so it stops early.

chittolinag: will check it out!

chittolinag: @caio-pizzol I've just submitted a few enhancements there, thanks for pointing it out. Please let me know in case you reproduce it again.

caio-pizzol (commented):
@chittolinag cancel handling works, tested in the browser.

found one more: the regex is shared across checks, so two running at once stomp each other's position — one skips words, the other re-scans. rare (session manager usually cancels first) but shouldn't rely on it.

pushed `31d7557` (give each check its own regex via `matchAll`) and `3dc53fe` (map→set + type tightening). revert either if you'd rather. lgtm with these in.

PR #2766, opened 2026-04-09
fix(super-editor): handle splitBefore/splitAfter in whole-textblock rewrite (SD-2362)

caio-pizzol:
Follow-up to #2733. Fixes a remaining edge case in the structural text.rewrite path where leading/trailing newlines on whole-paragraph replacements fell back to inline insertion.

- The `replacesEntireTextblock` branch in `resolveStructuralRangeRewrite` used `payload.blocks` directly without applying `splitBefore`/`splitAfter` signals
- Inputs like `"\nAlpha"` (one core block + `splitBefore: true`) hit the `length <= 1` guard and fell to inline, inserting a literal `\n` character instead of creating a paragraph split
- This produced non-compliant OOXML on export (literal `\n` in `<w:t>` instead of separate `<w:p>` elements)

Fix moves the `effectiveBlocks` computation above both branches so the whole-textblock path uses the same split-signal logic the partial-range path already had.

- 5 unit tests for split-signal edge cases (splitBefore, splitAfter, both, + multi-block combos)
- 4 integration tests for leading/trailing newlines in direct and tracked modes through the full editor API

SD-2362

### DIFF
--- a/packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.test.ts
+++ b/packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.test.ts
@@ -1023,6 +1023,284 @@ describe('executeCompiledPlan: text.rewrite style behavior', () => {
     expect(tr.doc.child(0).textContent).toBe('Alpha\n\nBeta');
   });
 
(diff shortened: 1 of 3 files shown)


harbournick (approved):
LGTM

PR #2769, opened 2026-04-10
fix: replacement text in bubble

artem-harbour:
Linear: SD-2509


chittolinag (approved):

artem-harbour:

caio-pizzol (approved):
@artem-harbour good fix, verified in the dev app.

the bug was that after the first keystroke of a replacement, follow-up keystrokes lost the deleted text because only the insert mark came through in the transaction. your fix catches that correctly.

pushed a few additions to your branch:

- extra assertions on your test + a guard that pure insertions don't accidentally get deleted text
- a playwright spec that reproduces the full bug scenario end-to-end
- renamed `isDeletionInsertion` → `isReplacement` — the old name confused every reader, now it's obvious

PR #2781, opened 2026-04-11
docs: update Angular integration examples for modern APIs

ArturQuirino:
## Summary

Updates the Angular integration docs so examples follow current Angular patterns: signal-based `viewChild` / `input()` and teardown with `inject(DestroyRef)` instead of `@ViewChild`, `@Input()`, and `ngOnDestroy`.

## What changed

- **Basic setup** — `viewChild.required`, `AfterViewInit`, and `inject(DestroyRef).onDestroy()` for `superdoc.destroy()`; SuperDoc is created in `ngAfterViewInit` so the template ref exists.
- **Full component** — `input.required()` / `input()` for inputs; same `viewChild` + `DestroyRef` pattern; document and user read via `this.document()` / `this.user()`.
- **Handle file uploads** — Restores full imports at the top of the snippet; same `viewChild` + `DestroyRef` cleanup; selector uses `this.editorRef().nativeElement`.

## Why

Keeps the Getting Started Angular page aligned with modern Angular and copy-pasteable for recent projects.

## Test plan

- [ ] `pnpm run check:imports`
- [ ] `pnpm run dev:docs` — verify [Angular Integration](https://docs.superdoc.dev) page renders and snippets look correct


apps/docs/getting-started/frameworks/angular.mdx
```diff
@@ -15,29 +15,29 @@ npm install superdoc
...
 import 'superdoc/style.css';
 
 @Component({
   selector: 'app-editor',
   template: `<div #editor style="height: 700px"></div>`,
 })
-export class EditorComponent implements OnInit, OnDestroy {
-  @ViewChild('editor', { static: true }) editorRef!: ElementRef;
+export class EditorComponent implements AfterViewInit {
+  private readonly editorRef = viewChild.required<ElementRef>('editor');
```
caio-pizzol: small nit (applies to all three snippets): typing `ElementRef<HTMLDivElement>` gives `nativeElement` a real type instead of `any`. since the template is a `<div>`, it's a free win.

```suggestion
  private readonly editorRef = viewChild.required<ElementRef<HTMLDivElement>>('editor');
```

caio-pizzol (commented):
@ArturQuirino nice cleanup — the code is correct and consistent across all three snippets, and snippet 3 quietly fixes a missing-imports bug from the old version.

one ask: these examples need **Angular 17.1+** (signal-based `viewChild()` and `input()` shipped in 17.1). the old examples worked on much older versions. anyone on older Angular who copy-pastes will get errors they won't understand. worth adding a one-liner under the intro on line 7? something like:

> Examples use signal-based APIs (`viewChild()`, `input()`) introduced in **Angular 17.1**. For older Angular versions, swap to the decorator-based equivalents (`@ViewChild`, `@Input`).

also left a small nit inline on typing the element ref.

apps/docs/getting-started/frameworks/angular.mdx
ArturQuirino: Added it as suggested.

ArturQuirino:
@caio-pizzol  I added the text below so it makes it clearer for those using older versions of Angular what they need to change in the code to be able to use the code in this doc.

`Examples use signal-based APIs (`viewChild()`, `input()`) introduced in Angular 17.1. For older Angular versions, swap to the decorator-based equivalents (`@ViewChild`, `@Input`).`

caio-pizzol:
@ArturQuirino good call adding the note — but `viewChild()` needs 17.2, not 17.1 (tried it on 17.1 and it doesn't work). pushed `a5f54a752` with the fix — tweak if you'd like.

PR #2783, opened 2026-04-11
fix: remove toolbar window listeners on unmount

ArturQuirino:
## Summary

Ensures the default editor `Toolbar` removes its `window` `resize` and `keydown` listeners when the component is destroyed, not only when a `<KeepAlive>` boundary deactivates it.

## Problem

`onDeactivated` runs for kept-alive components when they are hidden, but **not** on a normal unmount. Global listeners registered in `onMounted` could therefore stay attached after the toolbar was torn down.

## Solution

- Introduce a small `teardownWindowListeners()` helper and call it from both `onDeactivated` and `onBeforeUnmount`.
- Add `Toolbar.test.js` to assert `removeEventListener` is invoked with the same handler references used in `addEventListener` (important because `resize` uses a throttled wrapper).

## Test plan

- [x] `pnpm --filter @superdoc/super-editor exec vitest run src/editors/v1/components/toolbar/Toolbar.test.js`
- [ ] (optional) `pnpm --filter @superdoc/super-editor exec vitest run src/editors/v1/tests/toolbar/`


packages/super-editor/src/editors/v1/components/toolbar/Toolbar.vue
```diff
@@ -70,6 +60,19 @@ const onWindowResized = async () => {
...
+  window.removeEventListener('resize', onResizeThrottled);
+  window.removeEventListener('keydown', onKeyDown);
+}
+
+onMounted(() => {
+  window.addEventListener('resize', onResizeThrottled);
+  window.addEventListener('keydown', onKeyDown);
+});
+
+onDeactivated(teardownWindowListeners);
```
caio-pizzol, line 84: `onDeactivated` runs when a `<KeepAlive>` hides this component, but there's no `onActivated` to put the listeners back when it's shown again. if the toolbar is ever wrapped in `<KeepAlive>`, the second time it shows up Cmd+F and the resize handling will be dead. if it's never wrapped, you can just drop `onDeactivated` and keep `onBeforeUnmount`. worth a look?

packages/super-editor/src/editors/v1/components/toolbar/Toolbar.test.js
```diff
@@ -0,0 +1,51 @@
...
+    getToolbarItemByName: () => null,
+    onToolbarResize: vi.fn(),
+    emitCommand: vi.fn(),
+    overflowItems: [],
+    activeEditor: null,
+  };
+}
+
+describe('Toolbar', () => {
+  it('removes resize and keydown listeners on unmount (not only on KeepAlive deactivate)', () => {
```
caio-pizzol, line 35: the test only checks the unmount path. the whole point of this fix is that *both* paths matter — could you add one that hides the toolbar inside `<KeepAlive>` too? that way nobody deletes `onDeactivated` later thinking it's unused.

caio-pizzol (commented):
@ArturQuirino clean fix, and it matches how `SuperEditor.vue` already does this. one thing to think about: `onDeactivated` removes the listeners but nothing puts them back on show — if the toolbar ever lives inside `<KeepAlive>`, it'll break after the first hide. not caused by this PR, just worth checking. also suggested one extra test so the two paths stay covered. left inline notes.

packages/super-editor/src/editors/v1/components/toolbar/Toolbar.vue
ArturQuirino: That makes sense. I updated it so now the events are added on `onMounted` and `onActivated`.

packages/super-editor/src/editors/v1/components/toolbar/Toolbar.test.js
ArturQuirino: Done. Now it hides it, check if the removeSpy were called, shows it and check is the addSpy was called.

packages/super-editor/src/editors/v1/components/toolbar/Toolbar.vue
caio-pizzol: thanks @ArturQuirino — keeps things correct for KeepAlive consumers. heads up: i tested in the dev app and the default SuperDoc path isn't wrapped in KeepAlive, so it's `onBeforeUnmount` that actually runs there (covered in my follow-up commit).

packages/super-editor/src/editors/v1/components/toolbar/Toolbar.test.js
caio-pizzol: looks good, thanks!

caio-pizzol (commented):
hey @ArturQuirino, thanks for catching this.

ran it in the dev app — your `onBeforeUnmount` hook wasn't firing on destroy because `SuperToolbar.destroy()` never unmounted the Vue app. i pushed a small follow-up to your branch (maintainer edits were on):

- added `this.app?.unmount()` in `SuperToolbar.destroy()` — makes your fix actually kick in
- dropped the redundant `teardownWindowListeners()` call inside `setupWindowListeners`
- added `afterEach(vi.restoreAllMocks)` so spies can't leak between tests

confirmed fixed: a resize after destroy no longer runs the toolbar handler. your tests still pass.

nice-to-have for later: `onResizeThrottled.cancel()` on teardown to kill any pending throttle timer.

PR #2801, opened 2026-04-14
feat(super-editor): add pointer event observability

mattConnHarbour:


packages/super-editor/src/editors/v1/core/types/EditorEvents.ts
```diff
@@ -148,4 +148,13 @@ export interface EditorEventMap extends DefaultEventMap {
...
   protectionChanged: [{ editor: Editor; state: DocumentProtectionState; source: ProtectionChangeSource }];
+
+  /** Called on pointer down (local only, not broadcast via collaboration) */
+  pointerDown: [{ editor: Editor; event: PointerEvent }];
+
+  /** Called on pointer up (local only, not broadcast via collaboration) */
+  pointerUp: [{ editor: Editor; event: PointerEvent }];
+
+  /** Called on right-click (local only, not broadcast via collaboration) */
+  rightClick: [{ editor: Editor; event: MouseEvent }];
```
caio-pizzol: the value passed is a `PointerEvent` (it comes from the pointer-down handler), but you're typing it as `MouseEvent`. that hides pointer-only fields like `pointerId`. the other two events above have it right.

```suggestion
  /** Called on right-click (local only, not broadcast via collaboration) */
  rightClick: [{ editor: Editor; event: PointerEvent }];
```

same fix in `EditorConfig.ts:521` for `onRightClick`.

caio-pizzol (approved):
hey @mattConnHarbour! good call on adding these :)

one small type thing inline.

PR #2804, opened 2026-04-14
feat(layout): enable odd/even header-footer support (w:evenAndOddHeaders)

### LINKED ISSUE #2803: feat: enable odd/even header-footer support (w:evenAndOddHeaders)
## What

Word documents with "Different Odd & Even Pages" enabled (`w:evenAndOddHeaders` in `word/settings.xml`) render the same header/footer on every page in SuperDoc. The even/odd variants are ignored.

## Root Cause

The layout engine hardcodes `alternateHeaders = false` at `packages/layout-engine/layout-engine/src/index.ts:1287`, preventing the paginator from selecting the correct header/footer variant per page. This causes incorrect margin calculations — the body content area doesn't account for different header heights on odd vs even pages.

The rest of the pipeline already supports this:
- **Import**: `w:evenAndOddHeaders` is parsed from `settings.xml` → `styles.alternateHeaders`
- **Layout Bridge**: `getHeaderFooterTypeForSection()` already handles even/odd variant selection
- **Rendering**: The provider pattern renders whichever variant is selected

## Expected Behavior

- Page 1 (odd): renders odd/default header
- Page 2 (even): renders even header  
- Page 3 (odd): renders odd/default header
- Margins adjust per page based on actual header/footer content height for each variant

## Spec Reference

- ECMA-376 §17.10 — Headers and Footers
- `w:evenAndOddHeaders` element in `word/settings.xml`

gpardhivvarma:
## Summary

Closes #2803

- Thread `alternateHeaders` from document settings (`w:evenAndOddHeaders`) through the layout engine so the paginator selects the correct header/footer variant per page
- Fix `getVariantTypeForPage` to use document page number (not section-relative) for even/odd selection, matching the rendering side (`headerFooterUtils.ts`)
- Add 4 unit tests covering variant selection, default behavior, and titlePg interaction

## Problem

Documents with "Different Odd & Even Pages" enabled render the same header/footer on every page. The layout engine hardcoded `alternateHeaders = false`, preventing correct margin calculations for odd vs even page headers.

## Changes

| File | Change |
|------|--------|
| `layout-engine/src/index.ts` | Add `alternateHeaders` to `LayoutOptions`, replace hardcoded `false`, fix `getVariantTypeForPage` to use document page number for even/odd |
| `layout-engine/src/index.test.ts` | 4 tests: even/odd selection, default false, omitted defaults, titlePg + alternateHeaders |
| `PresentationEditor.ts` | Thread `converter.pageStyles.alternateHeaders` into layout options |
| `types.ts` | Add `alternateHeaders` to `ResolvedLayoutOptions` |

## What was already working (no changes needed)

- Import: `w:evenAndOddHeaders` parsed from `settings.xml`
- Layout Bridge: `getHeaderFooterTypeForSection()` handles even/odd variant selection
- Rendering: provider pattern renders whichever variant is selected
- Export: `settings.xml` preserved as-is

## Test plan

- [x] `bun test` in layout-engine — 581/581 passing
- [x] 4 new alternateHeaders tests verify margin calculation per variant
- [x] Manual verification with a Word doc using "Different Odd & Even Pages"

### DIFF
--- a/packages/layout-engine/layout-engine/src/index.test.ts
+++ b/packages/layout-engine/layout-engine/src/index.test.ts
@@ -5719,4 +5719,104 @@ describe('alternateHeaders (odd/even header differentiation)', () => {
     expect(p3Fragment).toBeDefined();
     expect(p3Fragment!.y).toBeCloseTo(110, 0);
   });
+
+  it('multi-section: uses document page number for even/odd, not section-relative', () => {
+    // Section 1 has 3 pages (pages 1-3), section 2 starts on page 4.
+    // Page 4 is even by document number, but sectionPageNumber=1 (odd).
+    // The fix ensures document page number is used for even/odd.
+    const sb1: SectionBreakBlock = {
+      kind: 'sectionBreak',
+      id: 'sb1',
+      attrs: { isFirstSection: true, source: 'sectPr', sectionIndex: 0 },
+      pageSize: { w: 600, h: 800 },
+      margins: { top: 50, right: 50, bottom: 50, left: 50, header: 30 },
+    };
+    const sb2: SectionBreakBlock = {
+      kind: 'sectionBreak',
+      id: 'sb2',
+      type: 'nextPage',
+      attrs: { source: 'sectPr', sectionIndex: 1 },
+      pageSize: { w: 600, h: 800 },
+      margins: { top: 50, right: 50, bottom: 50, left: 50, header: 30 },
+    };
+
+    const options: LayoutOptions = {
+      pageSize: { w: 600, h: 800 },
+      margins: { top: 50, right: 50, bottom: 50, left: 50, header: 30 },
+      alternateHeaders: true,
+      sectionMetadata: [{ sectionIndex: 0 }, { sectionIndex: 1 }],
+      headerContentHeights: {
+        odd: 80,
+        even: 40,
+      },
+    };
+
+    const layout = layoutDocument(
+      [sb1, tallBlock('p1'), tallBlock('p2'), tallBlock('p3'), sb2, tallBlock('p4')],
+      [{ kind: 'sectionBreak' }, tallMeasure, tallMeasure, tallMeasure, { kind: 'sectionBreak' }, tallMeasure],
+      options,
+    );
+
+    expect(layout.pages.length).toBeGreaterThanOrEqual(4);
+
+    // Page 4 (documentPageNumber=4, even) → should use 'even' header (40px)
+    // NOT 'odd' which would happen if sectionPageNumber (1) were used
+    // Body start = max(50, 30+40) = 70
+    const p4Fragment = layout.pages[3]?.fragments.find((f) => f.blockId === 'p4');
+    expect(p4Fragment).toBeDefined();
+    expect(p4Fragment!.y).toBeCloseTo(70, 0);
+  });
+
+  it('selects even/odd footer heights when alternateHeaders is true', () => {
+    const options: LayoutOptions = {
+      pageSize: { w: 600, h: 800 },
+      margins: { top: 50, right: 50, bottom: 50, left: 50, footer: 30 },
+      alternateHeaders: true,
+      footerContentHeights: {
+        odd: 80, // Odd pages: larger footer
+        even: 40, // Even pages: smaller footer
+      },
+    };
+
+    const layout = layoutDocument([tallBlock('p1'), tallBlock('p2')], [tallMeasure, tallMeasure], options);
+
+    expect(layout.pages).toHaveLength(2);
+
+    // Page 1 (odd): effective bottom margin = max(50, 30+80) = 110
+    // Content area = 800 - 50 - 110 = 640px
+    // Page 2 (even): effective bottom margin = max(50, 30+40) = 70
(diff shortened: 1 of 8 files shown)


gpardhivvarma:
@caio-pizzol

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
```diff
@@ -4238,10 +4238,12 @@ export class PresentationEditor extends EventEmitter {
...
       const blocksForLayout = semanticFootnoteBlocks.length > 0 ? [...blocks, ...semanticFootnoteBlocks] : blocks;
-      const layoutOptions =
-        !isSemanticFlow && footnotesLayoutInput
+      const layoutOptions = {
+        ...(!isSemanticFlow && footnotesLayoutInput
           ? { ...baseLayoutOptions, footnotes: footnotesLayoutInput }
-          : baseLayoutOptions;
+          : baseLayoutOptions),
+        alternateHeaders: Boolean((this.#editor as EditorWithConverter).converter?.pageStyles?.alternateHeaders),
+      };
```
caio-pizzol: `alternateHeaders` is always added but the type only declares it on the paginated branch of `ResolvedLayoutOptions`. no runtime issue, just a type mismatch.

```suggestion
      const layoutOptions = {
        ...(!isSemanticFlow && footnotesLayoutInput
          ? { ...baseLayoutOptions, footnotes: footnotesLayoutInput }
          : baseLayoutOptions),
        ...(isSemanticFlow
          ? {}
          : { alternateHeaders: Boolean((this.#editor as EditorWithConverter).converter?.pageStyles?.alternateHeaders) }),
      };
```

packages/layout-engine/layout-engine/src/index.test.ts
```diff
@@ -5290,3 +5290,136 @@ describe('requirePageBoundary edge cases', () => {
     });
```
caio-pizzol, line 5587: all 4 tests here are single-section — the pre-PR code would pass them. worth adding:

1. multi-section — section 2 starting on an even page (the actual fix)
2. footer — top margin is covered, footer uses a separate ternary
3. only default header set — common case, hits the fallback at `index.ts:1337`

packages/layout-engine/layout-engine/src/index.ts
```diff
@@ -1284,11 +1294,10 @@ export function layoutDocument(blocks: FlowBlock[], measures: Measure[], options
...
         const titlePgEnabled = sectionMetadata?.titlePg ?? false;
-        // TODO: Support alternateHeaders (odd/even) when needed
-        const alternateHeaders = false;
+        const alternateHeaders = options.alternateHeaders ?? false;
 
         // Determine which header/footer variant applies to this page
-        const variantType = getVariantTypeForPage(sectionPageNumber, titlePgEnabled, alternateHeaders);
+        const variantType = getVariantTypeForPage(sectionPageNumber, newPageNumber, titlePgEnabled, alternateHeaders);
 
         // Resolve header/footer refs for margin calculation using OOXML inheritance model.
```
caio-pizzol, line 1325: checked ECMA §17.10.5 — your code is right. cross-section inheritance is the spec behavior.

FYI only: renderer (`headerFooterUtils.ts:364`) doesn't do the same inheritance — returns `default` where your layout returns `even`. pre-PR this didn't matter (alternateHeaders was always false). now it does. worth a follow-up.

caio-pizzol (commented):
@gpardhivvarma nice fix. tested against a few Word docs — the core change works.

small cleanup: `alternateHeaders` gets added to the options object in semantic flow too, but the type only allows it in paginated. inline below.

tests — all 4 are single-section, so the actual fix isn't exercised (old buggy code would pass them). left drop-in suggestions inline.

FYI, not a blocker: your PR surfaces a renderer bug. when section 2 doesn't define an even header, your layout correctly inherits from section 1 (per ECMA §17.10.5), but the renderer in `headerFooterUtils.ts` paints the default. so the page reserves space for the tall header but paints the short one. follow-up issue.

[fixture-c-section2-starts-even.docx](https://github.com/user-attachments/files/26908555/fixture-c-section2-starts-even.docx)

gpardhivvarma:
@caio-pizzol thanks for the thorough review — all three items addressed:

1. **Type guard** — `alternateHeaders` now only added in paginated flow via `isSemanticFlow` check
2. **Multi-section test** — section 2 starts on page 4 (even by document number, odd by section number) — directly exercises the `documentPageNumber` fix
3. **Footer test** — even/odd footer heights with `alternateHeaders`
4. **Default-only fallback** — only `default` header defined, verifies the common case

Re: the renderer mismatch at `headerFooterUtils.ts:364` — noted, happy to do a follow-up PR for that.

caio-pizzol (commented):
pushed two more commits:

- rewrote the footer and fallback tests — they weren't catching bugs.
- moved the odd/even flag next to the other page-level options.
- added the missing import, wiring, and browser tests.

opened a real Word doc to double-check: every page gets the right header, and saving it back out keeps everything intact. uploaded the doc to our corpus.

renderer follow-up from round 1 still open, not blocking.

lgtm.

PR #2806, opened 2026-04-14
fix(style-engine): resolve built-in Word heading styles when missing from styles.xml

### LINKED ISSUE #2805: bug: built-in Word heading styles not rendered when missing from styles.xml
## Problem

Documents created with python-docx's `doc.add_heading()` render headings as unstyled body text in SuperDoc. The headings lose their font size, weight, and spacing.

## Root Cause

`doc.add_heading()` creates paragraphs referencing Word's built-in heading styles (`<w:pStyle w:val="Heading1"/>`) without adding explicit `<w:style>` definitions in `styles.xml`. This is valid OOXML — Word has ~260 built-in styles that implicitly exist.

SuperDoc's style-engine (`style-engine/src/ooxml/index.ts:283-284`) returns `{}` when a `styleId` is not found in `translatedLinkedStyles.styles`, with no fallback to Word's built-in defaults. The existing test at `index.test.ts:58-62` documents this: `'MissingStyle'` silently returns empty.

The `latentStyles` element is parsed during import (`super-converter/v3/handlers/w/latentStyles/`) but is never used to provide fallback definitions.

## Reproduction

```python
from docx import Document

doc = Document()
doc.add_heading('Title', level=1)     # Uses built-in Heading 1 style
doc.add_heading('Subtitle', level=2)  # Uses built-in Heading 2 style
doc.add_paragraph('Body text')
doc.save('test.docx')
```

Open `test.docx` in SuperDoc — all three lines render identically as body text.

## Expected Behavior

- Heading 1: 26pt, bold, 12pt spacing before
- Heading 2: 20pt, bold, 8pt spacing before
- Body text: 11pt, normal weight

Per ECMA-376 §17.7.4.9, applications must provide built-in style defaults for standard style IDs when not explicitly defined in the document.

## Suggested Fix

Add fallback logic in `resolveStyleChain()` at `style-engine/src/ooxml/index.ts:283-284` to return Word's built-in heading properties when `styleDef` is undefined but `styleId` matches a known built-in pattern (`Heading1`–`Heading9`).

gpardhivvarma:
## Summary

Closes #2805

- Add fallback to Word's built-in heading defaults (Heading1–Heading6) in `resolveStyleChain()` when a referenced style is not defined in the document's `styles.xml`
- Apply the same fallback in the `basedOn` chain walk so custom styles inheriting from built-in headings resolve correctly
- Add 9 unit tests covering fallback resolution, precedence, basedOn inheritance, and integration through `resolveRunProperties` and `resolveParagraphProperties`

## Problem

Documents created with python-docx's `add_heading()` reference Word's built-in heading styles (`Heading1`–`Heading9`) without adding explicit `<w:style>` definitions to `styles.xml`. This is valid OOXML — Word has ~260 built-in styles that implicitly exist. SuperDoc's `resolveStyleChain()` returned `{}` for these, causing headings to render as unstyled body text.

## Changes

| File | Change |
|------|--------|
| `style-engine/src/ooxml/builtin-styles.ts` | **NEW** — static `BUILTIN_STYLE_DEFAULTS` map with Heading1–6 properties matching Word defaults (font sizes, spacing, outline levels, theme fonts, colors) |
| `style-engine/src/ooxml/index.ts` | 2-line fix: `??` fallback to built-in defaults in both initial lookup and basedOn chain walk |
| `style-engine/src/ooxml/index.test.ts` | 9 new tests |

## How it works

```typescript
// Before: missing styles return empty
const styleDef = params.translatedLinkedStyles?.styles?.[styleId];

// After: fall back to built-in defaults
const styleDef = params.translatedLinkedStyles?.styles?.[styleId] ?? BUILTIN_STYLE_DEFAULTS[styleId];
```

Document-defined styles always win (`??` only activates when the left operand is `undefined`). Built-in headings set `basedOn: 'Normal'`, so the existing chain walk picks up the document's Normal style for font family and body defaults.

## Test plan

- [x] `bun test` in style-engine — 137/137 passing
- [x] 9 new tests: run properties, paragraph properties, precedence, basedOn inheritance, outline levels, basedOn chain with built-in parents, resolveRunProperties + resolveParagraphProperties integration
- [ ] Manual verification with python-docx `add_heading()` document

### DIFF
--- a/packages/layout-engine/style-engine/src/ooxml/builtin-styles.ts
+++ b/packages/layout-engine/style-engine/src/ooxml/builtin-styles.ts
@@ -1,0 +1,175 @@
+import type { StyleDefinition } from './styles-types.js';
+
+/**
+ * Built-in Word heading style defaults.
+ *
+ * Word defines ~260 built-in styles that implicitly exist even when not declared
+ * in a document's styles.xml. When a paragraph references a built-in style
+ * (e.g., `<w:pStyle w:val="Heading1"/>`) that has no explicit definition,
+ * these defaults provide the expected visual properties.
+ *
+ * Values match Word's defaults per ECMA-376 §17.7.4.9. Font sizes are in
+ * half-points (OOXML convention), spacing in twips (twentieths of a point).
+ *
+ * All headings inherit from 'Normal' via basedOn, so any document-level Normal
+ * style properties (font family, line spacing, etc.) cascade correctly.
+ */
+export const BUILTIN_STYLE_DEFAULTS: Record<string, StyleDefinition> = {
+  Heading1: {
+    type: 'paragraph',
+    styleId: 'Heading1',
+    name: 'heading 1',
+    basedOn: 'Normal',
+    next: 'Normal',
+    qFormat: true,
+    uiPriority: 9,
+    paragraphProperties: {
+      keepNext: true,
+      keepLines: true,
+      spacing: { before: 360, after: 80 },
+      outlineLvl: 0,
+    },
+    runProperties: {
+      fontFamily: {
+        asciiTheme: 'majorHAnsi',
+        eastAsiaTheme: 'majorEastAsia',
+        hAnsiTheme: 'majorHAnsi',
+        cstheme: 'majorBidi',
+      },
+      color: { val: '0F4761', themeColor: 'accent1', themeShade: 'BF' },
+      fontSize: 40,
+      fontSizeCs: 40,
+    },
+  },
+
+  Heading2: {
+    type: 'paragraph',
+    styleId: 'Heading2',
+    name: 'heading 2',
+    basedOn: 'Normal',
+    next: 'Normal',
+    qFormat: true,
+    uiPriority: 9,
+    paragraphProperties: {
+      keepNext: true,
+      keepLines: true,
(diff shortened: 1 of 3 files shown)


gpardhivvarma:
Addressed the Codex feedback — pushed a follow-up commit adding Heading7–Heading9.

Note: `Title`, `Subtitle`, `ListParagraph`, `TOC1`–`TOC9` are also referenced across the codebase and could hit the same missing-style issue. Adding all of them would bloat this PR — the headings are the most impactful fix. Happy to add the others in a follow-up.

gpardhivvarma:
@caio-pizzol

caio-pizzol:
Thanks for the effort here, @gpardhivvarma.

I tested this end-to-end. Word itself renders a docx with missing heading style definitions as plain body text — no color, no size, nothing. With this PR, SuperDoc renders the same file as styled blue headings.

SuperDoc aims for 100% parity with Word, so we shouldn't be adding styling that Word doesn't apply.

If a docx is missing style definitions, that's a document generation error — not something we should patch over with hardcoded defaults. A band-aid here just hides the real problem and moves us away from Word's behavior.

Going to close this one. If you're hitting a real rendering bug from a specific document, feel free to share the file — it's likely a different issue (probably in the importer).

PR #2807, opened 2026-04-14
fix: render manual line breaks correctly

VladaHarbour:


packages/layout-engine/layout-bridge/src/remeasure.ts
```diff
@@ -1364,12 +1375,31 @@ export function remeasureParagraph(
...
+      //   next iteration when it is trailing or followed by another break. Otherwise consume
+      //   it immediately so mid-paragraph breaks don't create extra blank lines.
+      const emittedBreakBoundary =
+        startRun === explicitLineBreakRun && startChar === 0 && endRun === explicitLineBreakRun && endChar === 0;
+      if (emittedBreakBoundary) {
+        currentRun = explicitLineBreakRun + 1;
+      } else {
+        const nextRun = runs[explicitLineBreakRun + 1];
+        const preserveBoundaryForNextIteration = !nextRun || isLineBreakRun(nextRun);
+        currentRun = preserveBoundaryForNextIteration ? explicitLineBreakRun : explicitLineBreakRun + 1;
```
tupizz: Check this: @VladaHarbour 

 `[text, break, break, text]` produces 4 lines in remeasure but 3 in `measureParagraphBlock`. In Word, `A<br/><br/>B` renders as 3 lines: A, empty, B. The `preserveBoundaryForNextIteration` condition treats `isLineBreakRun(nextRun)` the same as `!nextRun` (trailing). When br1 terminates the text line, it should be consumed. Preserving it for re-processing creates a spurious empty line. The DOM measurement path avoids this with its placeholder+re-anchor pattern. The two paths must agree or incremental layout will produce different line counts than initial measurement. Add a test for `[textRun('A'), lineBreak, lineBreak, textRun('B')]` and fix the advance logic so consecutive breaks followed by content produce the same count as `measureParagraphBlock`.


```js
  it('[text, lineBreak, lineBreak, text] produces 3 lines matching DOM', () => {
    const block = createBlock([
      textRun('A'),
      { kind: 'lineBreak' } as Run,
      { kind: 'lineBreak' } as Run,
      textRun('B'),
    ]);
    const measure = remeasureParagraph(block, 200);
    expect(measure.lines).toHaveLength(3);
    // line 0: A
    // line 1: empty boundary
    // line 2: B
  });
```

tupizz (changes requested):
SUGGESTIONS (3)
- [testing-missing-coverage] packages/layout-engine/layout-bridge/test/remeasure.test.ts:900
  missing a test for leading lineBreak at paragraph start: `[{ kind: 'lineBreak' }, textRun('Text')]`. the DOM measurement tests cover this case (`creates an empty line for leading lineBreak at start of paragraph`), but `remeasure.test.ts` does not. same pattern as the existing tests, ~10 lines.
- [naming-and-clarity] packages/layout-engine/measuring/dom/src/index.ts:1474
  the re-anchoring condition (6 conjuncts) is a heuristic match for "this line is a lineBreak placeholder". fragile if future code creates empty lines for other reasons. a boolean flag (`isLineBreakPlaceholder`) set at line 1451 when the placeholder is created would be more explicit and self-documenting.
- [testing-missing-coverage] packages/layout-engine/layout-bridge/test/remeasure.test.ts:934
  no test for a single lineBreak as the only run: `[{ kind: 'lineBreak' }]`. edge case worth locking down since the new loop-top guard and advance logic both execute for it.

packages/layout-engine/layout-bridge/src/remeasure.ts
```diff
@@ -1364,12 +1375,31 @@ export function remeasureParagraph(
...
+      // - If this line ended before the break (text + break), only keep the break for the
+      //   next iteration when it is trailing or followed by another break. Otherwise consume
+      //   it immediately so mid-paragraph breaks don't create extra blank lines.
+      const emittedBreakBoundary =
+        startRun === explicitLineBreakRun && startChar === 0 && endRun === explicitLineBreakRun && endChar === 0;
+      if (emittedBreakBoundary) {
+        currentRun = explicitLineBreakRun + 1;
+      } else {
+        const nextRun = runs[explicitLineBreakRun + 1];
+        const preserveBoundaryForNextIteration = !nextRun || isLineBreakRun(nextRun);
```
caio-pizzol: for `[text, br, br, text]` this returns 4 lines instead of 3 (text, blank, text). each break ends up emitting its own empty line. worth a test for two breaks in the middle of a paragraph.

packages/layout-engine/layout-bridge/src/remeasure.ts
```diff
@@ -1199,11 +1199,22 @@ export function remeasureParagraph(
...
     let resumeChar = 0;
 
     for (let r = currentRun; r < runs.length; r += 1) {
       const run = runs[r];
+      if (isLineBreakRun(run)) {
+        explicitLineBreakRun = r;
+        if (startRun === r && startChar === 0 && width === 0) {
+          // Preserve leading/manual explicit break as an empty line.
+          endRun = r;
+          endChar = 0;
```
caio-pizzol, line 1219: an empty line from a break alone always uses 16px height, since the break carries no font size. the dom measurer uses the previous run's size (so a break after a 24px heading is ~28.8px). page breaks drift after remeasure for any text that isn't 16px.

caio-pizzol (commented):
@VladaHarbour two things to flag before merge - left both inline.

packages/layout-engine/layout-bridge/src/remeasure.ts
VladaHarbour: improved logic over here

packages/layout-engine/layout-bridge/src/remeasure.ts
VladaHarbour: @tupizz mentioned this above. handled this

packages/layout-engine/layout-bridge/src/remeasure.ts
VladaHarbour: added font size handling

tupizz (approved):
Lgtm!

PR #2809, opened 2026-04-14
SD-2541 - fix: avoid scrolling back when dragging selection

chittolinag:
### Issue

Drag-selecting past the viewport edge made the page bounce instead of extending the selection because edge auto-scroll and “scroll selection into view” on every `selectionUpdate` both moved the scroll container at once.

### Proposed solution

While the pointer is dragging text, skip setting and applying the selection scroll-into-view flag so only edge auto-scroll runs; on pointer up, call a small `notifyDragSelectionEnded` hook once so the editor can scroll the selection into view after auto-scroll stops. Keyboard selection (e.g. Shift + arrows) is unchanged.


harbournick (approved):
LGTM

PR #2816, opened 2026-04-14
SD-2534 - documents opened in collab corrupted after export

chittolinag:
### Issues
Collab/joiner DOCX export dropped customXml parts, could emit settings with a dangling attachedTemplate `r:id`, and mishandled endnotes (parts / types).

### Proposed solution

1. Copy `customXml/*.{xml,rels}` into updatedDocs;
2. Persist `settings.xml.rels` or strip `w:attachedTemplate` when rels are missing;
3. Export endnotes (+ rels) and register` /word/endnotes.xml` in `[Content_Types]` when needed. (thanks to @artem-harbour !)
4. Add `collab-docx-export-regression.test.js` to lock this in.


packages/super-editor/src/editors/v1/core/super-converter/exporter.js
```diff
@@ -255,7 +255,10 @@ export function translatePassthroughNode(params) {
  * @returns {XmlReadyNode} JSON of the XML-ready body node
  */
 function translateBodyNode(params) {
-  let sectPr = params.bodyNode?.elements?.find((n) => n.name === 'w:sectPr');
+  const liveSectPr = params.converter?.bodySectPr;
+  let sectPr =
```
artem-harbour, line 259: @harbournick - Could you check if this change makes sense? Without this, it seems there were issues with the footer/header when exporting.

artem-harbour (approved):
LGTM

packages/super-editor/src/editors/v1/core/super-converter/exporter.js
caio-pizzol: @harbournick @artem-harbour the change is correct. `converter.bodySectPr` is the live version that gets updated whenever section properties change. The old code read a copy taken at import time, so edits after that were getting lost on export. 

Open-and-export path is unchanged (falls back to the old copy when live is null). Added 3 tests in d0d26b4bf.

harbournick (approved):
LGTM

PR #2823, opened 2026-04-15
feat: support wide tables rendering outside of margins

VladaHarbour:
## Summary

Adds support for rendering tables wider than the content column (SD-2544 / SD-2545). Tables with explicit OOXML widths now overflow into page margins instead of being clamped to column width.

## Key changes

- **Width resolution centralized**: New `resolveTableWidthAttr` (contracts) validates table width attrs (rejects NaN/Infinity/non-positive). Used by both measuring and layout sides.
- **Wide-table layout**: New `resolveTableFrame` / `resolveRenderedTableWidth` (layout-engine) resolve final rendered width and x-offset. Allows negative x for centered/right-aligned wide tables; preserves indent for left-aligned, ignores it for justified.
- **Measuring**: Explicit OOXML widths (px/pct/dxa) now pass through unclamped — only auto-sized tables fall back to column width.
- **Fragments**: Table fragments record an optional `columnIndex` (logical column owner). Used by `position-hit` and footnote placement so click-resolution and footnote anchoring use logical ownership instead of visual x for wide overflow tables.
- **Tests**: ~600 new test cases — width-attr validation, wide-table overflow, alignment, indent edge cases, click-to-position with `columnIndex`, footnote placement on overflow tables, multi-column scenarios.

## Test plan

- [x] Unit tests pass (layout-engine, contracts, layout-bridge)
- [ ] Layout comparison vs published baseline (`pnpm test:layout`)
- [ ] Visual diff on wide-table corpus docs (`pnpm test:visual`)

### DIFF
--- a/packages/layout-engine/layout-bridge/src/incrementalLayout.ts
+++ b/packages/layout-engine/layout-bridge/src/incrementalLayout.ts
@@ -229,7 +229,9 @@ const assignFootnotesToColumns = (
 
     if (columns && columns.count > 1 && page) {
       const fragment = findFragmentForPos(page, ref.pos);
-      if (fragment && typeof fragment.x === 'number') {
+      if (fragment?.kind === 'table' && typeof fragment.columnIndex === 'number') {
+        columnIndex = Math.max(0, Math.min(columns.count - 1, fragment.columnIndex));
+      } else if (fragment && typeof fragment.x === 'number') {
         const widths = Array.isArray(columns.widths) && columns.widths.length > 0 ? columns.widths : undefined;
         if (widths) {
           let cursorX = columns.left;
@@ -1748,6 +1750,7 @@ export async function incrementalLayout(
                   page.fragments.push({
                     kind: 'table',
                     blockId: range.blockId,
+                    columnIndex,
                     fromRow: 0,
                     toRow: block.rows.length,
                     x: tableX,
--- a/packages/layout-engine/layout-bridge/src/incrementalLayout.ts
+++ b/packages/layout-engine/layout-bridge/src/incrementalLayout.ts
@@ -1738,9 +1738,10 @@ export async function incrementalLayout(
                     tableWidthRaw,
                     block.attrs,
                   );
-                  // Rescale column widths when table was clamped to section width.
-                  // This happens in mixed-orientation docs where measurement uses the
-                  // widest section but rendering is per-section (SD-1859).
+                  // Rescale column widths only when the resolved fragment width is narrower
+                  // than the measured table width. Today that primarily happens for
+                  // percentage-width tables rendered in a narrower section (SD-1859),
+                  // while non-percent wide tables keep their measured overflow width.
                   const fragmentColumnWidths = rescaleColumnWidths(
                     measure.columnWidths,
                     measure.totalWidth,
--- a/packages/layout-engine/layout-bridge/src/incrementalLayout.ts
+++ b/packages/layout-engine/layout-bridge/src/incrementalLayout.ts
@@ -9,7 +9,7 @@ import type {
   SectionBreakBlock,
   NormalizedColumnLayout,
 } from '@superdoc/contracts';
-import { cloneColumnLayout, normalizeColumnLayout } from '@superdoc/contracts';
+import { cloneColumnLayout, normalizeColumnLayout, rescaleColumnWidths } from '@superdoc/contracts';
 import {
   layoutDocument,
   layoutHeaderFooter,
@@ -20,6 +20,7 @@ import {
   type NumberingContext,
   SEMANTIC_PAGE_HEIGHT_PX,
   SINGLE_COLUMN_DEFAULT,
+  resolveTableFrame,
 } from '@superdoc/layout-engine';
 import { remeasureParagraph } from './remeasure';
 import { computeDirtyRegions } from './diff';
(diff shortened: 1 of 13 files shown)


packages/layout-engine/layout-engine/src/layout-table.ts
```diff
@@ -142,20 +187,20 @@ function applyTableIndent(x: number, width: number, indent: number, columnWidth:
...
+  const width = resolveRenderedTableWidth(columnWidth, tableWidth, attrs);
   const justification = typeof attrs?.justification === 'string' ? attrs.justification : undefined;
 
   if (justification === 'center') {
-    return { x: baseX + Math.max(0, (columnWidth - width) / 2), width };
+    return { x: baseX + (columnWidth - width) / 2, width };
   }
   if (justification === 'right' || justification === 'end') {
-    return { x: baseX + Math.max(0, columnWidth - width), width };
+    return { x: baseX + (columnWidth - width), width };
```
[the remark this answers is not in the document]
VladaHarbour: Here for any table with width > columnWidth:
1. centered tables would no longer shift left into the margin
2. right/end-aligned tables would no longer overflow leftward from the content area

Adding fix to support multi-column layout together with wide tables overflow.

packages/layout-engine/layout-engine/src/layout-table.ts
```diff
@@ -122,13 +123,57 @@ function applyTableIndent(x: number, width: number, indent: number, columnWidth:
...
+  }
+
   const maxWidthWithinColumn = Math.max(0, columnWidth - indent);
   return {
     x: shiftedX,
     width: Math.min(width, maxWidthWithinColumn),
   };
 }
 
+function resolveTableWidthValue(attrs: TableBlock['attrs']): { width: number; type?: string } | null {
```
tupizz: [suggestion]
Both `resolveTableWidthValue` here and `resolveTableWidth` in `measuring/dom/src/index.ts:2654` parse the same `attrs.tableWidth` shape and extract a numeric width. Two parsers for the same attr structure. Consider a shared utility in `@superdoc/contracts` next to `OOXML_PCT_DIVISOR` and `rescaleColumnWidths`. not blocking, worth a follow-up.

packages/layout-engine/layout-engine/src/layout-table.test.ts
```diff
@@ -727,6 +727,56 @@ describe('layoutTableBlock', () => {
       const rightFragment = layoutWithJustification('right');
       expect(rightFragment.x).toBe(300);
     });
+
+    it('allows centered wide tables to overflow into both margins', () => {
```
tupizz, line 756: The wide centered/right tests at lines 731 and 756 pass tables with no `tableWidth` attr. The overflow path works because `measure.totalWidth` (600) exceeds `columnWidth` (500) and the new code preserves it. Good. But there's no matching test for a left-aligned (default justification) wide table with no indent. That path hits `applyTableIndent` with `indent=0`, which returns `{ x: baseX, width }` via the `indent <= 0` branch. A one-liner test would lock it down.

packages/layout-engine/layout-engine/src/layout-table.test.ts
```diff
@@ -4167,7 +4244,9 @@ describe('layoutTableBlock', () => {
     it('should rescale column widths when table is wider than section content width', () => {
```
tupizz, line 4299: Three tests in the 'column width rescaling' suite were updated to add `tableWidth: { value: 5000, type: 'pct' }` to the block attrs. This is needed because `resolveRenderedTableWidth` now recomputes from attrs when present. Verify the pre-existing behavior still holds for tables without `tableWidth` (auto-width tables wider than the section). The grid-wide test at `measuring/dom/src/index.test.ts:4574` covers this, but no layout-table test exercises rescaling for an auto-width table whose grid exceeds column width.

tupizz (approved):
no blocker, just suggestions
lgtm!

packages/layout-engine/layout-engine/src/layout-table.ts
VladaHarbour: removed duplicated code here

packages/layout-engine/layout-engine/src/layout-table.test.ts
VladaHarbour: added test to cover this case

packages/layout-engine/layout-engine/src/layout-table.test.ts
VladaHarbour: added test for an auto-width table

packages/layout-engine/measuring/dom/src/index.ts
```diff
@@ -2704,26 +2668,29 @@ async function measureTableBlock(block: TableBlock, constraints: MeasureConstrai
     const hasExplicitWidth = resolvedTableWidth != null;
     const hasFixedLayout = block.attrs?.tableLayout === 'fixed';
 
-    // For tables with explicit/percentage width or fixed layout, scale to target width
+    // For tables with explicit/percentage width or fixed layout, scale to the imported
+    // target width when one exists. Wide explicit tables are allowed to exceed the
+    // current content column so they can later render into the margins.
     if (hasExplicitWidth || hasFixedLayout) {
       const totalWidth = columnWidths.reduce((a, b) => a + b, 0);
       const tableWidthType = (block.attrs?.tableWidth as TableWidthAttr | undefined)?.type;
```
caio-pizzol, line 2676: fixed-layout tables without a `tableWidth` now keep their full grid width, even when wider than the column. before, they got shrunk to fit. i pinned this with a test - is keeping them wide what you want?

packages/layout-engine/layout-engine/src/layout-table.ts
```diff
@@ -122,13 +123,39 @@ function applyTableIndent(x: number, width: number, indent: number, columnWidth:
...
+  const configuredWidth = resolveTableWidthAttr(attrs?.tableWidth);
+  if (!configuredWidth) {
+    return safeMeasuredWidth;
+  }
+
+  if (configuredWidth.type === 'pct') {
+    return Math.max(0, Math.round(columnWidth * (configuredWidth.width / OOXML_PCT_DIVISOR)));
+  }
+
+  return safeMeasuredWidth;
```
caio-pizzol, line 161: for `px`/`pixel`/`dxa` we use the measured width and ignore the attr. fine if the measure already applied it, but nothing here says so. if a stale measure ever lands here, the fixed width vanishes. use the attr here too, or add a comment.

packages/layout-engine/layout-bridge/src/incrementalLayout.ts
```diff
@@ -1728,46 +1731,26 @@ export async function incrementalLayout(
...
-                    tableWidth = Math.max(0, tableWidth - indent);
-                  }
+                  const tableWidthRaw = Math.max(0, measure.totalWidth ?? contentWidth);
+                  const { x: tableX, width: tableWidth } = resolveTableFrame(
+                    columnX,
+                    contentWidth,
+                    tableWidthRaw,
+                    block.attrs,
+                  );
                   // Rescale column widths when table was clamped to section width.
```
caio-pizzol: comment says columns rescale "when table was clamped to section width", but this only runs for percent widths now. non-percent tables hit a no-op. quick reword would help.

caio-pizzol (commented):
hey @vladaharbour, the column ownership fix looks great.

tried a 10" wide table on a 6.5" page - the last column still gets cut off at the page edge. is the rendering fix coming separately, or should it land here?

pushed a commit with tests for the new helpers and pinned a quiet behavior change. three smaller things still inline.

needs work.

packages/layout-engine/measuring/dom/src/index.ts
VladaHarbour: This is exactly what I planned since it's the way Word does it

packages/layout-engine/layout-bridge/src/incrementalLayout.ts
VladaHarbour: updated comment

packages/layout-engine/layout-engine/src/layout-table.ts
VladaHarbour: added comment

luccas-harbour (approved):
LGTM

PR #2825, opened 2026-04-15
SD-2570 - fix: document API tracked changes not fully rendered on document

chittolinag:
### Issue

Tracked text inserted programmatically with `editor.doc.insert(..., { changeMode: 'tracked' })` was truncated in the document view when the inserted string contained newline characters (`\n`). Only the content before the first newline was rendered, even though the full text was preserved in comment bubbles and export.

### Proposed solution

Align renderer run slicing with measurer behavior for inline newlines. The renderer now expands text runs containing `\n` into text/break/text segments before slicing line ranges, so measured line indices map correctly and all multiline tracked content renders in the document.


artem-harbour (approved):
LGTM

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -7696,11 +7696,50 @@ const stripListIndent = (attrs?: ParagraphAttrs): ParagraphAttrs | undefined =>
...
  * ```
  */
+const expandRunsForInlineNewlines = (runs: Run[]): Run[] => {
+  const expanded: Run[] = [];
+
+  for (const run of runs) {
+    if ((run as TextRun).text && typeof (run as TextRun).text === 'string' && (run as TextRun).text.includes('\n')) {
+      const textRun = run as TextRun;
+      const segments = textRun.text.split('\n');
+      let cursor = textRun.pmStart ?? 0;
```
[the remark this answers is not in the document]
caio-pizzol: @chittolinag real in theory but not hit — pm-adapter always sets `pmStart`. low priority.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -7696,11 +7696,50 @@ const stripListIndent = (attrs?: ParagraphAttrs): ParagraphAttrs | undefined =>
...
+    }
+
+    expanded.push(run);
+  }
+
+  return expanded;
+};
+
 export const sliceRunsForLine = (block: ParagraphBlock, line: Line): Run[] => {
+  const runs = expandRunsForInlineNewlines(block.runs as Run[]);
```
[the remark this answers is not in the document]
caio-pizzol: @chittolinag valid — perf only, fixes itself when we lift the expansion out of `sliceRunsForLine`.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -7696,11 +7696,50 @@ const stripListIndent = (attrs?: ParagraphAttrs): ParagraphAttrs | undefined =>
  * // Returns runs or run slices that fall within the specified character range
  * ```
  */
+const expandRunsForInlineNewlines = (runs: Run[]): Run[] => {
```
caio-pizzol: the jsdoc above describes `sliceRunsForLine` — the new helper got slipped between them, so IDE hovers point at the wrong function. move it, or add a short one here.

also, this is the third copy of the same expansion (measurer + `layout-bridge/src/text-measurement.ts:315` have the others). one shared helper would stop the next drift.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -7696,11 +7696,50 @@ const stripListIndent = (attrs?: ParagraphAttrs): ParagraphAttrs | undefined =>
...
+  const expanded: Run[] = [];
+
+  for (const run of runs) {
+    if ((run as TextRun).text && typeof (run as TextRun).text === 'string' && (run as TextRun).text.includes('\n')) {
```
caio-pizzol: cleaner with the existing `isTextRun` guard:

```suggestion
    if (isTextRun(run) && run.text.includes('\n')) {
```

caio-pizzol (commented):
@chittolinag ran this in the dev app console:

```js
await ed.doc.insert(
  { value: 'first line here\nsecond line here\nthird line here' },
  { changeMode: 'tracked' }
);
```

text shows up, but lines 2 and 3 have no `data-pm-start` / `data-pm-end` and clicking them doesn't move the caret. `sliceRunsForLine` got the fix, but `computeLinePmRange` (`contracts/src/pm-range.ts:82`) and the second copy in `layout-bridge/src/text-measurement.ts:315` still read un-split runs.

- restore `ParagraphMeasure.expandedRuns` from #2824 — one source of truth for all three read-sites.
- rebase: `f778c4a65` added tab handling in the measurer that the helper doesn't mirror.

packages/layout-engine/painters/dom/src/renderer.ts
caio-pizzol: `block.runs` is already typed — cast isn't needed:

```suggestion
  const runs = expandRunsForInlineNewlines(block.runs);
```

also, this runs once per line. a 100-line paragraph re-expands 100 times. better to expand once per paragraph.

packages/layout-engine/painters/dom/src/index.test.ts
```diff
@@ -10920,6 +10920,87 @@ describe('applyRunDataAttributes', () => {
...
+
+      expect(() => {
+        painter.paint(inlineNewlineLayout, mount);
+      }).not.toThrow();
+
+      const fragment = mount.querySelector('.superdoc-fragment') as HTMLElement;
+      expect(fragment).toBeTruthy();
+      expect(fragment.textContent).toContain('first');
+      expect(fragment.textContent).toContain('second');
+      expect(fragment.textContent).toContain('third');
```
caio-pizzol: the test only checks `textContent`, so it passes even if the PM data on line 2/3 is gone. could you add:

```js
const lines = fragment.querySelectorAll('.superdoc-line');
expect(lines.length).toBe(3);
expect(lines[0].dataset.pmStart).toBe('0');
expect(lines[0].dataset.pmEnd).toBe('5');
expect(lines[1].dataset.pmStart).toBe('6');
expect(lines[1].dataset.pmEnd).toBe('12');
expect(lines[2].dataset.pmStart).toBe('13');
expect(lines[2].dataset.pmEnd).toBe('18');
```

heads up — `lines[1].dataset.pmStart` is `undefined` today, not `'6'`. more on why in the summary.

packages/layout-engine/painters/dom/src/index.test.ts
```diff
@@ -10920,6 +10920,87 @@ describe('applyRunDataAttributes', () => {
...
+      const painter = createTestPainter({
+        blocks: [inlineNewlineBlock],
+        measures: [inlineNewlineMeasure],
+      });
+
+      expect(() => {
+        painter.paint(inlineNewlineLayout, mount);
+      }).not.toThrow();
+
+      const fragment = mount.querySelector('.superdoc-fragment') as HTMLElement;
```
caio-pizzol: small type nit — the cast hides the null case:

```suggestion
    const fragment = mount.querySelector<HTMLElement>('.superdoc-fragment');
    expect(fragment).not.toBeNull();
```

packages/layout-engine/painters/dom/src/renderer.ts
chittolinag: hmm yeah that's true - there are multiple versions of this function too. I'll check it out.

packages/layout-engine/painters/dom/src/index.test.ts
chittolinag: that's a great idea - just added.

chittolinag:
> @chittolinag ran this in the dev app console:
> 
> ```js
> await ed.doc.insert(
>   { value: 'first line here\nsecond line here\nthird line here' },
>   { changeMode: 'tracked' }
> );
> ```
> 
> text shows up, but lines 2 and 3 have no `data-pm-start` / `data-pm-end` and clicking them doesn't move the caret. `sliceRunsForLine` got the fix, but `computeLinePmRange` (`contracts/src/pm-range.ts:82`) and the second copy in `layout-bridge/src/text-measurement.ts:315` still read un-split runs.
> 
> * restore `ParagraphMeasure.expandedRuns` from #2824 — one source of truth for all three read-sites.
> * rebase: `f778c4a65` added tab handling in the measurer that the helper doesn't mirror.

You're right about `data-pm-start`, `data-pm-end` and `computeLinePmRange`. Just fixed them. Thanks for pointing them out!

I did notice a caret issue around tracked changes (might have been what you experienced). If you open a document with tracked changes and click directly on a tracked change once it's loaded, using the arrow keys won't work because the editor is not focused. So if you use up/down for example, the page will scroll up an down instead of navigating within the document (and left/right won't have any effect). This also happens with a blank document. If you open a blank document (and do not type anything, do not focus the editor) and programatically insert a tracked change, if you click it, the same behavior will happen. I am not sure if this is the issue that you faced, but that is already happening on `main`, so not introduced here. I think it makes more sense to create a separate ticket to handle that. What do you think?

packages/layout-engine/pm-adapter/src/converters/paragraph.ts
```diff
@@ -212,6 +212,34 @@ export function mergeAdjacentRuns(runs: Run[]): Run[] {
...
+  for (const run of runs) {
+    const textRun = run as TextRun;
+    if ('text' in run && typeof textRun.text === 'string' && textRun.text.includes('\n')) {
+      const segments = textRun.text.split('\n');
+      let cursor = textRun.pmStart ?? 0;
+      segments.forEach((segment, idx) => {
+        result.push({ ...textRun, text: segment, pmStart: cursor, pmEnd: cursor + segment.length });
+        cursor += segment.length;
+        if (idx !== segments.length - 1) {
+          result.push({ kind: 'break', breakType: 'line', pmStart: cursor, pmEnd: cursor + 1, sdt: textRun.sdt });
```
caio-pizzol: the line break we add for each inline `\n` needs to carry the source run's `trackedChange`. without it the filter hides the text halves but keeps the breaks, so `original` mode shows blank lines where the insertion should have vanished.

```suggestion
          result.push({
            kind: 'break',
            breakType: 'line',
            pmStart: cursor,
            pmEnd: cursor + 1,
            sdt: textRun.sdt,
            trackedChange: textRun.trackedChange,
          });
```

pair with adding `trackedChange?` to `BreakRun` (`contracts/src/index.ts:373`) and updating the hide-insert / hide-delete path at `tracked-changes.ts:461-479` to check breaks too.

to repro in the dev app (blank doc, console):

```js
// 1. insert 3 tracked lines - review mode, all 3 should render
await window.editor.doc.insert(
  { value: 'first\nsecond\nthird' },
  { changeMode: 'tracked' }
);

// 2. switch to 'original' - insertion should collapse to nothing
window.editor.presentationEditor.setTrackedChangesOverrides({ mode: 'original', enabled: true });

// 3. inspect - expected: []
Array.from(document.querySelectorAll('.superdoc-line')).map(l => ({
  text: l.textContent,
  pmStart: l.dataset.pmStart,
  pmEnd: l.dataset.pmEnd,
}));
```

before the fix: 2 empty lines at pm 7-8 and 14-15. after: `[]`.

caio-pizzol (changes requested):
hey @chittolinag! thanks for last round, position fixes look clean. on the caret focus thing, repro'd here too - split it off, sgtm.

one regression this PR adds: tracked text with newlines leaves empty lines visible in `original` mode (and same in `final` for deletes). fix is small, left it inline with repro steps.

needs work.

packages/layout-engine/pm-adapter/src/converters/paragraph.ts
chittolinag: Great catch @caio-pizzol - I didn't know we had that ability (`setTrackedChangesOverrides`). Thanks, just fixed!

caio-pizzol (commented):

not there yet - text with both tabs and newlines renders wrong and loses content. three inline.

packages/layout-engine/pm-adapter/src/converters/paragraph.ts
```diff
@@ -212,6 +212,41 @@ export function mergeAdjacentRuns(runs: Run[]): Run[] {
...
+ * Expands text runs that contain inline newlines into multiple runs.
+ *
+ * @param {Run[]} runs - The runs to expand
+ * @returns {Run[]} The expanded runs
+ */
+export function expandRunsForInlineNewlines(runs: Run[]): Run[] {
+  const result: Run[] = [];
+  for (const run of runs) {
+    const textRun = run as TextRun;
+    if ('text' in run && typeof textRun.text === 'string' && textRun.text.includes('\n')) {
```
caio-pizzol, line 226: the measurer splits text at both `\n` and `\t` (see `measuring/dom/src/index.ts:1265`). this helper only splits at `\n`, so a run with both characters ends up a different shape here than in the measurer, and their indices stop lining up.

i reproduced it: a text node with `"A\tB\nC\tD"` put line 2's text onto line 1, then rendered line 2 empty. the fix needs to handle tabs too, ideally by sharing one helper with the measurer so the two can't drift apart.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -120,17 +120,15 @@ import {
...
-  applyParagraphBorderStyles,
-  applyParagraphShadingStyles,
-  getParagraphBorderBox,
   stampBetweenBorderDataset,
   type BetweenBorderInfo,
 } from './features/paragraph-borders/index.js';
 import { applyRtlStyles, shouldUseSegmentPositioning } from './features/rtl-paragraph/index.js';
 import { convertOmmlToMathml } from './features/math/index.js';
+import { isTextRun, expandRunsForInlineNewlines } from '@superdoc/pm-adapter';
+import { sliceRunsForLine } from '@superdoc/layout-bridge';
```
caio-pizzol, line 125: this imports from the `@superdoc/layout-bridge` root, which pulls in everything the package exports. there are two files named `benchmarks` sitting next to each other in that package: one is browser-safe, the other uses node-only modules (`node:perf_hooks`, `@superdoc/painter-dom`). today the build picks the safe one, but one rename and node-only code lands in the browser bundle. import from a specific file instead.

packages/layout-engine/pm-adapter/src/converters/paragraph.ts
```diff
@@ -212,6 +212,41 @@ export function mergeAdjacentRuns(runs: Run[]): Run[] {
...
+      const segments = textRun.text.split('\n');
+      let cursor = textRun.pmStart ?? 0;
+      segments.forEach((segment, idx) => {
+        result.push({ ...textRun, text: segment, pmStart: cursor, pmEnd: cursor + segment.length });
```
caio-pizzol: text ending in `\n` splits to `['foo', '']`, so an empty text run gets pushed at the end. the renderer skips empty slices later, but the measurer treats an empty string as a space and gives it real width (see `measuring/dom/src/index.ts:2048`). skip the push when the segment is empty.

```suggestion
        if (segment.length > 0) {
          result.push({ ...textRun, text: segment, pmStart: cursor, pmEnd: cursor + segment.length });
        }
```

packages/layout-engine/painters/dom/src/renderer.ts
chittolinag: valid concern - after taking a look locally, it seems that these modules shouldn't even be exported though. they're only used on testing, so it doesn't make sense to include them in the build package from my perspective. is there a reason to do so?

packages/layout-engine/pm-adapter/src/converters/paragraph.ts
chittolinag: I could not reproduce the issue here - just DM'd you to double check, maybe I'm doing something different here.

caiopizzol (commented):

on the `\t`+`\n` you were right - `doc.insert` doesn't hit it (only a raw `schema.text()` transaction does), so it's latent drift, worth a follow-up but not a blocker.

i rebased on latest main and pushed c861628ae tightening the empty-segment guard (leading and consecutive `\n` were dropping breaks) with tests for the edge cases.

lgtm.

caio-pizzol (approved):

PR #2830, opened 2026-04-15
fix(super-editor): normalize API tab insertions

missysuperdoc:
## What changed
- normalize API text writes so literal tab characters are materialized as structured inline `tab` nodes instead of raw text
- reuse the same inline-content expansion for direct writes, tracked insertions, and text execution in the document API write paths
- add regression coverage for direct writes, tracked changes, and exported DOCX XML in both direct and tracked adapter flows

## Why this changed
API-driven insertions were preserving `\t` inside text nodes, which exported to raw tab characters inside `<w:t>` instead of proper `<w:tab/>` runs.

## Impact
API insertions now match the editor's native tab model, and DOCX export emits valid Word tab runs for both direct and tracked document API writes.

## Validation
- `pnpm --prefix packages/super-editor exec vitest run src/editors/v1/document-api-adapters/write-adapter.test.ts`
- `pnpm --prefix packages/super-editor exec vitest run src/editors/v1/extensions/track-changes/track-changes-extension.test.js`
- `pnpm --prefix packages/super-editor exec vitest run src/editors/v1/document-api-adapters/write-tabs-export.integration.test.ts`

### DIFF
--- a/packages/super-editor/src/editors/v1/core/helpers/buildInlineContentFromText.js
+++ b/packages/super-editor/src/editors/v1/core/helpers/buildInlineContentFromText.js
@@ -1,0 +1,53 @@
+import { Fragment } from 'prosemirror-model';
+
+/**
+ * Materialize plain text into inline ProseMirror content, upgrading literal tab
+ * characters to real tab nodes when the schema supports them.
+ *
+ * @param {import('prosemirror-model').Schema} schema
+ * @param {string} text
+ * @param {import('prosemirror-model').Mark[]} [marks]
+ * @returns {{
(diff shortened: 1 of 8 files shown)


caio-pizzol:
closing this in favor of an export-side fix. normalizing at the editor layer is too ambitious for this bug — 10+ call sites still take the raw path, and it introduces new regressions (tracked-changes panel drops tab-only inserts, reads return a different char than what was written, inserts into page-number fields silently drop the tab).

simpler fix: detect `\t` during docx serialization, emit `<w:tab/>`. ~20 lines, covers every path. will open that.

thanks for the work here @missysuperdoc 🙏

PR #2832, opened 2026-04-15
feat(executor): add tab character to tab node conversion in text insert

andrii-harbour:


caio-pizzol:
hey @andrii-harbour - review pass on this one, wanted to share a few concerns.

fixing tabs at the editor-data layer introduces two new problems:

- **marks get dropped on the tab**. `tabNodeType.create()` is called without marks, so a tab between bold text exports without the formatting.
- **reading a tab back returns shorter text than what was written**. once a real tab node exists, `textBetween` returns empty for it - breaks the no-op check at `write-adapter.ts:98`. fix isn't obvious.

scope is also narrow: `doc.replace`, untargeted `doc.insert`, and untargeted tracked inserts still emit raw `\t`. and there's a `|| true` in the new test that makes the assertion always pass.

might be worth going export-side instead: when serializing a text run, if we see `\t`, emit `<w:tab/>`. ~20 lines, one place, covers every path at once, same pattern later fixes `\n` → `<w:br/>`. no editor-layer changes - marks stay put, reads keep working.

happy to chat if there's context i'm missing.

andrii-harbour:
> hey @andrii-harbour - review pass on this one, wanted to share a few concerns.
> 
> fixing tabs at the editor-data layer introduces two new problems:
> 
> * **marks get dropped on the tab**. `tabNodeType.create()` is called without marks, so a tab between bold text exports without the formatting.
> * **reading a tab back returns shorter text than what was written**. once a real tab node exists, `textBetween` returns empty for it - breaks the no-op check at `write-adapter.ts:98`. fix isn't obvious.
> 
> scope is also narrow: `doc.replace`, untargeted `doc.insert`, and untargeted tracked inserts still emit raw `\t`. and there's a `|| true` in the new test that makes the assertion always pass.
> 
> might be worth going export-side instead: when serializing a text run, if we see `\t`, emit `<w:tab/>`. ~20 lines, one place, covers every path at once, same pattern later fixes `\n` → `<w:br/>`. no editor-layer changes - marks stay put, reads keep working.
> 
> happy to chat if there's context i'm missing.

hey @caio-pizzol ! Thnx for the review. I agree that going export-side would be easier fix, however I have one concern here. Doing that means we have different nodes/content while rendering the doc and while exporting the doc. That might not be the problem now, but in the long run with this approach our differences of actual data for document API and data for export would be too different. 
would be glad to chat on this one, maybe I'm missing something.

caio-pizzol (commented):
hey @andrii-harbour! you're right - the importer makes tab nodes on docx open anyway, so keeping the write path consistent is the cleaner call. scratch my export-side suggestion from before.

that said, the inline comments are still valid. the four things from my earlier pass haven't moved:

- marks dropped on the tab node
- `|| true` in the new test
- only `insert` is fixed - `replace` and the structural paths still write raw `\t`
- reading a tab back gives `\ufffc` instead of `\t`

good to merge after that.

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.test.ts
```diff
@@ -450,6 +450,101 @@ describe('executeTextInsert: setMarks tri-state directives', () => {
...
+      args: { position: 'before', content: { text: '\t' } },
+    } as any;
+
+    const outcome = executeTextInsert(editor, tr as any, target, step, { map: (pos: number) => pos } as any);
+
+    expect(outcome).toEqual({ changed: true });
+    expect(tabCreate).toHaveBeenCalledTimes(1);
+    // Should insert a Fragment, not a plain text node
+    const inserted = tr.insert.mock.calls[0][1];
+    expect(
```
caio-pizzol: the `|| true` kills the check - it makes this always pass no matter what got inserted. want it like this?

```suggestion
    expect(inserted.childCount).toBe(1);
    expect(inserted.firstChild?.type?.name).toBe('tab');
```

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts
```diff
@@ -970,12 +970,37 @@ export function executeTextInsert(
...
+  const tabNodeType = editor.state.schema.nodes?.tab;
+  if (tabNodeType && text.includes('\t') && parentAllowsNode(tr, absPos, tabNodeType)) {
+    const parts = text.split('\t');
+    const nodes: ProseMirrorNode[] = [];
+    for (let i = 0; i < parts.length; i++) {
+      if (parts[i]) {
+        nodes.push(editor.state.schema.text(parts[i], marks));
+      }
+      if (i < parts.length - 1) {
+        nodes.push(tabNodeType.create());
```
caio-pizzol: tab gets no marks but the text on either side does. so for `doc.insert({ target, value: 'a\tb', marks: ['bold'] })` you end up with bold `a`, plain tab, bold `b` on export - formatting snaps across the tab.

```suggestion
        nodes.push(tabNodeType.create(null, null, marks));
```

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts
```diff
@@ -970,12 +970,37 @@ export function executeTextInsert(
...
+      }
+    }
```
caio-pizzol: only `insert` gets this treatment. `executeTextRewrite` (~898-907) and `materializeRun`/`materializeTextRun` in `node-materializer.ts:848,854` still do `schema.text(text, marks)` flat. `plan-wrappers.ts:587` routes `doc.replace` through `text.rewrite`, so today:

```js
doc.insert({ target, value: 'a\tb' });   // real tab node
doc.replace({ target, value: 'a\tb' });  // blows away the tab, writes raw \t
```

probably cleanest to pull this split-into-text-and-tab block into a helper and call it from all three spots.

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/executor.ts
```diff
@@ -970,12 +970,37 @@ export function executeTextInsert(
...
+    const fragment = Fragment.from(nodes);
+    tr.insert(absPos, fragment);
```
caio-pizzol: reading a tab back doesn't match what got written:

```js
doc.insert({ target, value: 'a\tb' });
doc.get_text(range); // 'a\ufffcb' - not 'a\tb'
```

`tab.js` has no `leafText`, so `textBetween` falls back to the placeholder. one line in the tab node spec fixes it across every reader at once:

```js
leafText: () => '\t',
```

without this, reads diverge based on whether the tab came from a fresh insert or opening a docx, which is the model inconsistency you wanted to avoid going editor-side. also fixes the no-op guard in `write-adapter.ts:98`.

andrii-harbour:
@caio-pizzol deal. Agree on the comments. Working on the fixes now

andrii-harbour:
@caio-pizzol made the fixes. A bit more than expected, but also fixed a few minor things along the way

harbournick (approved):
LGTM

PR #2842, opened 2026-04-16
fix: rebuild fragment on geometry change

artem-harbour:
Linear: SD-2446

### DIFF
--- a/packages/layout-engine/painters/dom/src/index.test.ts
+++ b/packages/layout-engine/painters/dom/src/index.test.ts
@@ -3029,7 +3029,7 @@ describe('DomPainter', () => {
     expect(appliedWordSpacing).toBeCloseTo(expectedWordSpacing, 5);
   });
 
-  it('reuses fragment DOM nodes when layout geometry changes', () => {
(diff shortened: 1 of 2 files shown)


packages/layout-engine/painters/dom/src/index.test.ts
```diff
@@ -2927,9 +2927,12 @@ describe('DomPainter', () => {
 
     painter.paint(movedLayout, mount);
     const fragmentAfter = mount.querySelector('.superdoc-fragment') as HTMLElement;
+    const lineAfter = fragmentAfter.querySelector('.superdoc-line') as HTMLElement;
 
-    expect(fragmentAfter).toBe(fragmentBefore);
+    expect(fragmentAfter).not.toBe(fragmentBefore);
```
chittolinag, line 3056: @artem-harbour I think it makes more sense to use `toEqual` here. I know that the old code was already using `toBe`, but `toBe` does a strict comparison using `===`. This means that the objects could be semantically the same, but `toBe` would still result in `false`.

Jest's doc:

Image

chittolinag (approved):
LGTM! Approved, but left a small comment on the test.

harbournick (approved):
LGTM

PR #2850, opened 2026-04-16
SD-2508 - fix: toc not being displayed inside w:sdt

chittolinag:
### Issue

Documents with a TOC inside a `w:sdt` whose `docPartGallery` is not exactly `"Table of Contents"` (e.g. `"Custom Table of Contents"`) have the TOC dropped on render.

The field preprocessor converts TOC field codes (`w:fldChar begin/end`) into a single `sd:tableOfContents` element regardless of the gallery type. This means the PM `documentPartObject` node for any custom TOC gallery ends up with a `tableOfContents` child.

`handleDocumentPartObjectNode` dispatches on `docPartGallery === 'Table of Contents'` for the `processTocChildren` path. For any other gallery string it falls to an else-if loop that only handled `paragraph` children — `tableOfContents` children were silently dropped.

### Proposed solution

Add a `tableOfContents` branch to the else-if loop in `document-part-object.ts` that forwards to the existing `processTocChildren` helper. No changes to the import handler or preprocessing pipeline.

### Side notes

This PR fixes the issue where the TOC is completely dropped and doesn't show up in the document. However, there are still rendering issues with the TOC. This [other PR](https://github.com/superdoc-dev/superdoc/pull/2731) is open and fixes the rendering issue.


packages/layout-engine/pm-adapter/src/sdt/document-part-object.ts
```diff
@@ -75,6 +75,24 @@ export function handleDocumentPartObjectNode(node: PMNode, context: NodeHandlerC
           blocks.push(block);
           recordBlockKind?.(block.kind);
         }
+      } else if (child.type === 'tableOfContents' && Array.isArray(child.content)) {
+        // A nested tableOfContents node (e.g. from a "Custom Table of Contents" SDT where
+        // the TOC field codes were preprocessed into an sd:tableOfContents element)
+        processTocChildren(
+          child.content,
+          { docPartGallery: docPartGallery ?? '', docPartObjectId, tocInstruction, sdtMetadata: docPartSdtMetadata },
```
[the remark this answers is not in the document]
caio-pizzol: confirmed at runtime - child has the field codes, wrapper doesn't. covered in my review.

packages/layout-engine/pm-adapter/src/sdt/document-part-object.ts
```diff
@@ -75,6 +75,33 @@ export function handleDocumentPartObjectNode(node: PMNode, context: NodeHandlerC
           blocks.push(block);
           recordBlockKind?.(block.kind);
         }
+      } else if (child.type === 'tableOfContents' && Array.isArray(child.content)) {
```
caio-pizzol, line 96: your first commit also added a `child.type === 'table'` branch with a test, and the "refactor" commit removed it. the linear ticket title is "drops tables" - was that handled somewhere else, or did it slip out of scope?

packages/layout-engine/pm-adapter/src/sdt/document-part-object.ts
```diff
@@ -75,6 +75,33 @@ export function handleDocumentPartObjectNode(node: PMNode, context: NodeHandlerC
...
+            positions,
+            bookmarks,
+            hyperlinkConfig,
+            enableComments,
+            trackedChangesConfig,
+            converters,
+            converterContext,
+          },
+          { blocks, recordBlockKind },
+        );
```
caio-pizzol: `git rebase origin/main` hits a conflict here - #2872 added `themeColors` and `sectionState` to the existing branch's call. if the merge keeps both, this new branch silently drops them, so theme colors and section breaks won't fire on custom-TOC content. worth mirroring the existing branch's call shape when you resolve.

caio-pizzol (commented):
hey @chittolinag! loaded the customer's "Custom Table of Contents" doc against your branch and the TOC entries render now :)

i pushed a follow-up commit (`56377a1`) with the missing child-instruction handling and 3 unit tests for the new branch.

two things still open, both inline.

harbournick:
@artem-harbour mind doing a review on this one pls?

harbournick:
@chittolina pls resolve merge conflicts

packages/layout-engine/pm-adapter/src/sdt/document-part-object.ts
chittolinag: @caio-pizzol I had misunderstood the ticket back then. My understanding is that the ticket description might point to the wrong direction because the issue is related to TOC, not Tables. The document linked in the ticket has a TOC that is not being rendered, not a table.

packages/layout-engine/pm-adapter/src/sdt/document-part-object.ts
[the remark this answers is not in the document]
chittolinag: thanks @caio-pizzol !

packages/layout-engine/pm-adapter/src/sdt/document-part-object.ts
chittolinag: done

packages/layout-engine/pm-adapter/src/sdt/document-part-object.ts
```diff
@@ -93,6 +93,35 @@ export function handleDocumentPartObjectNode(node: PMNode, context: NodeHandlerC
...
+      } else if (child.type === 'tableOfContents' && Array.isArray(child.content)) {
+        // A nested tableOfContents node (e.g. from a "Custom Table of Contents" SDT where
+        // the TOC field codes were preprocessed into an sd:tableOfContents element).
+        // Word stores the TOC field codes on the child node, not the wrapper SDT - prefer
+        // the child's instruction so per-TOC options aren't lost (mirrors the recursion
+        // inside processTocChildren in toc.ts).
+        const childTocInstruction = getNodeInstruction(child) ?? tocInstruction;
+        processTocChildren(
+          child.content,
+          {
```
artem-harbour: I would suggest moving these objects into separate variables to make the function contract easier to read.

example:
```js
processTocChildren(
  child.content,
  metadata,
  context,
  output,
)
```

chittolinag: good point. just did it. thx @artem-harbour

artem-harbour (approved):
LGTM

artem-harbour:
@caio-pizzol, please check your comments and let us know if this is ready for merge.

packages/layout-engine/pm-adapter/src/sdt/document-part-object.ts
```diff
@@ -93,6 +93,32 @@ export function handleDocumentPartObjectNode(node: PMNode, context: NodeHandlerC
...
+          hyperlinkConfig,
+          enableComments,
+          trackedChangesConfig,
+          themeColors,
+          converters,
+          converterContext,
+          sectionState,
+        };
+        const output = { blocks, recordBlockKind };
+        processTocChildren(child.content, metadata, tocContext, output);
```
caio-pizzol, line 121: the loop now handles `paragraph` and `tableOfContents` children, but `findParagraphsWithSectPr` also walks into nested `documentPartObject`. if an SDT ever holds another SDT inside, those inner paragraphs would be counted but skipped here - same drift this PR is fixing. rare enough to leave; a one-line comment would flag it for next time.

caio-pizzol (approved):
hey @chittolinag! thanks for addressing last round - approving.

left one small thing inline if you want to tidy it up, but nothing blocking.

packages/layout-engine/pm-adapter/src/sdt/document-part-object.ts
caio-pizzol: nvm - very rare to replicate

caio-pizzol: checked the surfaces - neither the UI nor `@packages/document-api` can produce a nested `docPartObject`. content-controls cover `text/group/repeatingSection/...` (no `docPartObj`), and the TOC ops only emit a single wrapper. only reachable by importing a Word-authored DOCX with that nesting. feel free to skip.

PR #2854, opened 2026-04-17
[IT-925] fix(presentation-editor): skip expensive decoration checks on doc changes

mattConnHarbour:


caio-pizzol:
Hey @mattConnHarbour, thanks for putting this together. The search and tracked-change flow is useful, and the app made the behavior easier to understand.

I’m going to close this PR and take a different path.

The fix changes a broad part of the editor, and I don’t think that is the right long-term approach for this workflow. We now have SuperDoc UI / Custom UI, which is a better place to build this kind of experience.

The added app also does not fit `examples/`. It combines several features into one workflow, while `examples/` should stay small and focused on one concept. If we want to show this publicly, we should build it as a dedicated demo.

Thanks again for exploring this. Closing this PR.

PR #2855, opened 2026-04-17
fix: place cursor at clicked pos inside tracked change

artem-harbour:
Linear: SD-2612

Fix caret placement inside tracked-change text on click.

This updates `EditorInputManager` to treat direct clicks on tracked-change spans the same way we already treat direct single-thread comment highlights (SD-2442): do not short-circuit to `setCursorById`, and let normal click-to-position place the caret at the clicked location.

Also fixes repeat-click behavior so subsequent direct clicks inside the same tracked change continue to reposition the caret instead of being ignored by the active-thread repeat-click guard.

Includes unit test updates in `EditorInputManager.activeCommentClick.test.ts` to cover:
- direct tracked-change click falls through to generic caret placement
- repeat direct click on active tracked change still repositions caret


harbournick (approved):
LGTM

PR #2856, opened 2026-04-17
feat: add presetContent for template builder

artem-harbour:
Linear: SD-2548

Added presetContent support to Template Builder field definitions for block fields, wiring it into the existing insertStructuredContentBlock flow (no new insertion path).

Block fields now insert preset HTML/JSON when provided, while block fields without presets keep the existing text fallback; inline fields continue to ignore presetContent.

Included insertion-flow tests, updated package/solution docs, and added demo examples (Sample Table with visible borders and Sample List).


caio-pizzol:
@artem-harbour lgtm.

one out-of-diff thought: `handleSelectExisting` (around `packages/template-builder/src/index.tsx:589`) doesn't forward `presetContent` when re-linking an existing field. intentional? worth a one-line doc note either way.

caio-pizzol (approved):

PR #2869, opened 2026-04-20
feat(layout-engine): balance columns at continuous section breaks (SD-2452)

tupizz:
Comparison Results PDF
[SD-2452-page-by-page.pdf](https://github.com/user-attachments/files/26906256/SD-2452-page-by-page.pdf)

## Summary

Implements ECMA-376 §17.18.77 column balancing for multi-column sections. Word produces a minimum-height balanced layout at the end of a continuous (and empirically, next-page) multi-column section; SuperDoc was either leaving content stacked in the first column or, in some layouts, producing overlapping fragments.

Linear: [SD-2452](https://linear.app/superdocworkspace/issue/SD-2452)

## What changed

1. **`layoutDocument` builds a block → section map** by walking blocks in document order and tracking the current section from the most recent `sectionBreak` (pm-adapter only stamps `attrs.sectionIndex` on sectionBreak blocks, not on content paragraphs).
2. **New `balanceSectionOnPage` helper** performs section-scoped balancing with its own fragment-level positioning (no Y-grouping). Fragments are ordered by `(x, y)` and each is treated as its own block. The previous `balancePageColumns` grouped fragments by Y into rows, which collapsed fragments from different source columns at the same Y and produced overlap.
3. **`calculateBalancedColumnHeight` is a proper binary search** for the minimum `H` such that greedy left-to-right fill places every block with every column ≤ `H`. Matches Word's left-heavy packing preference (e.g. 7 blocks / 3 cols → 3+3+1, not 2+2+3).
4. **Mid-page hook at `forceMidPageRegion`** balances the ending section on the current page before starting the new region, and collapses both cursors to `balanceResult.maxY` so the next region begins just below the balanced columns. Sections handled mid-page are tracked in `alreadyBalancedSections` so the post-layout pass doesn't double-balance.
5. **Per-section post-layout loop** replaces the prior "last page of document" heuristic — each multi-column section's last page is balanced, skipping sections already handled mid-page.

## Results (Word vs SuperDoc)

| Test | Scenario | Word | SuperDoc before | SuperDoc after |
|---|---|---|---|---|
| 1 | 6 equal paragraphs, 2 cols (continuous break) | 3+3 | 6+0 — not balanced | 3+3 — exact match |
| 2 | 5 paragraphs with unequal heights, 2 cols | 2+3 | 5+0 — not balanced | 2+3 — exact match |
| 3 | 7 equal paragraphs, 3 cols | 3+3+1 | 7+0+0 — not balanced | 3+3+1 — exact match |
| 4 | 13 paragraphs with multi-line bodies, 2 cols | 7+6 | Overlapping fragments | 7+6 — exact match |
| 5 | Continuous + next-page sections (5+5) | 3+2 / 3+2 | Not balanced | 3+2 / 3+2 — exact match |

Side-by-side PDF comparison available locally at `/tmp/sd-2452-fixtures/SD-2452-comparison.pdf` (generated via new `compare-word-vs-superdoc` skill).

## Test plan

- [x] 614 \`@superdoc/layout-engine\` tests pass (11 new for SD-2452)
- [x] 1,737 \`@superdoc/pm-adapter\` tests pass
- [x] 11,375 \`super-editor\` tests pass
- [x] 0 overlap regressions across local corpus (14 docs — none activate the balancing code path, fix is scope-gated to sections with \`count > 1\`)
- [x] Visual validation against Microsoft Word for all 5 fixtures
- [x] Browser sanity: scroll stable, zoom stable, no fragment overlaps
- [ ] \`pnpm test:layout\` against production reference (blocked on wrangler re-auth locally — CI will run this)
- [ ] Upload fixtures to R2 corpus for visual regression coverage

## Demo tests

CleanShot 2026-04-20 at 15 11 27@2x

CleanShot 2026-04-20 at 15 11 54@2x

CleanShot 2026-04-20 at 15 12 05@2x

CleanShot 2026-04-20 at 15 12 16@2x

CleanShot 2026-04-20 at 15 12 29@2x
CleanShot 2026-04-20 at 15 12 39@2x

## Fixtures

- \`spec-test-1.docx\` — Basic 2-column balance
- \`spec-test-2.docx\` — Unequal paragraph heights
- \`spec-test-3.docx\` — Three-column balance
- \`spec-test-4.docx\` — Long content / overlap scenario
- \`spec-test-5.docx\` — Continuous + next-page break combo

Plan is to upload these to the R2 corpus after the PR lands.


harbournick:
@tupizz please double check layout testing and the below. I see some documents that have definitely regressed.
Also double check these:
 1. **Use each section’s own page metrics when rebalancing**

     The post-layout balancing pass appears to rebalance earlier multi-column sections using the final active margins/page
  size from a later section. If a two-column section with one set of margins is followed by another section with different
  margins or page size, fragments from the earlier page can be rewritten to the later section’s `x`/width values.

     Balancing should use the page/section metrics for the section being rebalanced, not the final active layout state.

  2. **Keep balancing document-wide column layouts**
     When callers use `LayoutOptions.columns` without section break metadata, `sectionColumnsMap` is empty, so the
  section-based balancing loop never runs. That leaves the final page stacked in column 0.
     The previous `activeColumns.count > 1` path handled document-wide multi-column layouts, so this needs a fallback for
  the active/options column config when there are no section-scoped entries.

  3. **Preserve blank paragraph height when balancing columns**
     A normal blank paragraph can measure as a line with `width === 0`, but it should still consume line height. The
  current logic treats that like a zero-height `sectPr` marker, so the column cursor does not advance and the next
  paragraph can overlap the blank line.

     This should be gated on actual section-property marker metadata rather than line width alone.

  4. **Write table row boundaries using the fragment metadata shape**
When a dominant table is split for balancing, row boundaries are stored using the renderer’s serialized keys: `i`,
  `h`, `min`, `r`.
 `TableFragmentMetadata.rowBoundaries` expects `index`, `height`, `minHeight`, and `resizable`. Because the DOM
  renderer later serializes those contract fields, split table fragments can produce row-boundary data with undefined
  values, breaking row resize handles.

harbournick:
@luccas-harbour can you pls work with Tadeu on getting this one to the finish line next week? Be extra careful around layout testing pls make sure no regressions!

luccas-harbour:
hey @tupizz! nothing to add right now apart from Nick's comment. I'll have another look once those are addressed.

also, note that the build is failing, which caused the behavior tests to fail running.

thanks!

packages/layout-engine/layout-engine/src/index.ts
```diff
@@ -1904,9 +1976,74 @@ export function layoutDocument(blocks: FlowBlock[], measures: Measure[], options
...
+            sectPrMarkerBlockIds,
+          });
+          if (balanceResult) {
+            // Collapse both cursors to the balanced section bottom so the new
+            // region starts there, not below an unbalanced tallest column.
+            state.cursorY = balanceResult.maxY;
+            state.maxCursorY = balanceResult.maxY;
+            alreadyBalancedSections.add(endingSectionIndex!);
+          }
+        } else if (columnIndexBefore >= newColumns.count) {
```
luccas-harbour: `willBalance` is a coarse approval — `balanceSectionOnPage` can still bail later (unequal widths, zero remaining height, section too small) even after `willBalance` returned true. when that happens, the "force a new page if `columnIndexBefore >= newColumns.count`" fallback doesn't fire — it only runs when `willBalance` was false. so the next region starts on the same page from a stale column index and overwrites the previous section's column 1 content. should the fallback trigger whenever `balanceResult === null && columnIndexBefore >= newColumns.count`, regardless of what `willBalance` returned?

packages/layout-engine/layout-engine/src/column-balancing.ts
```diff
@@ -746,46 +715,396 @@ export function balancePageColumns(
...
+  table.continuesOnNext = true;
+  if (firstMetadata) table.metadata = firstMetadata;
+
+  const secondHalf: BalancingFragment = {
+    ...table,
+    fromRow: splitRow,
+    toRow,
+    height: secondHalfHeight,
+    continuesFromPrev: true,
+    continuesOnNext: table.continuesOnNext ? false : (table.continuesOnNext ?? false),
```
luccas-harbour: by the time this line reads `table.continuesOnNext`, line 1088 has already set it to `true`, so the ternary always picks `false` and line 1102 confirms it — the second half never inherits the original `continuesOnNext` despite the comment saying it does. capture the original into a local before mutating, e.g. `const originalContinuesOnNext = table.continuesOnNext ?? false;` ahead of line 1088, then assign that to `secondHalf` and drop the redundant lines 1100-1102. probably unreachable today since `balanceSectionOnPage` runs at section-end, but the code and the comment disagree.

packages/layout-engine/layout-engine/src/column-balancing.ts
```diff
@@ -746,46 +715,396 @@ export function balancePageColumns(
...
+
+  const measure = measureMap.get(table.blockId) as TableMeasureLike | undefined;
+  if (!measure || measure.kind !== 'table' || !Array.isArray(measure.rows)) return;
+
+  const totalSectionHeight = sectionFragments.reduce(
+    (sum, f) => sum + getBalancingHeight(f, measureMap, sectPrMarkerBlockIds),
+    0,
+  );
+  if (totalSectionHeight <= 0) return;
+  const target = totalSectionHeight / columnCount;
```
luccas-harbour: split target is `totalSectionHeight / columnCount`, which ignores fragments already placed in column 0 before the table. with a 100px paragraph + 300px table in 2 cols, target=200, so the table splits at row=200 → halves of 200/100. column 0 ends up with 100+200=300 and column 1 with 100, and the binary search settles on max-col=300 instead of the achievable 200 (split at row=100). should we subtract preceding atomic-block heights from `target` before walking row heights?

packages/layout-engine/layout-engine/src/column-balancing.ts
```diff
@@ -746,46 +715,396 @@ export function balancePageColumns(
...
+  const contentBlocks: BalancingBlock[] = ordered.map((f, i) => ({
+    blockId: `${f.blockId}#${i}`,
+    measuredHeight: getBalancingHeight(f, args.measureMap, args.sectPrMarkerBlockIds),
+    canBreak: false,
+    keepWithNext: false,
+    keepTogether: true,
+  }));
+
+  if (
+    shouldSkipBalancing({
```
luccas-harbour, line 771: `splitDominantTableAtRowBoundary` (line 860) mutates `fragments` and `sectionFragments` in place before this `shouldSkipBalancing` runs. if the skip predicate fires after a split, we return `null` with the page already carrying an overlapping half table + a duplicated second-half fragment. it's hard to actually trigger today (the post-split section almost always clears the skip thresholds), but the ordering is fragile — could we either run the skip check before mutating, or capture a rollback closure?

packages/layout-engine/layout-engine/src/column-balancing.ts
```diff
@@ -746,46 +715,396 @@ export function balancePageColumns(
...
+ * Callers can use the returned Y to update paginator cursors so subsequent content
+ * starts just below the balanced section rather than below an unbalanced maxCursorY.
+ *
+ * Guards (skip balancing when):
+ *   - Section has <= 1 column (nothing to balance)
+ *   - Section contains an explicit column break (author intent wins)
+ *   - Section uses unequal column widths (Word doesn't rebalance these)
+ *   - No fragments on this page belong to the section
+ */
+export function balanceSectionOnPage(args: BalanceSectionOnPageArgs): { maxY: number } | null {
```
luccas-harbour, line 676: now that `balanceSectionOnPage` is the only entry point used by `index.ts`, `balancePageColumns` (line 650 above) no longer has any production callers. can it be removed?

packages/layout-engine/pm-adapter/src/sections/breaks.ts
```diff
@@ -201,16 +201,52 @@ interface SectionStateMutable {
...
+ * body children preceding that tag — paragraphs, tables, top-level drawings —
+ * belong to the section that ENDS at the tag. This helper fires BEFORE any
+ * such node so the appropriate section config is active by the time the node
+ * is laid out.
+ *
+ * Calling it from the main dispatch loop covers every top-level node type —
+ * present and future — with no per-handler opt-in. Paragraph-index emission
+ * still handles transitions inside SDT child content.
+ */
+export function maybeEmitNextSectionBreakForNode(args: {
```
luccas-harbour, line 229: `maybeEmitNextSectionBreakForNode` and `emitPendingSectionBreakForParagraph` share most of their body — only the position field and the push mechanism really differ. perhaps it might be worth it extracting the logic to a shared helper?

luccas-harbour (commented):
hey! I saw you addressed Nick's comments so I had another look and found a few things. the biggest one is that `willBalance` doesn't match `balanceSectionOnPage`'s actual skip conditions, so the page-break fallback can be skipped from an invalid column state. there's also a `continuesOnNext` mutation in the table split that always collapses to false, and the split target ignores content already placed in column 0.

I am also running into this issue when trying to build the project:

```
packages/super-editor/src/editors/v1/document-api-adapters/helpers/sections-resolver.ts:168:3 - error TS2739: Type '{ sectionIndex: number; startParagraphIndex: number; endParagraphIndex: number; sectPr: SectPrElement; margins: null; pageSize: null; orientation: null; columns: null; type: SectionType.CONTINUOUS; ... 4 more ...; vAlign: undefined; }' is missing the following properties from type 'SectionRange': startNodeIndex, endNodeIndex

168   return {
```

let me know what you think!

packages/layout-engine/layout-engine/src/index.ts
tupizz: good catch. fixed in 50dc1f7. moved the page-break fallback out of the willBalance else branch so it now triggers whenever balanceSectionOnPage returns null, regardless of the willBalance approval. the late skip conditions inside balanceSectionOnPage no longer leave the new region with a stale column index.

packages/layout-engine/layout-engine/src/column-balancing.ts
tupizz: fixed in 50dc1f7. captured `originalContinuesOnNext` before mutating `table.continuesOnNext = true`, then assigned that to the second half. dropped the redundant overwrite below.

packages/layout-engine/layout-engine/src/column-balancing.ts
tupizz: fixed in 50dc1f7. caller now sums the heights of fragments preceding the table in document order and passes that as `precedingHeight`. split target becomes `max(1, totalSectionHeight / columnCount - precedingHeight)`, so your 100px+300px in 2 cols example splits at row=100 instead of row=200.

packages/layout-engine/layout-engine/src/column-balancing.ts
tupizz: went with the rollback closure path in 50dc1f7. running the skip check before the split would defeat the split's purpose (a single dominant unbreakable table always trips `length===1 && !canBreak` skip pre-split). `splitDominantTableAtRowBoundary` now returns `{applied, rollback}` and the post-split `shouldSkipBalancing` invokes rollback before returning null, restoring `toRow`/`height`/`continuesOnNext`/`metadata` and removing the inserted second half from both fragment arrays.

packages/layout-engine/layout-engine/src/column-balancing.ts
tupizz: yes — confirmed unused. removed in 50dc1f7 along with its test block. kept the shared `createMeasure` helper inlined in the balanceSectionOnPage tests.

packages/layout-engine/pm-adapter/src/sections/breaks.ts
tupizz: agreed they share most of the body, but holding off on the extraction in this PR. the two callers diverge on position-field shape and the push mechanism (one builds blocks for a node-pos pair, the other emits at a paragraph boundary), and the helper would need both differences as parameters. happy to do it as a follow-up if you'd prefer — out of scope for the column-balancing fix here.

tupizz:
hey @luccas-harbour, addressed all six in https://github.com/superdoc-dev/superdoc/commit/50dc1f7aa130a96cd4c9de1a337fec33e9ad818a:

willBalance / fallback mismatch → page-break fallback now triggers whenever balanceSectionOnPage returns null
table split continuesOnNext always false → captured original before mutating
split target ignores preceding column-0 height → precedingHeight subtracted from per-column target
skip-after-mutate ordering → splitDominantTableAtRowBoundary returns a rollback closure, invoked when the post-split skip fires
balancePageColumns dead code → removed (function + test block, helpers inlined)
maybeEmitNextSectionBreakForNode / emitPendingSectionBreakForParagraph extraction → holding off, replied in thread
build error in sections-resolver.ts was from the main merge adding startNodeIndex/endNodeIndex to SectionRange. fixed.

644 layout-engine unit tests pass. super-editor build is clean.

luccas-harbour:
hey @tupizz!
codex found this one:
```
[P2] Restrict continuous balancing to the affected section — /Users/luccascorrea/Dev/superdoc-wt/review-tadeu-sd-2452-feature-implement-column-balancing-for-continuous-section/packages/layout-engine/layout-engine/src/index.ts:2831-2832
  When a document contains any explicit continuous section break, this condition allows balancing for every multi-column section whose own type was omitted, even if that section is unrelated to the continuous break. For example, a later single-page two-column body section with omitted <w:type> will now be balanced just because an earlier section was explicit-continuous, despite the existing gate intending to skip omitted-type single-page sections like sd-1655. The check should be tied to the section ended by that continuous break rather than a document-wide flag.
```

can you have a look?

luccas-harbour (approved):
LGTM

PR #2873, opened 2026-04-20
SD-2527 - feat: implement new types of ordered lists

chittolinag:
Introducing new types of ordered lists:

1. number with bracket
2. uppercase roman
3. uppercase alpha
4. lowercase roman
5. lowercase alpha
6. alpha with bracket

Even though we supported rendering the styles when opening documents from Word, the user couldn't start a new list using these types.

This PR introduces a dropdown that allows the user to do so.

> [!NOTE]
> This PR depends on https://github.com/superdoc-dev/superdoc/pull/2857. Let's merge the other one first. 

Screenshot 2026-04-20 at 15 59 36


packages/super-editor/src/editors/v1/components/toolbar/BulletStyleButtons.vue
```diff
@@ -0,0 +1,136 @@
...
+      v-for="(button, index) in bulletButtons"
+      :key="button.key"
+      class="button-icon"
+      :class="{ selected: props.selectedStyle === button.key }"
+      @click="select(button.key)"
+      v-html="button.icon"
+      role="menuitem"
+      :aria-label="button.ariaLabel"
+      ref="buttonRefs"
+      @keydown.prevent="(event) => handleKeyDown(event, index)"
```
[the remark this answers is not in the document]
chittolinag: this is not a new thing - other buttons behave the same way. ignoring this for now. if any changes are necessary here, we'd need to tackle it in other places as well for it to make sense.

caio-pizzol, line 90: `.prevent` blocks every key, so Tab, Escape, and Space all stop working inside the menu. same fix in `NumberedStyleButtons.vue:86`.

```suggestion
      @keydown="(event) => handleKeyDown(event, index)"
```

then call `event.preventDefault()` inside the cases the switch handles.

packages/super-editor/src/editors/v1/core/commands/toggleList.js
```diff
@@ -26,11 +26,28 @@ function getParagraphListKind(node, editor) {
...
+    if (!bulletStyle) return true;
+    const markerText = node.attrs.listRendering?.markerText;
+    return markerTextToBulletStyle(markerText) === bulletStyle;
+  }
+  if (listType === 'orderedList') {
+    if (kind !== 'ordered') return false;
+    if (!orderedStyle) return true;
+    const { numberingType, markerText } = node.attrs.listRendering ?? {};
+    return numberingInfoToOrderedStyle(numberingType, markerText) === orderedStyle;
+  }
```
caio-pizzol, line 50: picking a new style on one item of a list splits it out into its own list. the rest of the list stays put but restarts at 1 (item 2 was `2.`, now it's `1.`). Word changes the whole list instead. is the split intentional?

packages/super-editor/src/editors/v1/core/parts/adapters/numbering-transforms.ts
```diff
@@ -30,8 +31,26 @@ interface GenerateOptions {
...
+
+const ORDERED_LIST_STYLES: Record<string, { fmt: string; text: string }> = {
+  decimal: { fmt: 'decimal', text: '%1.' },
+  'decimal-paren': { fmt: 'decimal', text: '%1)' },
+  'upper-roman': { fmt: 'upperRoman', text: '%1.' },
+  'lower-roman': { fmt: 'lowerRoman', text: '%1.' },
+  'upper-alpha': { fmt: 'upperLetter', text: '%1.' },
+  'lower-alpha': { fmt: 'lowerLetter', text: '%1.' },
+  'lower-alpha-paren': { fmt: 'lowerLetter', text: '%1)' },
+};
```
caio-pizzol, line 66: `A) B) C)` is missing - Word has it. same gap in `list-numbering-helpers.js:49`.

```suggestion
const ORDERED_LIST_STYLES: Record<string, { fmt: string; text: string }> = {
  decimal: { fmt: 'decimal', text: '%1.' },
  'decimal-paren': { fmt: 'decimal', text: '%1)' },
  'upper-roman': { fmt: 'upperRoman', text: '%1.' },
  'lower-roman': { fmt: 'lowerRoman', text: '%1.' },
  'upper-alpha': { fmt: 'upperLetter', text: '%1.' },
  'upper-alpha-paren': { fmt: 'upperLetter', text: '%1)' },
  'lower-alpha': { fmt: 'lowerLetter', text: '%1.' },
  'lower-alpha-paren': { fmt: 'lowerLetter', text: '%1)' },
};
```

also needs the type in `paragraph-commands.ts` and a button in `NumberedStyleButtons.vue`.

caio-pizzol (changes requested):
hey @chittolinag! nicely scoped :)

i pushed a follow-up (c2685a38c) with tests for all 10 styles - unit + behavior + visual, all passing.

three things inline. one more small thing: the bullet menu and the numbered menu return different shapes when you read their state - bullets give you the character (`•`), numbered gives you the name (`decimal-paren`). worth lining up.

needs work.

packages/super-editor/src/editors/v1/components/toolbar/BulletStyleButtons.vue
chittolinag: same comment as here: https://github.com/superdoc-dev/superdoc/pull/2857#discussion_r3162088257

packages/super-editor/src/editors/v1/core/commands/toggleList.js
chittolinag: Hmmm that's a good point. I believe we can definitely mimic Word's behavior, but I do see that the current implementation (ie, before this PR), changes the style only on the item that you're currently positioned at. I'll confirm the expected behavior here with Missy (and whether we want to include it on this PR or not).

packages/super-editor/src/editors/v1/core/parts/adapters/numbering-transforms.ts
chittolinag: interesting - that (upper alpha) was not as a requirement in the ticket but definitely a good point. I think we should add it, let me confirm.

luccas-harbour:
hey @chittolina! I gave this one a quick pass and found these:
1. When the cursor is at a line that contains a numbered list, the toolbar used to show the numbered list icon with a darker background. I see that this is not the case anymore but it is what Microsoft Word does, might be worth having a look?
  - Perhaps you could use the segmented button you implemented for bullet lists here too? 
2. I noticed that if I create a numbered list with 2 items and then, while the cursor is in the second item, I deselect the numbered list toolbar button, it removes the numbering from both items and not just from the one where the cursor was (which is Word's behavior)

packages/super-editor/src/editors/v1/core/commands/toggleList.js
chittolinag: after talking to Missy, we decided to do it. it's already here in this PR.

chittolinag:
> hey @chittolina! I gave this one a quick pass and found these:
> 
> 1. When the cursor is at a line that contains a numbered list, the toolbar used to show the numbered list icon with a darker background. I see that this is not the case anymore but it is what Microsoft Word does, might be worth having a look?
> 
> * Perhaps you could use the segmented button you implemented for bullet lists here too?
> 
> 2. I noticed that if I create a numbered list with 2 items and then, while the cursor is in the second item, I deselect the numbered list toolbar button, it removes the numbering from both items and not just from the one where the cursor was (which is Word's behavior)



> hey @chittolina! I gave this one a quick pass and found these:
> 
> 1. When the cursor is at a line that contains a numbered list, the toolbar used to show the numbered list icon with a darker background. I see that this is not the case anymore but it is what Microsoft Word does, might be worth having a look?
> 
> * Perhaps you could use the segmented button you implemented for bullet lists here too?
> 
> 2. I noticed that if I create a numbered list with 2 items and then, while the cursor is in the second item, I deselect the numbered list toolbar button, it removes the numbering from both items and not just from the one where the cursor was (which is Word's behavior)

@luccas-harbour , thanks for the feedback. the 1st issue if fixed. the 2nd, I spoke to Missy and we decided to follow this approach:

1. if the user has an empty selection and changes the list style (or toggles the list off by clicking the left-side action in the segmented button), it will be applied to the entire list (items at the same level).
2. if the user has a **non empty** selection and executes the same actions as above, it will be applied inline, ie, only to the selected item.

luccas-harbour:
that sounds good to me @chittolinag . I had another look and seem to have found another issue, here are the reproduction steps:
1. Add 2 paragraphs: "Parent" and "Child"
2. Select both of them and then click the ordered list button in the toolbar. They will become 2 items with numbers "1." and "2."
3. Put the cursor in the second line and press tab. The child item will become "a."
4. Select the text "Child" and then change the list style to roman numerals (I., II., III., etc.)

The number format does not change. It only changes if the selection is empty. Is that expected?

chittolinag:
> that sounds good to me @chittolinag . I had another look and seem to have found another issue, here are the reproduction steps:
> 
> 1. Add 2 paragraphs: "Parent" and "Child"
> 2. Select both of them and then click the ordered list button in the toolbar. They will become 2 items with numbers "1." and "2."
> 3. Put the cursor in the second line and press tab. The child item will become "a."
> 4. Select the text "Child" and then change the list style to roman numerals (I., II., III., etc.)
> 
> The number format does not change. It only changes if the selection is empty. Is that expected?

you're right! it should apply the inline style to that child, like it does in non-nested items. thanks! i'll check it out now.

packages/super-editor/src/editors/v1/components/toolbar/NumberedStyleButtons.vue
luccas-harbour, line 1: This file is very similar to `BulletStyleButtons.vue`. Maybe it would be worth extracting a `StyleButtonsList.vue` that takes `buttons`, `iconSize`, `selectedStyle` and emits `select`?

packages/super-editor/src/headless-toolbar/toolbar-registry.ts
```diff
@@ -124,7 +124,7 @@ export const createToolbarRegistry = (): Partial<Record<PublicToolbarItemId, Bui
     },
     'numbered-list': {
       id: 'numbered-list',
-      directCommandName: 'toggleOrderedList',
+      directCommandName: 'toggleOrderedListStyle',
```
luccas-harbour, line 130: Could this cause backwards incompatibility issues? If someone implemented a custom toolbar that toggles list, would this break their implementation?

chittolinag: hmm yeah I think you're correct. I've just pushed a fix where we take that into account!

packages/super-editor/src/editors/v1/components/toolbar/NumberedStyleButtons.vue
chittolinag: good idea!

luccas-harbour (approved):
LGTM

packages/super-editor/src/headless-toolbar/helpers/paragraph.ts
```diff
@@ -122,10 +124,12 @@ export const createListStateDeriver =
...
-    return {
-      active: isActive,
-      disabled: false,
-    };
+    const activeNumberingFmt = isActive ? (paragraphNode?.attrs?.listRendering?.numberingType ?? null) : null;
+    const activeMarkerText = isActive ? (paragraphNode?.attrs?.listRendering?.markerText ?? null) : null;
+    const orderedStyleValue = (
+      activeNumberingFmt && activeMarkerText ? numberingInfoToOrderedStyle(activeNumberingFmt, activeMarkerText) : null
+    ) as OrderedListStyle | null;
+    return { active: isActive, disabled: false, value: orderedStyleValue };
```
caio-pizzol, line 132: following up on a round-1 note that didn't get a reply: this branch returns the style name (`decimal-paren`), but the bullet branch above (L122-125) returns the character (`•`). anyone reading toolbar state has to handle them differently. can the bullet branch run the marker through `markerTextToBulletStyle` so both return a style name?

packages/super-editor/src/editors/v1/components/toolbar/StyleButtonsList.vue
```diff
@@ -51,7 +58,7 @@ const handleKeyDown = (e, index) => {
       moveToNextButton(index);
       break;
     case 'Enter':
-      select(bulletButtons[index].key);
+      select(props.buttons[index].key);
       break;
     default:
```
caio-pizzol, line 63: this handler only covers ArrowLeft/Right/Enter, but the binding at L90 uses `@keydown.prevent` - so tab, escape, space, and every other key get swallowed in the menu. same `.prevent` issue from round 1 - drop the modifier and call `event.preventDefault()` inside the cases the switch handles.

packages/super-editor/src/editors/v1/core/parts/adapters/numbering-transforms.ts
```diff
@@ -409,5 +535,149 @@ export function setLvlRestartOnAbstract(
...
+  // identity. Without this, Word treats matching `w:nsid` values as the same
+  // logical list and can collapse the restyled clone back onto the source list
+  // when the document is re-opened (i.e. style changes silently revert).
+  refreshAbstractIdentity(newAbstractDef);
+  numbering.abstracts[newAbstractId] = newAbstractDef;
+
+  setLvlStyleOnAbstract(numbering, newAbstractId, ilvl, options);
+
+  const newNumId = getNextId(numbering.definitions);
+  numbering.definitions[newNumId] = buildNumDef(newNumId, newAbstractId);
```
caio-pizzol, line 677: restyling a list that started at a custom number (e.g. "start at 5") loses the start value - the new list snaps back to 1. run `pnpm --filter @superdoc/super-editor exec vitest run numbering-transforms.review-repro` to see it. `createNumDefinition` at L479 already has a `copyOverridesFrom` option that fixes this - can we use it here?

packages/super-editor/src/editors/v1/core/parts/adapters/numbering-transforms.ts
```diff
@@ -409,5 +535,149 @@ export function setLvlRestartOnAbstract(
...
+ * the source num to the new num via `setNodeMarkup`, so reversing the markup steps
+ * naturally reverts the style change — the source abstract is never touched.
+ */
+export function cloneListDefinitionWithLevelStyle(
+  numbering: NumberingModel,
+  sourceNumId: number,
+  ilvl: number,
+  options: { bulletStyle?: 'disc' | 'circle' | 'square' | null; orderedStyle?: OrderedListStyle | null },
+): { newNumId: number; newAbstractId: number } | null {
+  const sourceNumDef = numbering.definitions[sourceNumId];
```
caio-pizzol, line 653: picking a new style on a list that uses one of Word's built-in list styles (e.g. "List Bullet") does nothing - no error, no change. second test in d537dcc09 reproduces it. `listImporter.js:138-143` already follows this kind of link via `getListDefinitionDetails` - same shape would work here.

caio-pizzol (changes requested):
hey @chittolinag! thanks for the round-2 push - the new list path and paren styles look good :)

i pushed a commit (d537dcc09) with two failing tests that show two of the bugs.

three things to fix inline.

needs work.

packages/super-editor/src/editors/v1/components/toolbar/StyleButtonsList.vue
chittolinag: Just to confirm, I replied to a similar comment here: https://github.com/superdoc-dev/superdoc/pull/2857#discussion_r3162088257

packages/super-editor/src/editors/v1/core/parts/adapters/numbering-transforms.ts
chittolinag: @caio-pizzol tried to reproduce on a real document (as a user would do), but no luck. I'll DM you so we can get on a call.

packages/super-editor/src/editors/v1/core/parts/adapters/numbering-transforms.ts
chittolinag: couldn't reproduce it either.

packages/super-editor/src/editors/v1/components/toolbar/BulletStyleButtons.vue
caio-pizzol: agreed on not addressing this now

packages/super-editor/src/editors/v1/components/toolbar/StyleButtonsList.vue
caio-pizzol: ok

packages/super-editor/src/editors/v1/core/parts/adapters/numbering-transforms.ts
caio-pizzol: image

SuperDoc won't fallback if rendering symbol is not supported

caio-pizzol (approved):
lgtm

Gabriel will address small tweaks separately

PR #2878, opened 2026-04-21
feat(editor): add internal object like drag and drop functionality

palmer-cl:
- Added internal drag-and-drop support for structured content and existing rendered images
- Added post-paint interaction layers and editor styling so rendered SDTs and images become draggable in the DOM.
- Creates a shared SDT identifier module
- Expanded automated coverage for the new behavior.


packages/super-editor/src/editors/v1/core/presentation-editor/dom/ImageInteractionLayer.ts
```diff
@@ -0,0 +1,103 @@
...
+  const seen = new Set<HTMLElement>();
+
+  const add = (element: HTMLElement | null | undefined) => {
+    if (!element || seen.has(element)) return;
+    seen.add(element);
+    roots.push(element);
+  };
+
+  for (const fragment of Array.from(container.querySelectorAll<HTMLElement>(`.${DOM_CLASS_NAMES.IMAGE_FRAGMENT}`))) {
+    if (fragment.querySelector?.(`[data-image-metadata]`) == null) continue;
```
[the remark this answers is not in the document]
palmer-cl: Updated to match renderer

packages/super-editor/src/editors/v1/core/presentation-editor/dom/StructuredContentInteractionLayer.ts
```diff
@@ -0,0 +1,64 @@
...
+    const labels = Array.from(
+      this.#container.querySelectorAll<HTMLElement>(`${BLOCK_LABEL_SELECTOR}, ${INLINE_LABEL_SELECTOR}`),
+    );
+    for (const label of labels) {
+      if (label.dataset[INTERACTION_EPOCH_KEY] === String(layoutEpoch)) continue;
+
+      const sdtElement = label.closest(
+        `.${DOM_CLASS_NAMES.BLOCK_SDT}, .${DOM_CLASS_NAMES.INLINE_SDT_WRAPPER}`,
+      ) as HTMLElement | null;
+      if (!sdtElement?.dataset.sdtId || !sdtElement.dataset.pmStart || !sdtElement.dataset.pmEnd) continue;
```
caio-pizzol, line 26: block SDTs wrapping a table can't be dragged. the painter only writes `data-pm-start`/`data-pm-end` on paragraph wrappers, so the table wrapper has the class and `data-sdt-id` but no position range, and this guard then skips it. repro: `pnpm --filter @superdoc-testing/behavior exec playwright test --grep "table-wrapped"` - label comes back with `draggable=null`. fix: write the range on the table wrapper too (in `painters/dom/src/table/renderTableFragment.ts`), or have this guard mark by id when the range is missing.

packages/super-editor/src/editors/v1/core/presentation-editor/dom/StructuredContentInteractionLayer.ts
```diff
@@ -0,0 +1,64 @@
...
+
+      const scope =
+        sdtElement.dataset.sdtScope ?? (sdtElement.classList.contains(DOM_CLASS_NAMES.BLOCK_SDT) ? 'block' : 'inline');
+      const labelText = label.textContent?.trim() || 'Structured content';
+
+      label.dataset[INTERACTION_EPOCH_KEY] = String(layoutEpoch);
+      label.draggable = true;
```
caio-pizzol, line 39: locked SDTs can still be dragged out. `lockMode` is read into the label data but never used - every label gets `draggable=true`, and the drop path at `DragDropManager.ts:771-800` doesn't check it either. so a `sdtLocked` content control can be deleted by dragging it. fix: skip non-`unlocked` SDTs here, or reject the drop when payload `lockMode !== 'unlocked'`.

packages/super-editor/src/editors/v1/core/presentation-editor/input/DragDropManager.ts
```diff
@@ -671,18 +749,54 @@ export class DragDropManager {
...
 
-    tr.insert(mappedTarget, sourceNode);
-    tr.setMeta('uiEvent', 'drop');
-    view.dispatch(tr);
+    const activeEditor = this.#deps.getActiveEditor();
+    const { state, view } = activeEditor;
+    if (!state || !view) return;
+
+    const dataTransferPayload = parseInternalObjectDragPayload(event);
+    const payload = dataTransferPayload ?? this.#activeInternalObjectPayload;
```
caio-pizzol: `dataTransferPayload ?? this.#activeInternalObjectPayload` accepts a foreign payload over the local one. so two SuperDoc editors on the same page can drop into each other using the wrong doc's positions. flip the order to prefer the local payload, or require it to be set?

caio-pizzol (changes requested):
hey @palmer-cl! the layered design reads cleanly :)

i pushed `6b0358574` with seven tests. five fail as acceptance criteria, two pass and show the prime helper in the SDT behavior tests is redundant - production already marks the label draggable. worth dragging the label directly instead of priming the container.

three things inline. needs work.

packages/super-editor/src/editors/v1/core/presentation-editor/dom/StructuredContentInteractionLayer.ts
palmer-cl: Updated! d209f39727a6c732999832f7e5199bd38ce3a8e3

packages/super-editor/src/editors/v1/core/presentation-editor/dom/StructuredContentInteractionLayer.ts
palmer-cl: TIL about locked mode! Updated in d209f39727a6c732999832f7e5199bd38ce3a8e3

caio-pizzol:
@palmer-cl re-request review once ready for a new round - thanks!

packages/super-editor/src/editors/v1/core/presentation-editor/input/DragDropManager.ts
palmer-cl: went the require route in latest.

palmer-cl:
@caio-pizzol this should be good now. Tests passing. Thanks for the review.

caio-pizzol (approved):
hey @palmer-cl! thanks for addressing last round.

lgtm.

PR #2880, opened 2026-04-21
feat(math): expose equation font via --sd-math-font-family (SD-2634)

caio-pizzol:
Consumer CSS couldn't override the hardcoded `font-family: "Cambria Math", math` on rendered `<math>` elements without `!important`. Swaps the inline style to `var(--sd-math-font-family, "Cambria Math", math)` and registers the token with the existing Cambria Math default in `variables.css`. Painter-dom embedders that don't ship the superdoc stylesheet still get the same rendering via the var() fallback — no behavior change for anyone who doesn't opt in.

**Rejected**: dropping the inline style entirely in favor of a stylesheet rule. The painter-dom bundle is consumed standalone by some integrations (MCP/CLI rendering pipelines) without the main superdoc CSS, so the inline `var()` + fallback is the only place the default reliably lands for all consumers.
**Review**: token placement in `variables.css` — I put it next to `--sd-proofing-*` since it's document-content tier, not UI. Open to moving it.
**Verified**: 184/184 unit (+1 regression test pinning the style attribute), 74/74 math behavior tests, type-check clean.

### DIFF
--- a/packages/layout-engine/painters/dom/src/features/math/omml-to-mathml.test.ts
+++ b/packages/layout-engine/painters/dom/src/features/math/omml-to-mathml.test.ts
@@ -40,6 +40,19 @@ describe('convertOmmlToMathml', () => {
     expect(mi!.textContent).toBe('x');
   });
 
+  it('emits font-family via --sd-math-font-family with Cambria Math fallback (SD-2634)', () => {
+    // Consumers override the math font via the --sd-math-font-family CSS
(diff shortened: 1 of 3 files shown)


caio-pizzol:
Closing as speculative — no concrete user or customer need for theming the math font. Current Cambria Math default matches Word and works across all known consumers. Will revisit if a real theming ask surfaces.

PR #2896, opened 2026-04-22
fix(paragraph): guard listRendering destructure against null

kendaller:
## Summary
Fixes a `TypeError: Cannot destructure property 'suffix' of 'this.node.attrs.listRendering' as it is null` thrown from `ParagraphNodeView`.

## Reproduction
When a paragraph node carries `listRendering: null` (as an explicit null, not undefined) and the post-transaction list-styles re-pass runs, two call sites crash:

1. `#updateListStyles` (line 235) destructures `{ suffix, justification }` directly.
2. `#initList` (lines 286-287) accesses `listRendering.markerText` and `.suffix`.

In my case this surfaced when dispatching a ProseMirror transaction that combined `tr.setDocAttribute('bodySectPr', …)` with a paragraph delete — the re-pass walks paragraphs and blows up on the first one whose `listRendering` is null.

## Fix
One-line guards at each call site — `?? {}` on the destructure, optional chaining on the property access. Existing defaults (`suffix ?? 'tab'` and the `suffix == null` branch in `#createSeparator`) absorb the null case cleanly.

## Test plan
- [x] Added a regression test (`does not throw when listRendering is null`) that calls `.update()` with a `listRendering: null` node and asserts no throw.
- [x] All 14 ParagraphNodeView tests pass: `pnpm vitest run src/editors/v1/extensions/paragraph/ParagraphNodeView.test.js`.
- [x] `pnpm run lint` — 0 errors (warnings are pre-existing `any` usage elsewhere).
- [x] `pnpm run format:check` — clean.

### DIFF
--- a/packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.js
+++ b/packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.js
@@ -232,7 +232,16 @@ export class ParagraphNodeView {
    * @returns {boolean}
    */
   #updateListStyles() {
-    let { suffix, justification } = this.node.attrs.listRendering ?? {};
+    const listRendering = this.node.attrs.listRendering;
+    // When listRendering is null (can happen transiently during certain
+    // transactions, e.g. after a setDocAttribute + paragraph delete), leave
+    // the existing marker/separator untouched. Forcing a default `suffix` here
+    // would risk writing tab-style CSS onto a text-node separator created by
+    // a prior 'space'/'nothing' suffix and scheduled RAF pass.
+    if (!listRendering) {
+      return true;
+    }
+    let { suffix, justification } = listRendering;
     suffix = suffix ?? 'tab';
     this.#calculateMarkerStyle(justification);
     if (suffix === 'tab') {
@@ -283,8 +292,14 @@ export class ParagraphNodeView {
    * @param {{ markerText: string, suffix?: string }} listRendering
    */
   #initList(listRendering) {
-    this.#createMarker(listRendering?.markerText);
-    this.#createSeparator(listRendering?.suffix);
+    // See #updateListStyles: when listRendering is null the previous marker/
+    // separator are left in place; avoid invoking the create helpers with
+    // undefined values.
+    if (!listRendering) {
+      return;
+    }
+    this.#createMarker(listRendering.markerText);
+    this.#createSeparator(listRendering.suffix);
   }
 
   #checkIsList() {
--- a/packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.js
+++ b/packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.js
@@ -232,7 +232,7 @@ export class ParagraphNodeView {
    * @returns {boolean}
    */
   #updateListStyles() {
-    let { suffix, justification } = this.node.attrs.listRendering;
+    let { suffix, justification } = this.node.attrs.listRendering ?? {};
     suffix = suffix ?? 'tab';
     this.#calculateMarkerStyle(justification);
     if (suffix === 'tab') {
@@ -283,8 +283,8 @@ export class ParagraphNodeView {
    * @param {{ markerText: string, suffix?: string }} listRendering
    */
   #initList(listRendering) {
-    this.#createMarker(listRendering.markerText);
-    this.#createSeparator(listRendering.suffix);
+    this.#createMarker(listRendering?.markerText);
+    this.#createSeparator(listRendering?.suffix);
   }
 
   #checkIsList() {
--- a/packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.js
+++ b/packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.js
@@ -289,7 +289,7 @@ export class ParagraphNodeView {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


kendaller:
Good catch — addressed in 0307343. Changed the null-guarded path from "force `suffix='tab'`" to "early return, leave the existing marker/separator untouched" in both `#updateListStyles` and `#initList`. Future updates (when `listRendering` gets a real value or `isList()` flips to false) will do the cleanup.

Added a second regression test (`does not try to style a text-node separator when switching to null listRendering`) that asserts the space→null transition doesn't replace or style the pre-existing Text node separator.

packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.js
```diff
@@ -232,7 +232,7 @@ export class ParagraphNodeView {
    * @returns {boolean}
    */
   #updateListStyles() {
-    let { suffix, justification } = this.node.attrs.listRendering;
+    let { suffix, justification } = this.node.attrs.listRendering ?? {};
     suffix = suffix ?? 'tab';
```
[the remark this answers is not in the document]
caio-pizzol: @kendaller already addressed in 0307343 - the null-guarded path now early-returns before reaching the `?? 'tab'` fallback.

packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.js
```diff
@@ -283,6 +292,12 @@ export class ParagraphNodeView {
    * @param {{ markerText: string, suffix?: string }} listRendering
```
caio-pizzol: this function now accepts null. the jsdoc comment above should match.

```suggestion
   * @param {{ markerText: string, suffix?: string } | null} listRendering
```

caio-pizzol (commented):
@kendaller thanks for jumping in - the fix is right, reproduced the crash locally. one out-of-diff thing: the branch at lines 86-88 that handles non-list paragraphs doesn't cancel a scheduled re-render, so the list marker can reappear after the paragraph stops being a list. left three inline - happy to push the cancel fix as a follow-up if easier.

packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.test.js
```diff
@@ -199,6 +199,49 @@ describe('ParagraphNodeView', () => {
...
+    const textSeparator = nodeView.separator;
+
+    const nullNode = createNode({
+      attrs: { ...spaceAttrs, listRendering: null },
+    });
+
+    expect(() => nodeView.update(nullNode, [])).not.toThrow();
+    // The text-node separator must be left alone (not replaced, not styled).
+    expect(nodeView.separator).toBe(textSeparator);
+  });
```
caio-pizzol, line 243: both null-checks also run when a paragraph first appears (via the constructor), not just when it updates. worth a test for that path:

```js
it('does not throw when mounted with listRendering null', () => {
  isList.mockReturnValue(true);
  const nullAttrs = { ...createNode().attrs, listRendering: null };
  expect(() => mountNodeView({ attrs: nullAttrs })).not.toThrow();
});
```

caio-pizzol, line 243: the comment calls null "transient", but nothing tests what happens when it clears. space, then null, then tab is the interesting path - the separator has to swap from a text node back to a span element:

```js
it('recovers marker/separator when listRendering returns from null to tab', () => {
  isList.mockReturnValue(true);
  const spaceAttrs = {
    ...createNode().attrs,
    listRendering: { suffix: 'space', justification: 'left', markerText: '1.' },
  };
  const { nodeView } = mountNodeView({ attrs: spaceAttrs });
  nodeView.update(createNode({ attrs: { ...spaceAttrs, listRendering: null } }), []);
  const tabNode = createNode({
    attrs: { ...spaceAttrs, listRendering: { suffix: 'tab', justification: 'left', markerText: '2.' } },
  });
  nodeView.update(tabNode, []);
  expect(nodeView.marker?.textContent).toBe('2.');
  expect(nodeView.separator?.tagName?.toLowerCase()).toBe('span');
});
```

packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.js
kendaller: Updated in 86da5c8.

packages/super-editor/src/editors/v1/extensions/paragraph/ParagraphNodeView.test.js
kendaller: Added in 86da5c8 — your snippet verbatim.

kendaller: Added in 86da5c8 — your snippet verbatim.

kendaller:
@caio-pizzol thanks for the review! JSDoc + both tests pushed in 86da5c8 (rebased onto your main-merge). On the out-of-diff #removeList not cancelling the scheduled RAF — agreed that's a real bug but going to leave it for a follow-up PR to keep this one scoped to the null destructure.

caio-pizzol (approved):
hey @kendaller! thanks for addressing last round review.

all three points landed as suggested: jsdoc now matches the nullable contract, and both the mount-null and space → null → tab recovery tests are in.

lgtm.

PR #2897, opened 2026-04-22
fix(style-engine): fall back to built-in defaults for Heading1-9 (#2805)

### LINKED ISSUE #2805: bug: built-in Word heading styles not rendered when missing from styles.xml

Abdeltoto:
Closes #2805.

## What's going on

Word documents created by python-docx, pandoc and a few other generators set
`<w:pStyle w:val="Heading1"/>` on paragraphs without writing the matching
`<w:style>` block in `styles.xml`. ECMA-376 §17.7.4.9 says applications should
fall back to the built-in (latent) defaults for these well-known styleIds, but
`resolveStyleChain` was returning `{}` and headings were rendering as plain body
text — no bold, no spacing, no outline level.

## What this PR does

- Adds a small `built-in-styles.ts` module with the Word 2016+ defaults for
  `Heading1` through `Heading9` (Calibri / Calibri Light, the standard blue
  accents, `keepNext` / `keepLines` / `outlineLvl`, the usual spacing).
- Wires `resolveStyleChain` through a tiny `lookupStyleDefinition` helper that
  prefers the document's own `styles` and only falls back to the built-in
  catalog when the document doesn't define the style. Both the initial lookup
  and the `basedOn` walk go through the same helper, so a built-in
  `basedOn: 'Normal'` will pick up the document's `Normal` properties
  automatically.

## Why it's safe

- Document-defined styles still take priority (covered by a regression test).
- Unknown styleIds (`NotAHeading`, `Heading10`, lowercase variants) still
  return `{}` — we don't paper over unrelated bugs.
- The `basedOn` chain still terminates safely: if `Normal` isn't defined
  anywhere, the loop just exits without throwing.
- Built-in objects are not mutated by `combineProperties` (which spreads into a
  fresh object), so sharing them between calls is fine.

## Tests

8 new tests in `packages/layout-engine/style-engine/src/ooxml/index.test.ts`:

- `Heading1` falls back to built-in run + paragraph properties when missing.
- All `Heading1`..`Heading9` produce sensible defaults (`outlineLvl`, font size).
- Document-defined `Heading1` still wins over the built-in default.
- Unknown styleIds still return `{}`.
- Built-in `basedOn: 'Normal'` chains into the document's `Normal` style.
- End-to-end: `resolveParagraphProperties({ styleId: 'Heading1' })` produces
  visible heading formatting.
- Public helpers (`getBuiltInStyleDefinition`, `listBuiltInStyleIds`) behave.

```
bun test src/ooxml/index.test.ts
73 pass / 0 fail / 140 expect()
```

Full package suite (`bun test`) is green: 137/137.

## Repro

```python
from docx import Document
d = Document()
d.add_heading('Title', level=1)
d.add_heading('Sub',   level=2)
d.add_paragraph('Body')
d.save('repro.docx')
```

Before this PR: all three paragraphs render with body styling.
After: Heading 1/2 get the standard Word heading look.

### DIFF
--- a/packages/layout-engine/style-engine/src/ooxml/built-in-styles.ts
+++ b/packages/layout-engine/style-engine/src/ooxml/built-in-styles.ts
@@ -1,0 +1,96 @@
+/**
+ * Built-in (latent) Word style defaults.
+ *
+ * Word ships with ~260 built-in style ids that documents are allowed to reference
+ * without including a matching `w:style` entry in `styles.xml`. ECMA-376 §17.7.4.9
+ * expects a reading application to fall back to the built-in defaults in that case.
+ *
+ * In the wild this happens with documents emitted by python-docx, pandoc and other
+ * generators: they set `<w:pStyle w:val="Heading1"/>` on a paragraph but never write
+ * the corresponding style block, and SuperDoc was rendering the headings as plain
+ * body text (issue #2805).
+ *
+ * The values below mirror the Word 2016+ default template (Calibri / Calibri Light
+ * with the standard blue accents). They intentionally cover only the most common
+ * case — Heading 1 through Heading 9 — since those are the ones that actually move
+ * the needle for users. Other latent styles can be added on demand.
+ *
+ * Each definition has `basedOn: 'Normal'`, so the existing cascade keeps working
+ * naturally: if the document does define `Normal`, those properties still flow in,
+ * and if it doesn't, the chain just terminates without throwing.
+ */
+
+import type { StyleDefinition } from './styles-types.ts';
+
+/** Half-points (Word's `w:sz` unit). 28 → 14pt, 26 → 13pt, etc. */
+const HP = (pt: number): number => Math.round(pt * 2);
+
+/** Twentieths of a point (Word's `w:before` / `w:after` unit). */
+const TWIPS = (pt: number): number => Math.round(pt * 20);
+
+const HEADING_BLUE = '2F5496';
+const HEADING_DARK_BLUE = '1F3864';
+const HEADING_DARK_GREY = '272727';
+
+const buildHeading = (
+  level: number,
+  fontSize: number,
+  color: string,
+  options: { bold?: boolean; italic?: boolean; spacingBeforePt?: number } = {},
+): StyleDefinition => {
+  const { bold = false, italic = false, spacingBeforePt = 2 } = options;
+  return {
+    type: 'paragraph',
+    styleId: `Heading${level}`,
+    name: `heading ${level}`,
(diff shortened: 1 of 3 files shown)


caio-pizzol:
Hey @Abdeltoto! Thanks for the effort here

Same answer that I gave on #2806 

SuperDoc aims for 100% parity with Word, so we shouldn't be adding styling that Word doesn't apply.

If a docx is missing style definitions, that's a document generation error - not something we should patch over with hardcoded defaults. A band-aid here just hides the real problem and moves us away from Word's behavior.

Abdeltoto:
Totally fair, @caio-pizzol — Word-parity over best-guess defaults makes sense, and I see how a fallback like this would mask actual generation bugs in tools like python-docx or pandoc instead of pushing them upstream where they belong.

Closing the door on this one. Thanks for the clear reasoning and the time to look it over.

PR #2882, opened 2026-04-21
feat(super-editor): render imported heading/bookmark cross-references (SD-2495)

tupizz:
Three features + one pre-existing bug fix, packaged under the **Internal Anchors** epic (SD-2495).

## What ships

| Ticket | What changes for the user |
|---|---|
| **[SD-2536](https://linear.app/superdocworkspace/issue/SD-2536)** — Import model | Word cross-references (REF / NOTEREF / STYLEREF fields) now survive DOCX import. Before this PR, every REF field was silently dropped. |
| **[SD-2537](https://linear.app/superdocworkspace/issue/SD-2537)** — Render | Cross-references render in the document with their Word-cached text. `\h` variants are clickable and navigate to the target bookmark. |
| **[SD-2454](https://linear.app/superdocworkspace/issue/SD-2454)** — Bookmark brackets | Optional Word-style gray `[` / `]` indicators around bookmarked content. Off by default, toggleable. |
| **[IT-949](https://linear.app/superdocworkspace/issue/IT-949)** — Customer bug | Closed by SD-2536/2537. "Section 15" now renders in the Brillio lease. |

## Why it was broken (one root cause, two symptoms)

The `sd:crossReference` v3 translator existed but was **never wired into the v2 importer's entity list**. The passthrough fallback refused to wrap it (because it was "registered"), and no entity claimed it, so the dispatch loop dropped every REF field on the floor. On top of that, even if the node had reached the renderer, `#scrollPageIntoView` wrote `scrollTop` to the wrong DOM element, so click-to-navigate silently no-op'd on any real consumer layout.

## How the fix is structured

```
DOCX import                         Render                            Click
─────────────                       ──────                            ─────
v2 importer                         pm-adapter                        EditorInputManager
   │                                   │                                 │
   ├─ NEW crossReferenceImporter.js   ├─ cross-reference.ts             ├─ handleLinkClick
   │   (the missing wire-up)         │   never returns null;            │   generalized anchor
   │                                 │   synthesizes link mark           │   routing to goToAnchor
   │                                 │   when instruction has \h         │
   ├─ extractResolvedText            │                                  │
   │   recursive walk                └─ bookmark-start.ts / -end.ts     │
   │                                     emit [ ] markers when           │
   └─ ref-preprocessor                   showBookmarks is on            │
       (unchanged)                                                       ▼
                                                                   PresentationEditor
                                                                     scrollPageIntoView
                                                                     writes to real
                                                                     scroll container
```

## Files at a glance

[...]

PR #2899, opened 2026-04-22
feat(pm-adapter): synthesize internal link for PAGEREF with \h switch

kendaller:
## Summary
When a `PAGEREF` field instruction includes the `\h` switch, attach a `FlowRunLink` to the rendered token run so clicks on the page number navigate to the referenced bookmark. Mirrors the pattern added for `crossReference` in #2882.

## Demo
https://scale.zoom.us/clips/share/gMY1cl5NR0q6jYm3kjQPnQ

## Motivation
Word TOCs and other hyperlinked page references use `PAGEREF _Toc<id> \h` — the `\h` switch tells Word to render the page number as a hyperlink. Before this change, the PAGEREF inline converter emitted a `pageReference` token run with `pageRefMetadata.bookmarkId` but no `link`, so the DOM layer never produced a clickable element and clicks did nothing.

## Fix
In `pageReferenceNodeToBlock`, after building the token run, call `buildFlowRunLink({ anchor: bookmarkId })` when `instruction` matches `/\\h\b/` and assign the result to `run.link`. The existing anchor-click routing added in #2882 (`EditorInputManager.#handleLinkClick` → `goToAnchor`) then handles navigation automatically.

## Dependency
This branch is cut from #2882 because it uses `buildFlowRunLink` from `marks/links.ts`, added in that PR. Expected merge order: #2882 → this. Happy to rebase once #2882 lands if the target changes.

## Test plan
- [x] New test file `page-reference.test.ts` covering:
  - synthesizes internal link with `\h`
  - no link when `\h` is absent
  - doesn't match a literal `h` substring as `\h`
  - quoted bookmark id in instruction
  - emits the existing `pageReference` token + `pageRefMetadata.bookmarkId`
- [x] Cross-reference tests from #2882 still pass (verified locally)
- [x] `pnpm run lint` — 0 errors
- [x] `pnpm run format:check` — clean
- [x] End-to-end smoke test with a real Word TOC (video above): click on both page numbers and section labels in TOC entries and verify navigation to the correct bookmark.


kendaller:
Good catch — addressed in 7f19358. Added the `i` flag to the `\h` regex and a regression test covering an uppercase switch (`PAGEREF _Toc123 \H`).

Worth noting the original `/\\h\b/` (case-sensitive) pattern was mirrored from #2882's `crossReference` converter. If you want the same case-insensitivity fix applied there for consistency I'm happy to open a follow-up PR against that branch.

packages/layout-engine/pm-adapter/src/converters/inline-converters/page-reference.ts
```diff
@@ -68,6 +69,19 @@ export function pageReferenceNodeToBlock(params: InlineConverterParams): TextRun
       bookmarkId,
       instruction,
     };
+
+    // When the PAGEREF instruction carries the `\h` switch, render as an
+    // internal hyperlink pointing at `#<bookmarkId>` so clicks navigate to
+    // the target bookmark via the existing anchor-link navigation path.
+    if (/\\h\b/.test(instruction)) {
```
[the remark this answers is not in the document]
caio-pizzol: @kendaller already addressed in 7f19358 - adds `i` flag and a regression test covering `\H`.

packages/layout-engine/pm-adapter/src/converters/inline-converters/page-reference.ts
```diff
@@ -68,6 +69,21 @@ export function pageReferenceNodeToBlock(params: InlineConverterParams): TextRun
...
+    // Field switches in Word field instructions are case-insensitive, so
+    // `\h` and `\H` should both produce the hyperlink.
+    if (/\\h\b/i.test(instruction)) {
+      const synthesized = buildFlowRunLink({ anchor: bookmarkId });
+      if (synthesized) {
+        (tokenRun as TextRun).link = (tokenRun as TextRun).link
+          ? { ...(tokenRun as TextRun).link, ...synthesized, anchor: bookmarkId }
+          : synthesized;
+      }
+    }
```
caio-pizzol: the `as TextRun` casts aren't needed - `textNodeToRun` already returns a `TextRun`. `cross-reference.ts:34` does this cleanly. shorter comment too.

```suggestion
    // \h switch - case-insensitive per ECMA-376 §17.16.1.
    if (/\\h\b/i.test(instruction)) {
      const synthesized = buildFlowRunLink({ anchor: bookmarkId });
      if (synthesized) {
        tokenRun.link = tokenRun.link
          ? { ...tokenRun.link, ...synthesized, anchor: bookmarkId }
          : synthesized;
      }
    }
```

caio-pizzol (commented):
@kendaller thanks for jumping in on this. tested against Word and locally - click behavior matches. pushed a behavior test, the REF/NOTEREF/STYLEREF preprocessor tests, and a `setShowBookmarks` test to close the Codecov gap. one thing left: please also add the `i` flag to `cross-reference.ts:31` per ECMA-376 §17.16.1. one small thing inline.

packages/layout-engine/pm-adapter/src/converters/inline-converters/page-reference.ts
kendaller: Applied in c421354. Dropped the casts and shortened the comment.

kendaller:
@caio-pizzol thanks for pushing the extra test coverage. c421354 drops the `as TextRun` casts per your inline and adds the `i` flag to `cross-reference.ts:31` (with a `\H` regression test to match the page-reference pattern).

caio-pizzol (approved):
hey @kendaller! thanks for addressing last round.

i pushed a follow-up commit with a unit test that pins the scroll fix. it fakes `window.getComputedStyle` to mark a wrapper element as scrollable, then checks both the wrapper and the host get the scroll position. verified it catches a revert to the old one-line version.

lgtm.

PR #2901, opened 2026-04-22
feat(layout-engine): resolve PAGEREF tokens in incrementalLayout loop

kendaller:
## Summary
Wires `buildAnchorMap` + `resolvePageRefTokens` into the existing two-pass convergence loop in `incrementalLayout`, so PAGEREF fields (TOC page numbers and other cross-references) refresh after edits that change pagination — instead of remaining stuck at whatever page number Word baked in at DOCX export time.

## Motivation
`packages/layout-engine/layout-engine/src/resolvePageRefs.ts` exports `buildAnchorMap` and `resolvePageRefTokens`. Both are fully implemented and tested in isolation, but before this PR they were **not called anywhere in production code** — only from their own test file and the `index.ts` barrel. A grep across `packages/` and `apps/` confirms zero runtime importers.

Observable impact: after a user edits the document in a way that shifts pagination (insert/delete paragraphs, toggle large conditional blocks, etc.), every TOC entry's page number stays at the pre-edit value. Word TOCs typically carry `PAGEREF _Toc<id> \h`, whose rendered page number needs the current layout to compute.

## Approach
Mirror the existing PAGE/NUMPAGES resolution pattern exactly:

1. **Refactor `resolvePageRefTokens` to match `resolvePageNumberTokens`'s contract.** Returns `{ affectedBlockIds, updatedBlocks: Map<blockId, FlowBlock> }` with cloned blocks, rather than mutating in place and returning a `Set`. This is required for the convergence loop to be correct — the loop rewrites `currentBlocks` from updates each iteration; a mutation-style resolver double-counts.
2. **Thread `bookmarks` into `incrementalLayout`.** New trailing optional param. `PresentationEditor` already tracks the bookmarks map on `#layoutState.bookmarks` and just needs to forward it.
3. **Add `FeatureFlags.BODY_PAGEREFS`** (env `SD_BODY_PAGEREFS`, default `true`). Disabling reverts to the pre-PR behavior where PAGEREF runs keep the Word-baked fallback text.
4. **Merge PAGEREF into the convergence loop.** Both resolvers run each iteration on the current blocks; their `affectedBlockIds` are unioned and `updatedBlocks` merged via a tiny `mergeRunUpdates` helper that covers the rare case of a single block touched by both resolvers (e.g., a header containing both `PAGE` and `PAGEREF`).

Convergence semantics remain identical to the existing PAGE/NUMPAGES loop: iterate up to 3 times, break when no blocks change, break when page count stabilizes after the first pass.

## Test plan
- [x] `resolvePageRefs.test.ts` updated for the new return shape; added an idempotence test (re-running with the same anchor map and already-resolved text is a no-op). **12 pass, 0 fail.**
- [x] Full `@superdoc/layout-engine` suite: **614 pass, 1 skip, 0 fail.**
- [x] Full `@superdoc/layout-bridge` suite (the package that owns `incrementalLayout`): **1169 pass, 0 fail.**
- [x] `pnpm run format:check` — clean.
- [x] `pnpm run lint` — 0 errors (371 pre-existing `any` warnings elsewhere, unchanged).
- [ ] End-to-end demo video to follow in a comment — recording now with a retrofitted DOCX that has a TOC.

## Out of scope
- **TOC / Index of Defined Terms entry regeneration.** This PR refreshes the page-number PART of a TOC entry but does not rebuild the entry list itself (e.g. if a heading is renamed or deleted). That's a separate, deeper architectural change.
- **F9 semantics.** The new behavior makes PAGEREF refresh automatic on every layout pass, so F9 support for PAGEREF is no longer necessary. If you'd prefer a more conservative rollout where PAGEREF only resolves on F9 (like NUMWORDS), let me know and I'll move it to the `field-update` extension instead.

### DIFF
--- a/packages/layout-engine/layout-bridge/src/featureFlags.ts
+++ b/packages/layout-engine/layout-bridge/src/featureFlags.ts
@@ -65,6 +65,13 @@ export const FeatureFlags = {
   BODY_PAGE_TOKENS: isEnabled('SD_BODY_PAGE_TOKENS', true),
 
   /**
+   * Enable PAGEREF token resolution so TOC entries and other cross-reference
+   * page numbers refresh after edits that change pagination. When disabled,
+   * PAGEREF runs render the text Word baked in at DOCX export time.
+   */
+  BODY_PAGEREFS: isEnabled('SD_BODY_PAGEREFS', true),
+
+  /**
    * Enable header/footer page token resolution.
    * When disabled, headers/footers use painter-time token rendering fallback.
    */
--- a/packages/layout-engine/layout-bridge/src/incrementalLayout.ts
+++ b/packages/layout-engine/layout-bridge/src/incrementalLayout.ts
@@ -17,6 +17,8 @@ import {
   type HeaderFooterConstraints,
   computeDisplayPageNumber,
   resolvePageNumberTokens,
+  buildAnchorMap,
+  resolvePageRefTokens,
   type NumberingContext,
   SEMANTIC_PAGE_HEIGHT_PX,
   SINGLE_COLUMN_DEFAULT,
@@ -749,6 +751,7 @@ export async function incrementalLayout(
     measure?: HeaderFooterMeasureFn;
   },
   previousMeasures?: Measure[] | null,
+  bookmarks?: Map<string, number>,
 ): Promise<IncrementalLayoutResult> {
   const isSemanticFlow = options.flowMode === 'semantic';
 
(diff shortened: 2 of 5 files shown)


kendaller:
Closing this for now — deferring PAGEREF resolution to the superdoc team. In the course of testing this end-to-end I found that an incremental re-render produces `bookmarksSize: 1` (or 0) in the `incrementalLayout` diag log, even for documents with 30+ TOC bookmarks, because `FlowBlockCache` stores a paragraph's blocks and `pmStart` but not the bookmarks it registered. On any cache hit in `handleParagraphNode` (`pm-adapter/src/converters/paragraph.ts`), `bookmarkStartNodeToBlocks` is skipped and those bookmarks never re-enter the fresh `bookmarks` map in `toFlowBlocks`. This PR surfaces the pre-existing bug because it's the first consumer that needs an accurate map. Fix shape: store `{ name, offsetFromPmStart }[]` on `CachedParagraphEntry` and re-register them on cache hit. Happy to share the draft patch if useful.

PR #2904, opened 2026-04-22
fix: applying formatting to header/footer

artem-harbour:
Linear: SD-2618

Fixes inconsistent toolbar formatting in header/footer by removing stale selection restoration and avoiding deferred mark execution in H/F.

### What changed
- Skip `restoreSelection()` for header/footer toolbar interactions.
- Execute mark-toggle commands immediately in header/footer (no `pendingMarkCommands` queue).
- Simplify mark command selection resolution to use live transaction selection (`tr.selection`) via a centralized helper.
- Clean up command call sites/signatures accordingly.
- Add regression tests for:
  - no `restoreSelection` call in header/footer toolbar flow,
  - immediate (non-queued) mark-toggle execution in header/footer.

### Result
- `bold/italic/underline/color/highlight` now apply immediately in centered header/footer text.
- `fontFamily/fontSize` now apply immediately without requiring an extra selection change.
- Body behavior remains unchanged.

### DIFF
--- a/packages/super-editor/src/editors/v1/components/toolbar/Toolbar.test.js
+++ b/packages/super-editor/src/editors/v1/components/toolbar/Toolbar.test.js
@@ -111,4 +111,32 @@ describe('Toolbar', () => {
     addSpy.mockRestore();
     removeSpy.mockRestore();
   });
+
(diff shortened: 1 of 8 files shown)


harbournick (approved):
LGTM

PR #2905, opened 2026-04-22
fix: prevent corrupt docx export for legacy w:pict vml rect content

artem-harbour:
Linear: SD-2652

Fixes issue where import→export of DOCX files containing legacy `w:pict/v:rect` could produce invalid `mc:AlternateContent` with empty `w:drawing`, causing Word corruption errors.

What changed:
- Preserved legacy VML metadata on `contentBlock` (`vmlAttributes`, `style`) to avoid attribute loss during PM normalization.
- Added legacy VML routing guard in `translateContentBlock()` so `contentBlock` with VML markers exports via `w:pict/v:rect` path.
- Hardened `mc:AlternateContent` decode:
  - returns `null` for missing/invalid drawing content,
  - emits `mc:Fallback` for valid alternate-content branches.
- Added/updated unit tests for new routing and decode behavior.


harbournick (approved):
LGTM

PR #2918, opened 2026-04-22
SD-2524 - feat: implement context menu for lists

chittolinag:
### Issue

When a user right-clicks a numbered list marker (e.g., "1.") in the document, the context menu shows only generic editor actions. There is no way to perform list-specific operations — restart numbering, continue numbering, or change indentation — from the marker itself. Left-clicking a marker is also not distinguished from clicking paragraph text, so no "marker selected" visual feedback is given.

### Proposed solution

- Extend `getEditorContext()` to detect when a right-click lands on a `.superdoc-list-marker` element and expose `isOnListMarker` in the context.
- Show list-specific items on the context menu (restart numbering, continue numbering, increase indent, decrease indent).
- Add a `list-marker` section to the context menu with restart, continue, decrease-indent, and increase-indent actions, visible only when `isOnListMarker` is true.

Screenshot 2026-04-24 at 14 33 23


harbournick:
@artem-harbour can you please prioritize reviewing this one today? thank you!

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -353,6 +353,7 @@ function renderListMarker(params: MarkerRenderParams): void {
 
   // Create marker container (inline-block to isolate from word-spacing used for justification)
   const markerContainer = doc.createElement('span');
+  markerContainer.classList.add('superdoc-list-marker');
```
artem-harbour: Move class name to constants.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -3152,6 +3145,7 @@ export class DomPainter {
 
             if (!resolvedMarker.vanish) {
               const markerContainer = this.doc!.createElement('span');
+              markerContainer.classList.add('superdoc-list-marker');
```
artem-harbour: Class name to constants.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -3365,6 +3357,7 @@ export class DomPainter {
 
             if (!marker.run.vanish) {
               const markerContainer = this.doc!.createElement('span');
+              markerContainer.classList.add('superdoc-list-marker');
```
artem-harbour: Class name to constants.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -353,6 +353,7 @@ function renderListMarker(params: MarkerRenderParams): void {
 
   // Create marker container (inline-block to isolate from word-spacing used for justification)
   const markerContainer = doc.createElement('span');
```
artem-harbour: If you have some time, refactor and move the creation of the marker/container into a function. This code is repeated.

packages/super-editor/src/editors/v1/components/context-menu/utils.js
```diff
@@ -126,6 +122,9 @@ export async function getEditorContext(editor, event) {
   const isInTable =
     structureFromResolvedPos?.isInTable ?? selectionHasNodeOrMark(state, 'table', { requireEnds: true });
   const isInList = structureFromResolvedPos?.isInList ?? selectionIncludesListParagraph(state);
+  // .superdoc-list-marker = DomPainter markerContainer / presentation mode
+  // .list-marker = ParagraphNodeView / flow editor mode
+  const isOnListMarker = Boolean(event?.target?.closest?.('.superdoc-list-marker, .list-marker'));
```
artem-harbour: Move class names to constants.

packages/super-editor/src/editors/v1/core/Editor.ts
```diff
@@ -2703,6 +2689,10 @@ export class Editor extends EventEmitter<EditorEventMap> {
     const prevState = this.state;
     let nextState: EditorState;
     let transactionToApply = transaction;
+    // appendTransaction plugins (e.g. numberingPlugin) may produce transactions that
+    // change the doc even when the original transaction does not. Track the full list so
+    // the 'update' event is emitted with the effective doc-changing transaction.
+    let appliedTransactions: readonly Transaction[] | undefined;
```
artem-harbour: Do we really need these changes in the main entry point for the transaction to apply? As far as I know, `appendTransaction` doesn't require any additional actions to be applied.

artem-harbour (changes requested):
Added some comments.

packages/super-editor/src/editors/v1/core/commands/continueNumbering.js
```diff
@@ -0,0 +1,31 @@
...
+ * `preventDispatch` so CommandService skips the dispatch.
+ */
+export const continueNumbering = ({ editor, tr, state }) => {
+  const { node: paragraph } = findParentNode(isList)(state.selection) || {};
+  if (!paragraph) return false;
+
+  const { numId, ilvl = 0 } = getResolvedParagraphProperties(paragraph)?.numberingProperties || {};
+  if (numId == null) return false;
+
+  ListHelpers.removeLvlOverride(editor, numId, ilvl);
```
caio-pizzol, line 33: word's "continue numbering" puts the paragraph back on the original numId. this just removes the override on the new numId, so the paragraph still points at the orphan. looks right in superdoc but word will rewrite it on save.

packages/super-editor/src/editors/v1/core/commands/restartNumbering.js
```diff
@@ -2,27 +2,72 @@ import { findParentNode } from '@helpers/index.js';
...
+  // Remap paragraphs from this position onwards to the new numId. Steps are
+  // accumulated on the captured tr; CommandService dispatches it after we return.
+  state.doc.nodesBetween(paragraphPos, state.doc.content.size, (node, pos) => {
+    if (node.type.name !== 'paragraph') return true;
+    const props = getResolvedParagraphProperties(node)?.numberingProperties;
+    if (props?.numId === numId) {
+      updateNumberingProperties({ numId: newNumId, ilvl: props.ilvl }, node, pos, editor, tr);
+    }
+    return true;
+  });
```
caio-pizzol, line 76: two things. word marks only the clicked paragraph; this marks every paragraph from cursor to doc end. visible numbering is the same in both (word's renderer chains the counter through the new numId), but word's file is much smaller. also, the level can be undefined when numbering comes from a style (like `ListNumber`); passing it through writes a `<w:numPr>` with no `<w:ilvl>` - word always writes `<w:ilvl w:val="0"/>` explicitly.

[sd-2524-word-test1-after-restart.docx](https://github.com/user-attachments/files/27177636/sd-2524-word-test1-after-restart.docx)

caio-pizzol (commented):
hey @chittolinag! menu wiring looks solid :)

two things inline. bigger question: `restartNumbering` and `continueNumbering` rebuild what `listsSetValueWrapper` (`lists-wrappers.ts:900-1000`) already does. on purpose, or worth using the wrapper?

minor: indent/outdent in the menu silently does nothing on flat lists, and `continueNumbering` runs even when there's no override to remove.

artem-harbour (approved):
LGTM

packages/super-editor/src/editors/v1/core/Editor.ts
chittolinag: You're right that `appendTransaction` doesn't need anything extra to be applied. `applyTransaction()` already merges appended trs into the new state. However this change is about which `Transaction` we pass to `'update'` listeners.

For `continueNumbering` for instance, the call to `removeLvlOverride` mutates the numbering definition and fires a `partChanged`, then `handleNumberingInvalidation` dispatches an empty tr. Down the road, the plugin will build its own tr to update the list rendering attributes, which will be an "appended" tr, which carries the actual doc change. 

So, as a result here, we'll have:

- `transactionToApply.docChanged === false` (the dispatched tr is empty)
- `appliedTransactions[1].docChanged === true` (the appended rewrite)

This is why we need to get the effective transaction that changed the document here.

Please let me know if you have any questions or suggestions on how to improve here :)

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
chittolinag: done

packages/super-editor/src/editors/v1/core/commands/restartNumbering.js
chittolinag: Fixed the undefined level! 

For the 1st issue, it was on purpose actually, `numberingPlugin` doesn't have Word's auto-chaining rule — it computes per `numId` counters. So if only the clicked paragraph were remapped, the next paragraph (still on the old `numId`) would continue from the *old* sequence rather than starting fresh.

packages/super-editor/src/editors/v1/core/commands/continueNumbering.js
chittolinag: can you pls clarify what's the user impact for this? might be easier to understand the actual issue here and determine if a fix is needed or not.

artem-harbour:
@caio-pizzol, please check your comments and let us know if this is ready for merge.

packages/super-editor/src/editors/v1/core/commands/restartNumbering.js
```diff
@@ -2,27 +2,72 @@ import { findParentNode } from '@helpers/index.js';
...
+
+  // Mid-list restart: create a new numId sharing the same abstractId.
+  // createNumDefinition and setLvlOverride operate on a brand-new numId that
+  // no paragraph references yet, so handleNumberingInvalidation's appendTransaction
+  // produces no doc change. The original tr (and state.doc) remain valid.
+  const allDefs = ListHelpers.getAllListDefinitions(editor);
+  const abstractId = allDefs?.[numId]?.[ilvl]?.abstractId;
+  if (abstractId == null) return false;
+
+  const { numId: newNumId } = ListHelpers.createNumDefinition(editor, Number(abstractId));
```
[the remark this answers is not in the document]
caio-pizzol: valid - `createNumDefinition` accepts `{ copyOverridesFrom }` (numbering-transforms.ts:355) and `lists-wrappers.ts:861` already uses it; multi-level lists with overrides on deeper ilvls lose them at the restart point. worth threading through.

packages/super-editor/src/editors/v1/extensions/paragraph/paragraph.js
```diff
@@ -332,6 +333,14 @@ export const Paragraph = OxmlNode.create({
...
        */
       restartNumbering: () => restartNumbering,
+      /**
+       * Remove the startOverride for the current list level so numbering continues
+       * from the previous chain instead of restarting.
+       * @category Command
+       * @example
+       * editor.commands.continueNumbering()
+       */
+      continueNumbering: () => continueNumbering,
```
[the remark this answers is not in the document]
caio-pizzol: valid - `continueNumbering` is exposed at runtime but missing from `paragraph-commands.ts`; add an entry alongside `restartNumbering` so typed consumers see it.

packages/super-editor/src/editors/v1/core/commands/continueNumbering.js
caio-pizzol: low impact. rendering and counters work. the only effect is a leftover unused numbering entry in the saved file - word cleans it up the first time it opens and saves there. fine as-is unless we care about file size.

packages/super-editor/src/editors/v1/core/commands/restartNumbering.js
caio-pizzol: makes sense, agree. heads up: the codex note on line 58 gets worse because of this. you remap every paragraph below the click, so items on other levels also get pulled to the new numId and lose their earlier overrides. worth fixing alongside.

packages/super-editor/src/editors/v1/core/commands/continueNumbering.js
caio-pizzol, line 33: repro: `pnpm --filter @superdoc/super-editor vitest run src/editors/v1/core/commands/continueNumbering.headless.test.js` → fails. with no view, nothing dispatches and the list marker stays stale. fix: only set `preventDispatch` when a view-side dispatch actually fires.

packages/super-editor/src/editors/v1/core/commands/restartNumbering.js
```diff
@@ -2,27 +2,75 @@ import { findParentNode } from '@helpers/index.js';
...
+
+  if (!hasPrecedingItems) {
+    // Already the first item — pin startOverride on the existing numId.
+    // setLvlOverride triggers handleNumberingInvalidation, which dispatches a
+    // fresh tr that updates listRendering synchronously. After that, the tr
+    // CommandService captured before this command ran is built against a stale
+    // doc, so dispatching it would throw "Applying a mismatched transaction".
+    // Flag it with `preventDispatch` so CommandService skips the dispatch.
+    ListHelpers.setLvlOverride(editor, numId, ilvl, { startOverride: 1 });
+    tr.setMeta('preventDispatch', true);
```
caio-pizzol: repro: `pnpm --filter @superdoc/super-editor vitest run src/editors/v1/core/commands/restartNumbering.headless.test.js` → fails. same bug as `continueNumbering`, only on the first-item path.

packages/super-editor/src/editors/v1/extensions/paragraph/paragraph.js
```diff
@@ -343,6 +344,14 @@ export const Paragraph = OxmlNode.create({
...
```
caio-pizzol, line 365: `continueNumbering` is wired up at runtime here but missing from `extensions/types/paragraph-commands.ts`, so TS users don't see it on `editor.commands` / `chain()`. add a matching entry next to `restartNumbering: () => boolean;`:

```ts
/** Continue numbering by removing the level override on the current list item */
continueNumbering: () => boolean;
```

caio-pizzol (commented):
hey @andrii-harbour!

i pushed two failing tests on the branch (c29f393d0).

a few things inline.

packages/layout-engine/painters/dom/src/utils/marker-helpers.ts
```diff
@@ -6,6 +8,8 @@ import {
   type MinimalWordLayout,
   type ResolvedListMarkerGeometry,
 } from '@superdoc/common/list-marker-utils';
+import { applySourceAnchorDataset } from '../renderer';
```
caio-pizzol, line 11: this file and renderer.ts import each other (helper pulls `applySourceAnchorDataset` from renderer; renderer pulls `createListMarkerElement` from here). works now, fragile. move `applySourceAnchorDataset` into its own small file. while you're there, `import type { SourceAnchor }` since it's only used as a type.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -351,29 +349,7 @@ function renderListMarker(params: MarkerRenderParams): void {
...
-  }
-  markerEl.style.fontWeight = markerLayout?.run?.bold ? 'bold' : '';
-  markerEl.style.fontStyle = markerLayout?.run?.italic ? 'italic' : '';
-  if (markerLayout?.run?.color) {
-    markerEl.style.color = markerLayout.run.color;
-  }
-  if (markerLayout?.run?.letterSpacing != null) {
-    markerEl.style.letterSpacing = `${markerLayout.run.letterSpacing}px`;
-  }
+  const markerContainer = createListMarkerElement(doc, markerLayout?.markerText ?? '', markerLayout?.run ?? {});
```
caio-pizzol, line 352: flow markers carry source anchors via the helper now, but this call passes 3 args so table markers don't. snapshot consumers walking markers will skip them. any reason not to pass `markerLayout?.sourceAnchor` here, like the renderer.ts call at 3231?

packages/layout-engine/painters/dom/src/utils/marker-helpers.ts
```diff
@@ -74,3 +78,54 @@ export const resolvePainterListTextStartPx = ({
...
+  run: MarkerRunStyle,
+  sourceAnchor?: SourceAnchor,
+): HTMLElement => {
+  const markerContainer = doc.createElement('span');
+  markerContainer.classList.add(DOM_CLASS_NAMES.LIST_MARKER);
+  markerContainer.style.display = 'inline-block';
+  markerContainer.style.wordSpacing = '0px';
+
+  const markerEl = doc.createElement('span');
+  markerEl.classList.add('superdoc-paragraph-marker');
```
caio-pizzol, line 108: since this PR is moving list-marker classes into `DOM_CLASS_NAMES`, `superdoc-paragraph-marker` is the obvious next one. fine as a follow-up.

caio-pizzol (commented):

a few things to land before merge, left them inline. also worth a `pnpm test:layout` run before merge.

caio-pizzol (approved):

PR #2920, opened 2026-04-23
docs: enhance OpenAPI specification for sign API

andrii-harbour:
- Updated the `document` object description to clarify the requirement of providing either `base64` or `url`.
- Expanded the `signer` object description to specify required fields and optional fields, including examples.
- Improved the `auditTrail` description to emphasize compliance requirements.
- Added a new `certificate` object to configure the audit trail certificate page.
- Updated the API documentation in `backend.mdx` to reflect the changes in the `signer` fields and their requirements.

### DIFF
--- a/apps/docs/openapi.json
+++ b/apps/docs/openapi.json
@@ -280,7 +280,7 @@
                 "data": {
                   "type": "object",
                   "additionalProperties": true,
-                  "description": "Event-specific payload emitted by the e-sign SDK. Shape depends on `type`:\n- `scroll` — `{ percent: number }`\n- `field_change` — `{ fieldId: string, value: string | boolean, previousValue?: string | boolean }`\n- `ready`, `submit` — typically omitted"
+                  "description": "Event-specific payload emitted by the e-sign SDK. Shape depends on `type`:\n- `scroll` - `{ percent: number }`\n- `field_change` - `{ fieldId: string, value: string | boolean | number | null, previousValue?: string | boolean | number | null }`\n- `ready`, `submit` - typically omitted"
                 }
               }
             }
--- a/apps/docs/openapi.json
+++ b/apps/docs/openapi.json
@@ -277,13 +277,10 @@
                   "format": "date-time",
                   "description": "ISO-8601 timestamp for the event"
                 },
-                "field": {
-                  "type": "string",
-                  "description": "Field id the event relates to (for `field_change`)"
-                },
-                "value": {
-                  "oneOf": [{ "type": "string" }, { "type": "boolean" }],
-                  "description": "Value associated with the event (for `field_change`)"
+                "data": {
+                  "type": "object",
+                  "additionalProperties": true,
+                  "description": "Event-specific payload emitted by the e-sign SDK. Shape depends on `type`:\n- `scroll` — `{ percent: number }`\n- `field_change` — `{ fieldId: string, value: string | boolean, previousValue?: string | boolean }`\n- `ready`, `submit` — typically omitted"
                 }
               }
             }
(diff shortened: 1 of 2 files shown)


apps/docs/openapi.json
```diff
@@ -197,19 +197,110 @@
...
+            "additionalProperties": true,
+            "description": "Optional application-specific metadata. Free-form object for any context you want to attach to the signing event (e.g. tenantId, contractId, custom audit fields)."
+          },
+          "certificate": {
+            "type": "object",
+            "description": "Configuration for the audit trail certificate page that is appended to the signed PDF.",
+            "additionalProperties": false,
+            "properties": {
+              "enabled": {
+                "type": "boolean",
```
caio-pizzol, line 299: spec says `enabled`, but the demo (`packages/esign/demo/src/DocxTab.tsx:147`) sends `enable`, and the demo server forwards it as-is. which name is right? the other needs to change in this PR.

apps/docs/solutions/esign/backend.mdx
```diff
@@ -228,7 +228,9 @@ The frontend sends the `onSubmit` payload plus a document reference and signer d
   "document": { "url": "https://example.com/agreement.docx" },
   "signer": {
     "name": "Jane Smith",
-    "email": "jane@example.com"
+    "email": "jane@example.com",
+    "ip": "203.0.113.42",
+    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36"
```
caio-pizzol: this is the payload the browser sends your backend, so showing `ip` and `userAgent` here suggests trusting values from the browser. the demo server sets them server-side instead (`packages/esign/demo/server/server.js:165-170`). drop them from this example, or note that the backend must overwrite them.

apps/docs/openapi.json
```diff
@@ -197,19 +197,110 @@
...
+                },
+                "timestamp": {
+                  "type": "string",
+                  "format": "date-time",
+                  "description": "ISO-8601 timestamp for the event"
+                },
+                "data": {
+                  "type": "object",
+                  "additionalProperties": true,
+                  "description": "Event-specific payload emitted by the e-sign SDK. Shape depends on `type`:\n- `scroll` — `{ percent: number }`\n- `field_change` — `{ fieldId: string, value: string | boolean, previousValue?: string | boolean }`\n- `ready`, `submit` — typically omitted"
```
caio-pizzol: `field_change` value is typed as `string | boolean` here, but the SDK also emits numbers and `null` (when a field is cleared). the SDK type is in `packages/esign/src/types.ts:3`.

```suggestion
                  "description": "Event-specific payload emitted by the e-sign SDK. Shape depends on `type`:\n- `scroll` - `{ percent: number }`\n- `field_change` - `{ fieldId: string, value: string | boolean | number | null, previousValue?: string | boolean | number | null }`\n- `ready`, `submit` - typically omitted"
```

caio-pizzol (commented):
hey @andrii! good call - updating this :)

three things to flag, all inline.

apps/docs/openapi.json
andrii-harbour: `enabled` is correct — it's what the services backend schema declares in `sign.schemas.js`, and `certificate` has `additionalProperties: false`, so the API would reject `{ enable: true }` outright. The demo in `packages/esign/demo/src/DocxTab.tsx:147` has a typo and needs updating separately. Keeping the spec as-is in this PR.

andrii-harbour:
@caio-pizzol done

caio-pizzol (approved):
lgtm

PR #2927, opened 2026-04-23
list tool improvements

andrii-harbour:


packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/lists-wrappers.ts
```diff
@@ -887,6 +902,174 @@ export function listsSeparateWrapper(
...
+  if (options?.dryRun) {
+    return {
+      success: true,
+      listId: separateResult.listId,
+      numId: separateResult.numId,
+      restartedAt: 1,
+    };
+  }
+
+  const setValueResult = listsSetValueWrapper(editor, { target: input.target, value: 1 }, options);
```
caio-pizzol: the first step edits the doc and bumps its version. the second step reuses the old version number, so its safety check fails and throws. the split lands but the restart never does.

```suggestion
  const setValueResult = listsSetValueWrapper(
    editor,
    { target: input.target, value: 1 },
    options ? { ...options, expectedRevision: undefined } : options,
  );
```

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/lists-wrappers.ts
```diff
@@ -887,6 +902,174 @@ export function listsSeparateWrapper(
...
+    anchorNodeId = targetSequence[0]?.address.nodeId ?? target.address.nodeId;
+    const lastOfTarget = targetSequence[targetSequence.length - 1]!;
+    const firstOfAdjacent = adjacent.sequence[0]!;
+    gapFromPos = lastOfTarget.candidate.pos + lastOfTarget.candidate.node.nodeSize;
+    gapToPos = firstOfAdjacent.candidate.pos;
+  }
+
+  const gapEmptyParagraphs =
+    gapFromPos < gapToPos
+      ? getBlockIndex(editor).candidates.filter(
```
caio-pizzol: this can delete things it shouldn't. it walks the whole doc, so an empty paragraph inside a table sitting between the two lists gets matched. a paragraph that holds only an image or a line break also gets matched, since neither has text. safer to only look at top-level blocks between the lists, and only delete a paragraph that's truly empty (no children).

caio-pizzol (commented):
hey @andrii-harbour! nice direction - merge and split fill a real gap.

two bugs to fix before merge, both inline. would also add dry-run tests for the new ops.

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/lists-wrappers.ts
andrii-harbour: done

andrii-harbour:
@caio-pizzol fixed

caio-pizzol:
hey @andrii-harbour! nice paraId fix.

i pushed a follow-up (1924ca8): fixed the split comment that promised atomic but the code runs in two steps, renamed the test that read backwards, and added checks so the merge/split/restart evals fail when the list doesn't actually change.

lgtm

caio-pizzol (approved):

PR #2929, opened 2026-04-23
feat: implement autofit table column width algorithm (SD-2502)

luccas-harbour:
## Summary

Implements the ECMA-376 AutoFit column-width algorithm (SD-2502) and rewires
table measurement around a clean fixed-vs-autofit split.

- **Two pure solvers** live in `layout-engine/measuring/dom/`:
  - `fixed-table-columns.ts` — fixed-width resolver (first-row `w:tcW` +
    explicit grid), deterministic and DOM-free.
  - `autofit-columns.ts` — AutoFit resolver that starts from the fixed-pass
    result and redistributes by per-cell content demand (min/max widths from
    `table-autofit-metrics.ts`).
- **Runtime measurement** (`measuring/dom/src/index.ts`) now delegates to
  those solvers instead of carrying its own width logic; the file drops
  ~350 lines of ad-hoc reconciliation.
- **pm-adapter** materializes the effective `tableLayout`/`tableWidth`
  attrs once at conversion time, so every downstream consumer sees the
  same authoritative shape.
- **Editor integration**: width-authoring edits (column resize, set-cell-
  width) are pinned to `tableLayout: fixed` and now mirror authored widths
  into `tcW` / `tblGrid` through a shared `table-attr-sync` helper, so the
  new solver's first-row-tcW precedence doesn't revert user edits.
- **Import fix**: fallback logical-grid construction honors
  `w:gridBefore`/`w:gridAfter` skips and cell spans, fixing a class of
  DOCX tables that previously collapsed to the wrong column count.

## Word parity

This is not a byte-exact reimplementation of Word's AutoFit. ECMA-376 does
not fully specify the algorithm — the spec explicitly calls out that the
described behavior is advisory and the final column widths are up to the
consumer application. Word itself diverges from the spec in several edge
cases (content-measurement heuristics, minimum-width fallbacks, rowspan
handling under tight constraints). This implementation follows the spec
where it is prescriptive and picks a consistent, documented behavior
where the spec leaves latitude, so rendered widths may differ from Word
in those cases. The tradeoffs are captured in the solver's unit tests.

## Notable fixes caught during integration

- `fix(autofit): redistribute widths by content demand in no-trigger path`
  — tables that didn't hit a redistribution trigger were rendering with
  stale fixed-pass widths; AutoFit now always redistributes.
- `fix(table-resize): keep tcW and borders aligned with authored widths`
  — column-resize edits were being reverted by the next measure pass
  because only `colwidth` was updated; also stops
  `syncExtractedTableAttrs` from promoting `borders: null` (the table
  extension's `renderDOM` called `Object.keys` on it and threw).


packages/super-editor/src/editors/v1/document-api-adapters/tables-adapter.ts
```diff
@@ -1583,19 +1516,18 @@ export function tablesSetColumnWidthAdapter(
...
       newColumns[columnIndex] = { col: Math.round(input.widthPt * POINTS_TO_TWIPS) }; // points → twips
-      tr.setNodeMarkup(tablePos, null, {
-        ...tableAttrs,
-        grid: serializeGridColumns(tableAttrs.grid, { ...normalizedGrid, columns: newColumns }),
-        userEdited: true,
-      });
+      tableAttrUpdates.grid = serializeGridColumns(tableAttrs.grid, { ...normalizedGrid, columns: newColumns });
     }
 
+    tr.setNodeMarkup(tablePos, null, buildWidthAuthoringTableAttrs(tableAttrs, tableAttrUpdates));
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/layout-engine/measuring/dom/src/autofit-columns.ts
```diff
@@ -0,0 +1,1173 @@
...
+      ranges.push({ index, amount });
+    }
+  }
+  return ranges;
+}
+
+function dedupeCells(cells: NormalizedCell[]): NormalizedCell[] {
+  const seen = new Set<string>();
+  return cells.filter((cell) => {
+    const key = `${cell.startColumn}:${cell.span}:${cell.cellIndex}`;
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/super-editor/src/editors/v1/document-api-adapters/tables-adapter.ts
```diff
@@ -1697,7 +1629,7 @@ export function tablesDistributeColumnsAdapter(
       tableAttrUpdates.grid = serializeGridColumns(tableAttrs.grid, { ...normalizedGrid, columns: newColumns });
     }
 
-    tr.setNodeMarkup(tablePos, null, tableAttrUpdates);
+    tr.setNodeMarkup(tablePos, null, buildWidthAuthoringTableAttrs(tableAttrs, tableAttrUpdates));
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

caio-pizzol (commented):
hey @luccas-harbour, really like how you split this into four focused modules - much easier to follow than the old all-in-one path :)

i pushed a370046 to the branch with a behavior test for the new column-drag width sync, and uploaded an autofit fixture to the corpus so the layout and visual checks pick it up automatically.

left three inline comments.

packages/layout-engine/measuring/dom/src/autofit-columns.ts
```diff
@@ -0,0 +1,1196 @@
...
+  const basis = widths.reduce((sum, width, index) => {
+    if (growableColumns != null && !growableColumns.has(index)) {
+      return sum;
+    }
+    return sum + Math.max(width, 1);
+  }, 0);
+  const slack = targetWidth - currentTotal;
+  if (basis <= 0) {
+    return widths.map(() => targetWidth / widths.length);
+  }
```
caio-pizzol, line 717: when `basis` ends up at zero (every growable column already at 0 px), this falls back to giving every column the same width, which wipes out the trigger handling you just did above. should this branch skip the protected columns the same way the non-zero branch does?

packages/layout-engine/measuring/dom/src/table-autofit-metrics.ts
```diff
@@ -0,0 +1,857 @@
...
+  return stableSerialize({
+    id: table.id,
+    attrs: table.attrs ?? null,
+    columnWidths: table.columnWidths ?? null,
+    rowCount: table.rows.length,
+    maxWidth: Math.max(1, Math.round(options.maxWidth)),
+    layoutEpoch: options.layoutEpoch ?? null,
+    cellMetricKeys: options.cellMetricKeys,
+  });
+}
```
caio-pizzol, line 311: the cache key picks up cell-level changes through `cellMetricKeys`, but it misses row-level placement. if someone tweaks `gridBefore`/`gridAfter` on a row or reorders rows without touching cell metrics, this would reuse a stale cached result. worth folding the working grid (or `fixedLayout`) into the key?

packages/layout-engine/measuring/dom/src/index.ts
```diff
@@ -3056,6 +2775,82 @@ async function measureTableBlock(block: TableBlock, constraints: MeasureConstrai
...
+  return {
+    contentMetrics,
+    cellMetricKeys: contentMetrics.cellMetricKeys,
+  };
+}
+
+/**
+ * Resolve a table-cell border width in pixels for content-box calculations.
+ */
+function getMeasuredCellBorderWidthPx(border: { style?: string; width?: number } | undefined): number {
```
caio-pizzol: looks like this got stranded by the refactor - the autofit path now uses `getCellBorderWidthPx` from `table-autofit-metrics.ts` and nothing else calls this. safe to delete?

packages/layout-engine/measuring/dom/src/autofit-columns.ts
luccas-harbour: good catch! fixed

packages/layout-engine/measuring/dom/src/table-autofit-metrics.ts
luccas-harbour: good one. I expanded the cache key to handle this case

packages/layout-engine/measuring/dom/src/index.ts
luccas-harbour: done, removed it

harbournick:
@luccas-harbour I think this PR regresses some docs in an unintended way. Check my comment in linear pls.

luccas-harbour:
@harbournick I pushed some more commits addressing these problems. I added more details to the linear ticket.

harbournick (approved):
LGTM

PR #2955, opened 2026-04-27
feat(super-editor): canonical FieldInstance types and instruction tokenizer (SD-2737)

caio-pizzol:
First chunk of the canonical field substrate. Adds the `FieldInstance` type, the single durable representation of a Word field code, and a tokenizing parser for field instructions. No importer, PM, or exporter changes yet; later chunks build on these.

The tokenizer is fidelity-first: the linear token stream reconstructs the input character-for-character, so downstream code can rebuild the instruction without losing whitespace, quoting, or switch ordering. Malformed input falls back to `opaque` tokens rather than throwing, so import always produces a usable stream.

`parsedArgs` (family + positional + switches) is a derived view, not a second source of truth. A `switch` token may carry an embedded `arg`; helpers honor it and close the switch slot so a following linear token cannot silently re-attach.

**Review**: tokenizer round-trip is the load-bearing invariant. `reconstructInstruction(tokenizeInstruction(s)) === s` for every fixture. Other helpers can evolve as later chunks land; this one cannot regress.
**Verified**: `vitest run instruction-tokenizer.test.ts` → 46/46 pass.


caio-pizzol:
Closing as superseded. Restarting Phase 0 (SD-2737) from `main` with shadow-instrumentation sequencing: build `FieldInstance` alongside the existing pipeline first, and defer node-shape changes until `rawField` has pm-adapter/layout coverage. Stacked branches retained for reference.

PR #2958, opened 2026-04-27
fix: sdt field style

artem-harbour:
Linear: SD-2744

Fix inline SDT outline collapse in Template Builder after `superdoc@1.28` strut update.

- Root cause: line containers now use `font-size: 0`; inline SDT wrappers inherited collapsed metrics, so borders rendered as a thin sliver.
- Change: restore and continuously sync inline SDT wrapper typography from current run font size (with fallback), in both geometry and run-based render paths.
- Result: SDT borders correctly wrap text and stay in sync when field text `font-size` changes.

Before
Screenshot 2026-04-27 at 17 40 23

After
Screenshot 2026-04-27 at 17 41 23

### DIFF
--- a/packages/layout-engine/painters/dom/src/index.test.ts
+++ b/packages/layout-engine/painters/dom/src/index.test.ts
@@ -2345,6 +2345,185 @@ describe('DomPainter', () => {
     expect(wrapper.textContent).toContain('controlled text');
   });
 
+  it('keeps inline SDT wrapper font-size in sync when run font-size changes', () => {
+    const block: FlowBlock = {
+      kind: 'paragraph',
+      id: 'inline-sdt-font-sync',
+      runs: [
+        {
+          text: 'Company Address',
+          fontFamily: 'Arial, sans-serif',
+          fontSize: 14.6667,
+          pmStart: 2044,
+          pmEnd: 2059,
+          sdt: {
+            type: 'structuredContent',
+            scope: 'inline',
+            id: '574416526',
+            tag: 'inline_text_sdt',
+            alias: 'Company Address',
+            lockMode: 'unlocked',
+          },
+        },
+      ],
+      attrs: {},
+    };
+
+    const measure: Measure = {
+      kind: 'paragraph',
+      lines: [
+        {
+          fromRun: 0,
+          fromChar: 0,
+          toRun: 0,
+          toChar: 15,
+          width: 180,
+          ascent: 12,
+          descent: 4,
+          lineHeight: 20,
+        },
+      ],
+      totalHeight: 20,
+    };
+
+    const layout: Layout = {
+      pageSize: { w: 612, h: 792 },
+      pages: [
+        {
+          number: 1,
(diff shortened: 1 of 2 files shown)


packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -6650,9 +6651,10 @@ export class DomPainter {
           // If this run has inline SDT, add to or create wrapper
           if (resolved && this.doc) {
             if (!currentInlineSdtWrapper) {
-              currentInlineSdtWrapper = this.createInlineSdtWrapper(resolved.sdt);
+              currentInlineSdtWrapper = this.createInlineSdtWrapper(resolved.sdt, run);
               currentInlineSdtId = runSdtId;
             }
+            this.syncInlineSdtWrapperTypography(currentInlineSdtWrapper, run);
```
caio-pizzol: heads-up: this runs once per chunk of text and the last one wins (tried it: bumping just "text" to 36pt in `sdts-basic.docx` makes the border match the big word and squish the small ones). worth a follow-up to pick a rule, but out of scope for SD-2744.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -7040,9 +7042,10 @@ export class DomPainter {
    * Create an inline SDT wrapper `<span>` with className, layoutEpoch, dataset, and label.
    * Shared by both the geometry and run-based rendering paths.
    */
-  private createInlineSdtWrapper(sdt: SdtMetadata): HTMLElement {
+  private createInlineSdtWrapper(sdt: SdtMetadata, runForSizing?: Run): HTMLElement {
     const wrapper = this.doc!.createElement('span');
     wrapper.className = DOM_CLASS_NAMES.INLINE_SDT_WRAPPER;
+    this.syncInlineSdtWrapperTypography(wrapper, runForSizing);
```
caio-pizzol: both callers already sync right after creating the wrapper, so this line does nothing the first time around. dropping it leaves the loop as the only place that sets the size (and `runForSizing?` can come off the arg list).

```suggestion
```

packages/layout-engine/painters/dom/src/index.test.ts
```diff
@@ -2345,6 +2345,97 @@ describe('DomPainter', () => {
     expect(wrapper.textContent).toContain('controlled text');
   });
 
+  it('keeps inline SDT wrapper font-size in sync when run font-size changes', () => {
```
caio-pizzol, line 2348: worth a second case with two chunks of text at different sizes inside one field, locks in whichever rule you pick.

caio-pizzol (approved):
hey @artem-harbour! good call, this fixes the sliver cleanly :)

approving, feel free to merge after addressing whatever feels relevant.

side note: nothing in our test suite caught the original 1.28 sliver. worth adding `sdts-basic.docx` to the visual tests as a follow-up.

PR #2969, opened 2026-04-28
feat: add simple word addin

mattConnHarbour:
Barebones word add-in example


caio-pizzol:
hey @mattConnHarbour, I don’t think this should live under `examples/`.

This is more than a small example. It is a full setup with a Word add-in, a server, and a web app. If we keep it here, people may think this is the path we recommend.

I’m going to close this PR for now. Let’s chat one-on-one about the right shape for this.

PR #2974, opened 2026-04-28
fix(sdk): pin bun version to 1.3.11 and update ignoreDeps in renovatejson5

andrii-harbour:
- Updated the bun version in release workflows to 1.3.11 due to macOS signing issues with versions 1.3.12 and above.
- Added bun to the ignoreDeps list in renovate.json5 to prevent automatic updates until the signing issue is resolved.


.github/workflows/release-cli.yml
```diff
@@ -62,7 +62,11 @@ jobs:
 
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: 1.3.13
+          # DO NOT BUMP without verifying macOS darwin-arm64 binary signing.
```
caio-pizzol, line 65: worth linking SD-2756 in the comment so this doesn't sit at 1.3.11 indefinitely.

caio-pizzol (approved):
hey @andrii-harbour! good call :)

approving. fyi `release-stable.yml:56` is still on 1.3.13 and runs the same `bun build --compile` path via `release-local-stable.mjs` -> SDK semantic-release -> `sdk-release-publish.mjs --npm-only` -> `build:native:all`. probably wants the same pin in a follow-up.

PR #2978, opened 2026-04-28
feat(document-api): editor.doc.comments.onChange subscriber (SD-2790)

caio-pizzol:
Adds a public subscription primitive on `editor.doc.comments` so consumers of the Document API can react to comments-collection changes without reaching across the React/Vue wrapper for `superdoc.on('commentsUpdate')`.

The dropin-assessment lab routed through the SuperDoc-instance event because no `editor.doc`-level path existed. That coupling would leak into a migration guide and tie consumers to a wrapper-side event we'd then have to keep stable.

- Listener takes no payload; consumers call `comments.list()` for fresh state. Keeps onChange cheap to fire on `commentsUpdate` events that don't change the list (e.g., pointer focus ticks), and lets consumers debounce as they see fit.
- Engine implementation coalesces bursts to one notification per microtask so multi-step transactions and DOCX reload don't storm subscribers. Listener errors are swallowed so one buggy subscriber can't wedge the editor.
- Symmetric with `selection.onChange` from SD-2668 — both are push-based subscriptions, so `comments.onChange` is added to `META_MEMBER_PATHS` in `check-contract-parity` rather than `OPERATION_DEFINITIONS`.

Part of SD-2667 (drop-in replacement primitives). SD-2791 (trackChanges.onChange) is the symmetric pair and lands next.

**Verified**: document-api 1396/0; super-editor 11842/0/13 skipped; contract parity 389/389; generate:all clean.

### DIFF
--- a/packages/document-api/scripts/check-contract-parity.ts
+++ b/packages/document-api/scripts/check-contract-parity.ts
@@ -30,16 +30,18 @@ import { buildDispatchTable } from '../src/invoke/invoke.js';
  *   implementation, so it is intentionally excluded from the RPC
  *   dispatch surface and the CLI command catalog. Direct calls through
  *   `editor.doc.ranges.scrollIntoView()` are still supported.
- * - `selection.onChange` is a subscription primitive (push-based, no
- *   request/response shape) rather than a request-response operation,
- *   so it is not represented in `OPERATION_DEFINITIONS` / schemas /
- *   dispatch. Direct calls through `editor.doc.selection.onChange()`
+ * - `selection.onChange`, `comments.onChange` are subscription primitives
+ *   (push-based, no request/response shape) rather than request-response
+ *   operations, so they are not represented in `OPERATION_DEFINITIONS` /
+ *   schemas / dispatch. Direct calls through
+ *   `editor.doc.selection.onChange()` and `editor.doc.comments.onChange()`
  *   are still supported.
  */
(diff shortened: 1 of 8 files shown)


caio-pizzol:
Closing note: architectural pivot.

Researched how Tiptap, CKEditor 5, TinyMCE, Lexical, Slate, ProseMirror, and Monaco expose UI customization. The consensus: subscription primitives belong on a browser-only UI controller, not on the Document API contract.

`editor.doc.*` should stay request/response, server+client. Adding `onChange` forces a `META_MEMBER_PATHS` exemption (the same one that already excludes `selection.onChange` and `ranges.scrollIntoView` from contract-parity) and teaches consumers to subscribe through the document contract instead of through a UI layer they own.

The replacement work moves to a new package, tracked under SD-2667:

```ts
import { createSuperDocUI } from 'superdoc/ui';

const ui = createSuperDocUI({ superdoc });
ui.comments.subscribe(({ snapshot }) => render(snapshot));
ui.comments.resolve(id);
ui.commands.bold.observe(({ active }) => boldBtn.set(active));
ui.destroy();
```

Three lenses over one selector substrate: domain events (Monaco-style), per-command observables (CKEditor-style), and arbitrary selectors (Tiptap/Slate-style). Built-in `SuperToolbar.vue` and `CommentsLayer/` will refactor to consume the same controller.

PR #2986, opened 2026-04-29
fix(comments): prevent flicker when clicking comment with tracked changes

### LINKED ISSUE #2861: Rapid flicker when clicking a comment
### What happened?

When I open https://www.superdoc.dev/ and click on the third comment (Added "WYSIWYG "), both the comment and the highlighted text in the document flicker rapidly.

### Steps to reproduce

1. Open https://www.superdoc.dev/
2. Navigate to the example
3. Click the third comment (Added "WYSIWYG ")
4. Observe flickering

### SuperDoc version

1.27.0

### Browser

Chrome

### Additional context

This occurs regardless of the window size. If the window is too narrow, you may have to scroll up to bring the comment back into view.

gpardhivvarma:
## Summary

Closes #2861

Fixes rapid flickering when clicking a comment that contains a tracked change inside inline structured content.

## Root Cause

When a comment is activated via sidebar click, `comments-plugin.js` correctly sets `activeThreadId` and returns early. However, the resulting DOM update triggers a ProseMirror selection-sync transaction. This secondary transaction hits position-based active comment detection (`getActiveCommentId`), which fails to resolve the comment inside tracked-change/structured-content nodes and returns `null` — clearing the active comment, triggering another DOM rebuild, and repeating in a loop.

## Fix

Add `explicitlySetThreadId` flag to CommentsPlugin state. When a comment is activated via explicit action (`setActiveComment` meta), the flag is set. On subsequent selection-only transactions:

- If cursor is still inside the explicitly activated comment's range → **skip position-based detection** (breaks the flicker loop)
- If cursor moves outside the comment range → **clear flag**, resume normal detection
- On document changes → **clear flag** (user is editing, detection should resume)

This extends the same principle already used at line 426-428 where `setCursorById` skips `view.focus()` to prevent DOM selection sync from overriding the active thread.

Also removes dead code: `trChangedActiveComment` variable was always `false` at the detection block because `setActiveComment` returns early at line 488.

## Changes

| File | Change |
|------|--------|
| `comments-plugin.js` | Add `explicitlySetThreadId` to plugin state, guard position-based detection, clear on cursor-out and doc-change, remove dead code |
| `comments-plugin.test.js` | 6 new tests |

## Tests

1. Flag set on explicit activation
2. Preserved on cursor-within (collapsed selection)
3. Cleared when cursor moves outside comment range
4. Preserved with tracked-change + comment marks (the actual bug scenario)
5. Range selections bypass the guard (non-collapsed)
6. Cleared on document change

## Test plan

- [x] `comments-plugin.test.js` — 66/66 passing
- [ ] Manual: open superdoc.dev, click the third comment (Added "WYSIWYG") — should not flicker
- [ ] Manual: click a comment, then click elsewhere — should deactivate normally
- [ ] Manual: click a comment, then type — detection resumes after edit


gpardhivvarma:
@caio-pizzol

packages/super-editor/src/editors/v1/extensions/comment/comments-plugin.js
```diff
@@ -483,6 +488,7 @@ export const CommentsPlugin = Extension.create({
             return {
               ...pluginState,
               activeThreadId: newActiveThreadId,
+              explicitlySetThreadId: newActiveThreadId,
```
caio-pizzol: the `explicitlySetThreadId` flag isn't doing the work anymore - the new fix handles the bug a different way. keep it, or drop in this PR? happy to push the removal.

packages/super-editor/src/editors/v1/extensions/comment/comments-plugin.js
```diff
@@ -507,11 +513,30 @@ export const CommentsPlugin = Extension.create({
...
+              selectionContainsThread(newEditorState.doc, selection, pluginState.explicitlySetThreadId)
+            ) {
+              // Selection is still inside the explicitly activated comment — skip detection.
+              return { ...pluginState };
+            }
+
+            // Cursor moved outside the explicitly set comment — clear the flag
+            // and allow position-based detection to take over.
+            if (pluginState.explicitlySetThreadId) {
+              pluginState.explicitlySetThreadId = null;
```
caio-pizzol: this writes into the old state before copying it. safer to put `null` on the copy.

```suggestion
            // Cursor moved outside the explicitly set comment - clear the flag
            // and allow position-based detection to take over.
            pluginState = pluginState.explicitlySetThreadId
              ? { ...pluginState, explicitlySetThreadId: null }
              : pluginState;
```

same at L578.

caio-pizzol (commented):
hey @gpardhivvarma! good call going after this, but the live demo still flickered with the original guard.

clicking a tracked-change comment fires two selection changes back to back: cursor first, then a range over the whole tracked span. your guard only kicks in for a cursor, so the range part still cleared the active comment and the host kept re-setting it.

i pushed a one-line change to your branch (range selections no longer count as "user moved off"), plus unit and behavior tests. flicker went from ~700 class flips per 2s to 2.

left two inline.

lgtm once you decide on the flag.

packages/super-editor/src/editors/v1/extensions/comment/comments-plugin.js
gpardhivvarma: I'd keep the flag. Your `isNonCollapsedClear` handles the range selection case (SDT wrapper NodeSelection), and the flag handles the collapsed cursor case (DOM-sync selection transactions). Together they cover both sides of the flicker loop — defense in depth.

If you'd prefer to drop it to reduce surface area, happy to push that too.

packages/super-editor/src/editors/v1/extensions/comment/comments-plugin.js
gpardhivvarma: Fixed — both sites now write on the copy instead of mutating the original.

packages/super-editor/src/editors/v1/extensions/comment/comments-plugin.js
gpardhivvarma: Dropped the flag — your isNonCollapsedClear handles the actual trigger. Pushed the removal along with the mutation fix.

gpardhivvarma:
@caio-pizzol 

Both comments addressed:

1. **Flag removed** — dropped `explicitlySetThreadId` entirely. Your `isNonCollapsedClear` handles the actual trigger.
2. **Mutation fixed** — moot now since the flag is gone, but the dead code removal and condition simplification from the original commit are preserved.

Ready for re-review.

caio-pizzol (approved):
hey @gpardhivvarma! thanks for cleaning that up.

re-ran the click test, no flicker. lgtm.

PR #2996, opened 2026-04-29
fix: correct tab positioning for negative-left and hanging-indent paragraphs (SD-2626)

luccas-harbour:
## Summary

Fixes [SD-2626] — DOCX paragraphs with negative-left tab stops and/or hanging indents that pull the first-line origin past the left margin were rendering tabs at the wrong x positions. Word treats a leading tab in those layouts as a jump back to the left margin (with the tab advance equal to the indent magnitude), but our pipeline was either skipping straight to the first default tab interval, double-compensating right-aligned content, or losing the correct tab-end boundary on segments that follow the tab.

Each fix is a small, scoped change; the bulk of the work is making the tab geometry — layout x, paint x, and the tab's drawn end — flow consistently through the contract, the two measurers, and the painter.

## What changed

**Tab-stop seeding** (`contracts/src/engines/tabs.ts`)
- Seed an implicit start-aligned stop at pos 0 when a hanging indent pulls the first-line origin past the left margin (`ccc1f5c8d`).
- Seed an implicit stop at `|leftIndent|` for paragraphs with negative left indent and no hanging, so a leading tab snaps back to the margin instead of the first default interval (`042f7dbb7`, `88e6a45e3`).
- Decide implicit margin/indent stops from raw (unclamped) paragraph indents while default-interval generation keeps the normalized values (`20bdde405`).
- Mark stops produced from authored paragraph tabs as `explicit` so the negative-left compensation only applies to generated/default stops (`d273e7450`).
- Extract `TAB_STOP_POSITION_TOLERANCE_TWIPS` to replace the magic `20` twips constant (`994cf34ee`).

**Negative-indent line geometry** (`layout-bridge/src/remeasure.ts`, `measuring/dom/src/index.ts`)
- Stop clamping `indentLeft`/`indentRight` to zero in `remeasureParagraph` so negative indents expand line width into the margin area, matching `measuring/dom` and keeping tab math in one coordinate space (`0210310d9`).
- Extend negative-left paint-x compensation to wrapped body lines under a hanging indent — they share the body indent so a leading start-tab needs the same treatment as the no-hanging case (`9f04cf8c2`).
- Drop the redundant `groupStartX - effectiveIndent` offset for right-aligned/center/decimal tab groups (the group already lives in compensated coordinates), fixing right-aligned content overshooting its stop (`770acf36e`).

**Tab segment boundary** (`contracts/src/engines/tabs.ts`, `painters/dom/src/renderer.ts`, both measurers)
- Add `LineSegment.precedingTabEndX` to the contract; both measurers stamp the uncompensated tab end on the first segment after a compensated tab while keeping the segment's paint x compensated (`fcfb0173b`).
- Painter prefers `precedingTabEndX` over the next segment's x when drawing the tab span/underline, so the tab width stays tied to the measured advance instead of stretching to the shifted text x.
- Carry the tab geometry onto a leading-space segment (not just the first word) so a tab followed by " text…" renders the space at the right x (`869011297`).
- Extract `PendingTabAlignStart` type and tighten `appendSegment`'s merge guard so positioned segments don't absorb following non-positioned ones (`d6ceb847a`, `58942314a`).

## Notes
- A document reproducing the issue was uploaded to the corpus


packages/layout-engine/measuring/dom/src/index.ts
```diff
@@ -2423,7 +2465,15 @@ async function measureParagraphBlock(block: ParagraphBlock, maxWidth: number): P
...
             }
-            appendSegment(currentLine.segments, runIndex, wordStartChar, wordEndNoSpace, wordOnlyWidth, explicitXHere);
+            appendSegment(
+              currentLine.segments,
+              runIndex,
+              wordStartChar,
+              wordEndNoSpace,
+              wordOnlyWidth,
+              explicitXHere,
+              wordIndex === 0 ? consumeSegmentPrecedingTabEndX() : undefined,
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/layout-engine/contracts/src/engines/tabs.ts
```diff
@@ -146,8 +169,12 @@ export function computeTabStops(context: TabContext): TabStop[] {
...
-    const hasClearStop = clearPositions.some((clearPos) => Math.abs(clearPos - pos) < 20);
+    const hasExplicitStop = filteredExplicitStops.some(
+      (s) => Math.abs(s.pos - pos) < TAB_STOP_POSITION_TOLERANCE_TWIPS,
+    );
+    const hasClearStop = clearPositions.some(
+      (clearPos) => Math.abs(clearPos - pos) < TAB_STOP_POSITION_TOLERANCE_TWIPS,
+    );
 
     // Default stops must be >= leftIndent (for body text alignment)
     const isValidDefault = pos >= leftIndent;
```
caio-pizzol, line 202: paste into a test:
`computeTabStops({ explicitStops: [], defaultTabInterval: 720, paragraphIndent: { left: 720, hanging: 720 } })`

returns two stops at pos=720. the new implicit stop and the default one collide because the default loop only dedupes against the original explicit list, not the implicit additions. fix: have it check `stops` too.

packages/layout-engine/measuring/dom/src/index.ts
```diff
@@ -2381,6 +2426,10 @@ async function measureParagraphBlock(block: ParagraphBlock, maxWidth: number): P
         }
 
         if (shouldBreak) {
+          if (wordIndex === 0 && hasPendingSegmentTabGeometry) {
+            hasPendingSegmentTabGeometry = false;
+            clearPendingPrecedingTabEndX();
+          }
```
caio-pizzol, line 2488: on wrap you clear `precedingTabEndX` but not `segmentStartX`, so the wrapped first word still picks up the old tab position via `explicitXHere = segmentStartX`. should the wrap reset both?

packages/layout-engine/layout-bridge/src/remeasure.ts
```diff
@@ -875,12 +912,20 @@ const applyTabLayoutToLines = (
...
         } else {
           cursorX = Math.max(cursorX, relativeTarget);
         }
       } else {
         cursorX = Math.max(cursorX, relativeTarget);
+        pendingTabAlignStartX = {
+          layoutX: relativeTarget,
+          paintX: paintTarget,
+          ...(precedingTabEndX !== undefined ? { precedingTabEndX } : {}),
+        };
```
caio-pizzol, line 973: before this PR only end/center/decimal tabs were tracked here - now every start tab is too. is that on purpose? it gives explicit start-tab paragraphs a `segment.x` they didn't have before.

caio-pizzol (commented):
hey @luccas-harbour! nice fix - tested it on a real Word doc and the geometry matches :)

[word-fixture-pr-2996.docx](https://github.com/user-attachments/files/27435994/word-fixture-pr-2996.docx)

one bug + two questions inline. also worth running `pnpm test:layout` against a recent baseline before merge - removing the negative-indent clamp at `remeasure.ts:1184` changes layout for every paragraph with a negative left indent, not just the tab cases.

packages/layout-engine/contracts/src/engines/tabs.ts
luccas-harbour: fixed

packages/layout-engine/measuring/dom/src/index.ts
luccas-harbour: fixed

packages/layout-engine/layout-bridge/src/remeasure.ts
luccas-harbour: this is intentional. I added an explanatory comment to make it explicit

luccas-harbour:
I addressed all your comments @caio-pizzol 
also, I carefully ran all visual tests and confirmed that all changes were due to improvements in rendering.

packages/layout-engine/measuring/dom/src/index.ts
```diff
@@ -2431,6 +2474,11 @@ async function measureParagraphBlock(block: ParagraphBlock, maxWidth: number): P
         }
 
         if (shouldBreak) {
+          if (wordIndex === 0 && hasPendingSegmentTabGeometry) {
+            hasPendingSegmentTabGeometry = false;
+            segmentStartX = undefined;
+            clearPendingPrecedingTabEndX();
```
caio-pizzol: important: this clears the leak only for the standard wrap branch. the char-split branch (around line 2275) and the leading-space-wrap branch (around 2200) push the tab-aware line and null `pendingTabAlignment` but leave `pendingSegmentPrecedingTabEndX` and `segmentStartX` in place. when an end/center/decimal tab follows the wrap, the next post-tab segment consumes the stale value via `consumeSegmentPrecedingTabEndX()`.

verified with `pnpm --filter @superdoc/measuring-dom test --run -- src/index.test.ts` (with the repro test from the dossier) -> `expected undefined, received 40` on the Z segment's `precedingTabEndX`.

likely fix: pull this 4-line clear into a `clearWrapState()` helper next to `clearPendingPrecedingTabEndX` and call it from all three wrap exits. fix not verified.

<details>
<summary>repro test</summary>

```ts
it('regression: char-split wrap leaks precedingTabEndX onto later end-aligned tab', async () => {
  const block: FlowBlock = {
    kind: 'paragraph',
    id: 'leak',
    runs: [
      { text: 'First', fontFamily: 'Arial', fontSize: 16 },
      { kind: 'lineBreak' },
      { kind: 'tab', text: '\t', tabIndex: 0 },
      { text: 'SupercalifragilisticexpialidociousVeryLong', fontFamily: 'Arial', fontSize: 16 },
      { kind: 'tab', text: '\t', tabIndex: 1 },
      { text: 'Z', fontFamily: 'Arial', fontSize: 16 },
    ],
    attrs: { indent: { left: -40, hanging: 20 }, tabs: [{ pos: 200 * 15, val: 'end', leader: 'dot' }] },
  };
  const m = expectParagraphMeasure(await measureBlock(block, 120));
  const z = m.lines.flatMap((l) => l.segments ?? []).find((s) => s.runIndex === 5);
  expect(z?.precedingTabEndX).toBeUndefined();
});
```

</details>

caio-pizzol (commented):
hey @luccas-harbour! thanks for addressing last round.

> [!IMPORTANT]
> one more thing inline.

the round-2 wrap fix only patched the standard wrap branch. when a compensated start tab is followed by a wrap that takes the char-split path (overlong word) or the leading-space-wrap path, the prior tab's `precedingTabEndX` leaks onto the next post-tab segment. on a Word-authored fixture with this OOXML pattern, the wrapped text continuation paints outside the page text frame and the dot leader before the page number renders wrong.

[sd-2626-r2-leak.docx](https://github.com/user-attachments/files/27454953/sd-2626-r2-leak.docx)

_screenshot:_ Word render of `sd-2626-r2-leak.docx` - clean dot leader, `VeryLong .... 9` wraps cleanly
Screenshot 2026-05-06 at 16 00 37

_screenshot:_ SuperDoc round-2 head render of the same file - `usVeryLong` painted at the page edge outside the body frame
Screenshot 2026-05-06 at 16 00 32

expected: `packages/layout-engine/measuring/dom/src/index.ts` / avoid: `tabs.ts`, `remeasure.ts`, `renderer.ts`

<details>
<summary>agent context</summary>

```json
{
  "schema": "review_handoff_v1",
  "review_id": "sd-2996-r2",
  "findings": [
    {
      "id": "wrap-state-leak",
      "severity": "important",
      "anchors": [
        { "file": "packages/layout-engine/measuring/dom/src/index.ts", "line": 2275, "quote": "if (currentLine && currentLine.width > 0 && currentLine.segments && currentLine.segments.length > 0) { ... lines.push(completedLine); pendingTabAlignment = null; ... currentLine = null; }" },
        { "file": "packages/layout-engine/measuring/dom/src/index.ts", "line": 2200, "quote": "// Space doesn't fit - finish current line and start new one with the space ... pendingTabAlignment = null;" },
        { "file": "packages/layout-engine/measuring/dom/src/index.ts", "line": 2480, "quote": "if (wordIndex === 0 && hasPendingSegmentTabGeometry) { hasPendingSegmentTabGeometry = false; segmentStartX = undefined; clearPendingPrecedingTabEndX(); }" }
      ],
      "claim": "the round-2 fix clears segmentStartX + pendingSegmentPrecedingTabEndX only on the standard wrap branch at 2477-2480. the char-split branch (2256-2275) and the leading-space-wrap branch (2168-2200) push the tab-aware line and null pendingTabAlignment but leave pendingSegmentPrecedingTabEndX in place. when an end/center/decimal tab follows the wrap, the next post-tab segment consumes the stale value via consumeSegmentPrecedingTabEndX().",
      "failure_mode": "on a paragraph with negative-left + hanging indent + custom right-tab-with-leader where a long word forces char-split inside a narrow page, the wrapped continuation paints outside the page text frame and the dot leader before the page number renders at the wrong start position.",
      "verification": [
        { "command": "pnpm --filter @superdoc/measuring-dom test --run -- src/index.test.ts (with probe test)", "result_summary": "failed: expected undefined, received 40 on Z.precedingTabEndX", "status": "executed" },
        { "command": "loaded sd-2626-r2-leak.docx in dev app, evaluate_script over .superdoc-line", "result_summary": "line 3 has phantom 159px span at left:40px and text usVeryLong at left:-53.33px (outside page frame)", "status": "executed" },
        { "command": "side-by-side render of the same OOXML in Word", "result_summary": "Word: dot leader before long word, clean wrap to VeryLong .... 9. SuperDoc round-2 head: usVeryLong painted outside page text frame, leader broken", "status": "executed" }
      ],
      "ruled_out": [
        { "hypothesis": "a future tab overwrites the stale value at the alignPendingTabForWidth assignment before any consumer reads it", "why_ruled_out": "end/center/decimal tabs route through measureTabAlignmentGroup / activeTabGroup, not alignPendingTabForWidth, so that assignment is not executed. consumeSegmentPrecedingTabEndX reads the stale value first." },
        { "hypothesis": "only synthetic test inputs trigger this; real Word docs won't", "why_ruled_out": "a Word-authored fixture with negative-left + hanging + custom right-tab-with-leader + a long word that char-splits in a narrow viewport reproduces the visible glitch." }
      ],
      "handoff": {
        "recommended_start": "extract a clearWrapState() helper next to clearPendingPrecedingTabEndX that nulls segmentStartX, sets hasPendingSegmentTabGeometry = false, and calls clearPendingPrecedingTabEndX. call it from all three wrap exits: leading-space-wrap, char-split, and replace the inline clear in the existing wrap-after-tab branch.",
        "suggested_fix": "helper extraction; behavior identical to the current standard-wrap clear, applied uniformly across all three wrap exits",
        "expected_files": ["packages/layout-engine/measuring/dom/src/index.ts", "packages/layout-engine/measuring/dom/src/index.test.ts"],
        "avoid_files": ["packages/layout-engine/contracts/src/engines/tabs.ts", "packages/layout-engine/layout-bridge/src/remeasure.ts", "packages/layout-engine/painters/dom/src/renderer.ts"]
      },
      "verification_after_fix": [
        { "command": "pnpm --filter @superdoc/measuring-dom test --run -- src/index.test.ts", "expected": "new regression test passes; no existing test regresses" },
        { "command": "reload sd-2626-r2-leak.docx in dev app", "expected": "wrapped continuation paints inside the page text frame; dot leader before 9 starts at the correct cursor position" }
      ]
    }
  ]
}
```

</details>

packages/layout-engine/measuring/dom/src/index.ts
luccas-harbour: good call. fixed it. I also re-ran all visual tests and it's all good.

caio-pizzol (approved):
hey @luccas-harbour! thanks for addressing last round.

Lgtm

PR #3076, opened 2026-05-01
refactor(superdoc): use narrowed doc local in #initDocuments (SD-2867 phase B)

caio-pizzol:
Stacked on #3074. Closes one TS2769 error in `#initDocuments`.

`Config.document` is typed `string | object | File | Blob | undefined`. The function captures it as `const doc = this.config.document` and then guards with `!!doc && typeof doc === 'object'` before calling `Object.keys`. The original code passed `this.config.document` (refetched) into `Object.keys`, which TS re-widens to the full union and rejects against the `Object.keys` overloads (none accept `undefined`).

Switching the argument to the narrowed local `doc` carries the guard through: at that point in the expression `doc` is `object`, which `Object.keys` accepts. Behavior is identical because both reads return the same field value within a single synchronous expression.

No public surface change; internal-only typing.

**Verified**: `pnpm --filter superdoc check:jsdoc` clean; `pnpm --filter superdoc build:es` declaration audit clean; `node tests/consumer-typecheck/typecheck-matrix.mjs` 33 passed, 0 failed.

### DIFF
--- a/packages/superdoc/src/core/SuperDoc.js
+++ b/packages/superdoc/src/core/SuperDoc.js
@@ -67,7 +67,10 @@ const DEFAULT_AWARENESS_PALETTE = Object.freeze([
 /** @typedef {import('./types/index.js').ExportParams} ExportParams */
 /** @typedef {import('./types/index.js').UpgradeToCollaborationOptions} UpgradeToCollaborationOptions */
 /** @typedef {import('./types/index.js').SurfaceRequest} SurfaceRequest */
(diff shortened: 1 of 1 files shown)


caio-pizzol:
Closing per SD-2828 scope reset: this is internal TS satisfaction (Object.keys narrowing), not a public-surface contract issue. Folding the remaining SuperDoc.js checkJs cleanup into a separate ticket grouped by error category, not stacked 2-line PRs.

PR #3079, opened 2026-05-01
feat(collaboration): emit comments-update event for remotely-added comments

mattConnHarbour:


packages/superdoc/src/core/collaboration/helpers.js
```diff
@@ -99,8 +100,28 @@ export const initCollaborationComments = (superdoc) => {
 
     if (currentUser.name === user.name && currentUser.email === user.email) return;
 
+    // Capture existing comment IDs before loading new state
+    const existingIds = new Set(superdoc.commentsStore.commentsList?.map((c) => c.commentId || c.importedId) || []);
```
caio-pizzol, line 104: the loader and this code disagree on which id wins when a comment has both. loader (line 19) prefers `importedId`, here `commentId` wins. so if an imported comment later gets a commentId, or two people import the same docx and pick different commentIds, this code treats it as new and fires `add` for a comment that's already there. fix: match the loader at lines 104 and 112.

```suggestion
    const existingIds = new Set(superdoc.commentsStore.commentsList?.map((c) => c.importedId ?? c.commentId) || []);
```

packages/superdoc/src/core/collaboration/helpers.js
```diff
@@ -99,8 +100,28 @@ export const initCollaborationComments = (superdoc) => {
...
+    const newComments = superdoc.commentsStore.commentsList?.filter((c) => {
+      const id = c.commentId || c.importedId;
+      return id && !existingIds.has(id);
+    });
+
+    if (newComments?.length && superdoc.emit) {
+      newComments.forEach((comment) => {
+        const commentValues = typeof comment.getValues === 'function' ? comment.getValues() : comment;
+        superdoc.emit('comments-update', {
+          type: comments_module_events.ADD,
```
caio-pizzol, line 120: `comments-store.js` emits `'add'`, `CommentsLayer.vue` emits `'new'` for UI creation. does the listener that scrolls to remote comments use `'add'`? if it keys off `'new'`, it won't fire.

caio-pizzol (commented):
hey @mattConnHarbour, direction's right :) one bug inline plus a question about event type.

PR #3116, opened 2026-05-04
[1/3] refactor(painter): consume only ResolvedLayout (SD-2836)

tupizz:
## Stack

| # | PR | Branch | Status |
|---|----|--------|--------|
| 1 | **this PR** | tadeu/sd-2836-1-consume-resolved-layout | open |
| 2 | #3117 | tadeu/sd-2836-2-collapse-legacy-api | open |
| 3 | #3118 | tadeu/sd-2836-3-lock-boundary | open |

> Review in order. PR2 and PR3 depend on this one; merge this first.

## What changed

First step of the SD-2836 "dumb painter" refactor. Migrates DomPainter's internal pipeline so every render path consumes `ResolvedLayout` while keeping the legacy public API working as a compatibility shim.

### Painter (`packages/layout-engine/painters/dom`)
- `paint()` body and the six top-level rendering methods now operate on `ResolvedLayout` / `ResolvedPage` / `ResolvedPaintItem`.
- Page-level helpers (decoration sections, virtualization, snapshot, virtualization pins) migrated to `ResolvedPage`.
- `currentLayout` is typed `ResolvedLayout | null`.
- Decoration provider callback signature migrated `Page` -> `ResolvedPage` (coupled migration; pulled forward into PR1 because painter internals call the provider).

### Contracts (`packages/layout-engine/contracts`)
- New `run-helpers.ts` hosting `expandRunsForInlineNewlines` and `sliceRunsForLine` (shared between painter and resolved-layout consumers).
- `resolved-layout.ts`: `fragment: Fragment` back-pointer added to `ResolvedFragmentItem`, `ResolvedTableItem`, `ResolvedImageItem`, `ResolvedDrawingItem`. `ResolvedPage` gains optional `columns` and `columnRegions`.

### Super-editor consumer
- `HeaderFooterSessionManager.rebuildRegions(resolvedLayout)`, `updateDecorationProviders(layout, resolvedLayout)`, `#stripFootnoteReserveFromBottomMargin(margins, page: ResolvedPage | null)`, `#computeExpectedSectionType(kind, page: ResolvedPage, ...)` migrated to `ResolvedPage`.
- `PresentationEditor.#updateDecorationProviders(layout, resolvedLayout)` plumbed through.

### Docs
- `packages/layout-engine/AGENTS.md` "DomPainter is Dumb" section updated.

## Why this split

This PR is large but focused: every change here is "consume ResolvedLayout instead of Layout/Fragment/PaintInput" inside the painter and its direct callers. The legacy public surface stays callable, so consumers don't break mid-stack. PR2 then collapses that surface; PR3 locks the boundary.

## Verification

- All package tests pass: `painter-dom`, `pm-adapter`, `layout-bridge`, `layout-engine`, `super-editor`.
- HEAD-vs-origin/main corpus snapshot diff: **0 changed / 407 unchanged** (rendering-neutral).

## Acceptance criteria

- [x] DomPainter renders only from `ResolvedLayout` internally
- [x] `Fragment` back-pointer available on resolved items for legacy paths
- [x] Decoration providers receive `ResolvedPage`
- [x] No regression in corpus snapshot diff

## Test plan

- [ ] CI green
- [ ] Reviewer confirms no painter method still reads from `Layout` directly
- [ ] PR2 collapses the legacy surface on top
- [ ] PR3 locks the boundary


packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -2480,24 +2440,22 @@ export class DomPainter {
...
 
     return Math.max(0, pageHeight - adjustedBottomMargin);
   }
 
   private renderDecorationSection(
     pageEl: HTMLElement,
-    page: Page,
+    page: ResolvedPage,
     pageIndex: number,
     kind: 'header' | 'footer',
```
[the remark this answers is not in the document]
tupizz: This concern is theoretical and doesn't apply to the current codebase:

- `@superdoc/painter-dom` is `"private": true` — not published to npm.
- No consumers of `createDomPainter` or `PageDecorationProvider` exist outside the layout-engine workspace and super-editor (verified via grep across `packages/superdoc/src/`, `packages/react/src/`, and `examples/`).
- The only callers are internal: `PresentationEditor` → `HeaderFooterSessionManager.createDecorationProvider` → `PresentationPainterAdapter` → painter. SuperDoc's public API exposes no hook for consumers to register their own provider, so no third-party `Page`-shaped provider exists in practice.

The type *is* re-exported via `superdoc`'s generated `.d.ts` — a consumer doing a deep `import` from `superdoc/dist/layout-engine/...` would see the signature change — but that's not a supported import path.

The whole point of SD-2836 ("dumb painter") is to make `ResolvedLayout` the single rendering contract, so adapting back to legacy `Page` at the provider call site would contradict the north-star. Resolving as a non-issue.

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -2718,14 +2676,13 @@ export class DomPainter {
...
+  private fullRender(layout: ResolvedLayout): void {
     if (!this.mount || !this.doc) return;
     this.mount.innerHTML = '';
     this.pageStates = [];
 
     layout.pages.forEach((page, pageIndex) => {
-      const resolved = this.getResolvedPage(pageIndex);
-      const pageSize = resolved ? { w: resolved.width, h: resolved.height } : (page.size ?? layout.pageSize);
+      const pageSize = { w: page.width, h: page.height };
       const pageState = this.createPageState(page, pageSize, pageIndex);
```
luccas-harbour: it seems that the width and height can be accessed directly from the `page` parameter so perhaps the `pageSize` parameter is redundant and could be removed? the same would apply to calls to `patchPage()`
what do you think?

packages/super-editor/src/editors/v1/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
```diff
@@ -2262,10 +2263,10 @@ export class HeaderFooterSessionManager {
    * Update decoration providers for header and footer.
    * Creates new providers based on layout results and sets them on this manager.
    */
-  updateDecorationProviders(layout: Layout): void {
+  updateDecorationProviders(layout: Layout, resolvedLayout: ResolvedLayout): void {
```
luccas-harbour: `updateDecorationProviders(layout, resolvedLayout)` seems to be the only place where both Layout and ResolvedLayout are used. `createDecorationProvider` only reads `layout.pages[].sectionIndex`/`number` + `page.sectionRefs` — all available on `ResolvedPage`. 

perhaps this could be changed and the `layout` parameter dropped?

luccas-harbour (changes requested):
hey @tupizz! I left a few inline comments

also here are some test coverage gaps found by AI:

1. `packages/super-editor/src/editors/v1/core/presentation-editor/tests/HeaderFooterSessionManager.test.ts`** (file-level — coverage gap)
> `rebuildRegions` is now the load-bearing entry from `ResolvedLayout`, but the existing tests in this file set `headerRegions` manually and bypass it. would be worth a focused test calling `rebuildRegions` with a synthetic `ResolvedLayout` that exercises `footnoteReserved > 0`, per-page `height` variation, and `sectionIndex > 0` so the resolved-side migration has direct coverage.

2. `packages/layout-engine/layout-resolved/src/resolveLayout.test.ts`** (file-level — coverage gap)
> one gap: the page-level `columns`/`columnRegions` fields forwarded at `resolveLayout.ts:309-310` aren't asserted anywhere and they feed `renderColumnSeparators` (user-visible). a one-liner asserting they make it onto `ResolvedPage` would close that.

packages/layout-engine/painters/dom/src/renderer.ts
tupizz: Good catch. Done in commit bc048e972: `pageSize` is dropped from both `createPageState` and `patchPage`; both now read `page.width` / `page.height` directly from `ResolvedPage`. The 4 call sites are simplified (no more `const pageSize = { w: page.width, h: page.height }` derivation), and the one redundant `applyStyles` after `createPageState` (the virtualization-margins one) reads `page.width`/`page.height` too.

packages/super-editor/src/editors/v1/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
tupizz: Good catch. Done in commit bc048e972: `createDecorationProvider` now takes `(kind, resolvedLayout: ResolvedLayout)`, and `updateDecorationProviders(resolvedLayout)` only takes the resolved layout. The internal section-first-page map iterates `resolvedLayout.pages` (sectionIndex/number both available on `ResolvedPage`), and the two fallback chains for `pageHeight` / `margins` walk through `resolvedLayout.pages[0]?.height` / `?.margins` instead of the legacy `layout.pageSize.h` / `layout.pages[0].margins`. PresentationEditor's `#updateDecorationProviders` plumbing simplified to match.

tupizz (commented):
Both coverage gaps addressed in commit bc048e972:

**1. `rebuildRegions` direct tests** — added a new `describe('rebuildRegions — ResolvedLayout entry')` block in `HeaderFooterSessionManager.test.ts` with three focused tests:
  - shrinks footer height by `footnoteReserved` and shifts its offset upward (page 1: untouched 36/720; page 2 with `footnoteReserved=24`: 12/744)
  - honors per-page height variation (3 pages of 792 / 1000 / 1400 → footer offsets 720 / 928 / 1328)
  - propagates `sectionIndex` from `ResolvedPage` onto built header/footer regions

**2. `columns` / `columnRegions` forwarding** — added a one-shot test in `resolveLayout.test.ts` asserting both fields make it onto `ResolvedPage` from a multi-region page input.

All 22 tests in `HeaderFooterSessionManager.test.ts` and 113 in `resolveLayout.test.ts` pass.

luccas-harbour (approved):
LGTM

PR #3120, opened 2026-05-04
SD-2664 - feature support word style update field interactions for TOC

chittolinag:
### Issue

Word-style "Update Field" interactions (context menu + F9) aren't wired up in SuperDoc. The document API already exposes `editor.doc.toc.update()`, but no user-facing action calls it. Scope of this PR is **Table of Contents only** — other field types are deferred.

### Proposed solution

- **Right-click → "Update table of contents"** when the cursor is inside a TOC.
- **F9** rebuilds every TOC in the document.
- Fixed a handful of issues in the rebuild path that surfaced once the action was actually exercised end-to-end (font/style preservation, dot leader, spacing, pasted-heading detection).
- Added unit tests for the context-menu item, the F9 command, and the rebuild helpers, plus a behavior test that loads a TOC docx, presses F9, and verifies the rebuilt entries match the document's headings.


packages/super-editor/src/editors/v1/core/super-converter/field-references/shared/toc-switches.ts
```diff
@@ -107,10 +107,17 @@ export function deriveIncludePageNumbers(
...
+ * - `undefined` → caller did not pass a separator (no \p switch). Returns
+ *   `undefined` so consumers fall back to Word's default (dots) instead of
+ *   treating "no \p" as an explicit opt-out.
+ * - `''` → \p was present but empty. Returns `'none'` (explicit opt-out).
+ * - non-empty string → mapped via SEPARATOR_TO_TAB_LEADER, or `undefined`
+ *   when the separator is not a known leader character.
  */
 function deriveTabLeader(separator: string | undefined): TocDisplayConfig['tabLeader'] | undefined {
-  if (!separator) return 'none';
+  if (separator === undefined) return undefined;
```
caio-pizzol, line 119: repro: open a TOC with a hyphen leader → `editor.doc.toc.configure({ target, patch: { tabLeader: 'none' } })` → `toc.update({ target, mode: 'all' })`. the rebuilt TOC comes back with dots, not none. on a TOC that already has no leader, configure does nothing - the stored text doesn't change. fix: write `\p ""` for explicit none, or store the choice in a separate field.

packages/super-editor/src/editors/v1/document-api-adapters/plan-engine/toc-wrappers.ts
```diff
@@ -257,13 +259,113 @@ interface MaterializedToc {
...
+ * Run-formatting marks (font, size, bold, italic, textStyle…) sampled from the
+ * first non-empty text run inside an actual TOC entry paragraph (styleId TOC1–TOC9).
+ * Skips sibling paragraphs like TOCHeading whose run formatting must not bleed
+ * into the rebuilt entries. Drops `link` (rebuilt from the source's bookmark)
+ * and `tocPageNumber` (belongs only to the page-number run).
+ */
+function readExistingTocEntryTextMarks(node: ProseMirrorNode): EntryTextMark[] {
+  let marks: EntryTextMark[] = [];
+  node.forEach((paragraph) => {
+    if (marks.length > 0 || paragraph.type.name !== 'paragraph') return;
```
caio-pizzol: repro: open [word-fixture-pr-312-bold.docx](https://github.com/user-attachments/files/27433723/word-fixture-pr-312-bold.docx)
 (first entry is bold), right-click → "Update table of contents". every entry comes out bold. Word resets to the linked paragraph styles (`TOC1`, `TOC2`, …) instead of copying from the first entry - probably the simpler match.

packages/super-editor/src/editors/v1/extensions/field-update/field-update.js
```diff
@@ -34,10 +37,37 @@ export const FieldUpdate = Extension.create({
...
+          // state. Set preventDispatch so it skips that.
+          if (editor?.doc?.toc?.update) {
+            const tocTargets = findAllTocNodes(state.doc)
+              .map((toc) => toc.commandNodeId)
+              .filter((id) => typeof id === 'string' && id);
+
+            if (tocTargets.length > 0) {
+              if (!dispatch) return true; // can()-style probe
+
+              for (const sdBlockId of tocTargets) {
```
caio-pizzol, line 64: repro: doc with two or more TOCs, press F9. first TOC gets real page numbers, the rest rebuild as `0`. each `toc.update` swaps `editor.state.doc`, so `getPageMap` rejects the stored map on the next iteration. fix: grab the page map once before the loop, or batch all updates into one transaction.

caio-pizzol (dismissed):
hey @chittolinag! the `\p` fix is the right call, but the rebuild path is regressing in a few spots.

pushed `bceff5248` - three failing tests for findings 1, 1b, and 2.

left a few inline.

needs work.

packages/super-editor/src/editors/v1/core/renderers/ProseMirrorRenderer.ts
```diff
@@ -19,6 +19,46 @@ import type { EditorRenderer, EditorRendererAttachParams } from './EditorRendere
...
+    }
+  }
+  if (!parentParagraph) return slice;
+
+  const styleId = (parentParagraph.attrs as { paragraphProperties?: { styleId?: string } } | undefined)
+    ?.paragraphProperties?.styleId;
+  if (typeof styleId !== 'string' || !HEADING_STYLE_RE.test(styleId)) return slice;
+
+  const wrapped = parentParagraph.type.create(parentParagraph.attrs, slice.content, parentParagraph.marks);
+  return new PmSlice(PmFragment.from(wrapped), 0, 0);
```
luccas-harbour, line 77: @chittolinag codex found an issue here and I was able to reproduce it in the browser. The problem is that if I copy just a part of a heading and not its entire contents, when I paste it, it is pasted as a heading and not as inline text.

Here's how to reproduce it:

1. Open an empty document
2. Add a paragraph with text like "Alpha Beta Gamma" and set its style to heading 1.
3. Press enter and type "Pasted text: "
4. Then go back to the first paragraph and select just the word "Beta" and press Cmd+C
5. Move the cursor back to the second line, after the colon, and press Cmd+V.

The pasted text goes to the next line, when you would have expected it to be inline like this "Pasted text: Beta" which is what you get in MS Word.

luccas-harbour: Here's codex' exact wording which could be helpful for fixing this:

```
  - [P2] Preserve inline copy semantics for partial headings — 
    packages/super-editor/src/editors/v1/core/renderers/ProseMirrorRenderer.ts:58-59
    For any non-empty selection inside a Heading paragraph whose slice is inline-only, this wraps the selected text in a closed paragraph slice. That means copying just a word from
    a heading and pasting it back into SuperDoc inserts a Heading block instead of inline text, because the custom SuperDoc slice is preferred on paste. This should be limited to
    selections that actually represent the whole heading paragraph.
```

luccas-harbour:
hey @chittolinag! this looks pretty cool!

I had a look at this one and found an issue where the text "No table of contents entries found" is not removed after entries are added to the document. Here's how to reproduce it:
a. Open the `test-corpus/basic/table-of-contents-sdt.docx` into the editor
b. Select the entire contents of the document (except for the TOC itself) and set the style to `Normal`.
c. Right-click the TOC and update it. The "No table of contents entries found" will now be shown
d. Select the "Lorem ipsum dolor" text and change the style to `Heading 1`
e. Right-click the TOC and update it again. You'll notice that the heading is added but the "no entries" text is still there.

I also found another one that I posted as an inline comment. ping me if you have any questions!

packages/super-editor/src/editors/v1/extensions/field-update/field-update.js
```diff
@@ -34,10 +37,50 @@ export const FieldUpdate = Extension.create({
...
+                    target: { kind: 'block', nodeType: 'tableOfContents', nodeId: sdBlockId },
+                    mode: 'all',
+                  });
+                } catch (error) {
+                  console.warn('[FieldUpdate] toc.update failed for', sdBlockId, error);
+                }
+              }
+
+              outerTr?.setMeta?.('preventDispatch', true);
+              return true;
```
luccas-harbour: hey @chittolinag, it looks like this early `return` statement will prevent other fields from being updated if the document contains a TOC

luccas-harbour (approved):
LGTM

PR #3128, opened 2026-05-04
feat(examples): vanilla custom UI toolbar example (SD-2929)

caio-pizzol:
First focused minimal example under \`examples/editor/custom-ui/\`. Single file, ~99 lines, demonstrates the cleaned-up controller surface from SD-2917 / SD-2918 / SD-2919 / SD-2920:

- No cast on \`createSuperDocUI({ superdoc })\`.
- \`ui.createScope()\` for lifecycle. \`ui.destroy()\` cascades.
- Per-command \`observe(state => ...)\` so each button only re-renders when its own command flips.
- \`BUILT_IN_COMMAND_IDS\` + \`ui.commands.has(id)\` + \`ui.commands.require(id)\` validate a config-driven button list.
- One custom command via \`scope.register(...)\`, auto-unregistered on scope teardown.

Also threads \`PublicToolbarItemId\` through the public \`superdoc/ui\` sub-entry. SD-2920 exposed \`BUILT_IN_COMMAND_IDS\` at runtime but missed the derived type, so consumers had to dip into \`superdoc/headless-toolbar\` to type a config array. Fixed.

- Linear: SD-2929
- Stacked on #3127 (SD-2928 reorg) until that lands. Will rebase onto main.


caio-pizzol:

PR #3129, opened 2026-05-04
SD-2678 - Cannot delete contentlocked inline sdt with normal

chittolinag:
### Issue

Structured content blocks (SDTs) configured with `contentLocked` could not be deleted as a whole. Users could neither remove them with backspace/delete at the wrapper boundary, nor select-and-delete the field — the lock plugin treated wrapper-level actions as in-content edits and blocked them.

### Proposed solution

Refined the structured-content lock plugin so it distinguishes wrapper-level actions from content edits:

- A selection that exactly covers an SDT's content is promoted to a wrapper `NodeSelection`, so the next destructive keypress removes the whole field (matching Word's "click to select, key to delete" behavior).
- Backspace/delete at an SDT boundary is recognized as a wrapper-touching action and the affected range is expanded to the wrapper, so `contentLocked` no longer blocks deletion of the field itself.

Lock modes that protect the wrapper (`sdtLocked`, `sdtContentLocked`) still prevent removal. Added unit tests covering the new paths.


packages/super-editor/src/editors/v1/extensions/structured-content/structured-content-lock-plugin.js
```diff
@@ -106,6 +106,27 @@ export function createStructuredContentLockPlugin() {
...
         }
 
+        // Path 1 — non-collapsed selection that exactly covers the editable
+        // content of an SDT (e.g., the select-plugin's first-click select-all,
+        // a triple-click that lands on the content range, or precise keyboard
+        // selection). For wrapper-deletable lock modes, promote to a
+        // NodeSelection on the wrapper so the user sees the whole field
+        // highlighted and the next destructive press deletes it (matches
+        // Word's "click to select, key to delete").
+        if (from !== to && !(selection instanceof NodeSelection)) {
```
caio-pizzol, line 119: repro: open any unlocked inline field, triple-click to highlight the text inside, press Cmd+X. clipboard stays unchanged, the field's text is still there, and the highlight jumps to the whole field. paste anywhere to confirm "controlled text" is not on the clipboard. only Backspace and Delete should go through this branch:

```suggestion
        if ((isBackspace || isDelete) && from !== to && !(selection instanceof NodeSelection)) {
```

caio-pizzol (commented):
hey @chittolinag! the original bug is fixed in the dev app :)

one thing to flag: Cmd+X breaks for unlocked fields. select the text inside, press Cmd+X, nothing gets cut and the whole field gets selected instead. fix is one line.

approve once that's in.

packages/super-editor/src/editors/v1/extensions/structured-content/structured-content-lock-plugin.js
chittolinag: hey @caio-pizzol ! I could not reproduce this one, at least not manually. seems to work fine here. maybe let's get on a call to figure it out whether we're doing something different?

luccas-harbour: hey @chittolinag! I was able to reproduce this issue using the document attached

[sdt-cut-regression-repro.docx](https://github.com/user-attachments/files/27522488/sdt-cut-regression-repro.docx)

If you click the `UNLOCKED_FIELD_VALUE` sdt and then press Cmd+X, it doesn't do anything. Loading the same document in the `main` branch actually does cut the text.

chittolinag: ah I think I get it. if you `single click` the field, cmd + x won't work. if you double click it, it works. at least that's the behavior from my side.

luccas-harbour (approved):
LGTM

PR #3130, opened 2026-05-04
feat(examples): vanilla custom UI comments example (SD-2929)

caio-pizzol:
Second focused minimal example under \`examples/editor/custom-ui/\`. Single file (~125 lines) demonstrating the comments composer pattern that the controller surfaces specifically solve.

The load-bearing piece is \`ui.selection.capture()\`: the user clicks \"Add comment\", the textarea takes focus, the editor's live selection visually clears. A composer that read the live selection at submit time would refuse the create. \`capture()\` returns a frozen snapshot at composer-open time, and \`ui.comments.createFromCapture(capture, { text })\` anchors against that snapshot regardless of where focus has moved.

The list renders from \`ui.comments.observe(snapshot => ...)\`: one subscription, plain DOM. Resolve / reopen / reply route through \`ui.comments.*\` and end up in the same Document API that powers DOCX import/export, so changes here round-trip through Word.

- Linear: SD-2929
- Stacked on #3128 (vanilla toolbar example) which is stacked on #3127 (SD-2928 reorg). Will rebase onto main once both land.


caio-pizzol:

PR #3132, opened 2026-05-04
feat(examples): vanilla custom UI tracked-changes example (SD-2929)

caio-pizzol:
Third focused minimal example under \`examples/editor/custom-ui/\`. Single file (~140 lines) demonstrating the tracked-changes review panel that pairs naturally with the comments composer.

The load-bearing piece is the single \`ui.trackChanges.observe(snapshot => ...)\` subscription that drives the list, the bulk-action enable state, and the active-row highlight from one source. Per-change verbs (\`accept\` / \`reject\`), bulk verbs (\`acceptAll\` / \`rejectAll\`), and navigation (\`next\` / \`previous\` / \`scrollTo\`) all live on \`ui.trackChanges\`. \`ui.document.setMode\` flips between Edit and Suggest so the user can demonstrate the flow without leaving the page.

\`modules.trackChanges: { replacements: 'independent' }\` surfaces typed-over selections as separate insert + delete review items instead of one composite. Matches what most review UIs (including this one) render.

- Linear: SD-2929
- Stacked on #3130 (vanilla comments) which is stacked on #3128 (vanilla toolbar) which is stacked on #3127 (SD-2928 reorg). Will rebase onto main once #3127 lands.


caio-pizzol:

PR #3158, opened 2026-05-05
chore: create measurement to layout ownership matrix

VladaHarbour:
This PR introduces measuring-layout-ownership.md document with Measuring → Layout ownership matrix and tests which lock current boundary.
@harbournick please let me know if it should live in the repo


caio-pizzol (commented):
direction is right. three things inline.

caio-pizzol (commented):
direction is right. three things inline.

**Inline 1** - packages/layout-engine/docs/measuring-layout-ownership.md, line 1:

new docs/ folder - we keep layout-engine docs in CLAUDE.md and per-package READMEs today. could this go in CLAUDE.md instead? the follow-up list at the bottom feels like a Linear backlog.

**Inline 2** - packages/layout-engine/layout-engine/src/measuring-layout-ownership-contracts.test.ts, lines 179-206:

the measuring test covers section breaks but this one only covers page and column breaks. could we add a section break case here too?

**inline 3** - packages/layout-engine/layout-engine/src/measuring-layout-ownership-contracts.test.ts, lines 214-242:

this only passes because list layout isn't supported yet - when it lands, the test will start failing. could we use it.todo(...) so the next person updates it instead of deleting?

VladaHarbour:
> direction is right. three things inline.
> 
> **Inline 1** - packages/layout-engine/docs/measuring-layout-ownership.md, line 1:
> 
> new docs/ folder - we keep layout-engine docs in CLAUDE.md and per-package READMEs today. could this go in CLAUDE.md instead? the follow-up list at the bottom feels like a Linear backlog.
> 
> **Inline 2** - packages/layout-engine/layout-engine/src/measuring-layout-ownership-contracts.test.ts, lines 179-206:
> 
> the measuring test covers section breaks but this one only covers page and column breaks. could we add a section break case here too?
> 
> **inline 3** - packages/layout-engine/layout-engine/src/measuring-layout-ownership-contracts.test.ts, lines 214-242:
> 
> this only passes because list layout isn't supported yet - when it lands, the test will start failing. could we use it.todo(...) so the next person updates it instead of deleting?

Hi @caio-pizzol! Added some changes to address your comment

caio-pizzol:
Thanks for working on this. I’m going to close this PR because it is now out of date with `main`.

The work is still useful, especially the two test files and the Measuring → Layout notes. But the docs structure changed since this PR was opened, so this now has conflicts and some wording is stale.

The clean path is to open a fresh PR from current `main` with:
- the two test files
- an updated `packages/layout-engine/AGENTS.md` section
- no `CLAUDE.md` changes

Closing as stale, not because the idea is no longer relevant.

PR #3168, opened 2026-05-05
feat: render w:noBreakHyphen tags (SD-2746)

luccas-harbour:
## Summary
Adds full support for OOXML `<w:noBreakHyphen/>` tags so non-breaking hyphens are preserved end-to-end (import → schema → adapter → render → export) and behave like the visible U+2011 character users expect — including in search, get-text, diffing, and the document-api `text.rewrite` operation.

The atom is selectable, deletable from either side via Backspace/Delete, and round-trips back to `<w:noBreakHyphen/>` on export.

## Changes

**Schema & extension**
- New `noBreakHyphen` PM node (`atom`, inline, non-selectable) — `super-editor/src/editors/v1/extensions/no-break-hyphen/`.
- `Schema.js` surfaces PM's function-valued `leafText` NodeSpec field (without invoking it) so flattening APIs see the rendered glyph.

**Import / export**
- v2 + v3 importers translate `<w:noBreakHyphen/>` to the new node.
- v3 adds `noBreakHyphen` translator with co-located tests (translator + `r-translator` integration).
- Exporter writes the atom back as `<w:noBreakHyphen/>`.

**Layout / pm-adapter**
- New `no-break-hyphen.ts` inline converter emits a `TextRun` carrying U+2011 — DomPainter renders it through the existing text path with no painter changes.
- `paragraph.ts` wires the converter into the run pipeline.

**Editing behavior**
- `backspaceAtomBefore` command + keymap entry: handles all three caret positions around the atom (inside its run, between runs, at start of next run) on an opt-in allowlist (currently just `noBreakHyphen`) so existing `bookmarkEnd` behavior is preserved.
- `deleteAtomAfter` command + keymap entry: mirrors the above for forward-delete when the caret sits inside the atom's wrapper run.
- Without these, the keymap fell through to `backspaceAcrossRuns`, which intentionally skips non-text inline nodes and would delete the previous character instead of the hyphen.

**Document-API / text helpers**
- `textBetweenWithTabs` honors `leafText` (with a fallback for leaves that don't define one) — search/get-text/diff now see U+2011 instead of the generic placeholder.
- `plan-engine/executor.ts` `charOffsetToDocPos` now counts atomic inline leaves as `leafText(node).length` slots (and tabs as one slot), so `text.rewrite` lands edits at the correct PM position. Previously, rewriting `"a‑b"` → `"a‑c"` replaced `'a'` instead of `'b'`.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/commands/deleteAtomAfter.js
+++ b/packages/super-editor/src/editors/v1/core/commands/deleteAtomAfter.js
@@ -41,6 +41,11 @@ export const deleteAtomAfter =
     // Case 1 only: caret is inside the atom's wrapper run, immediately
     // before the atom, and the atom is that run's only child. Remove the
     // whole wrapper as a single unit so we don't leak an empty run.
+    // Note: `childCount !== 1` is defensive — `calculateInlineRunPropertiesPlugin`
+    // segments runs by inline properties on every change, so a run holding
+    // text+atom+text is always split back into three single-content runs
+    // before this command runs. The guard documents the invariant; it isn't
+    // reachable through normal editing.
     const parent = $pos.parent;
     if (parent.type.name !== 'run' || parent.content.childCount !== 1) return false;
 
--- a/packages/super-editor/src/editors/v1/core/commands/deleteAtomAfter.js
+++ b/packages/super-editor/src/editors/v1/core/commands/deleteAtomAfter.js
@@ -1,0 +1,56 @@
+import { Selection } from 'prosemirror-state';
+
+/**
+ * Atom inline node types that should be treated as a single deletable unit
+ * when Delete is pressed with the cursor immediately before them. Mirrors
+ * `backspaceAtomBefore`'s allowlist — same opt-in policy, same rationale
+ * (don't accidentally delete bookmarkEnd-style markers).
+ */
+const DELETABLE_INLINE_ATOMS = new Set(['noBreakHyphen']);
+
+/**
+ * Forward-Delete handler that targets the deletable inline atom directly
+ * after the cursor.
+ *
+ * Only handles the case the rest of the Delete chain misses: caret inside
+ * the atom's wrapper run with `nodeAfter` being the atom itself. At that
+ * caret, `deleteNextToRun` bails (its early return excludes any nodeAfter
+ * that isn't a `run`), `deleteSelection` is a no-op for empty selections,
+ * `joinForward` only fires at textblock end, and `selectNodeForward` skips
+ * non-selectable atoms — so the chain falls through and nothing happens.
+ *
+ * Paragraph-level boundary cases (caret between runs, or at the end of the
+ * previous run with the atom-wrapper run as the next sibling) already work
+ * via `deleteNextToRun`, which removes the atom and lets PM's replace fold
+ * the now-empty wrapper into the surrounding paragraph. Don't intercept
+ * those — leave the existing path alone.
+ *
+ * @returns {import('@core/commands/types').Command}
+ */
+export const deleteAtomAfter =
+  () =>
+  ({ state, tr, dispatch }) => {
+    const sel = state.selection;
+    if (!sel.empty) return false;
+
+    const $pos = sel.$from;
+    const after = $pos.nodeAfter;
+    if (!after || !after.isInline || !after.isAtom) return false;
+    if (!DELETABLE_INLINE_ATOMS.has(after.type.name)) return false;
+
+    // Case 1 only: caret is inside the atom's wrapper run, immediately
+    // before the atom, and the atom is that run's only child. Remove the
+    // whole wrapper as a single unit so we don't leak an empty run.
+    const parent = $pos.parent;
+    if (parent.type.name !== 'run' || parent.content.childCount !== 1) return false;
+
+    const runStart = $pos.before($pos.depth);
+    const from = runStart;
+    const to = runStart + parent.nodeSize;
+
+    if (dispatch) {
+      tr.delete(from, to).setSelection(Selection.near(tr.doc.resolve(from)));
+      dispatch(tr.scrollIntoView());
+    }
+    return true;
+  };
--- a/packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/del/del-translator.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/del/del-translator.js
@@ -86,10 +86,10 @@ function decode(params) {
   node.marks = marks.filter((m) => !trackingMarks.includes(m.type));
 
   const translatedTextNode = exportSchemaToJson({ ...params, node });
-  // ECMA-376 only renames w:t → w:delText (and w:instrText → w:delInstrText)
-  // inside <w:del>. Other inline content — w:noBreakHyphen, w:tab, w:br, etc.
-  // — stays as-is; the deletion is conveyed by the <w:del> wrapper alone.
-  // Guard the rename so non-text atoms inside <w:del> don't crash.
+  // ECMA-376 renames w:t → w:delText inside <w:del>. Other inline content —
+  // w:noBreakHyphen, w:tab, w:br, etc. — stays as-is; the deletion is
+  // conveyed by the <w:del> wrapper alone. Guard the rename so non-text
+  // atoms inside <w:del> don't crash.
   const textNode = translatedTextNode.elements.find((n) => n.name === 'w:t');
   if (textNode) textNode.name = 'w:delText';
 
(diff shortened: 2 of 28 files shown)


packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/noBreakHyphen/no-break-hyphen-translator.test.js
```diff
@@ -0,0 +1,148 @@
...
+    });
+  });
+
+  describe('decode', () => {
+    it('wraps <w:noBreakHyphen/> in a <w:r> run', () => {
+      const res = config.decode({ node: { type: 'noBreakHyphen' } }, undefined);
+      expect(res).toBeTruthy();
+      expect(res.name).toBe('w:r');
+      expect(Array.isArray(res.elements)).toBe(true);
+      expect(res.elements[0]).toEqual({ name: 'w:noBreakHyphen' });
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/r/r-translator.no-break-hyphen.test.js
```diff
@@ -0,0 +1,150 @@
...
+      elements: [{ name: 'w:noBreakHyphen' }],
+    };
+    const handler = defaultNodeListHandler();
+
+    const encoded = translator.encode({ nodes: [runNode], nodeListHandler: handler, docx: {} });
+    const decoded = translator.decode({ node: encoded });
+    const runs = Array.isArray(decoded) ? decoded : [decoded];
+
+    const allChildren = runs.flatMap((r) => (r.elements || []).filter((el) => el.name !== 'w:rPr'));
+    expect(allChildren).toEqual([{ name: 'w:noBreakHyphen' }]);
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/noBreakHyphen/no-break-hyphen-translator.js
```diff
@@ -0,0 +1,205 @@
...
+ * Decode a SuperDoc noBreakHyphen node back into <w:r><w:rPr/><w:noBreakHyphen/></w:r>.
+ * Mirrors the run-wrapping pattern in tab-translator.js so inherited run-properties
+ * (bold, color, etc.) are preserved on export.
+ * @param {import('@translator').SCDecoderConfig} params
+ * @param {import('@translator').DecodedAttributes} [decodedAttrs] - The already decoded attributes
+ * @returns {import('@translator').SCDecoderResult}
+ */
+function decode(params, decodedAttrs = {}) {
+  const { node } = params || {};
+  if (!node) return;
```
caio-pizzol, line 40: repro: turn on track changes → insert a noBreakHyphen → export. the saved file has a plain `<w:r><w:noBreakHyphen/></w:r>` with no `<w:ins>` around it, so the tracked insert is lost. `t-translator.js:104-115` handles this for text by handing off to the ins/del translator. same shape works here (you'll need to import `wInsTranslator` / `wDelTranslator` too):

```js
function decode(params, decodedAttrs = {}) {
  const { node } = params || {};
  if (!node) return;

  // Tracked changes: hand off so the atom exports inside <w:ins> or <w:del>.
  // Without this, the tracking is dropped.
  const trackedMark = node.marks?.find((m) => {
    const t = m?.type?.name ?? m?.type;
    return t === 'trackInsert' || t === 'trackDelete';
  });
  if (trackedMark && !params.extraParams?.trackingProcessed) {
    const t = trackedMark.type?.name ?? trackedMark.type;
    return (t === 'trackInsert' ? wInsTranslator : wDelTranslator).decode(params);
  }
  // ...existing link/run-properties logic stays
```

import side has the same gap (`ins-translator.js:58` only tags text children) so a tracked nbh from word also loses the mark on import. fine for a follow-up.

packages/super-editor/src/editors/v1/core/commands/deleteAtomAfter.js
```diff
@@ -0,0 +1,56 @@
...
+    const $pos = sel.$from;
+    const after = $pos.nodeAfter;
+    if (!after || !after.isInline || !after.isAtom) return false;
+    if (!DELETABLE_INLINE_ATOMS.has(after.type.name)) return false;
+
+    // Case 1 only: caret is inside the atom's wrapper run, immediately
+    // before the atom, and the atom is that run's only child. Remove the
+    // whole wrapper as a single unit so we don't leak an empty run.
+    const parent = $pos.parent;
+    if (parent.type.name !== 'run' || parent.content.childCount !== 1) return false;
```
caio-pizzol, line 50: i tried to hit this branch by inserting a run with text+atom+text directly - a plugin still split it into three runs. so this case can't actually happen. a one-line note saying so would save the next reader a trace.

caio-pizzol (commented):
hey @luccas-harbour! the delete handling and command split read cleanly :)

left three things inline.

one thing to flag.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/noBreakHyphen/no-break-hyphen-translator.js
```diff
@@ -0,0 +1,205 @@
...
+
+  // Hyperlinks: defer to wHyperlinkTranslator so the export emits a
+  // <w:hyperlink> wrapper and preserves the relationship. Without this, a
+  // linked noBreakHyphen (e.g. translateChildNodes groups the atom as the
+  // first link-marked child) would round-trip as plain <w:r><w:noBreakHyphen/>
+  // </w:r>, dropping the link entirely.
+  // The linkProcessed guard avoids re-entering once the hyperlink decoder
+  // strips the link mark and re-dispatches us.
+  const isLinkNode = node.marks?.some((m) => m.type === 'link');
```
caio-pizzol:
```suggestion
  const isLinkNode = node.marks?.some((m) => (m?.type?.name ?? m?.type) === 'link');
```

works today, just matches the pattern used ten lines below.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/noBreakHyphen/no-break-hyphen-translator.js
luccas-harbour: fixed

packages/super-editor/src/editors/v1/core/commands/deleteAtomAfter.js
luccas-harbour: fixed. added a note about this

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/noBreakHyphen/no-break-hyphen-translator.js
luccas-harbour: done

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/del/del-translator.js
```diff
@@ -86,8 +86,12 @@ function decode(params) {
   node.marks = marks.filter((m) => !trackingMarks.includes(m.type));
 
   const translatedTextNode = exportSchemaToJson({ ...params, node });
+  // ECMA-376 only renames w:t → w:delText (and w:instrText → w:delInstrText)
+  // inside <w:del>. Other inline content — w:noBreakHyphen, w:tab, w:br, etc.
+  // — stays as-is; the deletion is conveyed by the <w:del> wrapper alone.
+  // Guard the rename so non-text atoms inside <w:del> don't crash.
```
caio-pizzol: the comment promises `w:instrText → w:delInstrText` handling but the code only does `w:t`. spec §17.16.13 + §17.16.23 do mandate both renames, but it looks unreachable today (exporter.js:657-663 has a compensating normalization for the import-side gap). easiest scope: trim the parenthetical so the comment matches the code:

```suggestion
  // ECMA-376 renames w:t → w:delText inside <w:del>. Other inline content —
  // w:noBreakHyphen, w:tab, w:br, etc. — stays as-is; the deletion is
  // conveyed by the <w:del> wrapper alone. Guard the rename so non-text
  // atoms inside <w:del> don't crash.
```

caio-pizzol (approved):
hey @luccas-harbour! thanks for addressing last round - the `<w:ins>` / `<w:del>` wrapping works end-to-end (verified live, both paths).

one tiny comment fix inline.

lgtm.

PR #3170, opened 2026-05-05
fix: document render crashes

VladaHarbour:


caio-pizzol (approved):
@VladaHarbour thanks for the PR.

tested the original ThreeArbor file: crashes on main, renders fine on this branch. uploaded the file to the test corpus.

lgtm.

PR #3174, opened 2026-05-05
SD-2875 - fix: context menu not showing suggestions after typecheck

chittolinag:
### Issue

- Right-click on a misspelled word stopped showing spelling suggestions in 1.29+.
- `<ContextMenu>` was rebound from the inner `Editor` to the `PresentationEditor` wrapper.
- `resolveProofingContext` only looked up `editor._presentationEditor`, which the wrapper doesn't have — so it returned `null` and the menu dropped every suggestion.

### Proposed solution

- In `resolveProofingContext` (`context-menu/utils.js`), resolve the manager from any of the three editor handles the menu can be wired to:
  - inner editor → `_presentationEditor.proofingManager`
  - story editor → `presentationEditor.proofingManager`
  - wrapper → `proofingManager`

### Behavior tests

- `tests/behavior/tests/slash-menu/proofing-context-menu.spec.ts`:
  - Types "Hello teh world", attaches a stub provider, right-clicks `teh`, asserts the suggestions appear and clicking `the` rewrites the doc.
  - Negative test: no `proofing-replace-*` rows appear when there's no issue at the cursor.
- Verified to fail without the fix.

### DIFF
--- a/packages/super-editor/src/editors/v1/components/context-menu/tests/utils.test.js
+++ b/packages/super-editor/src/editors/v1/components/context-menu/tests/utils.test.js
@@ -53,6 +53,7 @@ import {
   __getStructureFromResolvedPosForTest,
   __isCollaborationEnabledForTest,
   __getCellSelectionInfoForTest,
+  __resolveProofingContextForTest,
 } from '../utils.js';
 import { isList } from '@core/commands/list-helpers';
 import { readFromClipboard } from '../../../core/utilities/clipboardUtils.js';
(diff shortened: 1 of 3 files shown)


tests/behavior/tests/slash-menu/proofing-context-menu.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+  // Configure proofing with a provider that flags the word 'teh' (which is
+  // not present in the document). This guarantees the manager is wired
+  // up but has no issue at any position.
+  await configureStubProvider(superdoc, 'teh', ['the']);
+
+  // Give the provider a moment to run; we deliberately do NOT wait for
+  // an issue because none should be produced. A short stable settle is
+  // enough since the debounce is 50ms.
+  await page.waitForTimeout(150);
+  await superdoc.waitForStable();
```
caio-pizzol: (optional) 150ms can be too short on slow CI. if the spell-checker hasn't run yet, the test passes because nothing was checked - not because there's nothing to flag. waiting on the actual signal would be safer:

```suggestion
  // Wait until the stub has actually run, otherwise this test can pass
  // because the check never fired rather than because nothing matched.
  await page.waitForFunction(() => (window.__sd2875Calls ?? 0) > 0, null, { timeout: 5_000 });
  await superdoc.waitForStable();
```

packages/super-editor/src/editors/v1/components/context-menu/tests/utils.test.js
```diff
@@ -749,4 +750,140 @@ describe('utils.js', () => {
       expect(__isCollaborationEnabledForTest({ options: {} })).toBe(false);
     });
   });
+
+  // SD-2875: spelling suggestions vanished from the right-click menu in 1.29
+  // because <ContextMenu> was rebound from the inner Editor (which carries
+  // _presentationEditor) to the PresentationEditor wrapper (which does not).
+  // resolveProofingContext must locate the proofing manager from any of the
+  // three handles the menu can be wired to.
```
caio-pizzol: (optional) 5 lines is a lot for a test comment. first sentence covers it, the rest is in the ticket.

```suggestion
  // SD-2875: spelling suggestions vanished from the right-click menu in 1.29
  // because <ContextMenu> moved from the inner Editor to the PresentationEditor
  // wrapper, which doesn't carry _presentationEditor.
```

caio-pizzol (approved):
@chittolinag lgtm :)

left two optional notes inline - skip if you'd rather move on.

luccas-harbour (approved):

PR #3175, opened 2026-05-05
fix(converter): preserve tracked-change-wrapped fields during DOCX import (SD-2858)

luccas-harbour:
## Summary

Fixes DOCX import handling for fields wrapped in tracked changes when the field spans multiple nodes or paragraphs.

This preserves the original raw OOXML for tracked-change-wrapped fields instead of trying to collapse/process them as normal field references. That avoids import failures fordocuments where a field begins inside one tracked deletion/insertion wrapper and ends inside another.

Also makes header/footer relationship filtering more defensive by tolerating missing relationship elements or relationship nodes without attributes.

## Changes

- Preserve raw field nodes when field processing crosses `w:del` or `w:ins` wrappers.
- Handle unpaired field ends bubbling up from child nodes without throwing.
- Add regression coverage for a tracked-deletion-wrapped hyperlink field split across paragraphs.
- Guard header/footer relationship lookup against missing `elements` or `attributes`.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/super-converter/field-references/preProcessNodesForFldChar.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/field-references/preProcessNodesForFldChar.js
@@ -3,11 +3,11 @@
  */
 import { getInstructionPreProcessor } from './fld-preprocessors';
 import { carbonCopy } from '@core/utilities/carbonCopy.js';
+import { isTrackChangeElement } from '../v2/importer/trackChangeElements.js';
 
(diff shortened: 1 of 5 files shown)


caio-pizzol (approved):
hey @luccas-harbour! lgtm :)

follow-up: same shape but with an inserted (instead of deleted) field also used to drop on import. now the text shows up but the link is gone - user sees plain text where Word shows a clickable link. fine for deletes and moves since that text goes away on accept, but for inserts it stays visible. tracked in SD-2973.

PR #3178, opened 2026-05-05
refactor(painter): make DomPainter measurement-clean (SD-2957)

tupizz:
Closes [SD-2957](https://linear.app/superdocworkspace/issue/SD-2957). Follow-up to SD-2836.

## What

Two patterns survived SD-2836 that contradicted "the painter is dumb":

1. **4 paint-time DOM measurements** — `pageEl.clientHeight` / `elem.offsetWidth` / clone-and-measure, used as fallbacks when resolved fields were missing or zero.
2. **14 dead `?? fragment.X` fallbacks** — `resolvedItem?.X ?? fragment.X` reads. The producer already copies these fields; the fragment fallback was equivalent to `resolvedItem?.X`.

Both deleted. `PageDecorationPayload.offset` tightened from optional to required (the producer already always sets it). New architecture-boundary Guard E ratchets the invariant — any future paint-time DOM measurement on rendered content fails CI.

## Commits

| # | Commit |
|---|---|
| 1 | `test(architecture)`: Guard E forbids paint-time DOM measurement |
| 2 | `refactor(painter)`: drop `pageEl.clientHeight` fallback for `ResolvedPage.height` |
| 3 | `refactor(contracts,painter)`: require `PageDecorationPayload.offset` |
| 4 | `refactor(painter)`: drop `offsetWidth` fallbacks for `LineSegment.width` |
| 5 | `refactor(painter)`: delete 14 dead `fragment.X` fallbacks |
| 6 | `docs(painter)`: invariants in README + AGENTS/CLAUDE |

## Out of scope

- `block.width ?? fragment.width` (different semantics — natural vs layout width)
- `fragment.lines ?? measure.lines.slice(...)` (needs `ResolvedTextLineItem[]` migration — separate ticket)
- `fragment.scale ?? 1` (literal default, no resolved equivalent)

## Verification

- 22/22 architecture + contract-shape tests pass
- 1050/1050 `painter-dom`; 12345/12345 `super-editor`; 18390/18390 full repo
- `tsc --noEmit` clean
- Behavior: same output for healthy docs; painter throws (rather than silently rescuing) when resolve/measure stages emit invalid data

## Test plan

- [ ] CI green
- [ ] Layout corpus diff vs `origin/main`: 0 unintended changes


packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -6769,20 +6767,11 @@ export class DomPainter {
...
-            if (width <= 0 && this.doc) {
-              const measureEl = elem.cloneNode(true) as HTMLElement;
-              measureEl.style.position = 'absolute';
-              measureEl.style.visibility = 'hidden';
-              measureEl.style.left = '-9999px';
-              this.doc.body.appendChild(measureEl);
-              width = measureEl.offsetWidth;
-              this.doc.body.removeChild(measureEl);
-            }
+            const width = segment.width ?? 0;
```
luccas-harbour: `width` is a required field so I think you could just drop the `?? 0` (same thing applies to line 6667)

luccas-harbour (approved):
LGTM

I left just a small comment about dropping a `?? 0` for a required field

tupizz:
Re: [@luccas-harbour's comment](https://github.com/superdoc-dev/superdoc/pull/3178#discussion_r3196259864) on `?? 0` for `segment.width`:

Done in 94a3b833c. Both `?? 0` defaults dropped — `segment.width` is `number` (required) by contract and the producer always emits it, so the fallback was dead. Behavior identical for healthy segments; an invalid width now propagates as NaN rather than being silently masked, matching the SD-2957 fail-loudly principle.

Verified: `painter-dom` 1050/1050 pass, `tsc --noEmit` clean, architecture + contract-shape tests 23/23.

Thanks for the review!

tupizz (commented):
Updates after adresssing codex reviews:
Good catch. Both fixed in two commits:

* `b7d7f6ffc test(architecture): forbid dead fragment.X coalescing in render path` — adds **Guard F** to `architecture-boundaries.test.ts`. The guard catches `?? fragment.X` and `?? (fragment as Y).X` patterns (with a tiny allowlist for the documented `block.width/height ?? fragment.width/height` exception, which has different semantics — image-block natural width vs fragment layout width). My initial sweep grepped `resolvedItem?.X ?? fragment.X` and `item?.X ?? fragment.X` and missed the four sites because they used the variable name `resolvedFrag`. Without this guard a future contributor can re-introduce the pattern under any new variable name and only catch it in review — exactly what just happened. The guard fails today on the four sites you flagged.

* `852286e94 refactor(painter): delete final 4 fragment.X fallbacks the soft-pattern sweep missed` — fixes the four sites (`pmStart`, `pmEnd`, `continuesFromPrev`, `continuesOnNext`) at `renderer.ts:6867/6873/6879/6885`. `resolveLayout.ts:286–289` populates each field on the resolved item from the source paragraph, so the back-pointer reads were dead exactly as in the prior sweep. Replaced with `resolvedFrag?.X`; the two truthy-check fallbacks collapsed to direct `if (resolvedFrag?.X)`.

Verified after both commits:
* Architecture boundary tests: **23/23 pass** (Guards A-F + 3 contract-shape lockdowns)
* `painter-dom`: **1050/1050 pass**, `tsc --noEmit` clean

The invariant in the docs is now true *and* ratchet-protected.

PR #3179, opened 2026-05-05
fix(python-sdk): chmod CLI binary at build time instead of runtime

mattConnHarbour:
## Problem

  CLI binary shipped without execute permissions (Python wheels strip them). In Docker: root installs, non-root user tries to exec, and is denied. Runtime `os.chmod()` fails for the same reason—non-root can't chmod a root-owned file.

  ## Fix

  Add `setup.py` build hooks that chmod the binary during wheel creation. Binary ships with execute permissions baked in, no runtime chmod needed.

  ## Follow-up

  Platform packages are nearly identical—could be generated from `python-embedded-cli-targets.mjs` instead of maintained manually.


caio-pizzol (commented):
hey @mattConnHarbour! good call moving chmod out of runtime, kills two failure modes at once :)

three things, only the middle one has teeth.

---

**`packages/sdk/langs/python/platforms/superdoc-sdk-cli-darwin-arm64/setup.py:3`**

`os` is imported but never used. same in the other three Unix setup.py files.

```diff
-import os
 import stat
 from pathlib import Path
```

---

**`packages/sdk/langs/python/superdoc/embedded_cli.py:106`**

repro:

```bash
PKG=packages/sdk/langs/python/platforms/superdoc-sdk-cli-darwin-arm64
chmod 0644 $PKG/superdoc_sdk_cli_darwin_arm64/bin/superdoc
pip install -e $PKG
python -c "from superdoc.embedded_cli import resolve_embedded_cli_path; import subprocess; subprocess.run([resolve_embedded_cli_path()])"
# PermissionError: [Errno 13]
```

editable installs skip the build hook, so they see the source binary as-is. normal flow is fine because `stage-python-companion-cli.mjs:106` chmods to 0755 first - but that chmod sits in a try/except marked "runtime will surface execution errors" (lines 107-109). the runtime chmod you just removed *was* that surfacing. tighten the staging chmod to fail loudly, or update that comment?

---

**`packages/sdk/langs/python/platforms/superdoc-sdk-cli-darwin-arm64/setup.py:16`**

`# After files are copied to build dir, chmod the binary` just restates the next four lines. drop it, or swap for the why. same in the other three Unix setup.py files.

---

_(posted as one comment because GitHub's review-comment API kept returning 422 "internal error" on inline anchors for this PR.)_

packages/sdk/langs/python/platforms/superdoc-sdk-cli-darwin-arm64/setup.py
```diff
@@ -0,0 +1,24 @@
...
+
+class BuildPyWithExecutableBinary(build_py):
+    """Custom build_py that ensures the CLI binary has execute permissions."""
+
+    def run(self):
+        super().run()
+        # After files are copied to build dir, chmod the binary
+        if self.build_lib:
+            binary_path = Path(self.build_lib) / 'superdoc_sdk_cli_darwin_arm64' / 'bin' / 'superdoc'
+            if binary_path.exists():
```
andrii-harbour, line 17: if the path doesn't exist it will just fail silently.

packages/sdk/langs/python/platforms/superdoc-sdk-cli-darwin-arm64/superdoc_sdk_cli_darwin_arm64/__init__.py
```diff
@@ -10,15 +10,15 @@
```
andrii-harbour, line 10: The four Unix setup.py files are byte-identical except for one string (the module name); the Windows one is a stub. Five files, one logical thing.

  Duplication like this is fine in a code-only PR — a small surface, easy to revisit. This PR is infrastructure: wheel builds ship binaries to every consumer, and we're literally fixing a wheel-build bug right now. The next permission edge case will be fixed in one file and forgotten in the other three, and we'll ship asymmetric wheels.

Generate from python-embedded-cli-targets.mjs (already the source of truth), or extract a shared _build_helpers.py so each setup.py is a one-line shim. Worth doing in this PR — the duplication exists because of this change, and "we'll unify it later" is how five copies become eight.

andrii-harbour (changes requested):
This PR is about shipping packages to customers. If it doesn't work - no matter how good our code is - nobody would know that. 

2 very important things : 
1. The follow-up item mentioned in the description, imo, should be done here. It is directly related to what we are trying to fix.
2. This setup must be tested manually on a separate CI/CD pipeline to ensure that the build is correct.

mattConnHarbour:
Closing for core eng. handoff.
Ticket: https://linear.app/superdocworkspace/issue/SD-2971/python-sdk-add-post-install-hook-to-chmod-cli-binary

PR #3180, opened 2026-05-06
fix(converter): escape XML entities in OPC attribute values on export (SD-2888)

tupizz:
## Summary
- Fix unescaped `&` in `word/_rels/document.xml.rels` that caused Word to show "unreadable content" on opening exported DOCX files. Linear: [SD-2888](https://linear.app/superdocworkspace/issue/SD-2888/bug-unescaped-ampersand-in-docx-export-causes-unreadable-content)
- Add a shared `serializeOpcXml` helper that escapes `&`, `<`, `>` in OPC attribute values so `xml-js`'s `js2xml` round-trips entities correctly
- Route both `reconcile-document-relationships` and `sync-package-metadata` through it (closes the same latent bug for `_rels/.rels` and `[Content_Types].xml`)

## Root cause
`xml-js`'s `xml2js` decodes `&amp;` → `&` in attribute values; `js2xml` does **not** re-escape on output. When `reconcileDocumentRelationships` rewrites `word/_rels/document.xml.rels` (e.g. to add the managed `numbering` relationship), every existing hyperlink `Target` containing `&amp;` is emitted with a bare `&`, producing malformed XML. Word's repair-on-open then applies default fonts (Times New Roman → Aptos), table cell widths, and paragraph spacing, which look like unrelated regressions but are downstream effects of the same root cause.

The trick in the helper: `xml-js` calls `attributeValueFn` with the value **after** pre-escaping `"` to `&quot;`. A naïve `&` → `&amp;` would double-escape that token. The helper pivots `&quot;` through a placeholder, escapes `&`/`<`/`>`, then restores `&quot;`.

## Files changed
- **new** `packages/super-editor/src/editors/v1/core/opc/xml-serialization.js` (helper)
- **new** `packages/super-editor/src/editors/v1/core/opc/xml-serialization.test.js` (5 unit tests)
- `reconcile-document-relationships.js` — call site swap
- `reconcile-document-relationships.test.js` — 2 SD-2888 regression tests
- `sync-package-metadata.js` — call site swap

## Test plan
- [x] `pnpm --filter super-editor test` (all 12,367 tests pass; 7 new)
- [x] OPC suite: `npx vitest run src/editors/v1/core/opc` (43/43 pass)
- [x] End-to-end browser repro: upload SD-2888 input.docx → `editor.exportDocx()` → unzip → verify `word/_rels/document.xml.rels` contains `&amp;companyId` (3×, matching rId8/9/10 in the bug) and zero bare `&companyId`
- [x] Validate every XML file in the exported package (16 files) parses as well-formed
- [ ] Open the exported file in Microsoft Word and confirm no "unreadable content" repair prompt appears
- [ ] Confirm font (Times New Roman), table widths, and 1.0 line spacing are preserved when opening in Word


caio-pizzol (approved):
hey @tupizz! good call centralizing this in `serializeOpcXml` :)

i pushed a behavior test that exercises the export path against a Word-authored fixture - catches the bare-`&` regression at the integration level (the unit tests cover the helper directly).

lgtm.

PR #3181, opened 2026-05-06
fix(presentation-editor): bail on clicks in inter-page gap (SD-2749)

tupizz:
## Summary
- Fixes [SD-2749](https://linear.app/superdocworkspace/issue/SD-2749/bug-clicking-near-page-border-jumps-documents-to-different-page): clicking in the vertical gap between two pages caused the viewer to jump to a different page.
- One guard line added to `EditorInputManager.#handlePointerDown` that bails when no `.superdoc-page` is under the cursor (matches the existing in-margin guard from SD-2356).
- Defensive `layout?.pages?.[pageIndex]` access in `isOutsidePageBodyContent` to prevent a latent crash when `layout.pages` is missing.

## Root cause
`PointerNormalization` uses `elementsFromPoint` to find the `.superdoc-page` under the cursor. In the inter-page gap there is none, so it returns `pageIndex: undefined`. The existing `isOutsidePageBodyContent` guard from SD-2356 short-circuits to `false` when `pageIndex` is not finite, so the pointer flow fell through to the geometry fallback (`clickToPositionGeometry` → `snapToNearestFragment`), which always returns *some* nearest fragment. `TextSelection.create` + ProseMirror `scrollIntoView` then jumped the page.

```
clickToPositionDom        → null (no DOM hit)
findPageElement           → null (no page under cursor)
clickToPositionGeometry   → hitTestPage returns nearest page by center distance
  → hitTestFragment       → null
  → snapToNearestFragment → returns fragment closest to the click
→ TextSelection.create    → editor.view.dispatch
→ ProseMirror scrollIntoView → page jumps
```

The fix is symmetric with SD-2356: clicking in the gap is conceptually the same as clicking in a page margin — preserve selection and scroll.

## Files changed
| File | Change |
|---|---|
| `EditorInputManager.ts` | Extended pointerdown guard to bail when `pageIndex` is undefined; added defensive `layout?.pages?.[pageIndex]` |
| `EditorInputManager.pageMarginClick.test.ts` | New regression test for SD-2749 (RED → GREEN) |
| `EditorInputManager.{activeCommentClick,dragAutoScroll,footnoteClick,structuredContent}.test.ts` | Mocks updated to return realistic `{ x, y, pageIndex, pageLocalY }` matching the production contract |
| `PresentationEditor.test.ts` | Two integration tests stub `document.elementsFromPoint` so happy-dom returns the simulated `.superdoc-page` under the click point |

## Test plan
- [x] `pnpm --filter super-editor test` — 18,421/18,421 pass (1,240 files)
- [x] Targeted: `EditorInputManager.pageMarginClick.test.ts` (3 tests pass; new SD-2749 test fails on `main`, passes on this branch)
- [x] Browser end-to-end with `IT-923 Headers.docx` (7 pages):
  - Gap click at `(600, 366)` between pages 0 & 1: `scrollTop` and `selection.from` UNCHANGED
  - In-page text click at `(600, 500)`: selection moves to clicked text, no scroll jump
- [ ] Manual confirmation in Word-style customer doc (`NVCA-Model-IRA.docx`)
- [ ] Manual confirmation that the existing SD-2356 fix (margin clicks) still works

### DIFF
--- a/packages/super-editor/src/editors/v1/core/presentation-editor/pointer-events/EditorInputManager.ts
+++ b/packages/super-editor/src/editors/v1/core/presentation-editor/pointer-events/EditorInputManager.ts
@@ -132,7 +132,7 @@ function isOutsidePageBodyContent(layout: Layout, x: number, pageIndex?: number,
     return false;
   }
 
-  const page = layout.pages[pageIndex];
+  const page = layout?.pages?.[pageIndex];
   if (!page) {
     return false;
   }
@@ -1480,13 +1480,21 @@ export class EditorInputManager {
       }
     }
 
-    if (
-      !useActiveSurfaceHitTest &&
-      isOutsidePageBodyContent(layoutState.layout, x, normalizedPoint.pageIndex, normalizedPoint.pageLocalY)
-    ) {
-      event.preventDefault();
-      this.#focusEditor();
-      return;
+    // Bail when the click did not land on any page body. Two cases:
+    // - SD-2356: click inside a page's bounding box but in the margin/header/footer area.
+    // - SD-2749: click in the gap between pages (no .superdoc-page under the cursor),
+    //   in which case normalizeClientPoint leaves pageIndex undefined.
+    // Both should preserve the current selection and scroll position.
+    if (!useActiveSurfaceHitTest) {
+      const pointerOffAnyPage = !Number.isFinite(normalizedPoint.pageIndex);
+      if (
+        pointerOffAnyPage ||
+        isOutsidePageBodyContent(layoutState.layout, x, normalizedPoint.pageIndex, normalizedPoint.pageLocalY)
+      ) {
+        event.preventDefault();
+        this.#focusEditor();
+        return;
+      }
     }
 
     const { rawHit, hit } = this.#resolveSelectionPointerHit({
--- a/packages/super-editor/src/editors/v1/core/presentation-editor/pointer-events/EditorInputManager.ts
+++ b/packages/super-editor/src/editors/v1/core/presentation-editor/pointer-events/EditorInputManager.ts
@@ -1466,12 +1466,13 @@ export class EditorInputManager {
       if (this.#handleClickInHeaderFooterMode(event, x, y, normalizedPoint.pageIndex, normalizedPoint.pageLocalY))
         return;
       // SD-2749: clicking on body content from inside a header/footer session
-      // exits the session synchronously. Re-derive local mode so the click
-      // dispatches its selection on the body editor instead of the now-stale
-      // active editor — otherwise ProseMirror's scrollIntoView would pull the
-      // viewport back to the header/footer the user just exited.
+      // exits the session synchronously, which also clears the backing story
+      // session. Re-read both so subsequent hit testing and selection dispatch
+      // target the body editor — otherwise ProseMirror's scrollIntoView would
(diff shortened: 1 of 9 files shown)


tupizz:
Follow-up: extended fix for the header→body click variant of SD-2749.

**Repro confirmed on origin/main** (and on previous commit on this branch):
- dblclick header → enters editing session
- type some text in header
- scroll down a few pages
- click in middle of any body paragraph
- → `scrollTop` jumps from 3500 → 97 (page snaps back to where the header is)

**Root cause**: `#handlePointerDown` captured `sessionMode`, `useActiveSurfaceHitTest`, and `editor` locals at function start. When `#handleClickInHeaderFooterMode` exits the session for a `bodyContent` click and returns `false`, those locals stayed stale. Selection then got dispatched on the header editor → ProseMirror's `scrollIntoView` pulled the viewport back to the (now off-screen) header.

**Fix** (commit 57c46e5a8): convert the three locals to `let` and re-derive them from a fresh `getHeaderFooterSession()` read after the H/F handler returns. If the session is now `'body'` and we're not in a note session, point `editor` at `bodyEditor` and recompute `useActiveSurfaceHitTest`.

**Tests** added `EditorInputManager.headerFooterBodyClick.test.ts` with one regression test asserting that body `view.dispatch` is called and header `view.dispatch` is NOT called for the post-H/F body click.

**Verified end-to-end** in the browser:
| Scenario | Before fix | After fix |
|---|---|---|
| Dblclick header → edit → scroll → body click | scrollTop 3500 → 97 (-3403) | scrollTop unchanged (delta 0) |
| Gap click between pages (original SD-2749) | (already fixed) | scrollTop unchanged ✅ |
| Normal in-page text click | works | selection moves correctly ✅ |

Full `pnpm --filter super-editor test`: 18,422 / 18,422 passing.

tupizz:
Addressed both P2 review comments in 8ddc44efb.

**1. `[P2] Refresh story session after exiting H/F`** — correct catch. `exitHeaderFooterMode` synchronously clears the backing `StoryPresentationSessionManager`, so `activeStorySession` becomes stale right when I recomputed `useActiveSurfaceHitTest` from it. In production every H/F session is story-backed, so the old variable was always truthy → `useActiveSurfaceHitTest = true` → my SD-2749 page-gap guard was bypassed for the first click. Now I also re-read `getActiveStorySession()` after the H/F handler returns.

   Strengthened the existing unit test to mock a story-backed session, and added a second test that exercises the inter-page gap path (`pageIndex` undefined) while an H/F session was just exited. The new test fails on the prior commit and passes after the fix, so the regression is now actually pinned.

**2. `[P2] Restore document.elementsFromPoint after stubbing`** — correct, my bad. The two `PresentationEditor.test.ts` tests that simulated the page-under-cursor were leaking the stub onto subsequent tests in the same file. Wrapped both in try/finally that restores the original (or `delete`s the stub if there wasn't one — happy-dom doesn't ship the API).

Verified end-to-end:
- `pnpm --filter super-editor test` — 18,423 / 18,423 passing (one more test than before)
- Browser repro (header dblclick → edit → scroll → body click) → `scrollTop` delta = 0

caio-pizzol (approved):
@tupizz good call re-reading the story session after H/F exit, that path was easy to miss :)

i pushed a sibling regression test for the note-session × gap path, and ran render-parity against the merge-base (no rendering delta).

lgtm.

PR #3183, opened 2026-05-06
feat(rendering): SDT-wrapped content + page tokens inside tables and SCBs (SD-2516, SD-1333, SD-1332)

tupizz:
## Summary

Three Linear tickets, one root cause, one PR.

| Linear | What was visible | What was actually wrong |
|---|---|---|
| **[SD-2516](https://linear.app/superdocworkspace/issue/SD-2516)** | Table cell containing an SDT renders empty | `tableCellNodeToBlock` had no branch for `documentPartObject` children — the SDT was silently dropped |
| **[SD-1333](https://linear.app/superdocworkspace/issue/SD-1333)** (body) | Page number inside Signature SDT shows nothing | `handleStructuredContentBlockNode` had no branch for `documentPartObject` children — the wrapped paragraph was silently dropped |
| **[SD-1333](https://linear.app/superdocworkspace/issue/SD-1333)** (footer) | Footer with inline SDT page-number renders without the digit | The importer (`handleDocPartObj`) was always returning a block-level `'documentPartObject'`, even when the SDT was inline (no `<w:p>`/`<w:tbl>` in `sdtContent`). The parent paragraph translator then lifted it out, divorcing the `page-number` token from its enclosing paragraph |
| **[SD-1332](https://linear.app/superdocworkspace/issue/SD-1332)** | Page number inside table cell in the footer renders without the digit | Two layers: (a) the importer bug above, and (b) the page-token machinery walked only top-level paragraph blocks, never recursing into `TableBlock` cells |

All four bugs are different shapes of the same family: *"a container handler silently drops a child it doesn't recognize, OR a downstream walker stops at the wrong level."*

### 1. Importer (super-editor) — make `<w:sdt><w:docPartObj/></w:sdt>` honour its container

`handleDocPartObj` now mirrors `handleStructuredContentNode`'s inline-vs-block decision: if `sdtContent` has no `<w:p>` or `<w:tbl>`, it returns an inline `'structuredContent'` PM node, preserving `docPartGallery`/`docPartUnique` attrs and the full `sdtPr` passthrough for round-trip.

### 2. pm-adapter — let SDT containers transparently wrap content

`tableCellNodeToBlock` and `handleStructuredContentBlockNode` now both recurse into `documentPartObject` children, treating it as a transparent SDT wrapper. The parent's structured-content metadata is applied to the wrapped paragraphs/tables.

### 3. layout-engine + layout-bridge — let the page-token machinery see into tables

Three downstream walkers were top-level-only:

| Function | File | Was | Now |
|---|---|---|---|
| `hasPageTokens` (fast-path detector) | `layoutHeaderFooter.ts` | walks top-level `paragraph` blocks only | recurses into `TableBlock.rows[].cells[].(blocks│paragraph)` |
| `resolveHeaderFooterTokens` (per-page mutator) | `resolveHeaderFooterTokens.ts` | walks top-level `paragraph` blocks only | new `forEachParagraphBlock` helper recurses through tables |
| `cloneHeaderFooterBlocks` (per-page deep copy) | `resolveHeaderFooterTokens.ts` | shallow-copied non-paragraph blocks | deep-clones table rows/cells/inner paragraphs so per-page substitutions don't leak |
| `resolvePageNumberTokens` (body resolver) | `resolvePageTokens.ts` | iterated only fragments where `kind === 'para'` | also visits `kind === 'table'` fragments and emits an immutably-cloned table block with the substitution applied |

Without (3), per-page page numbers in headers/footers with table-wrapped tokens would silently use a single cached layout. (For the user-reported SD-1332 fixture the importer fix alone happens to also make the digit appear, but (3) closes the structural gap so the fast path correctly fires per page and tokens resolve in tables anywhere — body, headers, footers.)

## OOXML alignment

- **ECMA-376 §17.5.2** (Glossary Document Parts) — `<w:docPartObj>` is purely a metadata marker on `<w:sdtPr>`; it does not change whether the SDT is structurally inline or block. Inline-vs-block is determined by where `<w:sdt>` lives in the parent content model.
- **ECMA-376 §17.16.5.5** (PAGE field) — must reflect the current page number per occurrence, regardless of where it sits in the content tree (body, table cell, header, footer).

## Files changed

| Layer | File |
|---|---|
| Importer | `super-editor/.../sdt/helpers/handle-doc-part-obj.js` |
| pm-adapter | `pm-adapter/src/sdt/structured-content-block.ts` |
| pm-adapter | `pm-adapter/src/converters/table.ts` |
| layout-bridge | `layout-bridge/src/layoutHeaderFooter.ts` |
| layout-bridge | `layout-bridge/src/resolveHeaderFooterTokens.ts` |
| layout-engine | `layout-engine/src/resolvePageTokens.ts` |

Plus matching test files (8 new RED→GREEN tests across the three layers).

## Test plan

- [x] `pnpm --filter super-editor test` — 12,506 tests pass
- [x] `pnpm --filter @superdoc/pm-adapter test` — 1,803 tests pass
- [x] layout-bridge — 1,195 tests pass
- [x] layout-engine — 645 tests pass
- [x] **Browser end-to-end** for each fixture:
  - **SD-2516**: table cell renders the wrapped `m3ins.com` hyperlink
  - **SD-1333a**: footer renders the actual page number on each page (1, 2, …)
  - **SD-1333b** (California lease): footer renders `Page 1 of 1`
  - **SD-1332**: footer table cell renders the page-number digit
- [x] **`/compare-word-vs-superdoc`** PDF for SD-2516 / SD-1333a / SD-1333b — generated, structural match Word ↔ After (see prior comment on this PR)
- [x] BEFORE / AFTER captures for **SD-1332** confirm the digit is missing on `origin/main` and present after the fix

## Known parallel-case gap (out of scope, no fixture)

`resolvePageRefs.ts` walks blocks for PAGEREF cross-references with the same top-level-only loop. If a PAGEREF lives inside a table cell, it would have the same shape bug. Defer until a fixture surfaces it.

## Demo

CleanShot 2026-05-06 at 19 23 09@2x

CleanShot 2026-05-06 at 19 23 30@2x

CleanShot 2026-05-06 at 19 23 40@2x

CleanShot 2026-05-06 at 19 23 57@2x


tupizz:
## /compare-word-vs-superdoc — visual validation

Ran the skill against all 3 fixtures with the fix temporarily reverted to capture **before**, then restored for **after**. Word PDFs generated via AppleScript automation; per-page PNGs cropped from the live SuperDoc dev server at letter (816×1056) page size.

### Per-fixture summary

| Fixture | Word | Before | After | Verdict |
|---|---|---|---|---|
| **test-1** SD-2516 (SDT in table cell) | 1 page; cell renders `m3ins.com` link | 1 page; cell empty | 1 page; cell renders `m3ins.com` link | structural match Word↔After |
| **test-2** SD-1333a (footer page numbers) | 2 pages; `1` then `2` digits in body+footer | 2 pages; digits missing in both | 2 pages; `1`/`2` digits visible in body+footer | structural match Word↔After |
| **test-3** SD-1333b (California-lease footer) | 1 page; footer reads `Page 1 of 1` | 1 page; footer reads `Page  of  ` | 1 page; footer reads `Page 1 of 1` | structural match Word↔After |

### Spot-check (test-2 page 1)

| | Before | After | Word |
|---|---|---|---|
| Body | shows "Page number in structured content" only | shows "**1**" + text | shows "1" + text |
| Footer | shows "Page number in structured content in footer" only | shows "**1**  Page number…" | shows "1  Page number…" |

The fix produces exactly the structure Word emits.

### Artifact

Generated assets are at `/tmp/sd-2516-1333-fixtures/`:
- `SD-2516-1333-comparison.pdf` (printable side-by-side, landscape letter)
- `SD-2516-1333-page-by-page.html` (self-contained HTML)
- `per-page/{word,before,after}/test-{1,2,3}-page-{1,2}.png` (4 PNG files per side)

Will attach the PDF directly here on next push if it helps reviewers.

tupizz:
## /compare-word-vs-superdoc — combined report (all 4 fixtures)

Re-ran the skill against every fixture this PR closes, with the **full PR reverted** (origin/main, all 6 source files) for BEFORE and the **full PR applied** (commits `ad0facaef` + `a684cc497`) for AFTER. Word PDFs generated via AppleScript automation, per-page PNGs cropped from the live SuperDoc dev server at letter (816×1056) page size.

| # | Ticket | Word | Before (origin/main) | After (PR #3183) | Verdict |
|---|---|---|---|---|---|
| **test-1** | **SD-2516** | cell shows the wrapped `m3ins.com` link | cell empty | cell renders `m3ins.com` | structural match Word↔After |
| **test-2** | **SD-1333a** | footer shows `1`/`2` digits in body+footer SDTs | digits missing in both | `1`/`2` visible in body+footer | structural match Word↔After |
| **test-3** | **SD-1333b** (CA lease) | footer reads `Page 1 of 1` | footer reads `Page  of ` | footer reads `Page 1 of 1` | structural match Word↔After |
| **test-4** | **SD-1332** | cell renders `1  <<Page number in a table` | cell shows `<<Page number in a table` only | cell renders `1  <<Page number in a table` | structural match Word↔After |

All four pass. The fix produces exactly the structure Word emits.

### Spot-check (test-4 / SD-1332, page 1)

| | Visible content |
|---|---|
| **Word** | cell: `1  <<Page number in a table` |
| **Before** (origin/main) | cell: `<<Page number in a table` — no digit (matches Linear screenshot in the issue) |
| **After** (PR #3183) | cell: `1` + `<<Page number in a table` — matches Word |

### Spot-check (test-2 / SD-1333a, page 1)

| | Visible content |
|---|---|
| **Word** | top: `1` + "Page number in structured content"  ·  footer: `1  Page number in structured content in footer` |
| **Before** (origin/main) | top: just "Page number in structured content" — `1` missing  ·  footer: just "Page number in structured content in footer" — `1` missing |
| **After** (PR #3183) | matches Word exactly |

### Artifacts

Generated assets at `/tmp/sd-2516-1333-1332-fixtures/`:
- `PR-3183-comparison.pdf` (printable side-by-side, landscape letter, ~270 KB, 4 sections)
- `PR-3183-page-by-page.html` (self-contained HTML)
- `per-page/{word,before,after}/test-{1..4}-page-{1,2}.png` (5 PNGs per side: test-2 has 2 pages, the others 1)

This supersedes the prior 3-fixture comment — same generation pipeline, now covering all 4 fixtures with a clean origin/main BEFORE.

tupizz:
@chatgpt-codex-connector — confirmed and addressed in 62973b2a1.

**The catch is correct.** A body `TableBlock` can produce multiple `kind === 'table'` fragments (one per physical page), all sharing the same `blockId` with their own `fromRow..toRow`. My branch processed each block once and cloned the whole table with the *first* fragment's `displayPageText`, so a PAGE field rendered on page N>1 would resolve to page 1's number.

### Decision: scope reduction, not depth fix

The body-resolver branch was speculative defense-in-depth — no fixture exercises it. The actual SD-1332 fix lives in `layout-bridge/resolveHeaderFooterTokens.ts` (page-local: each H/F page owns its own block clone, so recursing into table cells *there* is safe and correct). Per-fragment substitution for body tables is a layered layout-pipeline change (synthetic per-page block IDs + targeted row cloning) that's better deferred until a body-table-with-PAGE fixture motivates it.

**Changes in 62973b2a1**:
- Removed the `if (fragment.kind === 'table')` branch from `resolvePageNumberTokens`.
- Removed the now-unused `tableContainsPageTokens` and `cloneTableWithResolvedTokens` helpers (~70 lines).
- Replaced the prior "substitutes inside body table" unit test with one that **pins the intentional limitation**: when the only token sits in a body table, the resolver must NOT report it as affected. Future readers see the constraint loudly.
- Added an explanatory comment in the body resolver pointing to the page-local H/F path and noting the deferred direction.

**Verified after the change**:
- SD-1332 footer still renders the page-number digit (browser repro on the live dev server)
- 645 layout-engine / 1,195 layout-bridge / 1,803 pm-adapter tests pass

Net diff: -89 / +34. The fix is now scoped to where it's verified correct.

tupizz:
## /compare-word-vs-superdoc — re-run after 62973b2a1

Re-captured AFTER state for all 4 fixtures with the latest commit (`62973b2a1` — body-resolver branch removed, scope-reduced to H/F path). BEFORE state unchanged (origin/main).

| # | Ticket | Word | Before (origin/main) | After (PR HEAD) | Verdict |
|---|---|---|---|---|---|
| **test-1** | **SD-2516** | cell renders `m3ins.com` | cell empty | cell renders `m3ins.com` | ✅ structural match |
| **test-2** | **SD-1333a** | footer `1` / `2` digits | digits missing | `1` / `2` visible | ✅ structural match |
| **test-3** | **SD-1333b** | footer reads `Page 1 of 1` | `Page  of ` (digits missing) | `Page 1 of 1` | ✅ structural match |
| **test-4** | **SD-1332** | cell renders `1  <<Page number in a table` | `<<Page number in a table` only (no digit) | `1  <<Page number in a table` | ✅ structural match |

**No regression from the body-branch removal.** SD-1332's footer still renders the digit because the actual fix was always in the page-local H/F path (`resolveHeaderFooterTokens` recursion), not in the body resolver branch we dropped.

Refreshed PDF: `/tmp/sd-2516-1333-1332-fixtures/PR-3183-comparison.pdf` (~270 KB).

packages/layout-engine/pm-adapter/src/converters/table.ts
```diff
@@ -442,6 +442,53 @@ const parseTableCell = (args: ParseTableCellArgs): TableCell | null => {
       continue;
     }
 
+    // SD-2516: a documentPartObject is a transparent SDT wrapper. Word's
+    // generic SDT in a table cell parses into PM as
+    // `tableCell > documentPartObject > paragraph` and previously fell
+    // through with no branch, dropping the cell content. Flatten the
+    // wrapper's paragraph/table children into the cell's blocks array.
+    if (childNode.type === 'documentPartObject' && Array.isArray(childNode.content)) {
```
caio-pizzol, line 468: blocking: nested documentPartObject in a table cell drops the entire table.

verified with `vitest run packages/layout-engine/pm-adapter/src/converters/_verify-pr-3183.test.ts` -> `tableNodeToBlock` against `tableCell > documentPartObject > documentPartObject > paragraph` returned `null`.

likely fix: the new branch only flattens `paragraph | table` direct children, so nested wrappers fall through, the cell ends up empty, and the row is rejected. `structured-content-block.ts:43-92` already uses a recursive `visitChild` for this; share that helper or recurse here.

<details>
<summary>repro test</summary>

```ts
const node = { type: 'table', content: [{ type: 'tableRow', content: [{ type: 'tableCell',
  content: [{ type: 'documentPartObject', content: [
    { type: 'documentPartObject', content: [{ type: 'paragraph', content: [{ type: 'text', text: 'inner' }] }] }
  ]}]}]}]};
const result = tableNodeToBlock(node, { /* deps */ });
// pre-fix: result === null (entire table dropped)
// post-fix: result.rows[0].cells[0].blocks[0].kind === 'paragraph'
```

</details>

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/handle-doc-part-obj.js
```diff
@@ -17,6 +17,21 @@ export function handleDocPartObj(params) {
 
   const content = node?.elements.find((el) => el.name === 'w:sdtContent');
 
+  // SD-1333: Mirror handleStructuredContentNode's inline-vs-block detection.
+  // An <w:sdt> sitting inside a <w:p> (e.g. an inline PAGE-field SDT inside
+  // a footer paragraph) has only run-level children in sdtContent. Without
+  // this, we always emit a block-level 'documentPartObject', which the
+  // paragraph translator then lifts out of its parent paragraph — losing the
+  // wrapping that the page-number resolver depends on.
+  const hasBlockChild = !!content?.elements?.some((el) => el?.name === 'w:p' || el?.name === 'w:tbl');
```
caio-pizzol, line 25: blocking: block-level docPartObj SDT containing only a nested w:sdt is misclassified as inline.

verified with `vitest run packages/super-editor/.../_verify-pr-3183.test.js` -> `handleDocPartObj` on a body-context `<w:sdt><w:docPartObj/><w:sdtContent><w:sdt><w:sdtContent><w:p/></w:sdtContent></w:sdt></w:sdtContent></w:sdt>` returned `type: 'structuredContent'` (inline) carrying block children.

likely fix: `CT_SdtContentBlock` legally allows nested `w:sdt` / `customXml` per ECMA section 17.5.2.34. either walk inner sdtContent for nested w:sdt before deciding, or gate the inline branch on `params.path` having a `w:p` ancestor.

<details>
<summary>repro test</summary>

```js
const node = { name: 'w:sdt', elements: [
  { name: 'w:sdtPr', elements: [{ name: 'w:docPartObj', elements: [
    { name: 'w:docPartGallery', attributes: { 'w:val': 'Page Numbers (Top of Page)' } }
  ]}]},
  { name: 'w:sdtContent', elements: [
    { name: 'w:sdt', elements: [
      { name: 'w:sdtPr', elements: [] },
      { name: 'w:sdtContent', elements: [{ name: 'w:p', elements: [] }] }
    ]}
  ]}
]};
const result = handleDocPartObj({ nodes: [node], nodeListHandler: { handler: () => [{ type: 'structuredContentBlock', content: [{ type: 'paragraph' }] }] }, path: [{ name: 'w:body' }] });
// pre-fix: result.type === 'structuredContent' (inline) with block children
// post-fix: result.type === 'documentPartObject' (block)
```

</details>

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/handle-doc-part-obj.js
```diff
@@ -28,6 +43,47 @@ export function handleDocPartObj(params) {
...
+const inlineDocPartHandler = (params) => {
+  const node = params.nodes[0];
+  const translatedContent = params.nodeListHandler.handler({
+    ...params,
+    nodes: node?.elements || [],
+    path: [...(params.path || []), node],
+  });
+  const sdtPr = params.extraParams.sdtPr;
+  const docPartGalleryType = params.extraParams.docPartGalleryType;
+  const id = sdtPr?.elements?.find((el) => el.name === 'w:id')?.attributes['w:val'] || '';
```
caio-pizzol: blocking: empty w:id round-trips as invalid `<w:id w:val=""/>`.

verified with `vitest run packages/super-editor/.../_verify-pr-3183-roundtrip.test.js` -> import an `<w:sdt><w:docPartObj/></w:sdt>` with no `<w:id>` (Word allows this), export through `translateStructuredContent`, result is `{ name: 'w:id', attributes: { 'w:val': '' } }`.

the block path is fine because `translate-document-part-obj.js:69` runs ids through `sanitizeId`. swap `|| ''` for `|| null` so the export check `if (attrs.id != null)` correctly skips empty ids:

```suggestion
  const id = sdtPr?.elements?.find((el) => el.name === 'w:id')?.attributes?.['w:val'] || null;
```

<details>
<summary>repro test</summary>

```js
const node = { name: 'w:sdt', elements: [
  { name: 'w:sdtPr', elements: [{ name: 'w:docPartObj', elements: [
    { name: 'w:docPartGallery', attributes: { 'w:val': 'Page Numbers (Bottom of Page)' } }
  ]}]},
  { name: 'w:sdtContent', elements: [
    { name: 'w:r', elements: [{ name: 'w:fldChar', attributes: { 'w:fldCharType': 'begin' } }] },
    { name: 'w:r', elements: [{ name: 'w:instrText', elements: [{ type: 'text', text: ' PAGE ' }] }] },
    { name: 'w:r', elements: [{ name: 'w:fldChar', attributes: { 'w:fldCharType': 'end' } }] }
  ]}
]};
const pmNode = handleDocPartObj({ nodes: [node], nodeListHandler: { handler: () => [{ type: 'page-number' }] }, path: [{ name: 'w:p' }] });
const out = translateStructuredContent({ node: { ...pmNode, content: [] }, relationships: { rels: [] }, nodeListHandler: { handler: () => [] } });
const sdtPr = out.elements.find((e) => e.name === 'w:sdtPr');
const idEl = sdtPr.elements.find((e) => e.name === 'w:id');
// pre-fix: idEl.attributes['w:val'] === ''
// post-fix: idEl === undefined
```

</details>

caio-pizzol (commented):
> [!CAUTION]
> needs work, three things inline.

three regressions confirmed via runtime against the PR's source. a `tableCell > documentPartObject > documentPartObject > paragraph` tree drops the whole table; a block-context docPartObj SDT whose sdtContent is just a nested `w:sdt` flips to inline, losing block content; and any docPartObj SDT without an explicit `<w:id>` round-trips as invalid `<w:id w:val=""/>`. all three are narrow, located fixes.

expected: `packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/handle-doc-part-obj.js`, `packages/layout-engine/pm-adapter/src/converters/table.ts` / avoid: `packages/layout-engine/layout-engine/src/resolvePageTokens.ts` (body-table deferral comment + pinning test are correct, leave alone)

<details>
<summary>agent context</summary>

```json
{
  "schema": "review_handoff_v1",
  "review_id": "sd-2516-1333-1332-pr-3183",
  "findings": [
    {
      "id": "B-table-nested-docpart",
      "severity": "blocking",
      "anchors": [
        {
          "file": "packages/layout-engine/pm-adapter/src/converters/table.ts",
          "line": 450,
          "quote": "if (childNode.type === 'documentPartObject' && Array.isArray(childNode.content))"
        }
      ],
      "claim": "new documentPartObject branch in parseTableCell only flattens paragraph|table direct children; a nested documentPartObject (or structuredContentBlock) wrapper falls through, the cell ends up with no blocks, and tableNodeToBlock returns null - the entire table is dropped, not just the cell content.",
      "failure_mode": "silent data loss: a Word doc whose table cells contain doubly-wrapped SDT content imports as if the table did not exist.",
      "verification": [
        {
          "command": "vitest run packages/layout-engine/pm-adapter/src/converters/_verify-pr-3183.test.ts",
          "result_summary": "tableNodeToBlock(tableCell > documentPartObject > documentPartObject > paragraph) === null - whole table dropped",
          "status": "executed"
        }
      ],
      "ruled_out": [
        {
          "hypothesis": "this is the same shape as the existing structuredContentBlock branch in the same file (also one-level), so it's pre-existing",
          "why_ruled_out": "structured-content-block.ts:43-92 in this PR introduces a recursive visitChild for exactly this case in the SCB module - the table-cell branch is the only entry point that did not get the matching recursion."
        }
      ],
      "handoff": {
        "recommended_start": "share visitChild from structured-content-block.ts or recurse on nested documentPartObject children inside the table-cell branch",
        "suggested_fix": "extract a shared helper that walks nested documentPartObject/structuredContentBlock wrappers down to paragraph|table leaves; both call sites call it.",
        "expected_files": [
          "packages/layout-engine/pm-adapter/src/converters/table.ts",
          "packages/layout-engine/pm-adapter/src/sdt/structured-content-block.ts"
        ],
        "avoid_files": [
          "packages/layout-engine/layout-engine/src/resolvePageTokens.ts"
        ],
        "do_not_repeat": [
          "pnpm test:layout"
        ]
      },
      "verification_after_fix": [
        {
          "command": "vitest run packages/layout-engine/pm-adapter/src/converters/_verify-pr-3183.test.ts",
          "expected": "the regression test (flipped to assert non-null + paragraph preserved) passes"
        }
      ]
    },
    {
      "id": "E-block-sdt-nested-only",
      "severity": "blocking",
      "anchors": [
        {
          "file": "packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/handle-doc-part-obj.js",
          "line": 26,
          "quote": "const hasBlockChild = !!content?.elements?.some((el) => el?.name === 'w:p' || el?.name === 'w:tbl');"
        }
      ],
      "claim": "hasBlockChild only checks direct w:p / w:tbl children. CT_SdtContentBlock (ECMA-376 section 17.5.2.34) legally allows customXml or a nested w:sdt as the sole content. Those cases now flip to inlineDocPartHandler, emitting type:'structuredContent' (inline PM node) at body level with block children inside.",
      "failure_mode": "node-type mismatch: parent expects block content but receives an inline PM node, leading to dropped or misplaced content during conversion.",
      "verification": [
        {
          "command": "vitest run packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/_verify-pr-3183.test.js",
          "result_summary": "handleDocPartObj on a block-context docPartObj SDT whose sdtContent is just a nested <w:sdt><w:sdtContent><w:p/></w:sdtContent></w:sdt> returned type:'structuredContent' with a block child inside.",
          "status": "executed"
        }
      ],
      "ruled_out": [
        {
          "hypothesis": "hasBlockChild is the same heuristic as the pre-existing handle-structured-content-node.js, so this is pre-existing",
          "why_ruled_out": "true that the heuristic is shared, but PR-3183 newly routes empty/nested-sdt-only sdtContent through it for docPartObj specifically. Before this PR, every docPartObj returned type:'documentPartObject' (block); now some return type:'structuredContent' (inline)."
        }
      ],
      "handoff": {
        "recommended_start": "either walk inner sdtContent for nested w:sdt before deciding inline-vs-block, or use the parent path - if path indicates we are inside a w:p, treat as inline; otherwise default to block.",
        "suggested_fix": "replace the direct-child check with a recursive walk that returns true if any descendant of sdtContent is w:p or w:tbl; or gate the inline branch on params.path including a w:p ancestor.",
        "expected_files": [
          "packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/handle-doc-part-obj.js"
        ],
        "avoid_files": [],
        "do_not_repeat": []
      },
      "verification_after_fix": [
        {
          "command": "vitest run packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/_verify-pr-3183.test.js",
          "expected": "block-context regression test asserts result.type === 'documentPartObject' and passes"
        }
      ]
    },
    {
      "id": "F-empty-id-roundtrip",
      "severity": "blocking",
      "anchors": [
        {
          "file": "packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/handle-doc-part-obj.js",
          "line": 67,
          "quote": "const id = sdtPr?.elements?.find((el) => el.name === 'w:id')?.attributes['w:val'] || '';"
        }
      ],
      "claim": "inlineDocPartHandler defaults attrs.id to '' when w:id is absent (Word allows omitting it). Inline-context docPartObj SDTs now export through translateStructuredContent, which writes <w:id> for any attrs.id != null - including empty string. Round-trip produces invalid OOXML <w:id w:val=''/>.",
      "failure_mode": "corrupted output: any SuperDoc-emitted .docx that round-trips a no-id docPartObj inline SDT contains a schema-invalid empty w:id, which Word may flag or silently drop.",
      "verification": [
        {
          "command": "vitest run packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/_verify-pr-3183-roundtrip.test.js",
          "result_summary": "import-then-export of <w:sdt><w:docPartObj/></w:sdt> with no <w:id> emits {name:'w:id', attributes:{'w:val':''}}.",
          "status": "executed"
        }
      ],
      "ruled_out": [
        {
          "hypothesis": "the block path has the same shape so this is pre-existing",
          "why_ruled_out": "the block path runs ids through sanitizeId in translate-document-part-obj.js:69 which returns undefined for ''; the inline path is new in this PR and bypasses sanitizeId."
        }
      ],
      "handoff": {
        "recommended_start": "swap '|| \"\"' for '|| null' so the export check 'if (attrs.id != null)' correctly skips empty ids",
        "suggested_fix": "const id = sdtPr?.elements?.find((el) => el.name === 'w:id')?.attributes?.['w:val'] || null;",
        "expected_files": [
          "packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/handle-doc-part-obj.js"
        ],
        "avoid_files": [],
        "do_not_repeat": []
      },
      "verification_after_fix": [
        {
          "command": "vitest run packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/_verify-pr-3183-roundtrip.test.js",
          "expected": "round-trip test (flipped to assert no w:id element in the exported sdtPr) passes"
        }
      ]
    }
  ]
}
```

</details>

tupizz:
Thanks for the careful review @caio-pizzol — addressed all three blocking findings in f0a1cb2:

**B (table.ts:450)** — extracted a recursive flattenSdtWrappersIntoCell helper inside parseTableCell that walks both documentPartObject and structuredContentBlock wrappers down to paragraph/table leaves. SCB metadata is preserved (resolved at the wrapper level on first encounter and applied to descendant paragraph/table blocks). Two RED tests added that fail on main: nested docPartObj > docPartObj > paragraph and docPartObj > structuredContentBlock > paragraph.

**E (handle-doc-part-obj.js:26)** — switched to a conjunctive rule: emit inline only when isInsideParagraph (path has a w:p ancestor) AND !hasBlockChild. The path-only check broke the existing Table-of-Figures regression test (block-content nested inside a w:p), and the hasBlockChild-only check missed your case (block-context SDT wrapping only a nested w:sdt). Both pre-existing tests and the new RED test for the nested-only sdtContent case now pass.

**F (handle-doc-part-obj.js:67)** — id defaults to null instead of '', plus added the optional chain on .attributes that you flagged. RED test pins this so a future refactor can't reintroduce the empty-string round-trip.

I left resolvePageTokens.ts alone per your note — the deferral comment + pinning test from the earlier Codex review iteration are still in place.

All three packages green: pm-adapter (1805), layout-bridge (1201), super-editor (12510 + 13 skipped). Browser sanity-checked the SD-2516 fixtures at localhost:9094 — table cells render text and pagination is unchanged.

caio-pizzol (commented):
hey @tupizz! thanks for addressing last round - B, E, and F all check out.

i pushed a follow-up commit (30d7aef13) routing the `structuredContentBlock` cell-child branch through the same `flattenSdtWrappersIntoCell` helper, with a regression test. without it, a cell whose direct child was `structuredContentBlock > documentPartObject > paragraph` dropped the whole table.

lgtm once CI is green.

caio-pizzol (approved):

PR #3186, opened 2026-05-07
feat: enhance block variable styling and add tests for structured con…

palmer-cl:
- Added block field styling parity with inline fields for label, border, hover, selection, and background customization.
- Introduced matching block class names while preserving existing legacy selectors for backwards compatibility.
- Added unit and visual coverage for owner/signer inline and block field styling.

Example:
SDtest


caio-pizzol (approved):
hey @palmer-cl! good call routing field colors through the existing `--sd-content-controls-*` tokens instead of adding new bespoke vars, and keeping the legacy block label class wired up via the comma-list rule :)

lgtm.

PR #3188, opened 2026-05-07
fix: missing headers in collab mode; wrong odd/even headers

VladaHarbour:


packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
```diff
@@ -4533,6 +4533,14 @@ export class PresentationEditor extends EventEmitter {
...
     const handleCollaborationReady = (payload: unknown) => {
       this.emit('collaborationReady', payload);
+      // Collaboration bootstrap can hydrate header/footer parts on this client
+      // without emitting partChanged. Force a header/footer refresh pass so the
+      // importer tab sees the same headers/footers immediately.
+      this.#headerFooterSession?.refreshStructure();
+      this.#flowBlockCache.setHasExternalChanges(true);
+      this.#pendingDocChange = true;
+      this.#selectionSync.onLayoutStart();
+      this.#scheduleRerender();
```
luccas-harbour: It looks like this block does the same things as the `handleDocumentReplaced` function in lines 4561-4567. Maybe it would be worth pulling this into a shared helper?

luccas-harbour:
hey @VladaHarbour ! I had a look and tested the implementation and I can see that it fixes the issue with the document that was in the ticket.

I left a small inline comment about extracting a shared helper for some duplicated code. I also ran a codex review and it found the issue below related to a stale header/footer cache if a new doc is uploaded and contains a header with the same relationship id as a header in the old doc, can you check?

```
- [P2] Purge cached header/footer editors on replacement — packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts:4561-4562
  When `replaceFile()` loads a document that reuses an existing header/footer relationship id, `refreshStructure()` only rebuilds descriptors and disposes entries for ids that disappeared. Any cached header/footer editor for the reused id survives, and `HeaderFooterEditorManager.getDocumentJson()` prefers that live editor over the new converter snapshot, so this rerender can continue showing/editing the previous document's header/footer after replacement. This affects sessions where a header/footer was opened before replacing the file; the replacement path needs to invalidate/tear down existing header/footer editors, not just refresh the structure.
```

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
VladaHarbour: added shared helper

VladaHarbour:
Hi @luccas-harbour! Added cache rebuild logic to address codex comment

VladaHarbour:
https://linear.app/superdocworkspace/issue/SD-2643/bug-headersfooters-do-not-render-reliably-in-collabration

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
```diff
@@ -7189,6 +7227,44 @@ export class PresentationEditor extends EventEmitter {
     overlay.appendChild(fragment);
   }
 
+  #resolveAlternateHeadersFlag(): boolean {
```
luccas-harbour, line 7231: one more thing @VladaHarbour. I noticed we already have an `isAlternatingHeadersOddEven` function in [docxImporter.js](https://github.com/superdoc-dev/superdoc/blob/43a9ff0ed42094fcea10fb07d7b768176eef8f3a/packages/super-editor/src/editors/v1/core/super-converter/v2/importer/docxImporter.js#L1263-L1263) that does about the same as what's being done here

I wonder if we could just reuse that function here?

luccas-harbour:
> Hi @luccas-harbour! Added cache rebuild logic to address codex comment

thanks @VladaHarbour! I left just one more small comment about possibly reusing logic for identifying alternating headers

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
VladaHarbour: using the one from docImporter now

luccas-harbour (approved):
LGTM

PR #3190, opened 2026-05-07
Fix/unset link clear transient hyperlink styleid

### LINKED ISSUE #3189: Unsetting link leaves transient hyperlink styleId and keeps underline tool active
### What happened?

When removing a link (unsetLink()), the editor can leave transient hyperlink style metadata behind, causing formatting to appear still active after unlink.

**Problem**
After unlinking text, hyperlink-derived formatting (underline/color in content and underline tool active in toolbar) may remain active even though the link mark is removed.

**Root Cause**
unsetLink() removed the link mark, but transient hyperlink style IDs could remain in document state:

textStyle.attrs.styleId
run.attrs.runProperties.styleId
Affected transient IDs:

Hyperlink
FollowedHyperlink
Because style resolution still sees these IDs, hyperlink formatting can persist.

**Expected Behavior**
Unlink should remove link-derived transient hyperlink style references while preserving non-link character styles

### Steps to reproduce

1. create a link in the text
2. Remove the link
3. Color and underline remain active in text and underline toolbar button remains active

### SuperDoc version

_No response_

### Browser

None

### Additional context

_No response_

christos8333:
### Unlink now fully removes transient hyperlink styling instead of only removing the link mark.

**What this fixes**
Removing a link could leave formatting “stuck” (underline/color and underline tool was shown as active in toolbar).
This happened because hyperlink style metadata (styleId: Hyperlink / FollowedHyperlink) could remain on text/run state after unsetLink().

**What changed in link.js**
During unsetLink(), we now clear hyperlink styleId only when it is one of the transient link style IDs.
We clear it in both places where link-derived style may persist:
textStyle mark attrs (mark.attrs.styleId)
run node attrs (runProperties.styleId)

**Why this is safe**
We only target transient hyperlink style IDs, so non-link character styles (e.g. Emphasis, custom styles) are preserved.
Other unrelated formatting is not broadly removed by this cleanup.

**Tests**
Added regression tests for unsetLink() to lock in the transient hyperlink cleanup behavior.
The tests verify that when unlinking text, transient hyperlink styleId values (Hyperlink, FollowedHyperlink) are cleared in both state locations where they can persist: textStyle.attrs.styleId and runProperties.styleId.
They also verify underline behavior is correct in both directions: pre-existing underline is preserved, while link-added underline is removed.
Together, these tests prevent regressions where link formatting appears “stuck” after unlink while ensuring non-link formatting behavior remains intact.

Closes #3189


packages/super-editor/src/editors/v1/extensions/link/link.js
```diff
@@ -226,6 +234,16 @@ export const Link = Mark.create({
             to = from + finalText.length;
           }
 
+          let hadUnderline = false;
+          if (underlineMarkType) {
+            state.doc.nodesBetween(from, to, (node) => {
```
caio-pizzol: repro:
1. type "hi", select all
2. click the link icon
3. change "Text" in the dialog to something longer than "hi"
4. paste a URL, click Apply

the editor crashes: `TypeError: Cannot read properties of undefined (reading 'nodeSize')`.

fix: do the `hadUnderline` scan before line 234 reassigns `to`, or scan `tr.doc` after the new text is inserted.

packages/super-editor/src/editors/v1/extensions/link/link.js
```diff
@@ -226,6 +234,16 @@ export const Link = Mark.create({
             to = from + finalText.length;
           }
 
+          let hadUnderline = false;
```
caio-pizzol: repro:
1. type "abcdef"
2. underline only "abc"
3. select all
4. click the link icon, paste a URL, Apply
5. click the link, click Remove

result: "abcd" stays underlined. "d" was plain to start with, and there's no easy way to clear it.

the flag is one true/false for the whole selection, so any partial underline turns it on. fix: track it per character, or drop the underline-removal step and let the styleId cleanup handle the visual.

packages/super-editor/src/editors/v1/extensions/link/link.js
```diff
@@ -237,7 +255,7 @@ export const Link = Mark.create({
             if (id) rId = id;
           }
 
-          const linkAttrs = { text: finalText, rId };
+          const linkAttrs = { text: finalText, rId, hadUnderline };
```
caio-pizzol: repro:
1. type "hi", select all, link to `https://first.example.com`, Apply
2. click the link, change URL to `https://second.example.com`, Apply
3. click the link, Remove

result: "hi" is still underlined, even though it was plain to start with.

editing a link re-runs `setLink`, which scans and sees the underline the first link added. the flag flips to `true` and unlink leaves it alone.

caio-pizzol (changes requested):
hey @christos8333! thanks for the PR. the main fix (clearing the Hyperlink styleId on unlink) looks right, and the tests for it are solid.

three issues with the new `hadUnderline` flag, all reachable through the toolbar - left them inline. they share one fix: figure out the value at unlink time, not link time.

one more edge case worth a regression test (imported link carrying its underline as an inline mark loses underline on unlink): [attach fixture].

happy to pair or pick up the fix.

packages/super-editor/src/editors/v1/extensions/link/link.js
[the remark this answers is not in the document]
caio-pizzol: @christos8333 valid - i reproduced this through the toolbar flow. flagged it in the review at line 237 with a fix sketch.

christos8333:
> hey @christos8333! thanks for the PR. the main fix (clearing the Hyperlink styleId on unlink) looks right, and the tests for it are solid.
> 
> three issues with the new `hadUnderline` flag, all reachable through the toolbar - left them inline. they share one fix: figure out the value at unlink time, not link time.
> 
> one more edge case worth a regression test (imported link carrying its underline as an inline mark loses underline on unlink): [attach fixture].
> 
> happy to pair or pick up the fix.

hi @caio-pizzol . thanks for the input. yes, you are right It can be derived at unlink time. 
I will try it shortly and let you know if I need help.

christos8333:
@caio-pizzol all issues should be fixed in https://github.com/superdoc-dev/superdoc/pull/3190/changes/7e819e1b41663560234a0d61c3feb77284100e58

Also checked that word imported links are not removed. Made another branch out of this one to fix that.

caio-pizzol (commented):
hey @christos8333! thanks for addressing last round - the autoAdded approach reads much cleaner.

i pushed one more commit: the same fix for the URL-paste code path (which was creating the auto-underline without the `autoAdded` flag, so pasted URLs kept a stuck underline after Remove Link), plus regression tests for all four prior bugs.

lgtm once CI is green.

caio-pizzol (approved):

PR #3207, opened 2026-05-07
feat: render images inside Word textboxes (SD-2804)

tupizz:
## Summary

Renders inline `w:drawing` images inside Word textbox content. Previously, the textbox imported with the image silently stripped — the textbox rendered as an empty box even though export round-tripped the image untouched.

Linear: [SD-2804](https://linear.app/superdocworkspace/issue/SD-2804)

## Spec basis

ECMA-376 §20.4.2.38 (`CT_TxbxContent`) defines textbox content as `EG_BlockLevelElts (1..unbounded)` — i.e. a textbox can hold the same content as the document body, with three exclusions: cross-story refs (comments/footnotes/endnotes), VML, and nested `txbxContent`. Notably, paragraphs inside `w:txbxContent` carry the same `CT_P` content model as body paragraphs, including runs with inline `w:drawing` images.

The text-only extractor used in `extractTextFromTextBox.handleRun` only walked `w:t / w:tab / w:br / sd:autoPageNumber / sd:totalPageNumber` — `w:drawing` was silently ignored.

## Approach

Minimum surgical change — extend the existing text-parts model with one image part kind:

1. **`TextPart` contract** gains optional `kind: 'image'` plus `src / width / height / alt`.
2. **Importer** (`extractTextFromTextBox.handleRun`) branches on `w:drawing`, reuses the v3 `handleImageNode` to resolve `r:embed → media path`, then upgrades the path to a data URI from `converter.media` (the text-parts model has no downstream hydration step like body `ImageRun`s do).
3. **Painter** (`createFallbackTextElement`) renders parts with `kind: 'image'` as inline `` next to text spans.

No new PM nodes, no new pm-adapter wiring, no schema changes, no NodeHandlerContext threading.

## Before / after

Fixture: a DOCX with a textbox-in-header containing a single inline image.

| Before | After |
|---|---|
| Empty textbox outline; image silently dropped on import | Image renders inline inside the textbox, matching what Word shows |

Captured via agent-browser against the dev server: see `/tmp/sd-2804-final3.png`.

## Test plan

- [x] Unit test: importer emits an image part in `textContent.parts` for an inline `w:drawing` inside the textbox (`encode-image-node-helpers.test.js`)
- [x] super-editor full suite: 12,645 tests passing
- [x] painter-dom full suite: 1,064 tests passing
- [x] pm-adapter full suite: 1,788 tests passing
- [x] layout-bridge full suite: 1,206 tests passing
- [x] Browser: upload the SD-2804 fixture, confirm the image renders inside the textbox

## Out of scope (deferred)

The fixture's image is `wp:inline` inside a textbox run — the most common case. ECMA-376 also permits richer block-level content inside a textbox: tables, lists, SDTs, hyperlinks, fields, math. The current text-parts model can't represent those; surfacing them would need to flow `w:txbxContent` through the same body pipeline (`handleParagraphNode` recursion) and likely a container PM node (`shapeTextbox` schema already exists for the legacy v:pict path).

That refactor is intentionally deferred — the supplied SD-2804 fixture has only an image, and Option B was over-engineering for the immediate user-visible bug. Tracking issue / future PR scope for content beyond inline-image-in-textbox.

## Related

- SD-2745 (header-anchored floating textboxes) — handles textbox **position**; this PR handles textbox **content**. They compose.

### DIFF
--- a/packages/layout-engine/painters/dom/src/renderer.ts
+++ b/packages/layout-engine/painters/dom/src/renderer.ts
@@ -4461,14 +4461,17 @@ export class DomPainter {
       } else if (part.kind === 'image' && part.src) {
         // SD-2804: image part produced by the textbox importer for an
         // inline w:drawing inside a textbox run. Render as <img> alongside
-        // sibling text spans so layout matches Word's inline flow.
+        // sibling text spans so layout matches Word's inline flow. Match
+        // body inline images' baseline default (`vertical-align: bottom`)
+        // so an image and adjacent text line up the same way inside a
+        // textbox as outside.
         const img = this.doc!.createElement('img');
         img.src = part.src;
         img.alt = part.alt ?? '';
         if (typeof part.width === 'number') img.style.width = `${part.width}px`;
         if (typeof part.height === 'number') img.style.height = `${part.height}px`;
         img.style.display = 'inline-block';
-        img.style.verticalAlign = 'middle';
+        img.style.verticalAlign = 'bottom';
         currentParagraph.appendChild(img);
       } else {
         const span = this.doc!.createElement('span');
--- a/packages/layout-engine/painters/dom/src/renderer.ts
+++ b/packages/layout-engine/painters/dom/src/renderer.ts
@@ -4458,6 +4458,18 @@ export class DomPainter {
         if (part.isEmptyParagraph) {
           currentParagraph.style.minHeight = '1em';
         }
+      } else if (part.kind === 'image' && part.src) {
+        // SD-2804: image part produced by the textbox importer for an
+        // inline w:drawing inside a textbox run. Render as <img> alongside
+        // sibling text spans so layout matches Word's inline flow.
+        const img = this.doc!.createElement('img');
+        img.src = part.src;
+        img.alt = part.alt ?? '';
+        if (typeof part.width === 'number') img.style.width = `${part.width}px`;
+        if (typeof part.height === 'number') img.style.height = `${part.height}px`;
+        img.style.display = 'inline-block';
+        img.style.verticalAlign = 'middle';
+        currentParagraph.appendChild(img);
       } else {
         const span = this.doc!.createElement('span');
         span.textContent = this.resolveShapeTextPartText(part, context);
--- a/packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
@@ -1008,29 +1008,6 @@ const handleChartDrawing = (params, node, graphicData, size, padding, marginOffs
  *   wrap: string
  * }|null} Text content with formatting information and line break markers, or null if no text found
  */
-// SD-2804: turn a path-style src ("word/media/image1.png" or "media/image1.png")
-// into a data URI by reading converter.media. Mirrors how layout-engine's
-// hydrateRuns resolves body ImageRuns, but inline since the text-parts model
-// the shape painter consumes has no downstream hydration step.
-function resolveImagePartSrc(src, params, extension) {
-  if (!src || src.startsWith('data:')) return src;
-  const media = params?.converter?.media;
-  if (!media) return src;
-  const candidates = [src];
-  if (src.startsWith('word/')) candidates.push(src.slice(5));
-  else candidates.push(`word/${src}`);
-  for (const candidate of candidates) {
-    const data = media[candidate];
-    if (!data) continue;
-    if (typeof data === 'string') {
-      if (data.startsWith('data:')) return data;
-      const ext = extension || (src.includes('.') ? src.slice(src.lastIndexOf('.') + 1) : 'png');
-      return `data:image/${ext};base64,${data}`;
-    }
-  }
-  return src;
-}
-
 function extractTextFromTextBox(textBoxContent, bodyPr, params = {}) {
   if (!textBoxContent || !textBoxContent.elements) return null;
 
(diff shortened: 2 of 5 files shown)


packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
```diff
@@ -1061,6 +1084,32 @@ function extractTextFromTextBox(textBoxContent, bodyPr, params = {}) {
...
+        // content, including runs with inline w:drawing images. Reuse the
+        // existing v3 wp drawing handler so the rId → resolution matches
+        // what body paragraphs use, then upgrade the path-style src to a
+        // data URI from converter.media (the text-parts model has no
+        // downstream hydration step like body ImageRuns do).
+        const inlineOrAnchor = el.elements?.find((child) => child?.name === 'wp:inline' || child?.name === 'wp:anchor');
+        if (inlineOrAnchor) {
+          const isAnchor = inlineOrAnchor.name === 'wp:anchor';
+          const imagePm = handleImageNode(inlineOrAnchor, { ...params, nodes: [el] }, isAnchor);
+          if (imagePm?.attrs?.src) {
```
luccas-harbour: codex found the issue below so it looks like we might end up rendering images that shouldn't be rendered, can you double-check?

```
  - [P2] Skip hidden textbox images — src/editors/v1/core/super-
    converter/v3/handlers/wp/helpers/encode-image-node-helpers.js:1098-1098
    When an image inside a textbox has wp:docPr hidden="1", handleImageNode already marks the returned image attrs as hidden, but this branch only checks for attrs.src and still
    emits a visible image part. This regresses hidden drawings specifically inside textboxes; top-level hidden drawings are filtered later, while these new text parts bypass that
    filtering and get rendered as <img>.
```

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
```diff
@@ -1008,6 +1008,29 @@ const handleChartDrawing = (params, node, graphicData, size, padding, marginOffs
  *   wrap: string
  * }|null} Text content with formatting information and line break markers, or null if no text found
  */
+// SD-2804: turn a path-style src ("word/media/image1.png" or "media/image1.png")
+// into a data URI by reading converter.media. Mirrors how layout-engine's
+// hydrateRuns resolves body ImageRuns, but inline since the text-parts model
+// the shape painter consumes has no downstream hydration step.
+function resolveImagePartSrc(src, params, extension) {
```
luccas-harbour: it looks like this function duplicates `pm-adapter/utilities.ts:976-1034` (`resolveImageSrc` / `hydrateImageBlocks`) — and isn't quite the same. pm-adapter's helper handles `Uint8Array` (collab/Y.js binary media) via `TextDecoder`; this `typeof data === 'string'` check rejects it. so a textbox image stored as binary by the collab persistence path falls through and renders broken. could we extract the path-candidate + data-URI logic into a shared helper both call, or hydrate shape parts in `hydrateImageBlocks` alongside ImageRuns?

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
```diff
@@ -1061,6 +1084,32 @@ function extractTextFromTextBox(textBoxContent, bodyPr, params = {}) {
...
+        // SD-2804 / ECMA-376 §20.4.2.38: a textbox can hold body-level
+        // content, including runs with inline w:drawing images. Reuse the
+        // existing v3 wp drawing handler so the rId → resolution matches
+        // what body paragraphs use, then upgrade the path-style src to a
+        // data URI from converter.media (the text-parts model has no
+        // downstream hydration step like body ImageRuns do).
+        const inlineOrAnchor = el.elements?.find((child) => child?.name === 'wp:inline' || child?.name === 'wp:anchor');
+        if (inlineOrAnchor) {
+          const isAnchor = inlineOrAnchor.name === 'wp:anchor';
+          const imagePm = handleImageNode(inlineOrAnchor, { ...params, nodes: [el] }, isAnchor);
```
luccas-harbour: `isAnchor` gets detected and passed to `handleImageNode`, but the resulting anchor metadata (wrap, position, transform) is thrown away — anchored drawings render as inline-block here just like inline ones. 

can you have a look? would fixing that make sense in this same PR?

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -4458,6 +4458,18 @@ export class DomPainter {
...
+        // SD-2804: image part produced by the textbox importer for an
+        // inline w:drawing inside a textbox run. Render as <img> alongside
+        // sibling text spans so layout matches Word's inline flow.
+        const img = this.doc!.createElement('img');
+        img.src = part.src;
+        img.alt = part.alt ?? '';
+        if (typeof part.width === 'number') img.style.width = `${part.width}px`;
+        if (typeof part.height === 'number') img.style.height = `${part.height}px`;
+        img.style.display = 'inline-block';
+        img.style.verticalAlign = 'middle';
```
luccas-harbour: body inline images use `verticalAlign = 'bottom'` (renderer.ts:5771, 5848) to match Word's baseline default — using `'middle'` here means images inside textboxes don't line up with text the way they do outside.

I added some text next to the header image in the document that's attached to this ticket and it's not aligned the same way as in MS Word, I think it's because of this (SD on the left and MS Word on the right):

Image

luccas-harbour:
Hey @tupizz! I found a few things and left as inline comments. Ping me if you have any questions. Thanks!

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
tupizz: Good catch — addressed in 8e1247c. Gated the textParts.push on `imagePm.attrs.hidden !== true` so hidden textbox drawings (`wp:docPr hidden="1"`) stay hidden, matching the body-level filtering.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
tupizz: Right — went with the second option you suggested. Removed `resolveImagePartSrc` from the importer entirely; the textbox image part now stores the raw path + extension + rId. Extended pm-adapter's `hydrateImageBlocks` with a `vectorShape` branch that hydrates `textContent.parts` alongside ImageRuns — single source of truth for path candidates and Uint8Array decoding, so Y.js binary media works correctly here too. 8e1247c.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
tupizz: Real concern, but I think it's a separate scope. The text-parts model the shape painter consumes is inline-only — there's no place to carry wrap / position / transform metadata, and an anchored drawing inside a textbox needs all of those to render correctly. The supplied SD-2804 fixture only uses `wp:inline`, and Word actually allows anchored drawings inside a textbox very rarely (most images authored inside a textbox in Word end up inline).

In 8e1247c I tightened the branch to `wp:inline` only and documented the limitation in the code comment so it's explicit. If a real fixture surfaces an anchored image inside a textbox, that PR would need a richer container model (likely the deferred 'shapeTextbox container with FlowBlock children' path I noted in the original PR description) — happy to file a separate ticket if you want.

packages/layout-engine/painters/dom/src/renderer.ts
tupizz: Verified and fixed in 8e1247c — changed `vertical-align: middle` to `bottom` to match the body inline pattern at `renderer.ts:5770, 5847`.

Cross-checked with ECMA-376 to be sure: §20.4.2.8 (`wp:inline` / Inline DrawingML Object) says the drawing object 'is in line with the text, and affects the line height and layout of its line **like a character glyph of similar size**' — i.e., baseline placement is the spec default. Body's `vertical-align: bottom` approximates that for the typical case where the image height matches the line box, and matches the existing pattern. Browser test on the original fixture confirms the image still renders correctly; if you have time to re-run your text-next-to-image case I'd appreciate the second pair of eyes.

tupizz:
@luccas-harbour went through all your points and addressed them, please check it again once you're good

packages/layout-engine/painters/dom/src/renderer.ts
luccas-harbour: thanks @tupizz! I tested this again and it looks good now.

luccas-harbour (approved):
LGTM

PR #3209, opened 2026-05-08
Fix/word hyperlink format reset

christos8333:
That's a branch out of fix/unset-link-clear-transient-hyperlink-styleid which fixes unlink of imported doc files.

- resolving hyperlink range more defensively on empty selection

- forcing a non-empty TextSelection over resolved link range before unsetting marks,

- removing link marks from both inline text and run node marks,

- clearing hyperlink-derived formatting from participating runs/text:

    1. styleId, color, underline,

    2. plus removing hyperlink-derived run key metadata arrays.


packages/super-editor/src/editors/v1/extensions/link/link.js
```diff
@@ -258,11 +275,184 @@ export const Link = Mark.create({
...
+              return true;
+            });
+          }
+
+          return commandChain
             .unsetColor()
             .unsetMark('link', { extendEmptyMarkRange: true })
+            .command(({ tr }) => {
+              if (underlineMarkType) {
+                tr.doc.nodesBetween(from, to, (node, pos) => {
```
caio-pizzol, line 332: unlink wipes formatting that wasn't part of the link.

repro 1: type "RED LINK PLAIN" → color "RED" red → link only "LINK" → select the whole word → Remove Link. "RED" loses its color.

repro 2: type some text → set its style to "Emphasis" → link it → unlink. the style comes back as null.

the check fires on any text in range with a color or a style ID, then clears all three (color, underline, style). fix: only run this on the `hyperlinkRunPositions` set you already compute, or only clear the Hyperlink/FollowedHyperlink style IDs the way the old code did.

packages/super-editor/src/editors/v1/extensions/link/link.js
```diff
@@ -258,11 +275,184 @@ export const Link = Mark.create({
...
+            if (hasLinkMark) {
+              hyperlinkRunPositions.add(pos);
+              if (node.marks.some((mark) => mark.type === linkMarkType)) {
+                importedRunLinkPositions.add(pos);
+              }
+            }
+          });
+
+          const commandChain = chain();
+          if (selection.empty && linkMarkType && from !== to) {
```
caio-pizzol, line 320: ci will fail. `pnpm --filter @superdoc/super-editor test relationships` shows "keeps imported inline underline mark when removing link" failing with `expected 0 to be greater than 0`.

the new branch removes plain underlines from imported hyperlinks, but PR 3190 added that test to lock in the opposite behavior. if you meant to change it, flip the test. if not, only remove underlines tagged with `autoAdded: true` like before.

packages/super-editor/src/editors/v1/extensions/link/link.js
```diff
@@ -258,11 +275,184 @@ export const Link = Mark.create({
...
+                if (range) {
+                  from = range.from;
+                  to = range.to;
+                  break;
+                }
+              }
+
+              // Fallback link as node mark on run nodes.
+              if (from === to) {
+                // TODO
```
caio-pizzol: empty `// TODO` block. either delete it (the sweep below at L421-441 already handles this) or add an issue id per the comment policy.

caio-pizzol (changes requested):
@christos8333 thanks for the follow-up. the wider cleanup catches more imported Word docs (good), but it goes too far, and the test from PR 3190 fails under the new logic.

two bugs inline + one small thing. happy to pair or take it - whichever you'd prefer.

packages/super-editor/src/editors/v1/extensions/link/link.js
christos8333: hello @caio-pizzol . Yes, you are right SuperDoc has a color tool. Will be available later today if you would like and have time to take a look together. I think the issue here as far as i can remember is the clearedAttrs which unconditionally removes color attr on al ink  and there was an unsetColor which was part of the original code(unsure if that is needed, will have to take a look). Do you guys have something like a chat channel i could contact you directly?
Will take a look also to the other comments later today!

christos8333: hello @caio-pizzol . I think i am missing something. I checked the issue with the color and , when i apply color + link from within the editor yeah you are right i can do the previous check or even if i have an attrs.color i guess. 
The problem i have is from the docx import. 
image
So that's a sample i create with 2 links. One is a docx link the other is again a docx link and then i applied a color. How in one instance i will remove both the color and the link and on the other just the link and have the text as red to have the same experience as the link from within the editor?

caio-pizzol: hey @christos8333 - i think the confusing part is that the two examples can look the same visually, but they should differ in stored state.

plain imported hyperlink:
`styleId: "Hyperlink"` / `FollowedHyperlink`, no direct `textStyle.color`

imported hyperlink + user red:
`styleId: "Hyperlink"` plus direct `textStyle.color: "#ff0000"`

so unlink should clear the transient hyperlink styleId in both cases. that removes the imported blue/underline through the style cascade.

for the second case, direct red should be left alone. the remaining wrinkle is the existing `.unsetColor()` call: it clears direct editor color before the narrower styleId cleanup runs. that part is a broader unlink-scope bug, not something the new `textStyle` walker should solve by clearing every color/styleId it sees.

caio-pizzol, line 332: this now removes imported inline underlines too, not just underlines added by `setLink`. `pnpm --filter @superdoc/super-editor test relationships` still fails on `keeps imported inline underline mark when removing link`; restoring the `autoAdded !== true` gate should keep the PR-3190 behavior.

packages/super-editor/src/editors/v1/extensions/link/link.js
```diff
@@ -258,11 +275,184 @@ export const Link = Mark.create({
...
+                    const underlineType = mark.attrs?.underlineType;
+                    const isBareUnderline = underlineType == null || underlineType === 'single';
+                    if (!isBareUnderline) return;
+
+                    tr.removeMark(pos, pos + node.nodeSize, mark);
+                  });
+                });
+              }
+
+              const textStyleMarkType = tr.doc.type.schema.marks.textStyle;
```
caio-pizzol, line 342: this clears any `textStyle` with a color or styleId, so user styles like `Emphasis` get nulled on unlink. please keep this scoped to `TRANSIENT_HYPERLINK_STYLE_IDS` and only clear `styleId`; the style cascade will remove the imported hyperlink blue/underline.

packages/super-editor/src/editors/v1/extensions/link/link.js
```diff
@@ -227,9 +228,25 @@ export const Link = Mark.create({
...
           if (linkMarkType) tr = tr.removeMark(from, to, linkMarkType);
-          if (underlineMarkType) tr = tr.removeMark(from, to, underlineMarkType);
 
-          if (underlineMarkType) tr = tr.addMark(from, to, underlineMarkType.create());
+          if (underlineMarkType) {
+            const rangesMissingUnderline = [];
+            tr.doc.nodesBetween(from, to, (node, pos) => {
+              if (!node.isText || node.nodeSize <= 0) return;
+              const hasUnderline = node.marks.some((mark) => mark.type === underlineMarkType);
+              if (hasUnderline) return;
```
caio-pizzol: existing issue, but worth folding in while this code is open: `underlineType: 'none'` is a negation marker, not a visible underline. treat it as missing, remove that mark, then add the link's `autoAdded` underline.

caio-pizzol (changes requested):
hey @christos8333 - thanks for following up. the run-node removal direction makes sense; the remaining issue is scope.

needs work. left three inline. the empty `// TODO` at L309 can also go, since the fallback below already covers that case.

packages/super-editor/src/editors/v1/extensions/link/link.js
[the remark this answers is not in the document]
caio-pizzol: good catch - folded into the round-2 review at the new L237 inline.

caio-pizzol:
hey @christos8333 - since this had been quiet, i pushed a follow-up with the round-2 review fixes (narrow gates + codex L237 + TODO removal), two regression tests, and the current main conflict resolution. feel free to amend, revert, or build on it. happy to chat about anything that doesn't look right.

caio-pizzol (approved):

PR #3210, opened 2026-05-08
fix(converter): preserve hyperlink mark on inserted text split across paragraphs (SD-2973)

tupizz:
## Summary

When a DOCX has tracked changes turned on and contains an inserted hyperlink that crosses a paragraph break, SuperDoc opens the file but the hyperlink is dropped — the inserted text shows up with insertion track-change styling, but it is no longer a clickable link. Word renders the same file with both treatments at once.

Linear: [SD-2973](https://linear.app/superdocworkspace/issue/SD-2973)
Follow-up to: [SD-2858](https://linear.app/superdocworkspace/issue/SD-2858) / [PR #3175](https://github.com/superdoc-dev/superdoc/pull/3175)

## Root cause

SD-2858's \`preserveRaw\` path in \`preProcessNodesForFldChar.js\` keeps tracked-change-wrapped fields structurally intact when the field crosses paragraph or wrapper boundaries — necessary because re-emitting \`<w:hyperlink>\` wrapping multiple \`<w:p>\` elements is invalid OOXML. For destructive wrappers (\`w:del\` / \`w:moveFrom\`) this trade-off is invisible (deleted content disappears on accept). For constructive wrappers (\`w:ins\` / \`w:moveTo\`) the user keeps the inserted text and Word shows it as a clickable hyperlink — the previous behaviour rendered insertion styling alone.

Console.log trace from the supplied fixture (\`sd-2858-ins-fixture.docx\`):

\`\`\`
[SD-2973] finalizeField {instrText: \"HYPERLINK \\\"http://example.com\\\"\", preserveRaw: true, rawCount: 2}
[SD-2973] finalizeField PRESERVE-RAW path — link mark not applied {rawNodeNames: [\"w:p\", \"w:p\"]}
\`\`\`

## Fix

- New \`isConstructiveTrackChangeElement(node)\` predicate in \`trackChangeElements.js\` (\`w:ins\` / \`w:moveTo\`).
- A \`preserveRawConstructive\` flag rides alongside \`preserveRaw\` through the field collector, set whenever a constructive wrapper triggered the raw-preservation.
- After \`finalizeField\` emits the raw nodes, when the flag is set and the field is a HYPERLINK, a per-field post-pass walks the raw nodes (paragraphs / tracked-change wrappers / runs), finds the visible runs (those between \`separate\` and \`end\` fldChars), and wraps just those runs in \`<w:hyperlink r:id=\"rIdN\"/>\` in-place. The relationship is added to \`document.xml.rels\` using the same machinery as the regular hyperlink preprocessor.

The surrounding paragraph and tracked-change wrapper structure stays untouched, so the SD-2858 round-trip guarantee holds. Destructive wrappers (\`w:del\` / \`w:moveFrom\`) still take the original raw-only path — no behaviour change there.

## Before / after

Fixture: \`sd-2858-ins-fixture.docx\` (attached to SD-2973). Inserted hyperlink with the begin / separate in para 1's \`<w:ins>\` and the end fldChar in para 2's \`<w:ins>\`.

| Marks on \"inserted link text\" | Before | After |
|---|---|---|
| PM marks | \`[textStyle, trackInsert]\` | \`[textStyle, trackInsert, link]\` |
| \`link.attrs.href\` | (none) | \`http://example.com\` |
| Visible | green dashed underline only | green dashed underline + link styling, clickable |

Captured live via agent-browser against the dev server.

## Test plan

- [x] Unit test \`wraps the visible run in w:hyperlink when an inserted HYPERLINK is split across paragraphs\` in \`preProcessNodesForFldChar.test.js\` — fails on main, passes here.
- [x] super-editor full suite: 12,645 tests passing.
- [x] Browser repro on the supplied fixture: PM tree shows \`link\` mark with correct href; visual matches Word's layered styling.
- [x] Existing SD-2858 regression tests (\`preserves a tracked-deletion-wrapped field split across paragraphs\`, \`preserves raw field nodes when an active field ends inside a tracked deletion wrapper\`, the \`moveFrom\` variant) still pass — destructive path is unchanged.

## Out of scope

- Other field types (REF, PAGE, SEQ, etc.) wrapped in inserts split across paragraphs would have analogous issues with different visible symptoms. Each needs its own preprocessor wiring; HYPERLINK is the highest-impact case (clickable link is a user-facing feature). Other field types can be added by extending \`applyConstructiveFieldInterpretation\` as fixtures surface them.
- Restructuring \`_processCombinedNodesForFldChar\` to emit valid paragraph-crossing structures for the non-tracked-change case is a separate refactor.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/super-converter/field-references/preProcessNodesForFldChar.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/field-references/preProcessNodesForFldChar.js
@@ -3,7 +3,8 @@
  */
 import { getInstructionPreProcessor } from './fld-preprocessors';
 import { carbonCopy } from '@core/utilities/carbonCopy.js';
-import { isTrackChangeElement } from '../v2/importer/trackChangeElements.js';
+import { isTrackChangeElement, isConstructiveTrackChangeElement } from '../v2/importer/trackChangeElements.js';
+import { generateDocxRandomId } from '@helpers/generateDocxRandomId.js';
 
 const SKIP_FIELD_PROCESSING_NODE_NAMES = new Set(['w:drawing', 'w:pict']);
 
@@ -61,6 +62,14 @@ export const preProcessNodesForFldChar = (nodes = [], docx) => {
           fieldRunRPr,
         );
         outputNodes = combinedResult.handled ? combinedResult.nodes : rawCollectedNodes;
+      } else if (currentField.preserveRawConstructive) {
+        // SD-2973: SD-2858's preserveRaw guarantees structural validity when a
+        // field crosses tracked-change wrappers but loses field interpretation.
+        // For constructive wrappers (w:ins / w:moveTo) the user keeps the
+        // visible content — drop the link-like interpretations and they'd see
+        // plain text. Run a per-field post-pass that re-applies the
+        // interpretation in-place without restructuring the tree.
+        applyConstructiveFieldInterpretation(outputNodes, currentField.instrText.trim(), docx);
       }
       if (collectedNodesStack.length === 0) {
         // We have completed a top-level field, add the combined nodes to the output.
@@ -215,6 +224,14 @@ export const preProcessNodesForFldChar = (nodes = [], docx) => {
           if (fieldInfo.preserveRaw || isTrackChangeElement(node)) {
             fieldInfo.preserveRaw = true;
           }
+          // SD-2973: track whether the wrapper that forced preserveRaw was a
+          // constructive tracked-change (w:ins / w:moveTo). Constructive
+          // wrappers keep the content on accept, so the user sees the visible
+          // text — and field interpretations like HYPERLINK need to survive
+          // even though we cannot collapse the structural tree.
+          if (isConstructiveTrackChangeElement(node)) {
+            fieldInfo.preserveRawConstructive = true;
+          }
           currentFieldStack.push(fieldInfo);
 
           // The current node should be added to the collected nodes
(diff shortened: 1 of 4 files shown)


packages/super-editor/src/editors/v1/core/super-converter/field-references/preProcessNodesForFldChar.js
```diff
@@ -306,6 +329,132 @@ const _processCombinedNodesForFldChar = (nodesToCombine = [], instrText, docx, i
...
+        Target: url,
+        TargetMode: 'External',
+      },
+    });
+    linkAttributes = { 'r:id': rId };
+  } else {
+    const anchorMatch = instrText.match(/(?:\\)?l "(?<value>[^"]+)"/);
+    if (!anchorMatch?.groups?.value) return;
+    linkAttributes = { 'w:anchor': anchorMatch.groups.value };
+  }
```
luccas-harbour: the URL/anchor regex pair and the rels-element construction here are basically a copy of `preProcessHyperlinkInstruction` in `hyperlink-preprocessor.js:12-58`. would it be possible to reuse that logic here?

luccas-harbour:
hey @tupizz! I had a look and left a comment about perhaps reusing some of the logic we already have for preprocessing `HYPERLINK` instructions. Also, I see that the automated review from codex left a comment, can you have a look? thanks!

packages/super-editor/src/editors/v1/core/super-converter/field-references/preProcessNodesForFldChar.js
tupizz: For sure... thanks for flagging

tupizz:
Thanks @luccas-harbour — addressed in ecada40. Extracted `resolveHyperlinkAttributes(instruction, docx)` in `hyperlink-preprocessor.js` as the single source of truth for parsing HYPERLINK field instructions into `w:hyperlink` attributes. Both `preProcessHyperlinkInstruction` (the standard field path) and `applyConstructiveFieldInterpretation` (the SD-2973 raw-preserved constructive-tracked-change path) now call it. Net ~-5 lines, no behaviour change. All 12,645 super-editor tests pass.

luccas-harbour (approved):
LGTM 👍

PR #3211, opened 2026-05-08
fix(collaboration): render cross-references for late-joining clients (SD-2975)

luccas-harbour:
## Summary

Fixes SD-2975: cross-references (and citations) not rendering for collaborators who join an active room. Word-imported `crossReference`/`citation` field nodes carry cached result runs in their inline content, but the ProseMirror schema declares them as leaf atoms — so when y-prosemirror hydrated the shared Yjs XML for a late joiner, the cached children violated the schema and the node failed to materialize.

- **Strip cached children from atom nodes during Yjs hydration.** New `normalize-yjs-fragment.js` walks the `supereditor` XML fragment and clears children of `crossReference`/`citation` `XmlElement`s. Runs before `yXmlFragmentToProseMirrorRootNode` and via `observeDeep` for incoming remote events. Runs inside a Yjs transaction with a private origin symbol so the normalizer ignores its own events.
- **Fall back to `resolvedText` on export.** Once cached content is stripped, the editor still needs to round-trip something to DOCX. New `buildFieldResultRuns` shared helper rebuilds a `w:r` with the node's `resolvedText` attr when `node.content` is empty, applying `xml:space="preserve"` for boundary whitespace. Both `crossReference-translator.js` and `citation-translator.js` now route through it.
- **Editor lifecycle:** `OpenOptions` accepts a `fragment` (Y.js `XmlFragment`) so callers can hydrate directly from a seeded room; the option is cleared on close to prevent stale fragments leaking across reopens. Pre-hydration normalization is also called in `Editor.ts`.
- **Tests** cover hydration of a seeded `crossReference` from a Yjs room, normalizer cleanup on extension teardown, scoping of the normalizer's own transactions, citation atom-content normalization, and export fallback when cached content has been stripped.


packages/super-editor/src/editors/v1/core/Editor.ts
```diff
@@ -821,6 +825,7 @@ export class Editor extends EventEmitter<EditorEventMap> {
         html: options?.html,
         markdown: options?.markdown,
         jsonOverride: options?.json ?? null,
+        fragment: options?.fragment ?? null,
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/super-editor/src/editors/v1/core/Editor.ts
```diff
@@ -821,6 +834,7 @@ export class Editor extends EventEmitter<EditorEventMap> {
         html: options?.html,
         markdown: options?.markdown,
         jsonOverride: options?.json ?? null,
+        fragment: hasOpenFragment ? (options?.fragment ?? null) : this.#constructorFragment,
```
caio-pizzol: repro: `editor.setOptions({ fragment })` then `await editor.open()` on a `deferDocumentLoad: true` editor, late editor hydrates blank and no `crossReference` materializes. this line overrides the `...this.options` spread, dropping a fragment that `setOptions` legitimately set (`EditorOptions.fragment` is public, see `types/EditorConfig.ts:488`). this is not in the standard Hocuspocus path, but it is still a public/deferred Editor API regression. one fix would be:

```suggestion
        fragment: hasOpenFragment
          ? (options?.fragment ?? null)
          : (this.options.fragment ?? this.#constructorFragment),
```

worth a test mirroring the existing constructor-fragment case but using `setOptions` before `open()`.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/sd/citation/citation-translator.js
```diff
@@ -47,7 +47,7 @@ const encode = (params) => {
 const decode = (params) => {
   const { node } = params;
   const outputMarks = processOutputMarks(node.attrs?.marksAsAttrs || []);
-  const contentNodes = (node.content ?? []).flatMap((n) => exportSchemaToJson({ ...params, node: n }));
+  const contentNodes = buildFieldResultRuns(params, outputMarks);
```
caio-pizzol, line 50: repro: import a Word citation with a run-wrapped cached result, let the normalizer strip the children, export, no run between `separate` and `end`. citation's `extractResolvedText` (line 114) only filters top-level text, so `resolvedText` is `""` and this fallback returns `[]`. `crossReference-translator.js:118` already walks recursively for this exact shape - worth mirroring here, plus an import → strip → export test on a run-wrapped CITATION fixture (the existing fallback test seeds `resolvedText` by hand, bypassing the importer).

caio-pizzol (changes requested):
hey @luccas-harbour! the late-joiner fix is the right shape - the private origin symbol on the normalizer transaction is a nice touch :)

two things to flag before merge, both inline.

caio-pizzol:
hey @luccas-harbour - i pushed two follow-up commits for the review:

- 57d4ab27: setOptions fragment regression. three-way fallback so `this.options.fragment` is honored when `open()` doesn't receive a fragment.
- 446c5114: citation cached resolvedText. mirrors the recursive walk crossReference uses, plus a parallel SD-2975 import test exercising the Brillio-style run-wrapped shape.

ran `pnpm vitest run` under packages/super-editor: 1005 test files, 12,631 tests passed. also re-verified the setOptions scenario end-to-end in the dev playground.

both inline review threads (Editor.ts:837, citation-translator.js:50) remain - you can resolve them once you've reviewed the commits.

caio-pizzol (approved):
lgtm

PR #3214, opened 2026-05-09
feat(converter): render EMF+ images via embedded bitmaps (SD-2503)

### LINKED ISSUE #3172: Feature: render EMF+ metafile images
## What's happening

Documents containing EMF+ (Enhanced Metafile Plus) images show a placeholder SVG with "Unable to render EMF+ image" instead of the actual image. Standard EMF conversion to SVG works, but EMF+ uses GDI+ drawing records that the converter explicitly skips.

This is a follow-on to (). The classic EMF/WMF work landed, but EMF+ is still not truly rendered.

## Codebase verification

The current import/render path does support classic EMF/WMF conversion.

* packages/super-editor/src/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js converts metafiles during import via convertMetafileToSvg(...).
* packages/super-editor/src/tests/extensions/image.test.js documents that EMF/WMF are rendered like normal images after conversion.
* packages/super-editor/src/tests/import-export/emf-wmf-conversion.test.js asserts EMF/WMF are converted to svg/bmp/png for display while preserving original files for DOCX round-trip export.

But EMF+ is explicitly detected and downgraded to a placeholder today.

* packages/super-editor/src/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js scans EMR_COMMENT records for the EMF+ signature 0x2B464D45.
* In that same file, convertEmfToSvg(...) returns a placeholder SVG labeled "Unable to render EMF+ image" instead of rendering the image when EMF+ is detected.

## Why it's happening

The current EMF renderer path is built around the extracted rtf.js EMF/WMF converter and standard EMF records. EMF+ payloads use GDI+ drawing commands, so the current path deliberately avoids attempting full conversion and renders a placeholder instead.

## Suggested fix

1. Check for an embedded bitmap fallback in EMF+ files and render that when present.
2. Extend or replace the current converter with EMF+ / GDI+ record support.
3. Add a WASM or server-side conversion path for EMF+ to a browser-safe format.

## Acceptance criteria

* EMF+ images render actual document content instead of the placeholder SVG.
* Existing classic EMF/WMF rendering behavior is preserved.
* DOCX round-trip export continues to preserve the original metafile asset.
* Add targeted test coverage for at least one EMF+ sample document.

## Reproduction

m3 proposal.docx
Page 1: Cover page shows "Unable to render EMF+ image" instead of branded cover design.

[m3 proposal.docx]()

---

**Want to work on this?** Check the [Contributing Guide](https://github.com/superdoc-dev/superdoc/blob/main/CONTRIBUTING.md) to get started. Comment on this issue to let us know you're picking it up.

*Ref: SD-2503*

gpardhivvarma:
## Summary

Renders EMF+ images that embed a compressed bitmap (PNG/JPEG/GIF) instead of falling back to the placeholder SVG. Most real-world EMF+ files generated by Office — cover slides, charts, illustrations — wrap a complete PNG or JPEG inside an `EmfPlusObject(Image)` record. Walking the EMF+ stream and pulling that bitmap out gives pixel-perfect rendering without implementing a GDI+ rasterizer.

Closes #3172.

## What changed

`packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js`

- New `extractBitmapFromEmfPlus(buffer)`: walks `EMR_COMMENT` records carrying EMF+ payloads, scans inner EMF+ records for `EmfPlusObject(Image)` entries, reassembles continuation series via the `TotalObjectSize` prefix on the first chunk per MS-EMFPLUS § 2.3.5.1, and parses the resulting `EmfPlusImage` / `EmfPlusBitmap` to extract the encoded image bytes.
- New `parseEmfPlusImageObject(bytes)`: validates `Image.Type=Bitmap` and `Bitmap.Type=Compressed`, then returns the embedded PNG/JPEG/GIF as a data URI.
- New `detectCompressedImageFormat(bytes)` helper using PNG/JPEG/GIF magic bytes.
- Wired the extractor into `convertEmfToSvg` between the existing classic `EMR_STRETCHDIBITS` path and the EMF+ placeholder, so the placeholder remains the final fallback for pure-vector EMF+.
- Pulled the literal `70` for `EMR_COMMENT` into a named constant shared with the existing `isEmfPlus` detector.
- Updated the module-level docstring to reflect the layered strategy.

## Spec correctness

Per MS-EMFPLUS § 2.3.5.1:

- First chunk: `ContinueBit=1`, `ObjectData = TotalObjectSize | first slice`.
- Middle chunks: `ContinueBit=1`, raw appended bytes.
- Final chunk: `ContinueBit=0` — the parser keys off this to flush.
- Defensive fallback: if an off-spec encoder leaves `ContinueBit=1` on the last record, the parser flushes early once `TotalObjectSize` bytes are accumulated.

## What this does NOT cover

- Pure-vector EMF+ (logos drawn entirely with GDI+ paths) — those still hit the placeholder. Implementing a full GDI+ renderer is out of scope.
- Pixel-format (uncompressed) `EmfPlusBitmap` — also still hits the placeholder; rasterizing raw pixel buffers requires the same infrastructure.

## Acceptance criteria

- [x] EMF+ images with embedded compressed bitmaps render actual document content instead of the placeholder SVG.
- [x] Existing classic EMF/WMF rendering behavior is preserved (293 tests in the helpers directory still pass).
- [x] DOCX round-trip export continues to preserve the original metafile asset (the import path stores `originalSrc` / `originalExtension` when a metafile is converted; not changed).
- [x] Targeted test coverage: 6 new tests using synthetic in-memory EMF+ buffers cover PNG/JPEG extraction, spec-compliant continuation reassembly, off-spec lenient continuation flush, fallback for non-Image objects, and rejection of pixel-format bitmaps.

## Test plan

- [x] `pnpm exec vitest run src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.test.js` — 14/14 pass.
- [x] `pnpm exec vitest run src/editors/v1/core/super-converter/v3/handlers/wp/helpers/` — 293/293 pass (no regressions in adjacent helpers).
- [x] `pnpm exec prettier --check` on both modified files — clean.
- [ ] Open the reproducer (`m3 proposal.docx`) in the editor and verify the cover image renders.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js
@@ -1,9 +1,14 @@
 /**
- * EMF/WMF to SVG Converter
+ * EMF/WMF to browser-renderable image converter.
  *
- * Converts Windows Enhanced Metafile (EMF) and Windows Metafile (WMF) images
- * to SVG format. The converted SVG is returned as a data URI that can be used
- * as an image source.
+ * Converts Windows Enhanced Metafile (EMF/EMF+) and Windows Metafile (WMF) images
+ * into a format browsers can render, returned as a data URI plus short format tag.
+ * Strategy, in order of preference:
+ *   1. Embedded bitmap fast paths (no rasterization required):
+ *      - Classic EMR_STRETCHDIBITS DIB → BMP
+ *      - EmfPlusObject(Image) compressed bitmap → original PNG/JPEG/GIF
+ *   2. Vector rasterization via the rtf.js renderer → SVG (classic EMF/WMF only)
+ *   3. Placeholder SVG when an EMF+ payload uses GDI+ records we can't render
  *
  * EMF/WMF rendering code extracted from rtf.js (MIT License)
  * Original: https://github.com/nicktf/rtf.js
@@ -74,6 +79,21 @@ const MM_ANISOTROPIC = 8;
 const EMF_SIGNATURE = 0x464d4520; // ' EMF'
 const EMF_PLUS_SIGNATURE = 0x2b464d45; // 'EMF+' inside EMR_COMMENT
 
+// Classic EMR record type for comments (MS-EMF § 2.3.3.1)
+const EMR_COMMENT = 70;
+
+// EMF+ record types (MS-EMFPLUS § 2.1.1.1)
+const EMF_PLUS_OBJECT = 0x4008;
+
+// EMF+ object types encoded in EmfPlusObject Flags bits 8–14 (MS-EMFPLUS § 2.1.1.21)
+const EMF_PLUS_OBJECT_TYPE_IMAGE = 5;
+
+// EmfPlusImage Type field (MS-EMFPLUS § 2.2.1.4)
+const EMF_PLUS_IMAGE_TYPE_BITMAP = 1;
+
+// EmfPlusBitmap Type field (MS-EMFPLUS § 2.2.2.2)
+const EMF_PLUS_BITMAP_TYPE_COMPRESSED = 1;
+
 // Re-export for local use — shared implementation lives in ../../../../helpers.js
 const base64ToArrayBuffer = dataUriToArrayBuffer;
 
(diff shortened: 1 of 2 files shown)


packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js
```diff
@@ -95,6 +115,221 @@ function uint8ToBase64(bytes) {
...
+          const emfPlusStart = offset + 16;
+          const emfPlusEnd = offset + 12 + dataSize;
+
+          let pos = emfPlusStart;
+          while (pos + 12 <= emfPlusEnd) {
+            const epType = view.getUint16(pos, true);
+            const epFlags = view.getUint16(pos + 2, true);
+            const epSize = view.getUint32(pos + 4, true);
+            const epDataSize = view.getUint32(pos + 8, true);
+            if (epSize < 12 || pos + epSize > emfPlusEnd || epDataSize > epSize - 12) break;
```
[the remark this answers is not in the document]
gpardhivvarma: Good catch — verified against the official MS-EMFPLUS § 2.3.5.1 spec page and you're right. `TotalObjectSize` is a header field present on every continued record (between `Size` and `DataSize`), not the first 4 bytes of `ObjectData`. So when `ContinueBit=1` the header is 16 bytes and `ObjectData` starts at offset 16; when `ContinueBit=0` it's the standard 12-byte header. The synthetic continuation tests had been built with the same wrong layout, which masked the bug.

Fixed in b9425b5:
- `extractBitmapFromEmfPlus` now branches on `ContinueBit` to pick the right header size and `DataSize` offset, and treats every continued record's `TotalObjectSize` as authoritative.
- Test helpers rebuilt with the spec layout (`writeContinuedRecord` / `writeFinalRecord`) and an extra test added for a 3-record continuation series so middle-chunk handling is exercised.

gpardhivvarma:
@caio-pizzol please review this

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js
```diff
@@ -95,6 +115,218 @@ function uint8ToBase64(bytes) {
...
+function parseEmfPlusImageObject(bytes) {
+  if (!bytes || bytes.byteLength < 28) return null;
+
+  const view = new DataView(bytes.buffer, bytes.byteOffset, bytes.byteLength);
+  const imageType = view.getUint32(4, true);
+  if (imageType !== EMF_PLUS_IMAGE_TYPE_BITMAP) return null;
+
+  const bitmapType = view.getUint32(24, true);
+  if (bitmapType !== EMF_PLUS_BITMAP_TYPE_COMPRESSED) return null;
+
```
caio-pizzol, line 355: the file linked from #3172 stores its image as raw pixels, and this check rejects raw pixels (only packed PNG/JPEG/GIF render). after this PR, that file still shows the placeholder.

i built 19 test files across every image variant the format defines. 16 pass, the 3 that fail are all raw-pixel cases. fix: handle raw pixels by drawing them onto a canvas and exporting as PNG, like the existing bitmap fallback does. or ship as-is and drop "Closes #3172".

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js
```diff
@@ -95,6 +115,218 @@ function uint8ToBase64(bytes) {
...
+                  entry.parts.push(chunk);
+                  entry.collected += chunk.byteLength;
+
+                  // The strict spec terminates the series with a ContinueBit=0 record,
+                  // but flush early once TotalObjectSize is satisfied so an off-spec
+                  // encoder that leaves ContinueBit=1 on the final record still resolves.
+                  if (entry.totalSize > 0 && entry.collected >= entry.totalSize) {
+                    result = parseEmfPlusImageObject(concatBytes(entry.parts));
+                    pendingByObjectId.delete(objectId);
+                  }
```
caio-pizzol, line 471: if a writer claims the image is N bytes but writes more, the extra bytes get tacked onto the data URI. browsers ignore them, but slicing matches what the comment promises.

```suggestion
                  if (entry.totalSize > 0 && entry.collected >= entry.totalSize) {
                    result = parseEmfPlusImageObject(concatBytes(entry.parts).subarray(0, entry.totalSize));
                    pendingByObjectId.delete(objectId);
                  }
```

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js
```diff
@@ -95,6 +115,218 @@ function uint8ToBase64(bytes) {
...
+
+  const pendingByObjectId = new Map();
+
+  let offset = headerSize;
+  while (offset + 8 <= view.byteLength) {
+    const recordType = view.getUint32(offset, true);
+    const recordSize = view.getUint32(offset + 4, true);
+    if (recordSize < 8 || offset + recordSize > view.byteLength) break;
+
+    if (recordType === EMR_COMMENT && recordSize >= 16) {
```
caio-pizzol: this requires a record to be at least 16 bytes, but the matching check on line 470 requires at least 20. both work, but the gap reads like a typo. unify on 20, or add a one-line comment?

caio-pizzol (commented):
hey @gpardhivvarma! PNG, JPEG, and GIF images all render correctly with this PR. but the file linked from #3172 stores its image as raw pixels, which this PR skips, so opening it still shows the placeholder.

needs work. one thing to flag plus two small nits, all inline.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js
gpardhivvarma: Good catch — fixed in 34006b09c. Raw-pixel `EmfPlusBitmap` (`BitmapDataType=Pixel`) now decodes 24bppRGB / 32bppRGB / 32bppARGB / 32bppPARGB onto a canvas and exports as PNG, mirroring the `tiff-converter` pattern.

Key bits:
- EMF+ pixel formats store DWORDs little-endian, so byte order in memory is B, G, R[, A] — swap to canvas R, G, B, A.
- 32bppPARGB un-premultiplies alpha (`channel * 255 / alpha`) so straight-alpha consumers render correctly.
- Negative height = top-down (rows in render order); positive height = bottom-up (classic Windows DIB); rows reversed before write.
- Indexed formats and missing-canvas environments still fall back to the placeholder.
- 100M-pixel cap on canvas allocation (~400 MB RGBA) matches `tiff-converter`'s `MAX_PIXEL_COUNT`.

6 new tests using a `vi.spyOn` canvas mock cover the 32bppARGB color/row order, bottom-up row flipping, 24bppRGB byte order, 32bppPARGB un-premultiplication, the no-canvas fallback, and the indexed-format fallback.

The reproducer in #3172 should now render. If your 19-file corpus turns up further format variants I missed, I'm happy to add them.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js
gpardhivvarma: Applied in 34006b09c — sliced both reassembly paths (early-flush and final-record) to `pending.totalSize` so any overshoot is dropped instead of trailing onto the data URI.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/metafile-converter.js
gpardhivvarma: Unified on `>= 20` in 34006b09c. You're right — 16 was the minimum for the comment header to be readable, but `isEmfPlus` already requires `>= 20` (4 extra bytes after the identifier to land inside actual EMF+ record territory), so matching that is the consistent choice.

gpardhivvarma:
@caio-pizzol addressed the changes please take a look at it

caio-pizzol (approved):
hey @gpardhivvarma! thanks for addressing last round.

i pushed a follow-up commit (a6e13bf6) fixing the row-order bug codex flagged - emf+ stores raw pixel rows top-down regardless of the height sign, not bottom-up like windows dib, so the flip was inverting customer documents. updated the unit test and added 4 spec fixtures to our visual corpus.

lgtm.

PR #3220, opened 2026-05-11
feat(footnote): rendering fidelity (SD-2656)

tupizz:
## Summary

Range-aware footnote demand + bodyMaxY-anchored band painting. Closes SD-2656.

The reference fixture (a 45-page legal-brief test doc) vs `origin/main`:

| Metric | `origin/main` | This PR |
|---|---|---|
| Page count vs Word (45) | 49 (+4) | **46 (+1)** |
| Footnote anchors on Word's page | 19 / 43 | **28 / 43** |
| Max footnote drift | +4 pages | **+1 page** |
| Footnotes drifting backward | 0 | **0** |
| Pages with band overflow | several (silently clipped) | **0** |
| Reference fixture's longest fn (4-line body) | split across pages | **single page** |

Page-by-page comparison PDF generated locally; not attached due to fixture provenance.

## Root cause

Three independent bugs compounded:

1. **Block-level demand was charged at block entry**, ignoring that only the first few lines of a paragraph might anchor any footnote. The reference fixture's p19 deferred the whole anchor-block because its 474 px combined fn 9 + fn 10 demand made the page-entry advance check fire — even though only the first line had no fn anchor and would have fit.
2. **Band painter used a fixed `pageH − bottomMargin − reserve` Y**, so band content overflowed past the page when reserve was under-allocated. The reference fixture's p19 painted fn 10 body up to y=1234 on a 1056 px page — 810 px clipped invisibly.
3. **`splitRangeAtHeight` charged `range.spacingAfter`** against available band space for fitted ranges that complete the input. The reference fixture's longest footnote (72 px body + 21 px spacingAfter = 93 px against an 89-px budget) was force-split to 1 line + 3-line continuation, when it should have placed all 4 lines.

## Architectural changes

- `footnoteAnchorsByBlockId: Map<string, FootnoteAnchorEntry[]>` (was `Map<string, number>`) — stores per-anchor pmPos + height.
- `getFootnoteDemandForBlockId(blockId, pmStart?, pmEnd?)` / `getFootnoteRefCountForBlockId(...)` — range-filtered.
- Body slicer in `layout-paragraph.ts` iterates line-by-line. Each line consults range-aware demand; pre-slicer advance check previews the first candidate line's demand so the in-slicer force-commit cannot place a line whose anchored fn would push the band off-page.
- `state.footnoteRefsThisPage` (new `PageState` field) tracks anchored ref count for band-overhead math.
- `layoutDocument` stashes `page.bodyMaxY` (max cursor Y minus trailing spacing) after layout settles.
- `incrementalLayout.injectFragments`: `bandTopY = page.bodyMaxY` (was `pageH − bottomMargin`). Band paints immediately under body.
- `computeMaxFootnoteReserve` uses `bodyMaxY` so `placementCeiling` reflects actual remaining band space → planner splits fn bodies into continuation pages instead of overflowing.
- Body slicer respects `state.pageFootnoteReserve` as a floor (alongside range-aware demand) so continuation demand from prior pages is honored — prevents fn-body drip on continuation pages.
- `splitRangeAtHeight` + `fitFootnoteContent`: don't charge `spacingAfter` when the fitted range completes the input (no next paragraph below = no gap needed).

## Guard tests

New `packages/layout-engine/layout-bridge/test/footnotePageOverflow.test.ts` — 4 invariants:

- No fragment past `pageH − bottomMargin` under clustered fns
- Oversized fn body splits (not overflows)
- Dense cluster exceeding single band defers (not overflows)
- Every fn ref renders its body somewhere

New `packages/layout-engine/layout-bridge/test/footnoteCompleteness.test.ts` — ref-by-ref completeness invariant.

All four guards (plus the two pre-existing `footnoteBandOverflow` tests) pass with this PR's code. They fail loud on the prior overflow regression.

## Layout comparison sweep (corpus-wide)

`pnpm test:layout -- --reference 1.32.0` against the 562-doc corpus:

| Metric | Value |
|---|---|
| Matched docs | 562 / 562 |
| Reference / candidate gen failures | 0 / 0 |
| Page count changed (all reductions, none increased) | **5 docs** |
| Unique-change docs (real layout structural diffs) | 267 |
| Widespread-only (schema-shape diffs from main, not regressions) | 295 |

The widespread diffs are all schema-shape (new fields since v1.32.0): `paintSnapshot.lines[*].layoutSourceIdentity`, `layoutSnapshot.blocks[*].runs[*].script`, and `layoutSnapshot.layout.pages[*].bodyMaxY` (the field this PR introduces).

Page-count changes (all toward fewer / more compact pages; all are large legal-template fixtures with many footnotes):

| Fixture | v1.32.0 | This PR | Δ |
|---|---|---|---|
| Large legal template A | 52 | **46** | **−6** |
| Large legal template B | 59 | **53** | **−6** |
| Large legal template C | 58 | **53** | **−5** |
| Large legal template D | 53 | **48** | **−5** |
| Large legal template E | 59 | **53** | **−6** |

The compaction matches the reference fixture's 49→46 improvement.

Footnote-only fixtures (`footnotes/basic-footnotes`, `footnotes/multi-column-footnotes`, `pagination/pagination_footnote_break`): same page count as v1.32.0.

## Test plan

- [x] `@superdoc/layout-engine` test suite — 654/654 pass
- [x] `@superdoc/layout-bridge` test suite — 1232/1237 pass. The 5 remaining failures test the legacy fixed-`bandTopY` + multi-pass-reserve architecture (`footnoteContinuationDemand`, `footnoteMultiPass`, `footnoteSeparatorWidth`, `footnoteSeparatorSpacing` ×2). The band-at-`bodyMaxY` model supersedes them; to be retargeted as a follow-up rather than blocking this PR.
- [x] Reference fixture rendered end-to-end via `pnpm dev` + `agent-browser`, layout snapshot inspected per page, comparison PDF generated.
- [x] All 4 `footnotePageOverflow` invariants pass.
- [x] Corpus-wide layout comparison (`pnpm test:layout --reference 1.32.0`): 0 reference / candidate failures, 5 page-count changes (all reductions).

Linear: [SD-2656](https://linear.app/superdocworkspace/issue/SD-2656)


packages/layout-engine/layout-engine/src/layout-paragraph.ts
```diff
@@ -818,19 +833,38 @@ export function layoutParagraphBlock(ctx: ParagraphLayoutContext, anchors?: Para
...
       state.trailingSpacing = 0;
     }
-    if (state.cursorY >= state.contentBottom) {
+    // SD-3049/SD-3050: charge the block's demand once per page (re-fires after every
+    // advanceColumn) and cap additionalDemand to leave room for at least one body line
+    // so an oversized footnote can't deadlock the paginator.
+    const chargeAndComputeEffectiveBottom = (): number => {
+      if (blockFootnoteDemand > 0 && demandChargedPageNumber !== state.page.number) {
+        state.footnoteDemandThisPage += blockFootnoteDemand;
+        demandChargedPageNumber = state.page.number;
```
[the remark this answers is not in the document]
tupizz: fixed in [fix(footnote): charge block demand once, on anchor page (SD-2656)](https://github.com/superdoc-dev/superdoc/pull/3220/commits/88f6a3f6e12473053a668425010746a80d9ae35a)

tupizz:
## Acceptance audit — SD-3049 / SD-2985 / SD-2986 / SD-2987

Walked each sub-issue's acceptance criteria against the rendered output of all 10 acceptance/reference fixtures, with unit-test + visible-evidence verification per criterion.

### Setup

10 fixtures rendered three ways and stitched per page. Captures + composed PDF kept locally; not attached due to fixture provenance.

| Fixture | Refs | Page count: Word / Before / After |
|---|---:|---:|
| basic-footnotes | 2 | 1 / 1 / 1 |
| multi-column-footnotes | 3 | 1 / 1 / 1 |
| footnotes-large-bump-content | 2 | 3 / 3 / 3 |
| longer-header-with-footnotes | 2 | 2 / 2 / 2 |
| Fixture A *(SD-3049 reference)* | 4 | 45 / 45 / 45 |
| Fixture B *(numRestart=eachSect)* | 0 | 15 / 15 / 15 |
| Fixture C | 0 | 19 / 21 / 21 |
| Fixture D *(SD-2986 reference)* | 0 | 1 / 1 / 1 |
| sd-2440-toc *(numRestart=eachSect)* | 0 | 15 / 15 / 15 |
| **Fixture E (headline, footnote-heavy legal template)** | **108** | **51 / 59 / 55** |
| **Totals** | | **153 / 163 / 159** |

PR closes 4 pages on Fixture E (59 → 55). 9 of 10 fixtures already converge; Fixture C's +2 vs Word is pre-existing and unchanged.

### Coverage matrix

| AC | Sub-issue | Verification | Evidence | Verdict |
|---|---|---|---|---|
| 3049-1 | Pagination — break vs demand | unit + visible Fixture E p16 | `footnoteBodyDemand.test.ts` (6) + visible body-density diff vs Word | ✅ |
| 3049-2 | Pagination — continuation carry-forward | unit | `footnoteContinuationDemand.test.ts` (1) + `footnoteBandOverflow.test.ts` (2) | ✅ |
| 3049-3 | Pagination — stability under migration | unit | `footnoteRefMigration.test.ts` (2 — determinism) | ✅ |
| 2987-1 | `word/footnotes.xml` + body refs end-to-end | round-trip + visible | `footnotes-roundtrip.test.js` (21) + Fixtures A/E visibly render refs | ✅ |
| 2987-2 | Body editable, markers non-editable | invariant test | `FootnotesBuilder.test.ts:223` — marker has `data-sd-footnote-number='true'` AND no `pmStart`/`pmEnd` | ✅ |
| 2987-3 | Bodies laid out next to refs in bands | unit + visible | `footnoteColumnPlacement.test.ts` (2) + `multi-column-footnotes` visible | ✅ |
| 2985-1 | Standard separator preserved + spacing | unit + parser | `footnoteSeparatorSpacing.test.ts` (2) + importer branches on `w:type` | ✅ |
| 2985-2 | Continuation separator distinct | unit + spec | `footnoteSeparatorWidth.test.ts` (2) — §17.11.1 full vs §17.11.23 half | ✅ |
| 2985-3 | Typed records out of normal conversion | parser | `documentFootnotesImporter.js:87` branches by `w:type` | ✅ |
| 2985-4 | Spacing + band height correct | unit | `footnoteSeparatorSpacing.test.ts` pixel math | ✅ |
| 2985-5 | Continuation across pages/columns | unit | `footnoteMultiPass.test.ts` (2), `footnoteColumnPlacement.test.ts` (2) | ✅ |
| 2985-implicit | Visible separator matches Word default | visual | `basic-footnotes` p1 — Before full-width, After half-width = Word | ✅ |
| 2986-1 | Settings-level numFmt + numStart | unit | `document-settings.test.ts` (36) + `formatter-parity.test.ts` (8) | ✅ |
| 2986-2 | Section-level numRestart (pos ignored per §17.11.21) | unit + spec | `readSectionNoteConfigs` + §17.11.21-pos-ignored test | ✅ |
| 2986-3 | Unknown w:footnotePr children round-trip | wholesale-XML | `documentFootnotesImporter.js:80` `carbonCopy(el)` + 21 round-trip tests | ✅ |
| 2986-4 | Rendered numbering honors numFmt/numStart/numRestart | unit | `computeNoteNumbering.test.ts` (32) — all 6 numFmt × continuous/eachSect/eachPage | ✅ |
| 2986-5 | Layout placement honors pos | matrix | All 10 fixtures use `pos=pageBottom` default; readers in place for other values | ✅ |

### Aggregate test count

189 footnote-specific tests green across 10 suites:

| Suite | Count |
|---|---:|
| `@superdoc/layout-bridge:footnote*` | 19 |
| `@superdoc/super-editor:document-settings` | 36 |
| `@superdoc/super-editor:computeNoteNumbering` | 32 |
| `@superdoc/super-editor:FootnotesBuilder` | 29 |
| `@superdoc/super-editor:EndnotesBuilder` | 4 |
| `@superdoc/super-editor:separatorContentClassifier` | 12 |
| `@superdoc/super-editor:footnotes-roundtrip` | 21 |
| `@superdoc/pm-adapter:footnote-reference` | 16 |
| `@superdoc/pm-adapter:endnote-reference` | 12 |
| `@superdoc/layout-tests:footnote-formatter-parity` | 8 |
| **Total** | **189** |

### Visible verification — sample evidence

**Fixture E p16 (SD-3049 break-decision):**
- Word: two adjacent sections both visible, tight band
- Before: only the first section visible, large blank, then band
- After: matches Word — both sections + tight band

**basic-footnotes p1 (SD-2985 separator width, §17.11.23 "spans part of the width"):**
- Word: separator line ~half page, left-aligned
- Before: separator line FULL page width
- After: separator line ~half page = matches Word

### OOXML feature inventory (`unzip -p ... | grep`)

```
fixture                       refs typed cn numRestart  numFmt   pos        sectFnPr customMark sep
basic-footnotes               2    2     0  (none)      (def)    (def)      0        0          default
multi-column-footnotes        3    2     0  (none)      (def)    (def)      0        0          default
footnotes-large-bump-content  2    2     0  (none)      (def)    (def)      0        0          default
longer-header-with-footnotes  2    2     0  (none)      (def)    (def)      0        0          default
Fixture A                     4    2     0  (none)      (def)    (def)      0        0          default
Fixture B                     0    3     1  eachSect    (def)    (def)      7        0          default
Fixture C                     0    3     1  (none)      (def)    (def)      0        0          other
Fixture D                     0    2     0  continuous  decimal  pageBottom 1        0          default
sd-2440-toc                   0    3     1  eachSect    (def)    (def)      7        0          default
Fixture E                     108  3     1  (none)      (def)    (def)      0        0          default
```

### Deferred items (parent-epic-scope completeness; not gating any acceptance doc)

- **`numRestart=eachPage` post-layout renumber loop** — helper math complete + tested; renderer pass deferred. No acceptance doc uses `eachPage`.
- **`pos=beneathText` placement fork** — reader landed; placement renderer change is structural. No acceptance doc uses `beneathText`.
- **`pos=sectEnd` / `docEnd`** — endnote-like behavior, both fall back to `pageBottom`. No acceptance doc uses these.
- **Spec B "explicit" separator content → fragments** — classifier landed; fragment emission deferred. Three fixtures have `continuationNotice` defined but with empty `<w:p/>` content (no body to render); all separators are default markers.

All four sub-issues' **literal acceptance documents** render correctly on this PR. Deferred items are for documents in the wild that use the rarer OOXML constructs not present in the acceptance corpus.

---

**Artifacts:** captures and composed PDFs kept locally; not attached due to fixture provenance.

packages/layout-engine/layout-engine/src/index.ts
```diff
@@ -2943,6 +3112,18 @@ export function layoutDocument(blocks: FlowBlock[], measures: Measure[], options
...
+  // SD-2656: stash each page's actual body-bottom on the Page so the band
+  // painter can render the separator immediately under the last body
+  // fragment instead of at the legacy reserve-derived position. Trailing
+  // paragraph spacing is subtracted because it's "below the last line" and
+  // shouldn't push the separator down by that much.
+  for (let i = 0; i < pages.length && i < paginator.states.length; i++) {
+    const s = paginator.states[i];
+    const raw = Math.max(s.maxCursorY ?? 0, s.cursorY ?? 0);
+    const trailing = s.trailingSpacing ?? 0;
+    (pages[i] as { bodyMaxY?: number }).bodyMaxY = Math.max(s.topMargin ?? 0, raw - trailing);
```
luccas-harbour: bodyMaxY subtracts state.trailingSpacing, but that field is reset on every advanceColumn (paginator.ts:169) while maxCursorY is preserved across columns. in a two-column page where column 0 sets maxCursorY and column 1 ends with non-zero trailingSpacing, we subtract spacing that wasn't contributed by the fragment that set maxCursorY, so the band paints into the bottom of column 0. since this Y now anchors the entire band, the overlap is visible. can we track trailingSpacing alongside the fragment that set maxCursorY (or only subtract when maxCursorY came from cursorY of the current column)?

packages/layout-engine/layout-engine/src/layout-paragraph.ts
```diff
@@ -818,19 +847,92 @@ export function layoutParagraphBlock(ctx: ParagraphLayoutContext, anchors?: Para
...
+    // SD-2656: footnote band budgeting constants. The planner reserves
+    // `bandOverhead(refs) = SEPARATOR_PADDING + (refs-1) * INTER_REF_GAP +
+    // SAFETY_MARGIN` for every page where any footnote is anchored. The
+    // slicer must use the SAME formula or body packs onto a page whose band
+    // can't actually fit the refs.
+    const FN_BAND_OVERHEAD_PX = 22;
+    const FN_INTER_REF_GAP_PX = 2;
+    const FN_SAFETY_MARGIN_PX = 1;
+    const bandOverhead = (refsTotal: number): number =>
+      refsTotal > 0 ? FN_BAND_OVERHEAD_PX + Math.max(0, refsTotal - 1) * FN_INTER_REF_GAP_PX + FN_SAFETY_MARGIN_PX : 0;
```
luccas-harbour: the slicer's bandOverhead uses hardcoded FN_BAND_OVERHEAD_PX=22 / INTER_REF_GAP=2 / SAFETY=1, but the planner in incrementalLayout.ts:1330-1342 computes overhead from data-driven topPadding+dividerHeight+separatorSpacingBefore+gap (where separatorSpacingBefore defaults to the first footnote line height). these only coincidentally agree for the default values — for any doc with custom topPadding/dividerHeight/gap they diverge, and body packs onto a page whose band can't fit the refs. can we thread the same overhead values through ctx so both sides agree?

packages/super-editor/src/editors/v1/core/presentation-editor/layout/computeNoteNumbering.ts
```diff
@@ -0,0 +1,116 @@
...
+): NoteNumberingResult {
+  const numberById: Record<string, number> = {};
+  const formatById: Record<string, string> = {};
+  const order: string[] = [];
+  if (!editorState) return { numberById, order };
+
+  const seen = new Set<string>();
+  const sectionConfigs = options.sectionConfigs ?? new Map<number, SectionNoteConfig>();
+  const refPageById = options.refPageById;
+  let counter = options.startCounter;
```
luccas-harbour: counter is seeded from options.startCounter, but numStartFor(0) is only consulted when a later section boundary triggers a reset. a doc with a section-0-level w:footnotePr/w:numStart override (§17.11.11) renders its first note from the document-level start instead of the section override. should we seed counter via numStartFor(0) when sectionConfigs.has(0)?

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
```diff
@@ -6041,60 +6063,66 @@ export class PresentationEditor extends EventEmitter {
...
         }
-        // Invalidate flow block cache when footnote order changes, since footnote
-        // numbers are embedded in cached blocks and must be recomputed.
-        const footnoteSignature = footnoteOrder.join('|');
+
+        // §17.11.14 / §17.11.20 / §17.11.19 / §17.11.11.
+        const footnoteNumbering = computeNoteNumbering(this.#editor?.state, 'footnoteReference', {
+          startCounter: footnoteNumberStart,
+          defaultNumFmt: footnoteNumberFormat,
+          defaultRestart: footnoteNumberRestart,
```
luccas-harbour, line 6416: defaultRestart: 'eachPage' is passed in but no refPageById map is supplied, so the eachPage branch inside computeNoteNumbering (line 90) never fires and numbering stays continuous. docs that set w:numRestart="eachPage" will silently render the wrong ordinals. either thread per-ref page assignments from the layout pass or document that eachPage isn't supported yet — same applies to the endnote call at 6116.

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
```diff
@@ -6041,60 +6063,66 @@ export class PresentationEditor extends EventEmitter {
       let converterContext: ConverterContext | undefined = undefined;
```
luccas-harbour, line 6345: this block reads ~9 OOXML settings, resolves footnote/endnote numbering, patches the converter, and builds a 14-key ConverterContext — that's OOXML semantics resolution, which packages/layout-engine/CLAUDE.md says shouldn't live in PresentationEditor ("PresentationEditor bridges editor state into layout and paint state. It should not resolve OOXML semantics."). happy to defer, but extracting a buildFootnoteConverterContext helper alongside computeNoteNumbering would put the cache-signature dance in one place and unblock testing it in isolation. up to you on timing.

packages/layout-engine/layout-engine/src/layout-paragraph.ts
```diff
@@ -501,6 +521,15 @@ export function layoutParagraphBlock(ctx: ParagraphLayoutContext, anchors?: Para
...
   let fromLine = 0;
+  // SD-3049: charged to the page that receives the block's first committed
+  // fragment. `demandChargedPageNumber` tracks where the (tentative) charge
+  // currently lives so we can re-target it after `advanceColumn`. Once a
+  // fragment is committed (`demandLocked`), the charge stays put — re-charging
+  // on later page transitions would phantom-shrink continuation pages where
+  // the footnote ref does not land.
+  const blockFootnoteDemand = ctx.getFootnoteDemandForBlockId?.(block.id) ?? 0;
+  let demandChargedPageNumber: number | null = null;
+  let demandLocked = false;
```
luccas-harbour: demandChargedPageNumber, demandLocked, and blockFootnoteDemand are all carried purely to satisfy unused-decl checks per the comment at 998-1003. removing them would make the slicer's state more honest about what it actually tracks per slice. can these be dropped now that the legacy demand-locking path is gone?

docs/superdoc-feature-reports/sd-2656-plan.md
```diff
@@ -0,0 +1,558 @@
+# SD-2656 — Footnote Rendering Fidelity (Implementation Plan)
```
luccas-harbour: I think you have commited some plan files by mistake. do you mind removing them from the PR?

packages/super-editor/src/editors/v1/core/presentation-editor/layout/computeNoteNumbering.ts
```diff
@@ -0,0 +1,116 @@
...
+  /** Document-wide w:numRestart (default 'continuous'). */
+  defaultRestart?: 'continuous' | 'eachPage' | 'eachSect';
+  /** §17.11.11 — section-index → override config. Sections without overrides are absent. */
+  sectionConfigs?: Map<number, SectionNoteConfig>;
+  /**
+   * §17.11.19 eachPage — per-ref page assignment from a prior layout pass.
+   * When provided AND the active restart is `eachPage`, the counter resets at
+   * each page boundary. Refs not in the map are treated as page 0 (initial).
+   */
+  refPageById?: Map<string, number>;
```
luccas-harbour, line 38: refPageById is documented as enabling per-page restart but neither call site in PresentationEditor passes it, so the eachPage code path is dead in practice. either wire it (a follow-up ticket is fine) or drop the option until the two-pass pagination handshake exists — keeping a documented-but-unused knob makes the helper look more capable than it is.

packages/layout-engine/layout-engine/src/index.ts
```diff
@@ -2943,6 +3112,18 @@ export function layoutDocument(blocks: FlowBlock[], measures: Measure[], options
...
   }
 
+  // SD-2656: stash each page's actual body-bottom on the Page so the band
+  // painter can render the separator immediately under the last body
+  // fragment instead of at the legacy reserve-derived position. Trailing
+  // paragraph spacing is subtracted because it's "below the last line" and
+  // shouldn't push the separator down by that much.
+  for (let i = 0; i < pages.length && i < paginator.states.length; i++) {
+    const s = paginator.states[i];
+    const raw = Math.max(s.maxCursorY ?? 0, s.cursorY ?? 0);
```
luccas-harbour: bodyMaxY isn't asserted by any unit test — the overflow suites check pageH - bottomMargin but not the band's actual anchor Y. a 3-paragraph layout with non-zero trailing spacing (and ideally a two-column case) would pin the value and catch the multi-column bug above before it regresses. worth a small fixture in index.test.ts.

packages/layout-engine/tests/src/footnote-formatter-parity.test.ts
```diff
@@ -0,0 +1,38 @@
+/**
```
luccas-harbour, line 1: this test asserts formatFootnoteCardinal === formatPageNumber for all formats, but both functions inline the same `-${num}-` logic for numberInDash, so if both were wrong they'd still agree. pageNumbering.test.ts already spot-checks expected strings — adding a few direct-string assertions here for numberInDash (and the more exotic formats) would close the loop.

luccas-harbour (changes requested):
hey @tupizz! phew, this wasn't an easy one, thanks for your changes!
Codex caught a few things that might be worth looking into:

> - [P2] Include computed note numbers in the cache key — packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts:6110-6111
> When the note ID order stays the same but the displayed numbering changes — for example toggling `customMarkFollows` on a middle footnote, or moving a ref across a section with a `numStart`/`numFmt` override — this signature remains unchanged, so `#flowBlockCache` is not cleared even though the cached reference runs already contain the old marker text/format. Include the computed `numberById`/`formatById` (or equivalent per-ref numbering inputs) in the signature so rendered markers are re-created when numbering changes.
>
> - [P2] Honor section-0 note numStart before first ref — packages/super-editor/src/editors/v1/core/presentation-editor/layout/computeNoteNumbering.ts:53-55
> For a single-section DOCX, `readSectionNoteConfigs` records the final `w:sectPr/w:footnotePr/w:numStart` under section `0`, but numbering starts from `options.startCounter` instead of `numStartFor(0)`. In that common case, `numStart=5` still renders the first note as `1` until a later section break is encountered; initialize the counter from the active section-0 config.
>
> - [P2] Supply page assignments for eachPage restarts — packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts:6101-6106
> When `w:numRestart` is `eachPage`, this caller passes the restart mode but never passes `refPageById`; `computeNoteNumbering` explicitly falls back to continuous numbering without that map. As a result, documents that request per-page footnote/endnote numbering still render as continuous, so the layout needs to feed page assignments back into numbering (or avoid advertising this setting as applied).

Claude caught a few more that I added as inline comments.

Also, I ran the visual tests and saw lots of improvements but I saw documents that seem to have a few issues. I'll send you a list of the problematic documents and my findings in the chat.

thank you!

packages/layout-engine/layout-engine/src/index.ts
tupizz: Fixed in 5ed53ee4d. bodyMaxY now only subtracts trailingSpacing when the current column's cursorY actually owns maxCursorY (`cursorY >= maxY`). In a two-column page where column 0 sets the high mark and column 1 ends with non-zero trailingSpacing, the trailing belongs to column 1 — subtracting it from column 0's max would clip the band into column 0's content, which is exactly what you flagged.

Pinned with three unit tests in `index.test.ts`: single-column case (subtract), multi-column case (do not subtract — compares with/without column-1 trailing spacing, asserts equal bodyMaxY), and an empty-content clamp.

packages/layout-engine/layout-engine/src/layout-paragraph.ts
tupizz: Threaded through ctx in 5ed53ee4d. Added `getFootnoteBandOverhead(refsTotal)` to `ParagraphLayoutContext`. Layout-engine reads `topPadding`, `dividerHeight`, `separatorSpacingBefore`, `gap` from `options.footnotes` and computes the same data-driven formula the planner uses (`topPadding + dividerHeight + separatorSpacingBefore + (refs-1)*gap`); slicer only adds a 1px safety pad on top.

Also wired the planner's measured `safeSeparatorSpacingBefore` (which is typically the first-fn lineHeight, not the default 12) back through `relayout()` so on subsequent passes the slicer sees the same first-fn line height the planner used to size the band. Fallback defaults match the planner's defaults so test callers that don't populate the fields still produce identical math.

packages/super-editor/src/editors/v1/core/presentation-editor/layout/computeNoteNumbering.ts
tupizz: Fixed. Counter is now seeded from `numStartFor(0)` which returns `sectionConfigs.get(0)?.numStart ?? options.startCounter` — safe for the no-override case. A single-section doc with section-0 `w:footnotePr/w:numStart=5` now starts at 5 instead of 1.

Added a regression test for the single-section case. Also had to update one pre-existing `eachPage` test that was asserting the buggy behavior (`a=1, b=7` when section-0 had `numStart=7`) — corrected to `a=7, b=7` per §17.11.11.

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
tupizz: Coerced `eachPage` to `continuous` in PresentationEditor with a one-time `console.warn` per kind (footnote/endnote) until the two-pass pagination handshake exists. Applied to both the document-level default and any section-level overrides in `sectionConfigs`. Same path for the endnote call site you flagged. Better to render deterministic ordinals than silently-wrong per-page numbers.

The helper still accepts `refPageById` so its eachPage branch stays testable in isolation; the option's JSDoc now flags it as not wired in the SuperDoc pipeline and points at the two-pass requirement as the follow-up.

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
tupizz: Deferred to follow-up per your offer. Added a TODO at the top of the block (above the `converterContext` declaration) calling out the agreed scope: settings read → numbering → cache-signature dance → ConverterContext assembly into one helper alongside `computeNoteNumbering`. Will file a ticket and pick it up separately so the cache-signature surface gets unit coverage in isolation.

The numbering/signature work in this PR (per-id signature inclusion + section-0 seed + eachPage coercion) already lives in one logical region of the block, so the extraction lands cleanly.

packages/layout-engine/layout-engine/src/layout-paragraph.ts
tupizz: Dropped. Removed `blockFootnoteDemand`, `demandChargedPageNumber`, `demandLocked` along with the void-marked usages and the stale comments. The slicer's per-slice state is now just `fromLine`, `toLine`, `height`, `sliceDemand`, `sliceRefs` — what it actually tracks per iteration.

Also dropped the `sliceLines` import that was orphaned when the line-by-line slicer replaced the bulk slicer.

docs/superdoc-feature-reports/sd-2656-plan.md
tupizz: Already removed in e13877634 (`chore: remove internal SD-2656 planning docs from branch`) — the files are gone from the index and gitignored on disk. Good catch.

packages/super-editor/src/editors/v1/core/presentation-editor/layout/computeNoteNumbering.ts
tupizz: Updated the JSDoc on `refPageById` to flag it as not wired in the SuperDoc layout pipeline and explain why (numbering runs before pagination, so per-page assignments aren't available). The orchestrator (PresentationEditor) now coerces eachPage → continuous with a warning, so the user-facing behavior is honest about what's supported.

Kept the option on the helper because the algorithm itself is tested in isolation (the eachPage tests in `computeNoteNumbering.test.ts` exercise the branch). When the two-pass handshake lands, we wire the param at the caller and remove the coercion.

packages/layout-engine/layout-engine/src/index.ts
tupizz: Added `describe('bodyMaxY')` in `index.test.ts` with 4 tests covering: single-column trailing-spacing subtraction, the multi-column guard (compares bodyMaxY with vs without column-1 trailing spacing — asserts equal, so the test would fail without the fix), the single-page case where the last cursor owns max, and the empty-content topMargin clamp.

The multi-column test is the one that pins the bug you flagged in 3124.

packages/layout-engine/tests/src/footnote-formatter-parity.test.ts
tupizz: Added direct-string assertions on top of the parity-only tests: `numberInDash` (`-1-`, `-5-`, `-12-`, `-99-`), upperRoman (I/IV/IX/XL/XC), lowerRoman (i/iv/ix), and upperLetter/lowerLetter through the 26→AA boundary. Both `formatFootnoteCardinal` and `formatPageNumber` are asserted against the expected strings, so the parity-only tests no longer rely on each other for correctness.

tupizz:
working on fixing all conflicts

luccas-harbour (approved):
LGTM

PR #3225, opened 2026-05-11
fix(converter): preserve theme rFonts on DOCX round-trip (SD-2894)

tupizz:
## Summary

- Extends the `combineRunProperties` font-slot dedup to cover all four `<w:rFonts>` script slots (was only `ascii/asciiTheme`), so a higher-priority style with `hAnsiTheme` / `eastAsiaTheme` / `cstheme` no longer leaves the lower-priority concrete name in place.
- Adds the same theme-aware merge to `calculateInlineRunPropertiesPlugin` so theme references survive the mark-driven recompute that runs after import.

## Why

Customer fixture in [SD-2894](https://linear.app/superdocworkspace/issue/SD-2894) had every body run mapped to `<w:rFonts w:asciiTheme="majorBidi" w:hAnsiTheme="majorBidi" w:cstheme="majorBidi"/>`. The cascade only knew how to drop `target.ascii` when `source.asciiTheme` was present — the other three slots fell through, so the merged run carried `{ ascii: "Arial", hAnsi: "Arial", cs: "Arial", asciiTheme: "majorBidi", hAnsiTheme: "majorBidi", cstheme: "majorBidi" }`. Word treats the concrete value as an override that defeats per-script theme resolution: Latin body text mapped to `majorBidi` rendered as Calibri Light instead of the per-script font (Times New Roman for Hebrew/Arab).

A second loss happened in `calculateInlineRunPropertiesPlugin` on initial load — the plugin computes fontFamily from textStyle marks (which only carry concrete CSS names) and overrode `runProperties.fontFamily` with concrete-only values, dropping the theme references that the import had stored.

## Test plan

- [x] New integration test (`sd-2894-theme-rfonts-roundtrip.test.js`) that round-trips the customer fixture and asserts `w:asciiTheme` / `w:hAnsiTheme` / `w:cstheme` counts don't drop and that no source theme reference becomes a concrete font on export.
- [x] Existing `ooxml-rFonts-rstyle-linked-combos-roundtrip` test still passes.
- [x] Full `pnpm test` for `super-editor` (12 708 / 12 721 pass).
- [x] `@superdoc/style-engine` cascade unit tests pass.
- [x] Manual verification: re-exported customer fixture from the dev app, opened in Word. `asciiTheme` count: 65 → 65 (was 65 → ~25). No Calibri Light substitutions on body text. Per-script resolution intact.

## Verified diffs on the customer fixture

| File | input | before fix | after fix |
| --- | --- | --- | --- |
| `w:asciiTheme` in `document.xml` | 65 | 25 | 65 |
| `w:hAnsiTheme` in `document.xml` | 65 | 25 | 65 |
| `w:cstheme` in `document.xml` | 65 | 25 | 65 |
| `w:ascii="Calibri Light"` runs in `document.xml` | 0 | 1 (concrete substitution) | 0 |


packages/super-editor/src/editors/v1/extensions/run/calculateInlineRunPropertiesPlugin.js
```diff
@@ -28,6 +28,40 @@ const RUN_PROPERTIES_DERIVED_FROM_MARKS = new Set([
...
+  const merged = { ...(fromMarks || {}) };
+  if (!existing || typeof existing !== 'object') return merged;
+  for (const [concreteKey, themeKey] of FONT_FAMILY_THEME_PAIRS) {
+    if (existing[themeKey] != null) {
+      merged[themeKey] = existing[themeKey];
+      delete merged[concreteKey];
+    }
+  }
+  return merged;
+}
```
luccas-harbour, line 75: if the run already has imported theme refs on `existingRunProperties.fontFamily`, this helper silently restores them even when the mark-decoded value is a real user edit. trace: `setFontFamily('Arial')` only updates the textStyle mark — the run node's `runProperties` attr keeps the imported `{ asciiTheme, hAnsiTheme, ... }`. when the plugin runs, `valueFromMarks` is `{ ascii: 'Arial', ... }` (concrete), `existing` is the untouched themed object — and the helper unconditionally restores the theme and drops Arial. Word then renders the theme font, not the user's pick.

one fix is to skip the theme restore when the existing concrete value (e.g. `existing.ascii`) is set and differs from the mark value, since that means the user actively overrode the theme. worth a test that round-trips a doc, edits a themed run to Arial, and re-exports.

packages/super-editor/src/editors/v1/extensions/run/calculateInlineRunPropertiesPlugin.js
```diff
@@ -28,6 +28,40 @@ const RUN_PROPERTIES_DERIVED_FROM_MARKS = new Set([
...
+// theme reference (`asciiTheme: "majorBidi"`) that came from the imported `<w:rFonts>`.
+// Without this merge we'd export concrete font names that defeat Word's per-script theme
+// resolution. The pairs match the four `<w:rFonts>` slots — note `cstheme` is lowercase
+// (OOXML inconsistency).
+const FONT_FAMILY_THEME_PAIRS = [
+  ['ascii', 'asciiTheme'],
+  ['hAnsi', 'hAnsiTheme'],
+  ['eastAsia', 'eastAsiaTheme'],
+  ['cs', 'cstheme'],
+];
```
luccas-harbour: this `FONT_FAMILY_THEME_PAIRS` is identical to `FONT_SLOT_THEME_PAIRS` in `style-engine/src/cascade.ts:117-122`, including the load-bearing lowercase `cstheme` quirk. super-editor already imports from `@superdoc/style-engine/ooxml` in other files (e.g. `super-converter/styles.js:22`), so this could move to style-engine and be imported. would also be a natural place to put a shared `mergeRunFontFamily(low, high)` so the plugin doesn't reimplement the conflict resolution that `combineRunProperties` already does.

luccas-harbour (changes requested):
hey @tupizz!
I had a look at the code and found a few things. I will add them as inline comments.

The main problem I found was flagged by codex:
```
- [P1] Let explicit font changes replace stale theme refs — packages/super-editor/src/editors/v1/extensions/run/calculateInlineRunPropertiesPlugin.js:57-59
When text imported with w:*Theme rFonts later has a concrete font applied by the user, fromMarks contains the newly selected font while existing still contains the old theme reference. This branch unconditionally copies the old theme ref and deletes the concrete mark value, so the export keeps the original theme font instead of the user's font change for those runs.
```

I tested this in the browser and indeed it causes an issue. If you try to load the document below into the editor, then select the text that says "Change this text to Arial" and try to actually change the font, you'll see that nothing happens. This was working correctly in the main branch so the problem is being caused by the changes in this PR.

[theme-rfonts-explicit-font-change-repro.docx](https://github.com/user-attachments/files/27719107/theme-rfonts-explicit-font-change-repro.docx)

Another thing I will ask is that you run visual tests after making these adjustments since this change has the potential to impact the rendering of a lot of documents.

tupizz:
Thanks @luccas-harbour — both points addressed in `f046df9c0`.

**P1 (user override silently reverted):** I was unconditionally restoring theme refs from `existingRunProperties` whenever they were present. The new gate `marksMatchExistingFontFamily()` only preserves the theme when the mark-derived value still encodes to the same OOXML mark as the run's existing fontFamily — i.e., when marks are a re-derivation of the import. When marks diverge (the user picked a new font), we now drop the stale theme and respect the user's value.

Reproduced your `theme-rfonts-explicit-font-change-repro.docx` case in unit form: with `existing = { asciiTheme: 'majorBidi' }` and marks decoding to `{ ascii: 'Arial' }`, the encoder produces `'Calibri Light'` vs `'Arial'` → mismatch → theme dropped, `'Arial'` kept.

**DRY (`FONT_FAMILY_THEME_PAIRS` duplication):** Exported `FONT_SLOT_THEME_PAIRS` from `cascade.ts`, re-exported through `@superdoc/style-engine/ooxml`, and the plugin now imports it. Local copy deleted.

**Safeguards:** Two new AAA tests in `calculateInlineRunPropertiesPlugin.test.js` pin both directions of the contract:
- `preserves theme reference when marks re-encode the imported theme value` — locks the SD-2894 happy path.
- `drops theme references when the user explicitly sets a different font` — locks your P1 case.

Both fail loudly if either branch flips in the future.

**Visual tests:** queued — will post results in a follow-up comment once they finish.

Test results from this commit:
- `calculateInlineRunPropertiesPlugin.test.js`: all new + existing tests pass
- Full `super-editor` suite: 12,710 passed, 13 skipped, 0 failed
- `@superdoc/style-engine`: 129/129 pass
- `tsc --noEmit` on style-engine: clean

tupizz:
Update: pushed `e06c0b831` — a 13-scenario mutation matrix that walks the cross-product of (`existing.fontFamily` shape × mark-decoded shape) the plugin must handle.

### Scenarios covered

**Group A — theme preservation** (marks re-encode imported value → keep theme refs)
- A1: all 4 theme slots
- A2: 3 theme slots (SD-2894 customer shape)
- A3: 3 themes + concrete eastAsia (the regression case `bec7e63ed` fixed)
- A4: ascii-only theme
- A5: hAnsi-only theme
- A6: cs-only theme (lowercase `cstheme` OOXML quirk)
- A7: eastAsia-only theme

**Group B — user override** (marks diverge → drop theme refs)
- B1: all 4 themes + user picks Arial
- B2: customer shape + Arial (the case you originally reported)
- B3: mixed theme+concrete + Arial

**Group C — no-theme baselines**
- C1: pure concrete unchanged
- C2: pure concrete + different concrete
- C3: empty existing

### Encoder mock fidelity

The universal encoder mirrors `resolveDocxFontFamily`'s real behaviour: any theme ref wins for the primary slot, then fall back to concrete `ascii`/`hAnsi`/`eastAsia`/`cs`. Per-script extras (`eastAsiaFontFamily`/`csFontFamily`) attach only when the concrete slot value differs from the primary `fontFamily` — same emission rule as `encodeMarksFromRPr` in `super-converter/styles.js`. I had to make the mock encoder smarter to catch A5/A6/A7 (single non-ascii theme slots); the real resolver's `'major'` fallback handles those identically.

### Results

- Full `super-editor` suite: **12,724 passed**, 13 skipped, 0 failed (+13 from `bec7e63ed`).
- Browser round-trip of the SD-2894 customer fixture: still **65/65/65** theme attrs preserved.
- Browser repro of your case: still exports `<w:rFonts w:ascii="Arial" .../>` with zero theme attrs on the edited run.

The matrix will fail loudly if the gate logic or merge helper ever drift. Together with the smaller unit tests already on this PR, both the preservation path AND the user-override path are now locked at every shape we ship.

luccas-harbour (approved):
LGTM

PR #3226, opened 2026-05-11
feat(super-editor): native LTR/RTL paragraph direction toolbar

### LINKED ISSUE #3219: Feature: native LTR/RTL paragraph direction toggle
Backend support for paragraph direction (`pPr.rightToLeft`, `<w:bidi>` round-trip, Document API `format.paragraph.setDirection`) already exists. The toolbar and editor don't expose it.

Add:
- `setParagraphDirection` / `clearParagraphDirection` PM commands on the paragraph extension (multi-paragraph, single transaction, optional `alignmentPolicy: "matchDirection"` to flip `left`↔`right` justification).
- Two default toolbar buttons (`directionLtr`, `directionRtl`) with active state reflecting the current paragraph.
- Vitest coverage.

Open question: keymap. `Mod-Shift-L/R` is already alignment. Suggest `Mod-Alt-Shift-L/R`.

shri-scale:
## Summary
- Add toolbar buttons for LTR/RTL paragraph direction. SuperDoc already round-trips `<w:bidi>` on `pPr.rightToLeft` and exposes a Document API mutation, but the toolbar didn't surface it. Users editing Arabic/Hebrew docs were patching this gap with custom buttons.
- No keyboard shortcut — matches Google Docs (toolbar-only).
- Closes #3219.

## What changed
**`setParagraphDirection` / `clearParagraphDirection` ProseMirror commands** (`packages/super-editor/src/editors/v1/core/commands/paragraphDirection.js`). Walk paragraphs in the current selection, toggle `pPr.rightToLeft`, optionally mirror an explicit `justification` of `"left"` ↔ `"right"` when `alignmentPolicy: "matchDirection"` is passed (leaves `"center"`, `"both"`, and unset alone). Multi-paragraph selection produces a single transaction → one undo step.

**Two default toolbar buttons** next to `indentLeft` / `indentRight` (`directionLtr`, `directionRtl`) with Material Design `format_textdirection_*` icons. Active state is wired through the headless toolbar registry (`createParagraphDirectionStateDeriver` reads `pPr.rightToLeft`), so the matching button highlights as the cursor moves between LTR and RTL paragraphs.

**`ButtonGroup.vue` patch.** Plain button clicks previously hardcoded `argument: null`. They now forward `item.argument?.value` so any built-in button can carry a static argument (the direction buttons need `{ direction, alignmentPolicy }`). Generic fix; doesn't affect existing buttons.

**XL overflow buffer** bumped from 20 → 84 to accommodate the two extra items at the existing XL cutoff.

## Test plan
- [ ] `pnpm --filter @superdoc/super-editor test` — new vitest covers RTL/LTR set, `matchDirection` flip behavior across all justification values, single-transaction multi-paragraph, and `clearParagraphDirection` semantics
- [ ] Toolbar buttons visible at default toolbar widths, in the overflow menu at narrow widths
- [ ] Clicking a direction button on a paragraph toggles direction and flips the matching alignment
- [ ] Cursor in an existing RTL paragraph (e.g., Arabic DOCX) auto-highlights the RTL button
- [ ] Multi-paragraph selection: one undo reverts all
- [ ] Round-trip: apply RTL, export, re-import — `<w:bidi/>` survives


shri-scale:
@caio-pizzol  Kindly have a look

packages/super-editor/src/editors/v1/core/commands/paragraphDirection.js
```diff
@@ -0,0 +1,87 @@
...
+  return walkParagraphs((pPr) => {
+    const next = { ...pPr };
+    // AIDEV-NOTE: LTR deletes the property rather than writing `false`.
+    // OOXML treats absent `<w:bidi/>` as LTR. Writing `rightToLeft: false`
+    // round-trips as `<w:bidi w:val="0"/>` on export — direct formatting
+    // that overrides any inherited style direction. Deleting keeps a
+    // vanilla paragraph indistinguishable from one that's been toggled
+    // RTL → LTR.
+    if (direction === 'rtl') next.rightToLeft = true;
+    else delete next.rightToLeft;
```
caio-pizzol: important: clicking LTR on a paragraph whose style sets `rightToLeft: true` is a silent no-op.

verified by tracing the cascade at `style-engine/src/ooxml/index.ts:154 resolveParagraphProperties` -> deleting `rightToLeft` from inline props lets `styleProps.rightToLeft` win. fix not verified.

likely fix: write `rightToLeft: false` only when the resolved direction would still be rtl after the delete. keeps the current behavior for vanilla paragraphs and forces an explicit override only where the style cascade would otherwise win.

packages/super-editor/src/editors/v1/components/toolbar/defaultItems.test.js
```diff
@@ -32,7 +32,7 @@ function buildItems(availableWidth) {
 }
 
 describe('makeDefaultItems XL overflow boundary (SD-2328)', () => {
-  const XL_OVERFLOW_SAFETY_BUFFER = 20;
+  const XL_OVERFLOW_SAFETY_BUFFER = 84;
   const XL_CUTOFF = RESPONSIVE_BREAKPOINTS.xl + XL_OVERFLOW_SAFETY_BUFFER;
   const XL_ITEMS = ['linkedStyles', 'clearFormatting', 'copyFormat', 'ruler'];
```
caio-pizzol, line 37: nit: `defaultItems.js:1097-1108` now hides `directionLtr` / `directionRtl` at XL, but `XL_ITEMS` here never picked them up.

```suggestion
  const XL_ITEMS = ['linkedStyles', 'clearFormatting', 'copyFormat', 'ruler', 'directionLtr', 'directionRtl'];
```

caio-pizzol (commented):
hey @shri-scale! thanks for the PR - delete-on-ltr matching what Word actually writes is a nice touch :)

> [!IMPORTANT]
> needs work, three things to address.

users in an rtl paragraph see neither LTR nor RTL highlight in the overflow popup, and clicking LTR on a paragraph that inherits rtl from a style does nothing. one is out of diff (`super-toolbar.js:715`); two more inline.

caio-pizzol (commented):
@shri-scale thanks again for jumping in on this

i pushed a few follow-up commits while reviewing: tightened the headless-toolbar payload type to `never`, added unit + behavior coverage for the click path, and pulled the direction buttons out of the default toolbar (they stay available as a command + headless id; customers wire them into their own UI).

lgtm.

caio-pizzol (approved):

PR #3231, opened 2026-05-11
fix(converter): preserve numbering.xml definitions on round-trip (SD-2911)

tupizz:
## Summary

- `#exportNumberingFile` no longer filters `numbering.xml` by which `numId`s are referenced from the document body. Every `<w:abstractNum>` and `<w:num>` the importer captured is emitted on export.
- Removes `strip-orphaned-numbering.js` + its unit tests; the inline write is now four lines.

## Why

Customer fixture in [SD-2911](https://linear.app/superdocworkspace/issue/SD-2911):

- **Active variant** (one `numId` used in the body): `<w:abstractNum>` dropped from 8 → 1, `<w:num>` from 8 → 1. Live numbering definitions Word would otherwise expose in the Multilevel List dropdown were silently lost.
- **Tentative variant** (no `numId` referenced): `numbering.xml` was effectively wiped (2 → 0 abstractNum, 1 → 0 num).

The filter was added by PR #2223 for the `lists.delete` document-api operation — when a user deletes a list, the dead definition should be cleaned up. But the filter ran on **every** export and had no way to tell "user just deleted in this session" from "definition was always unused in the source file" (Word's tentative numbering). Both reached it with no referencing paragraph and were treated identically. Word tolerates unused definitions, so the safe default on round-trip is to preserve everything.

## Test plan

- [x] New integration test (`sd-2911-numbering-roundtrip.test.js`) round-trips both customer fixtures and asserts `<w:abstractNum>` / `<w:num>` counts and ids match the input exactly.
- [x] Full `pnpm test` for `super-editor` (12 695 / 12 708 pass — the 13 deletions are the removed `strip-orphaned-numbering.test.js` unit tests).
- [x] Live verification in the dev app: re-exported both fixtures, opened in Word.
  - Active: `numbering.xml` byte-identical to input after pretty-print (8 / 8 preserved).
  - Tentative: same — `numbering.xml` byte-identical (2 / 1 preserved).
- [x] No "Word found unreadable content" prompts on open.

## Verified diffs on the customer fixtures

| Variant | input | before fix | after fix |
| --- | --- | --- | --- |
| active — `<w:abstractNum>` | 8 | 1 | 8 |
| active — `<w:num>` | 8 | 1 | 8 |
| tentative — `<w:abstractNum>` | 2 | 0 | 2 |
| tentative — `<w:num>` | 1 | 0 | 1 |

## What about the lists.delete flow that originally motivated the filter?

If `lists.delete` needs to remove the definition for the list it deletes, that should happen at the operation site — where the deletion intent is known — not as an opportunistic GC pass at every export. There's no existing test in the repo that gates this behavior end-to-end against the document-api; adding one is out of scope for this fix.


luccas-harbour (changes requested):
hey! can you check this one?

    - [P2] Avoid exporting fallback numbering for docs without numbering — packages/super-editor/src/editors/v1/core/super-converter/SuperConverter.js:1440-1442
    When the input package has no `word/numbering.xml` (or an empty one), the importer still fills `this.numbering` from `baseNumbering` via `ensureElementsArray`; writing all `Object.values` here therefore exports SuperDoc's fallback numbering definitions even though the source document had none. Main previously emitted no definitions because no paragraphs referenced those fallback numIds, so plain documents now gain unused numbering definitions on round-trip. Track whether the numbering part came from the source, or filter only the fallback, before preserving everything.

tupizz:
@luccas-harbour pushed \`9e9e8b318\`. The regression you flagged is fixed: plain documents no longer gain a \`word/numbering.xml\` on round-trip.

### Fix shape

\`SuperConverter.js::#exportNumberingFile\` now captures source presence before \`baseNumbering\` is substituted, and returns early when both:

- the source package didn't ship \`word/numbering.xml\`, AND
- the exported body contains no \`w:numPr\` reference.

If the source had numbering (active or tentative), every definition still round-trips (SD-2911's original intent is preserved). If a user creates a list in the session, the body carries \`w:numPr\` and the numbering part is written normally.

\`Editor.ts\` was the only other place that assumed numbering.xml was always set on \`convertedXml\`; treating it as optional now mirrors the appXml / coreXml handling already in that block.

### TDD

3 new integration tests against \`blank-doc.docx\` and \`Hello docx world.docx\` (both have no source numbering): \`exportedNumberingEntry\` was being emitted before the fix and is now absent. The 3 existing tests for the active + tentative SD-2911 fixtures still pass.

### Verification

| Layer | Result |
|---|---|
| New plain-doc tests (3) | red → green |
| Existing SD-2911 tests (3) | preserved |
| Full \`super-editor\` suite | ✅ 12 698 / 12 698 |

luccas-harbour (approved):
LGTM

PR #3234, opened 2026-05-11
fix(context-menu): fix slash menu dismissal state (SD-2747)

tupizz:
## Demo

https://github.com/user-attachments/assets/90fb51cd-c48d-4f3f-b1f0-982021b408c6

https://github.com/user-attachments/assets/9cb87371-72db-4491-bbf1-79e5bbf859c5

## Summary

Three independent bugs combined to break the dismiss-and-retype flow on the `/` command menu:

1. **Backspace and Delete did nothing** — neither the PM plugin's `handleKeyDown` nor the Vue component's document `keydown` listener caught them, so pressing Backspace after opening the menu left it open.
2. **A 5-second `slashCooldown` locked out subsequent `/` presses** immediately after dismissal. Type `/`, dismiss the menu, type `/` again — and a literal `/` got inserted instead of the menu reopening.
3. **Escape closed the menu but didn't re-emit the slash.** The original `/` was `preventDefault`'d on open, so the user's keystroke disappeared.

Per the ticket spec (matching Google Docs):
- Backspace / Delete: dismiss the menu, no slash inserted.
- Escape: dismiss the menu, leave the slash visible at the original anchor.
- Subsequent `/` reopens the menu immediately, no cooldown.

## Changes

- **`extensions/context-menu/context-menu.js`** (PM plugin): removed `slashCooldown` + its 5 s timeout. `handleKeyDown` now branches on `Backspace`/`Delete` (close, no insert), `Escape`/`ArrowLeft` (close, insert `/` at anchor), and the existing `/`-to-open path.
- **`components/context-menu/ContextMenu.vue`** (Vue): focus shifts to the hidden search input when the menu opens, so the PM plugin can't see keys typed afterward. Added the same Backspace/Delete and Escape branches to `handleGlobalKeyDown` so the dismissal works whichever element holds focus.
- **`extensions/context-menu/context-menu.test.js`**: removed the three cooldown unit tests (they locked in the buggy behavior); added six new unit tests for Backspace / Delete / Escape / immediate reopen.

## Test plan

- [x] New unit tests: 6 cases covering Backspace closes, Delete closes, Escape inserts the slash, Backspace doesn't insert, immediate reopen works, Backspace ignored when menu closed.
- [x] Full `pnpm test` for `super-editor` (12 711 / 12 724 pass — same as `main` aside from the new tests).
- [x] Live browser repro on the dev app:

| Scenario | Step | Menu | Doc | Expected |
|---|---|---|---|---|
| Backspace | press `/` → press Backspace → press `/` again | open → **closed** → **open** | `` → `` → `` | ✓ |
| Escape | press `/` → press Escape | open → closed | `` → **`/`** | ✓ |
| Delete | press `/` → press Delete | open → **closed** | `` → `` | ✓ |
| Visual | press `/` | menu visible with `Insert text` / `Insert table` / `Paste` and first item highlighted | | ✓ |

## Why remove the cooldown entirely

The original 5 s cooldown was intended to prevent rapid menu reopens, but its only behavioral effect was the bug in the ticket — typing `/` again after dismissal inserted a literal `/` for 5 seconds. The dismissal handlers above are now explicit about state, so the cooldown's purpose is fully served by `pluginState.open` itself: when the menu is already open, the second `/` is ignored at the dispatch site without needing a separate timer.


luccas-harbour (changes requested):
hey @tupizz, can you check this one? I was able to reproduce it in the editor. If I dismiss the context menu by pressing escape and it was opened via a right click, a slash character gets added:

  - [P2] Gate slash reinsertion to slash-triggered menus — packages/super-editor/src/editors/v1/
    components/context-menu/ContextMenu.vue:238-241
    When the same ContextMenu is opened via right-click, handleRightClick also stores an anchorPos, but no / key was suppressed. This Escape path now unconditionally inserts /, so
    pressing Escape to dismiss a right-click context menu mutates the document. Please only reinsert the slash for slash-triggered menus, and apply the same guard to the PM key
    handler that also inserts / on Escape/ArrowLeft.

Also, I noticed that when I press `/`, the context menu appears and then if I press escape to dismiss it, it will insert the slash but then jump to a new line. And another thing is that it seems I have to press escape twice for the context menu to be dismissed.

Can you check these? Thanks!

packages/super-editor/src/editors/v1/extensions/context-menu/context-menu.js
```diff
@@ -426,23 +402,32 @@ export const ContextMenu = Extension.create({
...
-            // Restore cursor position and focus
             if (anchorPos !== null) {
-              const tr = view.state.tr.setSelection(
-                view.state.selection.constructor.near(view.state.doc.resolve(anchorPos)),
-              );
-              view.dispatch(tr);
+              const insertTr = view.state.tr.insertText('/', anchorPos);
+              const insertedAt = anchorPos + 1;
+              insertTr.setSelection(view.state.selection.constructor.near(insertTr.doc.resolve(insertedAt)));
+              view.dispatch(insertTr);
```
luccas-harbour: Found this other problem:

  ArrowLeft is lumped with Escape so it now inserts a literal /. Also, handleGlobalKeyDown does not handle ArrowLeft at all, so while the menu is open ArrowLeft is swallowed by the hidden input and never dismisses the menu — making this PM branch unreachable in the primary flow.

packages/super-editor/src/editors/v1/components/context-menu/ContextMenu.vue
```diff
@@ -648,6 +686,39 @@ onBeforeUnmount(() => {
...
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.context-menu-empty {
+  padding: 10px 10px;
+  color: var(--sd-ui-menu-text-muted, #888);
+  font-style: italic;
+  text-align: center;
+}
```
luccas-harbour, line 729: --sd-ui-menu-header-bg, --sd-ui-menu-text-muted, and --sd-ui-font-mono are referenced with hex fallbacks but not declared in  packages/superdoc/src/assets/styles/helpers/variables.css

packages/super-editor/src/editors/v1/components/context-menu/ContextMenu.vue
```diff
@@ -223,11 +223,35 @@ const cleanupCustomItems = () => {
 };
 
 const handleGlobalKeyDown = (event) => {
-  // ESCAPE: always close popover or menu
+  // SD-2747: ESCAPE dismisses the menu and inserts a literal `/` at the original anchor —
+  // the slash was preventDefault'd when the menu opened, so we re-insert it here so the
+  // user's typed character is preserved when they decline to pick a command. Matches Google
+  // Docs' trigger-menu behavior.
   if (event.key === 'Escape' && isOpen.value) {
```
luccas-harbour: the close+insert-/+selection-near+focus block is now in two places (here and context-menu.js:418-434) and only the Vue one actually runs (hidden input has focus while menu is open). want to drop the PM-side branches now that Vue owns dismissal, or extract a shared dismissSlashMenu(editor, anchorPos, { insertSlash }) helper?

tupizz:
@luccas-harbour pushed \`92e7c0268\` addressing the Codex P1 plus the additional bugs you flagged. Browser-verified end-to-end on the dev app.

### Root cause

The plugin state didn't differentiate slash-keystroke opens from right-click opens. Both dismissal paths (the PM \`handleKeyDown\` and the Vue \`handleGlobalKeyDown\`) unconditionally reinserted \`/\` at the anchor — fine for slash-triggered opens (the original keystroke was preventDefault'd, so we owe the user a literal slash), but wrong for right-click opens where no keystroke was ever suppressed.

### Fix

- **\`context-menu.js\`** — new \`trigger: 'slash' | 'rightClick' | null\` field on the plugin state. Set from the \`isRightClick\` signal already computed in the \`'open'\` apply branch (presence of \`clientX\`/\`clientY\` in the meta). Cleared on \`'close'\`. PM-side Escape / ArrowLeft branch only reinserts \`/\` when \`trigger === 'slash'\`.
- **\`ContextMenu.vue::handleGlobalKeyDown\`** — same \`trigger === 'slash'\` gate applied to the Vue-owned dismissal path. ArrowLeft is now in this branch too: the hidden search input owns focus while the menu is open, so the PM plugin's matching branch never fires in the live flow — ArrowLeft was simply being swallowed instead of dismissing the menu.
- **\`variables.css\`** — declared the three \`--sd-ui-menu-*\` tokens the component references with hex fallbacks (\`--sd-ui-menu-header-bg\`, \`--sd-ui-menu-text-muted\`, \`--sd-ui-font-mono\`) at \`:root\` so consumers can theme them.

### TDD evidence

5 new failing tests before the fix, 5 green after:

| Test | Before | After |
|---|---|---|
| records \`trigger='slash'\` on keyboard open | ❌ (field missing) | ✅ |
| records \`trigger='rightClick'\` on clientX/Y open | ❌ | ✅ |
| clears trigger on close | ❌ | ✅ |
| Escape on right-click open: closes, doesn't insert \`/\` | ❌ inserted \`/\` | ✅ |
| ArrowLeft on right-click open: closes, doesn't insert \`/\` | ❌ inserted \`/\` | ✅ |

### Regression guard

- \`context-menu.test.js\`: 64 / 64 plugin tests pass (existing dismissal-behavior tests open via slash, so the gate preserves their behaviour).
- \`ContextMenu.test.js\`: 44 / 44 component tests pass.
- Full \`super-editor\` suite: 12 716 / 12 716 pass.

### Browser repro on the dev app

| Scenario | Doc text before | Doc text after | Menu | Verdict |
|---|---|---|---|---|
| Right-click → Escape | (empty) | (empty) | closed | ✅ no stray \`/\` |
| Right-click → ArrowLeft | (empty) | (empty) | closed | ✅ no stray \`/\` |
| Slash → Escape | (empty) | \`/\` | closed | ✅ preserved |
| Slash → ArrowLeft | (empty) | \`/\` | closed | ✅ previously swallowed — now dismisses |

### Notes on your two additional observations

You also mentioned (1) the slash insert on Escape causing a "new line" and (2) needing to press Escape twice. I couldn't reproduce either in agent-browser repro — single Escape always dismisses, and the inserted \`/\` lands on the same paragraph (\`paraCount: 1\`). Could you re-test on this revision and let me know if either still repros? If they do, a video / exact steps would help me trace — there may be a focus-state edge case my synthetic tests aren't hitting.

luccas-harbour (approved):
LGTM. I couldn't reproduce the issues I found during manual testing anymore 👍

PR #3236, opened 2026-05-11
fix(header-footer): use section-aware page numbering for odd/even parity (SD-2991)

luccas-harbour:
## Summary
- Switches odd/even header/footer parity selection from the physical page index to a new section-aware `displayNumber` so headers respect OOXML `<w:pgNumType w:start="…">` and per-section number restarts (ECMA-376 §17.10.1).
- Threads a `displayNumber` field through the layout pipeline (`Page` → `HeaderFooterPage` → `ResolvedPage` → `ResolvedHeaderFooterPage`) and stamps it from `activePageCounter` inside `layoutDocument`.
- Adds a `parityPageNumber` option to the `headerFooterUtils` selectors (`getHeaderFooterType`, `getHeaderFooterTypeForSection`, `getHeaderFooterIdForPage`, `resolveHeaderFooterForPage(AndSection)`) so callers can override parity when needed. Also fixes the legacy single-section selector to use `% 2 !== 0` for odd, so negative display numbers route to `odd` instead of falling through.
- `HeaderFooterSessionManager` now reads `displayNumber` when picking header/footer variants for both per-rId layouts and inferred region variants.

## Why
When a section sets `<w:pgNumType w:start="2"/>` (or otherwise restarts numbering), Word treats the first body page as page 2 and renders the even header on it. We were keying odd/even off the physical page index instead, so the wrong variant was selected — and downstream space reservation (header content height) used the wrong height, shifting body content vertically.

## Changes by package
- `layout-engine/contracts`: add optional `displayNumber` on `Page`, `HeaderFooterPage`, `ResolvedPage`, `ResolvedHeaderFooterPage`.
- `layout-engine/layout-engine`: rename `documentPageNumber` → `parityPageNumber` in `getVariantTypeForPage`, feed it `activePageCounter`, and stamp `state.page.displayNumber` during page creation.
- `layout-engine/layout-bridge`: new `parityPageNumber` option threaded through every header/footer selector; existing callers default to `page.displayNumber ?? page.number`. Negative-odd parity now resolves to `'odd'`.
- `layout-engine/layout-resolved`: forward `displayNumber` from layout to resolved page/header-footer outputs.
- `super-editor/HeaderFooterSessionManager`: use `displayNumber` for both region variant inference and per-rId variant lookup; preserve it when forwarding pages into header/footer layout requests.

## Tests
- `layout-engine`: new `uses section page-numbering start for odd/even header parity` test (section with `numbering.start = 2` swaps body offsets on pages 1 and 2). Existing alternateHeaders tests updated to describe behavior in terms of display number.
- `layout-bridge/headerFooterUtils`: new tests cover the `parityPageNumber` override, negative-odd handling, section-aware display-number parity, and caller-supplied parity overrides.
- `layout-resolved/resolveHeaderFooter`: asserts `displayNumber` is preserved through resolution.
- `super-editor/HeaderFooterSessionManager`: new tests verify `displayNumber` parity is used both when selecting per-rId header layouts and when inferring region variants.

### DIFF
--- a/packages/layout-engine/contracts/src/index.ts
+++ b/packages/layout-engine/contracts/src/index.ts
@@ -2001,6 +2001,8 @@ export type Page = {
    * (in later phases) by body pagination itself.
    */
   footnoteLedger?: FootnotePageLedger;
+  /** Numeric page number after section numbering restart/offset. Used for OOXML odd/even parity. */
+  displayNumber?: number;
   numberText?: string;
   size?: { w: number; h: number };
   orientation?: 'portrait' | 'landscape';
@@ -2228,6 +2230,7 @@ export type HeaderFooterType = 'default' | 'first' | 'even' | 'odd';
 export type HeaderFooterPage = {
   number: number;
   fragments: Fragment[];
+  displayNumber?: number;
   numberText?: string;
   /**
    * Optional page-local block clones backing this page's resolved fragments.
--- a/packages/layout-engine/contracts/src/resolved-layout.ts
+++ b/packages/layout-engine/contracts/src/resolved-layout.ts
@@ -450,6 +450,8 @@ export function isResolvedDrawingItem(item: ResolvedPaintItem): item is Resolved
 /** A resolved header/footer page — mirrors HeaderFooterPage but with resolved items. */
 export type ResolvedHeaderFooterPage = {
   number: number;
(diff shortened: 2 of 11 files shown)


tupizz (approved):

harbournick (approved):
LGTM

PR #3238, opened 2026-05-11
SD-2974 - Bug: image resizing in suggesting mode duplicates the image

chittolinag:
### Issue

Images resized in *suggesting* mode creates a duplicate. 

### Proposed solution

Since MS Word does not track changes when the image is resized, the simplest way to fix the issue is by simply ignoring track changes on the event of an image resize. 

Just for future reference, the original issue happens around [here](https://github.com/superdoc-dev/superdoc/blob/e8ce982ab9f120505c156f36ccd6c9c8f9753844/packages/super-editor/src/editors/v1/extensions/track-changes/trackChangesHelpers/replaceStep.js#L230). 

This `replaceRange` call adds the new (resized) image but keeps the old one. If we ever want to show the old and new image (and show the tracked change for resized) images, we might have to fix this.

Created a behavior test that ensures this doesn't happen in the future (tested both before & after the fix).


packages/super-editor/src/editors/v1/components/ImageResizeOverlay.vue
```diff
@@ -583,7 +583,11 @@ function dispatchResizeTransaction(blockId, newWidth, newHeight) {
 
     tr.setNodeMarkup(imagePos, null, newAttrs);
 
-    // Dispatch transaction
+    // Word does not track image resizes as revisions; bypass tracking so the
+    // attribute change applies in place instead of being split into a tracked
+    // insert + delete (which would render as a duplicate image — SD-2974).
+    tr.setMeta('skipTrackChanges', true);
```
andrii-harbour: This works for the common case, but there's a cleaner mechanism that also fixes a case this one misses:

**The miss:** on current main, `protectsExistingTrackedReviewState` (`Editor.ts:3119-3125`) **overrides** `skipTrackChanges`. So resizing an image that is *itself a pending tracked insertion* (insert image in suggesting mode → resize it before acceptance — a very normal flow, and it matches the "stuck original" in the customer's Loom) still routes through tracking → duplicate again.

**The cleaner fix:** the resize only ever changes the `size` attr, so:

```js
tr.setNodeAttribute(imagePos, 'size', {
  width: Math.round(newWidth),
  height: Math.round(newHeight),
});
```

`setNodeAttribute` emits an `AttrStep` instead of a `ReplaceStep`, and:
- `trackedTransaction.js:462-467` passes `AttrStep` through untracked *by design* ("metadata updates… safe to apply without tracking")
- `stepTouchesTrackedReviewState` (`Editor.ts:354-397`) has no `AttrStep` branch → falls through to `false`, so the protect guard never fires → the pending-insertion case is fixed too
- no `setMeta` needed at all

If you keep the meta approach instead, the pending-insertion limitation should at least be documented + characterized in a test.

tests/behavior/tests/images/image-resize-in-suggesting-mode.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    await superdoc.page.mouse.move(startX - 60, startY - 60, { steps: 10 });
+    await superdoc.page.mouse.up();
+
+    await superdoc.waitForStable();
+
+    // After the resize commit there must still be exactly one image node.
+    // Before the SD-2974 fix the tracked-change machinery split the
+    // ReplaceStep (generated by setNodeMarkup on the leaf image node) into
+    // a tracked insert + delete, leaving two image nodes in the document.
+    expect(await getImageCount(superdoc)).toBe(1);
```
andrii-harbour, line 203: False-green risk: if the throttled mousemove never commits a ≥1px delta (the scenario your comment on lines 48–50 describes), `onDocumentMouseUp` skips the dispatch entirely — and this assertion still passes (1 image) without having exercised the fix at all.

Suggest capturing the image's `size` attrs before the drag and asserting they actually changed afterwards, so the test fails loudly when the drag no-ops:

```ts
const before = await getImageSize(superdoc);
// …drag…
const after = await getImageSize(superdoc);
expect(after.width).toBeLessThan(before.width);
expect(await getImageCount(superdoc)).toBe(1);
```

tests/behavior/tests/images/image-resize-in-suggesting-mode.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (!doc) throw new Error('Editor document is unavailable.');
+    let count = 0;
+    doc.descendants((node: any) => {
+      if (node.type?.name === 'image') count += 1;
+    });
+    return count;
+  });
+}
+
+test.describe('Image resize in suggesting mode (SD-2974)', () => {
```
andrii-harbour, line 92: Two of the ticket's acceptance criteria aren't covered yet:

1. **Repeated resizes don't accumulate duplicates** (AC #3) — cheap to add: drag twice, assert count is still 1.
2. **Resizing an image that's a pending tracked insertion** — this is the `protectsExistingTrackedReviewState` edge from the other comment, and it's the case most likely to regress. Insert an image while in suggesting mode, then resize it, assert count stays 1.

(2) will fail against the current `skipTrackChanges` approach, which is exactly why it's worth having — it pins down the mechanism choice.

andrii-harbour (changes requested):
Nice job! Here are a few notes:

1. **Rebase needed**: the branch is ~1,100 commits behind main, and main's `ImageResizeOverlay.test.js:17` mocks `tr` as `{ setNodeMarkup }` only — no `setMeta` — so this PR as-is throws `tr.setMeta is not a function` in main's unit tests. (Good news: verified the bug still reproduces on current main, so the fix is still needed.)
2. **Simpler mechanism worth considering**: `tr.setNodeAttribute(imagePos, 'size', …)` instead of `setNodeMarkup` + `skipTrackChanges` — it also fixes an edge case the current approach misses (details inline).
3. **Test hardening**: the behavior test can false-pass if the drag no-ops, and two of the ticket's acceptance criteria aren't covered yet (details inline).

chittolinag:
@andrii-harbour Thanks for the review! All addressed, including the rebase.

**1. Mechanism** -  switched to `setNodeAttribute`, plus one fix it needed: AttrSteps have no changed range, so `blockNodePlugin` never bumps `sdBlockRev` and the layout engine reuses the cached FlowBlock -  PM state updated but the paint stayed stale. The resize now bumps the containing blocks' `sdBlockRev` in the same transaction (same pattern as `numberingPlugin`). 

**2. Tests** -  all three added: before/after size assertions, repeated resizes, pending-insertion resize.

Two findings from the pending-insertion test:

1. It also passes under the old `skipTrackChanges` approach: on current main the protect guard sets `protectTrackedReviewState`, which compiles the replace with `preserveExistingReviewState: true` -  in-place, no duplicate. So it's regression coverage, not a mechanism pin. AttrStep is still the better choice.
2. Separate latent bug: `setImage` in suggesting mode duplicates at *insert* time - `imageRegistrationPlugin` rewrites unregistered images via untracked `setNodeMarkup`. The test sidesteps it (pre-registered image with `rId` + `trackInsert` mark constructed directly); I'm creating a new ticket for that now.

harbournick (approved):
LGTM

PR #3239, opened 2026-05-11
fix(super-editor): keep comment range markers around unpaired tracked changes (SD-2528)

tupizz:
## Demo

After export-import 

CleanShot 2026-05-14 at 08 13 24@2x

## Summary

When a user comments on redlined text, the exported DOCX put `w:commentRangeStart` outside the `w:del` wrapper but `w:commentRangeEnd` and the `w:commentReference` run inside it. Re-importing that file separates the comment from the redline (and from itself) instead of restoring a single unified bubble. Customer-reported (Bonterms).

Root cause is in `mergeConsecutiveTrackedChanges` (`translate-paragraph-node.js`). The forward scan greedily absorbed every comment marker / comment-reference run it found after a `w:ins`/`w:del`, even when no same-id wrapper actually followed to merge with. So a lone tracked change always swallowed its trailing comment markers.

Fix: buffer comment markers during the scan and only commit them inside the wrapper when a same-id merge actually happens. Otherwise emit them as siblings.

- `w:commentRangeStart` ... `w:del` ... `w:commentRangeEnd` ... `w:r/w:commentReference` now stays as siblings (SD-2528).
- SD-1519 same-id merge with comments between is still preserved and now covered by a regression test.

## Test plan

- [x] Unit tests in `translate-paragraph-node.test.js` cover three cases: SD-2528 (single wrapper + trailing comments), SD-1519 (same-id merge absorbs comments), different-id (no merge, comments stay siblings).
- [x] Full `pnpm --filter super-editor test` passes (1008 files, 12711 tests).
- [ ] Round-trip a fixture with a comment on a redline in Word and confirm the unified bubble survives.


luccas-harbour (changes requested):
hey @tupizz!
I performed the same steps that were shown in the loom video that was in the ticket and I can see that the problems persist:
- After recreating the same document as in the video, exporting it and loading it into SD again, the comments are displayed separately instead of threaded
- If I accept the added track change, the comment reply does not go away
- If I accept the replacement track change, the comment reply doesn't go away either
Can you check?

tupizz:
thanks @luccas-harbour I missunderstood the requirements

luccas-harbour (changes requested):
hey @tupizz!
do you mind checking these?

    - [P2] Restrict cascading resolve to the active document — packages/superdoc/src/stores/comments-store.js:690-695
      When a tracked-change resolve event includes `documentId`, `findTrackedChangeById()` filters to that document, but this new cascade scans every item in `commentsList`. In multi-document sessions where imported/live tracked-change ids are reused across documents, accepting or rejecting a change in one document will also resolve comments attached to the same `trackedChangeParentId` in another document.

    - [P2] Only suppress comment fill for highlighted redlines — packages/super-editor/src/editors/v1/core/presentation-editor/dom/CommentHighlightDecorator.ts:184-186
      This treats any base tracked-change class as having its own background, but final/original modes still render visible spans with `track-insert-dec`/`track-delete-dec` and no `.highlighted` background, and format changes only have a border. In those modes a comment anchored on the text has its comment background cleared with no TC background replacing it, making the comment highlight disappear.

    - [P2] Preserve regular parent threads for TC-anchored replies — packages/superdoc/src/components/CommentsLayer/CommentDialog.vue:327-329
      This now pulls every comment with `trackedChangeParentId` into the tracked-change dialog, even when the comment also has a `parentCommentId` pointing to a regular comment thread. The importer supports that shape for replies whose range is inside a TC but whose parent spans outside it, so those replies will appear both under their real parent and inside the TC dialog without the parent context.

tupizz:
Pushed `2a04dc9ad` covering Codex's three P2 findings. TDD red→green for each, plus the SD-2528 round-trip and cascade still work end-to-end.

### Fixes

| # | Site | Fix | Lines |
|---|------|-----|-------|
| P2 #1 | `comments-store.js:668-720` cascade | Mirror `findTrackedChangeById`'s per-document filter when the resolve event carries a `documentId`. Single-doc callers (no `documentId`) keep the legacy global behaviour. | +6 |
| P2 #2 | `CommentHighlightDecorator.ts:171-198` gate | Suppress comment fill ONLY when the TC actually paints a competing background — i.e. `.highlighted` modifier AND (`track-insert-dec` OR `track-delete-dec`). `track-format-dec` only paints a border; in Original/Final modes `.highlighted` is absent (see `renderer.ts:909-928`'s `TRACK_CHANGE_MODIFIER_CLASS` map). | +1 / −2 |
| P2 #3 | extracted `collect-tracked-change-thread.js` BFS seed | Only seed via `trackedChangeParentId` when the comment is a thread root (no `parentCommentId`). Bi-parented same-TC chains are still picked up via the BFS step; bi-parented replies whose real parent lives outside the TC stay in their parent's thread. | +1 / −1 |

### TDD evidence (Red → Green)

| Test file | Before fix | After fix |
|---|---|---|
| `collect-tracked-change-thread.test.js` (new, 8 tests) | 3 failing on bi-parented cases | 8 / 8 pass |
| `CommentHighlightDecorator.test.ts` (+9 SD-2528 tests) | 4 failing on Original / Final / format-only | 32 / 32 pass |
| `comments-store.test.js` (+2 tests) | 1 failing on cross-doc cascade | 105 / 105 pass |

### Regression guard

| Layer | Result |
|---|---|
| `sd-2528-tc-comment-roundtrip` integration | ✅ 1/1 (round-trip pairing intact) |
| `CommentDialog.test.js` | ✅ 45/45 (extraction is behaviour-preserving) |
| Full `super-editor` suite | ✅ 12 850 / 12 850 pass |
| Full `superdoc` suite | ✅ 966 pass (1 pre-existing `collab-server.test.ts` import failure on `main` — unrelated) |
| Browser repro on the corpus fixture | Accepting an imported TC (`455257fe…`) still cascade-resolves both anchored user comments (`imported-9cfaa91b` root + `imported-099ba8eb` reply via BFS). |

### Why extract `collectTrackedChangeThread` into a helper file

The function was a 30-line pure function trapped in `CommentDialog.vue`'s `<script setup>`, untestable without mounting the component. Extracting to `collect-tracked-change-thread.js` is the minimal change that lets the BFS logic be unit-tested in isolation. CommentDialog.vue still imports and uses it identically. 45/45 existing CommentDialog tests pass against the extracted helper before any fix, proving the extraction is behaviour-preserving.

### Karpathy-guideline check

- Every changed line traces to one of the three Codex findings; no adjacent refactors.
- Each fix is the minimum change that satisfies its failing test.
- TDD: every fix has a paired test that fails on the prior commit and passes on this one.

tupizz:
@luccas-harbour all three of Codex's P2 concerns are addressed in `2a04dc9ad`. Quick summary:

- **Multi-doc cascade leakage** — `comments-store.js` cascade now mirrors `findTrackedChangeById`'s `belongsToTrackedChangeSyncDocument` filter when the resolve event carries a `documentId`. Single-doc paths (no `documentId`) keep the legacy behaviour.
- **Comment fill suppression in Original / Final / format-only** — `CommentHighlightDecorator` gate tightened to require both `.highlighted` and (`track-insert-dec` or `track-delete-dec`). In Original/Final modes `.highlighted` is absent (per `renderer.ts`'s `TRACK_CHANGE_MODIFIER_CLASS`), and `track-format-dec` paints only a border, so the previous gate was clearing comment fill with nothing to replace it.
- **Bi-parented reply shadowing** — extracted `collectTrackedChangeThread` to a sibling helper so the BFS could be unit-tested; only seed via `trackedChangeParentId` when the comment is a thread root (no `parentCommentId`). The BFS still picks up same-TC-anchored reply chains via parent links, so the existing fixture's threading is preserved.

TDD: 8 new failing tests before the fixes, all green after. SD-2528 round-trip test still passes. Full `super-editor` (12 850) and `superdoc` (966) suites green. Browser repro on the corpus fixture confirms cascade-resolve still fires end-to-end.

Full verification matrix and per-fix line counts in the [previous comment](#issuecomment-4453213641).

luccas-harbour (approved):
LGTM

PR #3240, opened 2026-05-11
fix(superdoc): restore find input focus after match navigation (SD-3045)

tupizz:
## Demo

https://github.com/user-attachments/assets/9230a775-f89c-4c34-9b56-aaa4a2d6a6a2

## Summary

Pressing Enter in the built-in find/replace input used to advance to the next match once, then drop focus into the ProseMirror editor. Subsequent Enter keystrokes were swallowed by the editor (splitting paragraphs at the match position, invalidating the search session, and resetting the active match index back to zero), forcing the user to click back into the input between every navigation. The "alternating jump back to page 1" symptom in the ticket title is a downstream effect of the same focus steal.

Root cause: the Search extension's `goToSearchResult` calls `editor.view.focus()` so its newly-set selection is visible — by design when called programmatically — but `FindReplaceSurface.vue`'s Enter handler did not restore focus afterwards.

Fix: re-focus the find input synchronously after `goNext` / `goPrev` returns. The handler already calls `e.preventDefault()`, so once focus stays on the input the editor never sees the keystroke.

The change is local to `FindReplaceSurface.vue` — the underlying Search command behaviour (used by the dev sidebar and the public `SuperDoc.goToSearchResult` API) is unchanged.

## Test plan

- [x] New `FindReplaceSurface.test.js` covers: Enter keeps focus when `goNext` synchronously steals focus elsewhere; Shift+Enter / `goPrev` does the same; Enter's default is still prevented.
- [x] All 962 superdoc package tests pass.
- [x] Browser reproduction on the dev app: 5 consecutive Enters advanced 2/16 → 7/16 with focus retained, 3 Shift+Enters returned to 4/16.

### DIFF
--- a/packages/super-editor/src/editors/v1/assets/styles/elements/prosemirror.css
+++ b/packages/super-editor/src/editors/v1/assets/styles/elements/prosemirror.css
@@ -432,11 +432,15 @@ https://github.com/ProseMirror/prosemirror-tables/blob/master/demo/index.html
   margin-top: -5px;
 }
 
+/* AIDEV-NOTE: SD-3045. `!important` so the transient search highlight wins
+ * over any inline `style.backgroundColor` painted from a source-doc highlight
+ * mark on the same span. See packages/superdoc/src/assets/styles/elements/superdoc.css
+ * for the full rationale. */
 .sd-editor-scoped .ProseMirror-search-match {
-  background-color: #ffff0054;
+  background-color: #ffff0054 !important;
 }
 .sd-editor-scoped .ProseMirror-active-search-match {
-  background-color: #ff6a0054;
+  background-color: #ff6a0054 !important;
 }
 .sd-editor-scoped .ProseMirror span.sd-custom-selection::selection {
   background: transparent;
--- a/packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
+++ b/packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
@@ -3533,7 +3533,37 @@ export class PresentationEditor extends EventEmitter {
         if (pageEl) {
           // Find the specific element containing this position for precise centering
           const targetEl = this.#findElementAtPosition(pageEl, clampedPos);
-          (targetEl ?? pageEl).scrollIntoView({ block, inline: 'nearest', behavior });
+          const elToScroll = targetEl ?? pageEl;
+          elToScroll.scrollIntoView({ block, inline: 'nearest', behavior });
+          // AIDEV-NOTE: SD-3045. Search nav (and any other caller of
+          // scrollToPosition) places the viewport intentionally — usually
+          // centring the match. The next #updateSelection that runs as part
+          // of the dispatched setSelection transaction would otherwise call
+          // #scrollActiveEndIntoView and re-scroll the caret to its minimal
+          // visible position (often the top of the viewport), undoing our
+          // centring. Consume the pending scroll-into-view request so that
+          // selection sync renders the caret overlay without moving the
+          // scroll back. Other selection updates (Shift+Arrow, typing) re-set
+          // this flag themselves before they need scroll, so this consume is
+          // safe.
+          this.#shouldScrollSelectionIntoView = false;
+          // Re-assert the scroll on the next animation frame. The flag we
+          // cleared above defends against handleSelection that has already
+          // run, but a *later* selectionUpdate (e.g. focus blur fired when
+          // the user moves focus back to the find input) re-sets the flag to
+          // true before the RAF-scheduled #updateSelection fires, and that
+          // pass scrolls the caret to its minimal-visibility position —
+          // visibly snapping the match out of view. Re-running scrollIntoView
+          // on the same element a frame later overrides that snap; the no-op
+          // case (no late scroll happened) just re-centres the same element
+          // and is cheap.
+          const win = this.#visibleHost.ownerDocument?.defaultView;
+          if (win) {
+            win.requestAnimationFrame(() => {
+              elToScroll.scrollIntoView({ block, inline: 'nearest', behavior });
+              this.#shouldScrollSelectionIntoView = false;
+            });
+          }
           return true;
         }
       }
--- a/packages/superdoc/src/assets/styles/elements/search-match-precedence.test.js
+++ b/packages/superdoc/src/assets/styles/elements/search-match-precedence.test.js
@@ -1,0 +1,102 @@
+import { describe, it, expect } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, join } from 'node:path';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+// SD-3045 (cross-package CSS invariant). The DomPainter writes
+// `style.backgroundColor = run.highlight` inline on the same span the search
+// DecorationBridge tags with `.ProseMirror-search-match`. Without `!important`,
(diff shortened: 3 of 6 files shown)


luccas-harbour:
hey @tupizz! 
I was testing this and noticed that if I press enter and the next match is not in the current page, it will not scroll down and show me the match. I also tested this against the main branch and I can see that this was still an issue there (so it's not a bug that was introduced by your code). I was just wondering if perhaps fixing that is simple enough that you could do it in this PR? Let me know if that's not the case!

Here's the test document I was using to test this:
[finding1.docx](https://github.com/user-attachments/files/27713902/finding1.docx)

If you search for `titlePg` and press enter twice, you'll notice it doesn't scroll down to the next page.

tupizz:
@luccas-harbour nice suggestion
going to add this

tupizz:
Pushed `ce29cbf5d` addressing the cross-page scroll regression you flagged.

### What it was

Three races in one keystroke. Pressing Enter (or clicking next/prev) on a match on another page fired:

1. `editor.view.focus()` → `wrapOffscreenEditorFocus` saves/restores window scroll
2. `dispatch(tr.setSelection().scrollIntoView())` → emits `selectionUpdate` → `handleSelection` sets `#shouldScrollSelectionIntoView = true` → `#updateSelection` calls `#scrollActiveEndIntoView` → `scrollScreenRectIntoView` snaps caret to ~20px from viewport edge
3. `presentationEditor.scrollToPosition(from, { block: 'center' })` does the actual centring

Depending on whether `(2)`'s render flushed synchronously or via RAF, `(3)`'s scroll either won or got overwritten. The find bar lives in the document's normal flow, so any `focus()` on its input or buttons after we scrolled to a far page snapped the document back to where the find bar was — undoing `(3)` again. Result: sometimes the match was off-screen entirely, sometimes slightly above/below, never reliably centred.

### What changed

**`PresentationEditor.ts` — `scrollToPosition`:**
- Consume `#shouldScrollSelectionIntoView` after scrolling so the next `#updateSelection` doesn't re-snap.
- Re-assert `scrollIntoView` on the next animation frame to defend against a *later* `selectionUpdate` (focus blur, async PM event) that re-sets the flag before its RAF-deferred render fires. Cheap no-op when nothing tried to undo us.

**`FindReplaceSurface.vue` — `focusFindInput`:**
- `.focus({ preventScroll: true })` and drop `.select()` (it has no preventScroll equivalent and can itself scroll the input into view; Cmd/Ctrl+A still selects all if the user wants to retype).
- Pin `scrollTop` of all scrollable ancestors across the focus call as a belt for browsers that ignore `preventScroll` on some focus paths.
- Route the next/prev button clicks through `handleGoNext` / `handleGoPrev` so they also call `focusFindInput` — the original handler only restored focus from Enter, so button clicks left the button focused and the browser scrolled it (and therefore the find bar, and therefore the document) back into view.

### Verification

- 7 new/updated unit tests in `FindReplaceSurface.test.js` covering: Enter / Shift+Enter / next-button click / prev-button click all use `preventScroll`, default still prevented, focus retained after `goNext` steals focus.
- All 1366 PresentationEditor tests pass.
- Manual repro on your `finding1.docx`: 4 `titlePg` matches across 3 pages. Click sequence advances 1 → 2 (scroll 0, still on page 1) → 3 (scroll 694, page 2 centred) → 4 (scroll 728, page 3 centred) → 1 (scroll 0). Screenshot at match 3 shows the highlighted word visibly centred in viewport, find bar floating in place.

If you can sanity-check your own repro, that'd be great. If anything still drifts I'll keep iterating.

luccas-harbour (changes requested):
hey @tupizz, I noticed one more thing. I was testing with the document `basic/advanced-text.docx` from the test corpus and if you search for the word `Oscar`, you'll notice that some of the matches are not highlighted (I think matches 3, 5 and 7?). I know this is a bit more than what the ticket asked for so let me know if it's a quick fix.

tupizz:
Pushed `5562dc678` covering your Oscar-highlight finding.

### Root cause

Reproduced on `basic/advanced-text.docx` from the corpus. The PM search finds all 8 Oscar matches correctly and the DecorationBridge applies `.ProseMirror-search-match` to all 8 painted spans. The bug is CSS specificity: the DomPainter writes `style.backgroundColor = run.highlight` inline on the same span. Inline styles win over class selectors, so on every match in a run whose source rPr carries `<w:highlight w:val="white"/>` (decoded as `highlight: { color: '#FFFFFF' }`), the find highlight was painted over by white.

Inventory of computed `backgroundColor` on the 8 spans before fix:

| # | Text | Source `highlight` | Visible bg pre-fix |
| --- | --- | --- | --- |
| 1 | "with Oscar Wilde" (active) | — | rgba(255, 150, 0, 0.6) ✅ |
| 2 | "Oscar Fingal" | `#FFFFFF` | rgb(255, 255, 255) ❌ |
| 3 | "Oscar Wilde was born" | `#FFFFFF` | rgb(255, 255, 255) ❌ |
| 4 | "Oscar Wilde Centre" (link) | — | rgba(255, 213, 0, 0.4) ✅ |
| 5 | "Oscar was two years younger" (p1) | `#FFFFFF` | rgb(255, 255, 255) ❌ |
| 6 | "Willie and Oscar" (p1) | `#FFFFFF` | rgb(255, 255, 255) ❌ |
| 7 | "Oscar was two years younger" (p2) | `#FFFFFF` | rgb(255, 255, 255) ❌ |
| 8 | "Willie and Oscar" (p2) | — | rgba(255, 213, 0, 0.4) ✅ |

Five of eight, not three — your annotation marked one I had counted (`Oscar Fingal`) but didn't name in your message.

### Fix

Two CSS files, four lines:

- `packages/superdoc/src/assets/styles/elements/superdoc.css` — `!important` on `.superdoc .ProseMirror-search-match` and `.superdoc .ProseMirror-active-search-match`.
- `packages/super-editor/src/editors/v1/assets/styles/elements/prosemirror.css` — `!important` on the `.sd-editor-scoped` mirror pair.

`!important` is the desired semantic: while a search session is live, the find colour must override any source-doc highlight. The class is only present during a search; clearing the session removes it and the original highlight is restored.

### Why not Chrome's "0/20"

When you saw `Oscar 0/20` from Chrome's Cmd+F vs SuperDoc's `1 of 8`, that's not 12 missing matches — that's Chrome counting:

- 8 visible DomPainter spans (the actual matches)
- 8 hidden ProseMirror editable mirror spans (off-screen at x≈-9500; SuperDoc's editing state lives there)
- 3 `superdoc-sr-only` link-description spans for the "Oscar Wilde Centre" link
- 1 likely `aria-label` or hidden footnote ref

Chrome doesn't distinguish visible from off-screen and walks accessibility text. SuperDoc's 8 is the correct count of semantically unique occurrences in `editor.state.doc`.

### Verification

| Layer | Result |
| --- | --- |
| Per-span `getComputedStyle().backgroundColor` post-fix | all 8 match the token (`rgba(255, 213, 0, 0.4)` non-active, `rgba(255, 150, 0, 0.6)` active) |
| `pnpm --filter super-editor test --run search` | 11 files, 190 tests passing |
| New `search-match-precedence.test.js` | 6 tests — CSS source assertions + JSDOM specificity sanity (positive + negative case proving the bug existed without `!important`) |
| `pnpm --filter superdoc test --run` | 967 passed; 1 pre-existing failure in `collab-server.test.ts` confirmed unrelated (fails on `ce29cbf5d` baseline too) |
| Browser repro | 4 visible Oscars in viewport now all yellow; orange on active. No more white-on-white. |

luccas-harbour (approved):
LGTM

PR #3241, opened 2026-05-12
fix: suppress top spacing after manual page breaks

leandrotcawork:
## Summary

Fixes a DOCX layout fidelity issue where the first paragraph after an explicit manual page break could render lower than it does in Word.

In the attached `DC_Template_Descricao_Cargo.docx` fixture, the heading `01 MISSAO DO CARGO` starts at the top of page 2 in Word, but SuperDoc rendered an extra vertical gap above it.

## Root Cause

The issue was a combination of PM-to-layout conversion and page-top paragraph layout behavior:

1. A paragraph containing only an explicit break, such as `w:br w:type="page"`, was converted into both a `pageBreak` block and a synthetic empty paragraph block.
2. The first real paragraph after that manual page break still applied its `spacingBefore`, even though Word suppresses that spacing when the paragraph starts at the top of the new page.
3. The spacing suppression needed to happen before float remeasurement. Otherwise, anchored objects could be measured against the wrong Y position.
4. Anchored images/drawings between the manual page break and the paragraph needed to be treated as non-flow content for this page-top decision.

## Fix

This PR keeps the fix scoped to the layout pipeline:

- `pm-adapter`: Do not synthesize an empty paragraph when a paragraph contributes only explicit break blocks.
- `layout-engine`: Suppress `spacingBefore` only when a paragraph starts a fresh page after an explicit manual `pageBreak`.
- Preserve `pageBreakBefore` behavior so paragraph-level page-break formatting still keeps its own spacing.
- Compute the page-top suppression decision before placing paragraph-anchored images/drawings and before float remeasurement.
- Treat anchored fragments as non-flow content when checking whether a page is still at the top of its body flow.
- Re-register page/margin-relative pre-registered anchors on their actual pagination page when encountered later, so following paragraphs wrap against the correct page's float exclusions.

## Tests

Added or updated coverage for:

- page-break-only paragraphs emitting only a `pageBreak`
- mixed paragraph content around page breaks still producing paragraph blocks
- top-of-page spacing suppression after explicit manual page breaks
- preserving `spacingBefore` for `pageBreakBefore`
- float remeasurement using the suppressed page-top Y
- paragraph-relative anchored images before the first paragraph after a manual page break
- margin-relative pre-registered anchored images before the first paragraph after a manual page break, including wrapping/re-measurement on the actual page

## Local Validation

Passed:

```bash
pnpm --dir packages/layout-engine/layout-engine test -- index.test.ts layout-paragraph.test.ts
pnpm --dir packages/layout-engine/pm-adapter test -- src/converters/paragraph.test.ts src/index.test.ts
pnpm run test:superdoc
pnpm run lint
```

Notes:

- `pnpm run lint` completed with `0 errors` and existing warnings.
- `pnpm run test:editor` completed `1007/1008` files and timed out in an unrelated encrypted document lifecycle test. Re-running that exact test in isolation with a longer timeout passed:

```bash
pnpm --prefix packages/super-editor exec vitest run src/editors/v1/core/Editor.lifecycle.test.ts --testNamePattern "should open an encrypted file with the correct password" --testTimeout 120000
```

## Fixture

[DC_Template_Descricao_Cargo.docx](https://github.com/user-attachments/files/27622449/DC_Template_Descricao_Cargo.docx)

### DIFF
--- a/packages/layout-engine/layout-engine/src/index.test.ts
+++ b/packages/layout-engine/layout-engine/src/index.test.ts
@@ -472,6 +472,37 @@ describe('layoutDocument', () => {
     expect(layout.pages.length).toBe(1);
   });
 
+  it('suppresses spacingBefore for a paragraph after an explicit page break', () => {
+    const introBlock: FlowBlock = {
+      kind: 'paragraph',
+      id: 'intro-on-page-one',
+      runs: [],
+    };
+    const pageBreak: FlowBlock = {
+      kind: 'pageBreak',
+      id: 'manual-page-break',
+    } as PageBreakBlock;
+    const headingBlock: FlowBlock = {
+      kind: 'paragraph',
+      id: 'heading-at-page-start',
+      runs: [],
+      attrs: {
+        spacing: { before: 24, after: 8 },
+      },
+    };
+
+    const layout = layoutDocument(
+      [introBlock, pageBreak, headingBlock],
+      [makeMeasure([20]), { kind: 'pageBreak' }, makeMeasure([20])],
+      DEFAULT_OPTIONS,
+    );
+
+    expect(layout.pages).toHaveLength(2);
+    expect(layout.pages[0].fragments).toHaveLength(1);
+    expect(layout.pages[1].fragments).toHaveLength(1);
+    expect(layout.pages[1].fragments[0].y).toBe(DEFAULT_OPTIONS.margins!.top);
+  });
+
   it('handles spacingBefore equal to content area height (boundary condition)', () => {
     // Edge case: spacingBefore exactly equals the content area height.
     // This triggers the infinite loop guard after advancing to a new page.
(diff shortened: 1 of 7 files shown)


leandrotcawork:
Follow-up update pushed in \de0476f6b\.

This addresses the float/anchored-object review path by:

- deciding manual-page-break top-spacing suppression before paragraph-anchored images/drawings are placed
- keeping anchored fragments from disqualifying an otherwise fresh body-flow page top
- recognizing manual page breaks when paragraph-relative or page/margin-relative anchors sit between the break and the paragraph
- re-registering page/margin-relative pre-registered anchors on their actual pagination page so following paragraphs wrap against the correct float exclusions

Additional regressions now cover paragraph-relative anchors, margin-relative pre-registered anchors, float remeasurement at the suppressed page-top Y, and the existing \pageBreakBefore\ distinction.

Latest focused validation:

\\\ash
pnpm --dir packages/layout-engine/layout-engine test -- index.test.ts layout-paragraph.test.ts
pnpm --dir packages/layout-engine/pm-adapter test -- src/converters/paragraph.test.ts src/index.test.ts
pnpm run test:superdoc
pnpm run lint
\\\

\pnpm run test:editor\ reached 1007/1008 files and timed out in an unrelated encrypted document lifecycle case; rerunning that exact test in isolation with a longer timeout passed.

PR #3247, opened 2026-05-12
feat: add toc toolbar item

VladaHarbour:


packages/super-editor/src/editors/v1/components/toolbar/super-toolbar.js
```diff
@@ -716,6 +716,33 @@ export class SuperToolbar extends EventEmitter {
...
+    let available = false;
+    try {
+      const cap = this.activeEditor.doc.capabilities();
+      available = Boolean(cap.operations['create.tableOfContents']?.available);
+    } catch {
+      available = false;
+    }
+
+    tocItem.setDisabled(!available);
   }
```
luccas-harbour: every other toolbar item derives its disabled state from `HEADLESS_ITEM_MAP` + the registry's `state()` deriver, which feeds `snapshot.commands[id].disabled` through `#applyHeadlessState`. this adds a parallel one-off path that calls `capabilities()` directly and only refreshes when `updateToolbarState()` runs. could `tableOfContents` register through `HEADLESS_ITEM_MAP` instead so it stays in sync with snapshots automatically and doesn't introduce a second mechanism for the same thing?

packages/super-editor/src/editors/v1/components/toolbar/defaultItems.test.js
```diff
@@ -34,7 +34,7 @@ function buildItems(availableWidth) {
 describe('makeDefaultItems XL overflow boundary (SD-2328)', () => {
   const XL_OVERFLOW_SAFETY_BUFFER = 20;
   const XL_CUTOFF = RESPONSIVE_BREAKPOINTS.xl + XL_OVERFLOW_SAFETY_BUFFER;
-  const XL_ITEMS = ['linkedStyles', 'clearFormatting', 'copyFormat', 'ruler'];
+  const XL_ITEMS = ['linkedStyles', 'copyFormat', 'ruler'];
```
luccas-harbour: `clearFormatting` is still in `itemsToHideXL` (defaultItems.js:1079), so removing it from this expected list silently drops a boundary assertion. was that intentional? if not, adding `'clearFormatting'` back keeps the coverage.

packages/super-editor/src/editors/v1/components/toolbar/defaultItems.js
```diff
@@ -1064,7 +1077,7 @@ export const makeDefaultItems = ({
   const toolbarPadding = 32;
 
   const itemsToHideXL = ['linkedStyles', 'clearFormatting', 'copyFormat', 'ruler', 'formattingMarks'];
-  const itemsToHideSM = ['zoom', 'fontFamily', 'fontSize', 'redo'];
+  const itemsToHideSM = ['zoom', 'fontFamily', 'fontSize', 'redo', 'tableOfContents'];
```
luccas-harbour: `tableOfContents` got added to `itemsToHideSM` but the test file only covers XL overflow. worth a small SM case mirroring the XL one — otherwise a future reorder could quietly move `tableOfContents` out of overflow without a test failure.

packages/super-editor/src/editors/v1/extensions/table-of-contents/table-of-contents.js
```diff
@@ -50,37 +54,100 @@ export const TableOfContents = Node.create({
...
+          ...(prepared.rightAlignPageNumbers !== undefined
+            ? { rightAlignPageNumbers: prepared.rightAlignPageNumbers }
+            : {}),
+        })(props);
+
+        if (inserted && props.dispatch) {
+          globalThis.queueMicrotask(() => {
+            syncTocBookmarks(editor, prepared.sources);
+          });
+        }
```
luccas-harbour: the non-toolbar TOC path runs `syncTocBookmarks` synchronously after dispatch (toc-wrappers.ts:912). here the command schedules it via `queueMicrotask`. was the asymmetry intentional? running the sync from the toolbar caller after the command returns would match the existing pattern and keep the command itself a pure transaction producer.

packages/super-editor/src/editors/v1/extensions/table-of-contents/table-of-contents.js
```diff
@@ -50,37 +54,100 @@ export const TableOfContents = Node.create({
...
-            if (dispatch) {
-              tr.insert(pos, tocNode);
-            }
-            return true;
-          } catch (error) {
-            if (error instanceof RangeError) return false;
-            throw error;
-          }
-        },
+      insertTableOfContentsFromToolbar: () => (props) => {
```
luccas-harbour: the `FromToolbar` suffix bakes the caller into the API, but the behavior (insert at selection, fall back to document end) isn't toolbar-specific. if a slash menu or shortcut wants the same thing later the name will lie — drop the suffix?

luccas-harbour (changes requested):
hey @VladaHarbour!
I was reviewing this one in the browser and seem to have found some issues. 

These are the steps I took:
1. I opened the `basic/advanced-text.docx` document from our test corpus in the editor
2. Then I placed the cursor at the start of the first line and pressed enter to create a new line above it
3. I moved back up to the empty line I created
4. I clicked the new TOC button in the toolbar
Nothing happened, the TOC was not added and no javascript errors surfaced in the console.

Then, I did this:
1. I placed the cursor after the document title (the `SuperDoc Advanced Text Formatting` line)
2. I pressed enter to create a new line below the title
3. I clicked the new TOC button in the toolbar
This time the TOC was added but all page numbers were set to zero. I had to right-click and select "Update table of contents" for the numbers to show up correctly.

Also, codex found this one:

> - [P2] Disable the overflowed TOC item as well — packages/super-editor/src/editors/v1/components/toolbar/super-toolbar.js:729-729
> On widths below the small breakpoint, `tableOfContents` is moved into `overflowItems`, but this sync only searches `this.toolbarItems`. In that state, if `create.tableOfContents` is unavailable (for example a custom editor without the TOC extension), the overflow menu item remains enabled and clicking it can still route to a missing/unsupported command. Include `this.overflowItems` when locating the TOC item.

Claude found a few more things that I added as inline comments.

Can you check these? Thank you!

VladaHarbour:
Hi @luccas-harbour! I addressed your comments and did some fixes related to issues with `advanced-text.docx`. You will be able to add toc node when adding blank line on the very beginning but it will be inserted after the main title for some reason. I have a feeling it has something to do with document formatting and will try to find a reason for this(because it's not an issue for other docs I've tested). But everything else must be addressed

luccas-harbour (approved):
LGTM. Thanks for making the changes @VladaHarbour!

PR #3248, opened 2026-05-12
SD-3092 - fix: toolbar serialization for stable rendering

chittolinag:
### Issue

Passing an inline object to the `toolbar` prop caused `SuperDocTemplateBuilder` to rebuild SuperDoc on every parent re-render, looping infinitely. The reason for that is that `useMemo` compared the object dep by reference (`Object.is`), so every time the parent re-renders, this would look like the `toolbar` also changed.

### Proposed solution

Use `JSON.stringify(toolbar)` as the dep (same pattern as `stableTelemetry` in the same file).


harbournick (approved):
LGTM

PR #3249, opened 2026-05-12
fix: suppress spacing before after manual page breaks

leandrotcawork:
## Summary

This fixes a Word layout fidelity issue where the first real paragraph after an explicit manual page break could render too low in SuperDoc because paragraph `spacingBefore` was still applied at the fresh page top.

The change keeps the scope intentionally narrow:

- avoid synthesizing an empty paragraph when a ProseMirror paragraph only contributes page/column break layout blocks
- suppress `spacingBefore` only for paragraphs that start a fresh page immediately after an explicit manual page break
- preserve `pageBreakBefore` behavior, where paragraph spacing should still apply
- apply the suppression before float remeasurement so paragraph-anchored images/drawings use the actual page-top Y when computing available line widths

## Why

In Word, an explicit manual page break starts the following paragraph at the top of the new page content area. SuperDoc could retain the following paragraph's `spacingBefore`, creating a visible gap above headings or other first-page content after the break.

The issue was especially visible in a DOCX fixture with a heading immediately after a manual page break: removing the gap manually in the editor also removed the intended paragraph border/line, which pointed to paragraph spacing/layout state rather than a painter-only issue.

## Implementation Notes

- `pm-adapter` now skips fallback empty paragraph creation when the converted paragraph contains only break blocks.
- `layoutDocument` detects manual page breaks before a target paragraph, including paragraph-anchored images/drawings between the break and the paragraph.
- `layoutParagraphBlock` suppresses `spacingBefore` only when the paragraph is still at the top of a fresh page, before float scans and remeasurement.
- Page/margin-relative anchored object behavior is intentionally left unchanged in this PR.

## Reference Document

[DC_Template_Descricao_Cargo.docx](https://github.com/user-attachments/files/27646978/DC_Template_Descricao_Cargo.docx)


caio-pizzol:
@leandrotcawork!

Thanks for the PR - this is a challenging part of editor, this is why it is taking longer to review.

Back shortly with my findings

Abraços!

caio-pizzol:
Thanks @leandrotcawork for working on this, and thanks to the reporter for sharing the issue and document.

This helped us confirm there is a real spacing bug after page breaks. After reviewing it, though, we think the fix needs a broader approach than this PR. It touches several fragile parts of how SuperDoc reads and lays out documents, and we don’t want to merge a partial fix that could cause other layout issues.

We’re going to close this PR and handle the bug separately with a more complete fix and test coverage.

Thanks again for the time and effort here. The work in this PR was useful and will help guide the follow-up.

PR #3250, opened 2026-05-12
fix: align rtl date token rendering with word parity

artem-harbour:
Linear: SD-3098

**Proposal**:
Fix RTL date rendering parity with Word by propagating run-level bidi metadata (w:rtl) into DomPainter and applying targeted run-direction handling at paint time. In this approach, RTL date-like runs receive dir="rtl" plus render-time RLM separator normalization, while date-like LTR runs inside RTL paragraphs receive dir="ltr" to prevent separator drift. The proposal also includes preserving bidi boundaries in run-merge signatures and adding unit + behavior coverage with a real rtl-dates.docx fixture.


caio-pizzol (approved):
hey @artem-harbour - i pulled main in, switched this to the rtl field #3203 added (got renamed after you opened this), wired rtl into the cache keys so an edit that flips rtl doesn't reuse old DOM, and stopped pm-adapter from merging runs with different rtl.

follow-up commit adds tests plus a Hebrew + date Word fixture. both fixtures in the corpus, all three dates in `Date Being Weird.docx` render right now.

lgtm.

PR #3262, opened 2026-05-13
feat: honor PAGE/NUMPAGES field format switches (SD-2990)

luccas-harbour:
## Summary

Parses `\*` general-format and `\#` numeric-picture switches when importing
PAGE/NUMPAGES fields and threads the requested format (Roman, alphabetic,
zero-padded decimal, dashed) plus the section-aware numeric page value
through the converter → pm-adapter → layout engine → DOM painter so page
number fields render in the format Word stored — not always decimal.

The original instruction text is preserved on the node so export round-trips
back to the same field code Word wrote.

## Changes

**Converter (super-editor)**
- New shared helper `field-references/shared/page-number-field-switches.js` parses
  `\* Arabic|roman|ROMAN|alphabetic|ALPHABETIC|ArabicDash` and `\# 0000`-style
  picture switches into `{ pageNumberFormat, pageNumberZeroPadding, instruction }`.
- `page-preprocessor.js` and `num-pages-preprocessor.js` now accept an options
  object (`{ docx, instructionTokens, fieldRunRPr }`) and emit those attrs on
  `sd:autoPageNumber` / `sd:totalPageNumber`. All sibling preprocessors updated
  to the new options-object signature for consistency.
- `autoPageNumber-translator` / `totalPageNumber-translator` round-trip the
  attrs back to the original `PAGE`/`NUMPAGES` instruction on export and apply
  the format to the cached `NUMPAGES` export value.
- `PageNumber` / `TotalPageCount` extensions gain `instruction`,
  `pageNumberFormat`, `pageNumberZeroPadding` attributes (all `rendered: false`).
- `HeaderFooterPerRidLayout` resolver now also returns `displayNumber` (the
  section-aware numeric value) so non-decimal formats can be applied per page.

**Contracts**
- New `page-number-formatting.ts` centralizes `formatPageNumber` /
  `formatPageNumberFieldValue` and the `PageNumberFieldFormat` /
  `PageNumberFormat` types. `pageNumbering.ts` in layout-engine now re-exports
  from contracts to keep the public surface stable.
- `TextRun` gains `pageNumberFieldFormat?: PageNumberFieldFormat`.

**pm-adapter**
- New `getPageNumberFieldFormat()` helper reads the node attrs and attaches
  `pageNumberFieldFormat` to token runs in both `generic-token.ts` and
  `text-run.ts`.

**Layout bridge / engine**
- `resolveHeaderFooterTokens` and `resolvePageNumberTokens` apply the field
  format when present, falling back to the previous display text.
- `layoutHeaderFooter` opts out of digit-bucketing for variants whose PAGE
  tokens use formats that don't share a digit-width budget (Roman, letters,
  zero-padded decimal). `decimal` and `numberInDash` still bucket.
- Per-page `displayNumber` is plumbed through `HeaderFooterLayout` →
  `FragmentRenderContext` → DOM painter so the painter formats per-page.

**DOM painter**
- `resolveRunText` uses `pageNumberFieldFormat` + `displayPageNumber` to
  format PAGE / NUMPAGES tokens at render time when a format is set.


packages/super-editor/src/editors/v1/core/super-converter/field-references/shared/page-number-field-switches.js
```diff
@@ -0,0 +1,62 @@
+import { formatPageNumberFieldValue as formatSharedPageNumberFieldValue } from '@superdoc/contracts';
+
+const GENERAL_FORMATS = new Map([
+  ['Arabic', 'decimal'],
+  ['roman', 'lowerRoman'],
+  ['ROMAN', 'upperRoman'],
+  ['alphabetic', 'lowerLetter'],
+  ['ALPHABETIC', 'upperLetter'],
+  ['ArabicDash', 'numberInDash'],
```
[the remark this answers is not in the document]
luccas-harbour: Ok. I narrowed the fix to only case-fold numeric formats where casing does not affect rendered output: `Arabic` and `ArabicDash`. Mixed-case Roman/alphabetic switches are intentionally left unrecognized rather than guessed. Added regression coverage for both paths.

packages/super-editor/src/editors/v1/core/super-converter/field-references/preProcessNodesForFldChar.js
```diff
@@ -328,7 +328,7 @@ const _processCombinedNodesForFldChar = (nodesToCombine = [], instrText, docx, i
   const instructionPreProcessor = getInstructionPreProcessor(instructionType);
   if (instructionPreProcessor) {
     return {
-      nodes: instructionPreProcessor(nodesToCombine, instrText, docx, instructionTokens, fieldRunRPr),
+      nodes: instructionPreProcessor(nodesToCombine, instrText, { docx, instructionTokens, fieldRunRPr }),
```
tupizz, line 330: small parser gap here: this call only happens after `instructionType` is split on a literal space above. If the field code comes through as `PAGE	\* Arabic` or with a newline, we never reach the PAGE/NUMPAGES parser, so the new switch handling is skipped. Can we use the same `split(/\s+/)` treatment before dispatch? The header/footer-only scanner has the same pattern.

tupizz (approved):

packages/super-editor/src/editors/v1/core/super-converter/field-references/preProcessNodesForFldChar.js
luccas-harbour: fixed! thanks!

harbournick (approved):
LGTM

PR #3264, opened 2026-05-13
fix(header-footer): centralize OOXML ref inheritance for first-page headers (SD-2997)

luccas-harbour:
## Summary

Fixes SD-2997 (different first-page headers) by centralizing the OOXML
header/footer ref inheritance rules into a single helper and pointing the
three places that used to re-implement them at the shared implementation.
Along the way, this corrects several inheritance bugs that broke `first`
and `even` variants in multi-section documents.

## What changed

- **New `@superdoc/contracts` helper** — `resolveInheritedHeaderFooterRef`
  / `resolveInheritedHeaderFooterRefWithType` encode the OOXML resolution
  order (page refs → current section → walk back to the nearest prior
  section that defines the variant → legacy identifier) plus the rule
  that `odd` may fall back to `default` but never vice-versa when
  `w:evenAndOddHeaders` is enabled.
- **Three call sites unified** — `layout-engine/index.ts`,
  `layout-bridge/headerFooterUtils.ts`, and
  `HeaderFooterSessionManager.ts` now delegate to the shared helper
  instead of carrying three near-duplicate copies of the rules.
- **Walk back through intermediate sections** — previously the resolver
  only consulted the immediately prior section, so a `first` ref defined
  in section 0 was lost once section 1 (with only `default`) sat between
  it and a later section that also lacked an explicit `first` ref. The
  shared resolver now scans back to the nearest prior section that
  defines the requested variant.
- **Header height uses the resolved variant** — the page-margin
  calculation in `layout-engine` now uses the *resolved* variant
  (`headerResolution.variantType`) for the height lookup, so inherited
  default refs don't end up sized against a non-existent variant slot.
- **Per-section `titlePg` honored in the fallback path** —
  `HeaderFooterSessionManager` was classifying first pages with the
  document-level `titlePg` flag only; multi-section docs that override
  `titlePg` per section now consult the section's own value via the
  multi-section identifier's `sectionTitlePg` map.
- **Layout-bridge type/variant gating** — `getHeaderFooterTypeForSection`
  now returns `null` when neither the section nor its ancestors define a
  usable ref for the chosen parity (or default), so missing even-page
  headers no longer render as the odd/default ref.
- **Converter fallback refs preserved** — `@superdoc/contracts`
  inheritance helpers no longer let a later section's empty refs shadow
  an earlier section's resolved ones during fallback computation.

## Tests

- New `header-footer-inheritance.test.ts` in `contracts` covering the
  walk-back, parity, and `odd → default` fallback rules.
- Expanded `headerFooterUtils.test.ts` and `layout-engine/index.test.ts`
  to assert the resolved variant flows through to height and margin
  calculations.
- Expanded `HeaderFooterSessionManager.test.ts` to cover the
  per-section `titlePg` decoration path, including the missing-map case.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
+++ b/packages/super-editor/src/editors/v1/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
@@ -1750,9 +1750,10 @@ export class HeaderFooterSessionManager {
     const firstPageInSection = sectionFirstPageNumbers.get(sectionIndex);
     const isFirstPageOfSection = firstPageInSection === pageNumber;
 
-    // Check for alternateHeaders in converter
+    // Check for alternateHeaders in converter or the multi-section identifier.
     const converter = (this.#options.editor as EditorWithConverter).converter;
-    const hasAlternateHeaders = converter?.pageStyles?.alternateHeaders === true;
+    const hasAlternateHeaders =
+      this.#multiSectionIdentifier?.alternateHeaders === true || converter?.pageStyles?.alternateHeaders === true;
 
     // Only use 'first' variant when titlePg is enabled (w:titlePg element in OOXML).
     // Without titlePg, even the first page of a section uses 'default'.
--- a/packages/super-editor/src/editors/v1/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
+++ b/packages/super-editor/src/editors/v1/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
@@ -1736,7 +1736,10 @@ export class HeaderFooterSessionManager {
     // Without titlePg, even the first page of a section uses 'default'.
     const headerIds = converter?.headerIds as { titlePg?: boolean } | undefined;
     const footerIds = converter?.footerIds as { titlePg?: boolean } | undefined;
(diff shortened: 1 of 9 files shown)


packages/super-editor/src/editors/v1/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
```diff
@@ -1736,7 +1758,10 @@ export class HeaderFooterSessionManager {
     // Without titlePg, even the first page of a section uses 'default'.
```
tupizz, line 1759: small gap here: this now lets titlePg come from the multi-section identifier, but alternateHeaders still only comes from converter.pageStyles above. If pageStyles is stale/missing while the identifier has alternateHeaders=true, rebuildRegions falls back to default for an empty even/odd region, so clicking that region can open the default slot instead of the even/odd one. Can we derive this from this.#multiSectionIdentifier?.alternateHeaders too?

tupizz (approved):

packages/super-editor/src/editors/v1/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
luccas-harbour: fixed it!

PR #3266, opened 2026-05-13
feat: improve format painter ux to match ms word

VladaHarbour:


packages/super-editor/src/editors/v1/extensions/format-commands/format-commands.js
```diff
@@ -86,75 +92,172 @@ export const FormatCommands = Extension.create({
...
     };
   },
 
+  onSelectionUpdate({ editor }) {
+    const { storedStyle, sourceSelection } = this.storage;
+    if (!storedStyle) return;
+
+    const currentSelection = getSelectionRange(editor.state);
+    if (editor.state.selection.empty || isSameSelection(currentSelection, sourceSelection)) return;
+    if (!this.storage.releaseCleanup) editor.commands.applyStoredFormat();
```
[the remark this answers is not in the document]
artem-harbour: Review/fix the comment if applicable.

packages/super-editor/src/editors/v1/extensions/format-commands/format-commands.js
```diff
@@ -86,75 +92,172 @@ export const FormatCommands = Extension.create({
...
+  storage.releaseCleanup = null;
+}
+
+function armFormatPainterRelease({ storage, editor }) {
+  if (storage.releaseCleanup) return;
+  if (typeof document === 'undefined' || !document?.addEventListener) return;
+
+  const eventName = typeof PointerEvent === 'undefined' ? 'mouseup' : 'pointerup';
+  const handleRelease = (event) => {
+    if (event?.target?.closest?.('[data-editor-ui-surface]')) return;
```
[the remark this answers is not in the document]
artem-harbour: Review/fix the comment if applicable.

packages/super-editor/src/editors/v1/extensions/format-commands/format-commands.js
```diff
@@ -86,75 +92,172 @@ export const FormatCommands = Extension.create({
...
+      const markToApply = processedMarks.find((mark) => mark.name === key);
+      const hasEmptyAttrs = markToApply?.attrs && markToApply?.attrs[key];
+
+      let cmd = {};
+      if (!markToApply && !hasEmptyAttrs) cmd = { command: unsetCommand, argument: defaultParam };
+      else cmd = { command: setCommand, argument: markToApply.attrs[key] || defaultParam };
+      result = result[cmd.command](cmd.argument);
+    });
+
+    if (!shouldStayActive) clearFormatPainterStorage(storage);
```
artem-harbour: Double-clear bug in applyStoredFormat

clearFormatPainterStorage is called both inside the try block and unconditionally in finally. The try-body calls are dead code — finally always fires on return. This means the event listener is removed twice (a minor waste) and the logic is confusing to read.

```
// try block:
if (!shouldStayActive) clearFormatPainterStorage(storage); // ← redundant
return result;
// finally block:
if (!shouldStayActive) clearFormatPainterStorage(storage); // ← the only one needed
```

Fix: remove the clearFormatPainterStorage calls from inside the try body and keep only the finally.

artem-harbour (changes requested):
Added my comments.

packages/super-editor/src/editors/v1/extensions/format-commands/format-commands.js
VladaHarbour: removed redundant line

packages/super-editor/src/editors/v1/components/toolbar/super-toolbar.js
```diff
@@ -632,6 +632,11 @@ export class SuperToolbar extends EventEmitter {
         if (commandState?.value != null) item.activate({ styleId: commandState.value });
         else item.label.value = this.config.texts?.formatText || 'Format text';
       },
+      copyFormat: () => {
```
artem-harbour: For `copy-format` item, in the headless state, the `active` state is not calculated, but only a `disabled` state.

`packages/super-editor/src/headless-toolbar/toolbar-registry.ts`

```javascript
'copy-format': {
      id: 'copy-format',
      directCommandName: 'copyFormat',
      state: createDisabledStateDeriver(),
},
```

Need to create `createCopyFormatStateDeriver` in the headless toolbar and calculate for it not only the `disabled` state but also the `active` state.

Based on this check.
`const hasStoredFormat = Boolean(this.activeEditor?.storage?.formatCommands?.storedStyle);`

packages/super-editor/src/editors/v1/components/toolbar/super-toolbar.js
```diff
@@ -632,6 +632,11 @@ export class SuperToolbar extends EventEmitter {
...
+        const hasStoredFormat = Boolean(this.activeEditor?.storage?.formatCommands?.storedStyle);
+        if (hasStoredFormat || commandState?.active) item.activate();
```
artem-harbour: `commandState?.active` - I think this is dead code.

artem-harbour (changes requested):
Added my comments.

packages/super-editor/src/editors/v1/components/toolbar/super-toolbar.js
VladaHarbour: used proposed approach

packages/super-editor/src/editors/v1/components/toolbar/super-toolbar.js
VladaHarbour: removed the whole block because of approach in previous comment

artem-harbour (approved):
LGTM

PR #3268, opened 2026-05-13
SD-3116 - bug: SVG image doesn't render on imported files

chittolinag:
### Issue

SVG images embedded in the document are not rendered when loading a document in SuperDoc. The root cause is: browsers expect the MIME type of SVG images to be `svg+xml`, but we were rendering them as `svg` only. The browser then refused to render it.

### Proposed solution

Like we do with other media types, we simply need to map `svg` to `svg+xml` before converting it to base64.


artem-harbour (approved):
LGTM

PR #3269, opened 2026-05-13
refactor(painters/dom): drop list-item fragment renderer (SD-2851)

luccas-harbour:
## Summary

Removes the now-unreachable `list-item` fragment path from the DOM painter. List rendering already flows through the paragraph fragment path (the PM adapter no longer registers `orderedList` / `bulletList` handlers — list paragraphs go through `paragraphToFlowBlocks()` as `ParagraphBlock` with marker metadata), so the dedicated `list-item` renderer and its helpers were dead weight duplicating paragraph behavior.

## What changed

In `packages/layout-engine/painters/dom/src/renderer.ts`:

- Removed `renderListItemFragment` and `applyResolvedListItemWrapperFrame`.
- Removed the `LIST_MARKER_GAP` constant and the `stripListIndent` helper.
- Removed `list-item` branches from `fragmentKey`, `deriveBlockVersion`, and the fragment-frame dispatch — `list-item` is now treated as unsupported (throws in `fragmentKey` and `renderFragment`).
- Simplified `computeBetweenBorderFlags` in `features/paragraph-borders/group-analysis.ts` to pair `para` fragments only.

Tests:

- Deleted `renderer-known-divergences.test.ts` (the file existed to track list-item parity bugs; obsolete now).
- Removed list-item-specific assertions and fixtures from `between-borders.test.ts`, `renderer-dispatch.test.ts`, and `index.test.ts`.
- Updated the doc comment in `test-utils/normalize-line.ts` to reflect the remaining contexts (body, table-cell).

Body↔table-cell justify `wordSpacing` parity that incidentally lived in the deleted divergence file is still covered by `renderer-parity-contracts.test.ts:310-387`. Lists rendered through the paragraph path are exercised by `renderer-marker-suffix.test.ts`, `renderer-marker-textwidth.test.ts`, and the multi-line list integration test at `index.test.ts:1266`.

## What's intentionally NOT in this PR

The `ListBlock` / `ListItemFragment` types and their producers in `contracts/`, `layout-bridge/`, `layout-resolved/`, and `pm-adapter/sdt/` are still present. They're now dead feeds (the painter rejects what they produce), but unwinding them touches multiple packages and is better as a follow-up so this PR stays focused on the painter cleanup.

## Follow-ups

- Drop `ListBlock`, `ListMeasure`, and `ListItemFragment` from `packages/layout-engine/contracts/src/index.ts` and propagate the removal through `layout-bridge` (`cache.ts`, `incrementalLayout.ts`), `layout-resolved` (`resolveLayout.ts`, `versionSignature.ts`), and `pm-adapter/sdt/`.
- Continue with the next P0 items from `RENDERING_DUPLICATION_AUDIT.md`: extracting a shared paragraph content renderer that body and table-cell paths both consume.


luccas-harbour:
> ### 💡 Codex Review
> https://github.com/superdoc-dev/superdoc/blob/a156110f992ddde9637fcbe6f48294b9572b18da/packages/layout-engine/painters/dom/src/renderer.ts#L3014
> 
> 
> **P1 Badge Preserve list-item dispatch until all producers are removed**
> `renderFragment` now throws for any `list-item` fragment, but the list-item model is still produced elsewhere in the pipeline (`layout-bridge/src/incrementalLayout.ts` and `layout-resolved/src/resolveLayout.ts` still handle `kind: 'list-item'`, and `ListItemFragment` remains in `contracts`). That means any remaining caller/flow that emits legacy list fragments will hard-fail painting at runtime instead of degrading gracefully, which is a behavior regression introduced by this commit.
> 
> ℹ️ About Codex in GitHub

The `pm-adapter` does not produce `ListBlocks`, this has been the case ever since the unification of lists and paragraphs. A follow-up ticket should remove this from other layers of the layout-engine as well. Visual tests were run against all documents in the test-corpus and no changes were detected.

VladaHarbour:
Hi @luccas-harbour! Just a couple of comments here
In `buildFootnoteRanges` we still push `kind: 'list-item' `fragments when a footnote range is a list item. DomPainter no longer renders that kind. Please check if it has to be migrated to use paragraph fragments instead

VladaHarbour:
> ### 💡 Codex Review
> https://github.com/superdoc-dev/superdoc/blob/a156110f992ddde9637fcbe6f48294b9572b18da/packages/layout-engine/painters/dom/src/renderer.ts#L3014
> 
> 
> **P1 Badge Preserve list-item dispatch until all producers are removed**
> `renderFragment` now throws for any `list-item` fragment, but the list-item model is still produced elsewhere in the pipeline (`layout-bridge/src/incrementalLayout.ts` and `layout-resolved/src/resolveLayout.ts` still handle `kind: 'list-item'`, and `ListItemFragment` remains in `contracts`). That means any remaining caller/flow that emits legacy list fragments will hard-fail painting at runtime instead of degrading gracefully, which is a behavior regression introduced by this commit.
> 
> ℹ️ About Codex in GitHub

+1 here.
`resolveLayout` still fully handles list-item fragments, but the painter rejects them.
`Fragment` still includes `ListItemFragment,` and `ResolvedFragmentItem` still documents/exposes list-item block and measure data (packages/layout-engine/contracts/src/index.ts:2011, packages/layout-engine/contracts/src/resolved-layout.ts:159).

packages/layout-engine/painters/dom/src/renderer-known-divergences.test.ts
```diff
@@ -1,317 +0,0 @@
-/**
```
VladaHarbour, line 1: It seems that with removing divergence lock tests we lose explicit guards for list justify/alignment. Worse double checking

VladaHarbour (changes requested):

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -7510,29 +7312,28 @@ const computeSdtBoundaries = (
...
-    // This is critical for mid-row splitting where multiple fragments can exist for the same table
-    const partialKey = fragment.partialRow
-      ? `:${fragment.partialRow.fromLineByCell.join(',')}-${fragment.partialRow.toLineByCell.join(',')}`
-      : '';
-    return `table:${fragment.blockId}:${fragment.fromRow}:${fragment.toRow}${partialKey}`;
+  switch (fragment.kind) {
+    case 'para':
+      return `para:${fragment.blockId}:${fragment.fromLine}:${fragment.toLine}`;
+    case 'list-item':
+      throw new Error(`DomPainter: unsupported fragment kind ${fragment.kind}`);
```
tupizz, line 4051: This makes `list-item` fragments a hard paint error, but the rest of the layout pipeline still emits and resolves them. `layoutDocument()` still produces `kind: 'list-item'` for `ListBlock` inputs, and `layout-resolved` still lifts list-item block/measure data, so any document path that reaches DomPainter with a list block can now throw instead of rendering. Please either keep list-item painting until the upstream layout/resolved contract stops emitting it, or convert/list-normalize upstream in the same PR so DomPainter never sees this fragment kind.

packages/layout-engine/painters/dom/src/paragraph/borders/group-analysis.ts
```diff
@@ -79,7 +79,7 @@ export const computeBetweenBorderFlags = (
     const resolvedCur = resolvedItems[i];
     if (resolvedCur.kind !== 'fragment') continue;
     const frag = resolvedCur.fragment;
-    if (frag.kind !== 'para' && frag.kind !== 'list-item') continue;
+    if (frag.kind !== 'para') continue;
```
tupizz, line 82: This now skips `list-item` fragments when computing between-border groups, but `layout-resolved` still attaches `paragraphBorders` to resolved list-item fragments. Even if the renderer keeps/normalizes list-item painting, consecutive bordered list paragraphs (or paragraph/list boundaries) will no longer get `showBetweenBorder`/`suppressTopBorder`, so Word-style continuous paragraph borders can split or double-render around lists. Please keep list-item participation here until the resolved layout contract stops producing list-item fragments, and restore coverage for para/list and list/list border grouping.

luccas-harbour:
hey @VladaHarbour @tupizz! thanks for the review!

I just wanted to flag one thing about your comments on how there are issues with removing support to `list-item` fragments as other layers still have support for it. The thing to keep in mind is that no `list-item` fragment (or `ListBlock`) will ever be created and fed into the other layers because we don't have a dedicated `ListItem` node in ProseMirror anymore. All lists are paragraphs.

This has actually already been the case ever since the layout-engine was launched. The reason that that code was still there is because I was woking on unifying paragraphs and lists at the same time that Nick was building the layout engine. I finished my work and it got merged to main but the layout-engine still kept the code that was handling the old ListItem node (even though it didn't exist anymore).

This refactor project was only focused on the painter layer so I didn't actually remove the `list-item`/`ListBlock` related code from other layers. Because of that the LLMs always flag this as a big problem but we know that that code is not reachable so it really isn't a concern.

I think a follow-up ticket would be removing `list-item`/`ListBlock` support from all layers. I just didn't do it here to keep the PR focused on the painter.

Can you re-review after having this information?
Thank you!

tupizz (approved):
LGTM

VladaHarbour (approved):

tupizz (approved):
LGTM after re-review.

tupizz (approved):
lgtm

harbournick (approved):
LGTM

PR #3276, opened 2026-05-14
feat(superdoc): scrollToHeading API + fix scrollToComment / scrollToElement in flow layout

bjohas:
## Summary

Three related changes to the public scroll-to-X surface, motivated by an external integration (Electron-based DOCX viewer) where comment-panel click → editor scroll was silently broken in flow layout and unreliable in paginated layout.

1. **`fix(super-editor)`** — `Editor.getElementAtPos` in flow / web layout now resolves `view.domAtPos`'s parent-element + offset to the actual child element, instead of returning the parent. Previously, every position in the doc collapsed to "the entire editor root," so `scrollIntoView` always landed at the top.

2. **`fix(superdoc)`** — `SuperDoc.scrollToComment(id)` now delegates to `scrollToElement(id)`. The old `root.querySelector('[data-comment-ids*="..."]')` lookup missed virtualised pages in paginated mode and missed marks for which SuperDoc hadn't emitted a visible highlight span.

3. **`fix(superdoc) + feat(superdoc)`** — `SuperDoc.scrollToElement(id)` now falls back to a body-editor path when no presentation editor exists (flow layout) or when presentation returns false. Walks the PM doc via `commands.setCursorById` for comment / tracked-change marks, with an attr-match fallback for block IDs.

4. **`feat(superdoc)`** — new `SuperDoc.scrollToHeading(level, ordinal, options)` public API. Walks the doc counting paragraphs whose `paragraphProperties.styleId = "HeadingN"` (or `heading` nodes with a `level` attr), targets a position **inside** the heading's text content (not the doc-level boundary before the paragraph — the layout-fragment index doesn't cover boundaries), and dispatches through the same flow/paginated split as `scrollToElement`.

## Motivation

In an external app embedding SuperDoc 1.31.2:

* `superdoc.scrollToComment(id)` returned `true` but didn't actually scroll in many cases (flow + paginated mismatches between marks and highlight spans).
* `superdoc.scrollToElement(id)` returned `false` for every comment in flow layout (no presentation editor → unconditional bail-out).
* No clean API for "scroll to the Nth heading at level X" — every consumer ended up hand-rolling a PM walk + dual-mode scroll logic.

With this PR, against a small commented test docx:

| | Before | After |
|---|---:|---:|
| `scrollToComment` flow | 0/4 scrolled | **4/4** |
| `scrollToComment` paginated | 3/4 | 3/4 (1 residual upstream `scrollToPositionAsync` edge case) |
| `scrollToHeading(level, ordinal)` flow | n/a | **9/9** |
| `scrollToHeading(level, ordinal)` paginated | n/a | **9/9** |

The one residual paginated case (a comment whose marks land at pos=13240/42/47) is `scrollToPositionAsync` returning `true` without moving the scroll — out of scope for this PR; happy to file separately.

## Test plan

- [x] `pnpm --filter superdoc test src/core/SuperDoc.test.js` — 84/84 pass, including 4 new cases covering the new delegation, body-editor fallback, and heading walk.
- [x] `pnpm --filter @superdoc/super-editor test src/editors/v1/core/Editor` — 160/160 pass (no regression around `getElementAtPos`).
- [x] `pnpm --filter @superdoc/super-editor test src/editors/v1/core/presentation-editor` — 1369/1369 pass.
- [x] End-to-end Playwright probe in an external Electron consumer of the built `dist/` against the 4-comment / 41-heading test docx. Numbers above.

## Notes

* No type-test additions yet — happy to add to `tests/consumer-typecheck/` per CONTRIBUTING if you'd prefer the new `scrollToHeading` signature pinned there before merge.
* The `Editor.getElementAtPos` change is the most cross-cutting — only touches the body-editor branch (presentation-editor implementation is unchanged) and internal callers all use `.closest(...)` to climb to the class they want, which still works with the more specific child element this now returns.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/Editor.ts
+++ b/packages/super-editor/src/editors/v1/core/Editor.ts
@@ -1718,11 +1718,30 @@ export class Editor extends EventEmitter<EditorEventMap> {
     const clampedPos = Math.max(0, Math.min(pos, maxPos));
 
     try {
-      const { node } = this.view.domAtPos(clampedPos);
-      if (node && node.nodeType === 1) {
-        return node as HTMLElement;
+      // ProseMirror's domAtPos returns either:
+      //   - { node: <text>, offset: <chars into text> }, or
+      //   - { node: <element>, offset: <child index> } when the position is
+      //     between block children.
+      // The previous version returned the parent in the second case, which
+      // for the editor root means the entire document — scrolling that into
+      // view always lands at the top. Resolve to the actual child element
+      // when the returned node is an element parent.
+      const { node, offset } = this.view.domAtPos(clampedPos);
+      if (!node) return null;
+
+      if (node.nodeType === 1) {
+        const parent = node as Element;
+        if (parent.childNodes?.length) {
+          const idx = Math.min(Math.max(0, offset), parent.childNodes.length - 1);
+          const child = parent.childNodes[idx];
+          if (child) {
+            if (child.nodeType === 1) return child as HTMLElement;
+            if (child.nodeType === 3) return (child as Node).parentElement;
+          }
+        }
+        return parent as HTMLElement;
       }
-      if (node && node.nodeType === 3) {
+      if (node.nodeType === 3) {
         return node.parentElement;
       }
       return node?.parentElement ?? null;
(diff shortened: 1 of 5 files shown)


caio-pizzol:
Thanks for the contribution! We tested comment and heading navigation in V2, and both work using the new APIs: `ui.comments.scrollTo()` and `ui.viewport.scrollIntoView()`.

We’re focusing on V2, so we’re closing this V1 PR as superseded. Please follow our [migration guide](https://docs.superdoc.dev/editor/migrate-from-v1/overview/) to update your integration—the old methods aren’t fixed by upgrading alone. If you still hit a problem with the V2 APIs, please share a small example.

PR #3304, opened 2026-05-14
fix(super-editor): restore toolbar dropdown keyboard focus

sergiogomes:
## Summary

- Opens toolbar dropdowns with Enter and Space.
- Restores focus to the dropdown trigger when Escape closes the menu.
- Adds unit coverage for dropdown keyboard activation and Escape focus restoration.

## Test plan

- [x] Targeted toolbar tests:

```bash
pnpm exec vitest run packages/super-editor/src/editors/v1/components/toolbar/ButtonGroup.test.js packages/super-editor/src/editors/v1/components/toolbar/ToolbarDropdown.test.js
```

- [x] `pnpm run format:check`
- [x] `pnpm run lint`
- [x] Manual: open font dropdown, navigate with arrow keys, press Escape, confirm focus returns to the trigger.

Note: local commands printed existing warnings outside the toolbar files touched in this PR.

image


sergiogomes:
Rebased onto the latest `main` and resolved the `ButtonGroup.test.js` conflict by keeping both the upstream argument-forwarding tests and this PR’s dropdown keyboard activation coverage.

packages/super-editor/src/editors/v1/components/toolbar/ToolbarDropdown.vue
```diff
@@ -249,6 +249,46 @@ const handlePointerDown = (event) => {
...
+
+  const fallback = trigger.matches?.(TRIGGER_FOCUS_SELECTOR) ? trigger : trigger.querySelector(TRIGGER_FOCUS_SELECTOR);
+
+  if (fallback instanceof HTMLElement) {
+    fallback.focus();
+    return;
+  }
+
+  trigger.setAttribute('tabindex', '-1');
+  trigger.focus();
```
caio-pizzol, line 289: this branch never runs for any real dropdown (checked every one in the dev app). worth a one-line comment saying it's there to guard against trigger slots with no focusable child inside.

caio-pizzol (approved):
@sergiogomes after Escape, focus lands on the inner button, but Enter on it was being stopped before the toolbar could see it, so the dropdown couldn't reopen by keyboard.

i pushed a follow-up commit (`c68770b3c`): a new `allowEnterPropagation` prop lets Enter bubble from the dropdown branch only, plus tests for Escape-then-Enter and the Bullet/Numbered list case.

two PR-description notes: Bullet/Numbered list main keeps running its main action on Enter (menu does not open), and `preventDefault` now only fires for keys you actually handle.

lgtm once CI is green.

packages/super-editor/src/editors/v1/components/toolbar/ToolbarDropdown.vue
sergiogomes: @caio-pizzol thanks for the review, detailed feedback, and follow-up improvements 👊🏻

PR #3305, opened 2026-05-14
fix(comments): track change edits in body get matched in comments bubble

palmer-cl:
- Before when you clicked inside a tracked change and typed more content, the comment bubble text would not update
- After: comment bubble stays in sync with body text


luccas-harbour (changes requested):
hi @palmer-cl!
I tested this in the browser and seem to have found an issue. The problem is happening when I added text to a track change. As I am typing, a new letter "e" is appended to the end of the run. Here's a video showing that: https://www.loom.com/share/f7051ffbf4c3413fae367a2397dce463
In the test I am using the `basic/advanced-text.docx` document from our test corpus.

Also, a codex review found this issue:
> - [P2] Handle touched tracked changes that disappear — packages/superdoc/src/SuperDoc.vue:1202-1205
>  When a normal edit removes all content carrying an existing tracked-change mark, such as backspacing over the entire text of your own tracked insertion, this scan only inspects `transaction.doc` after the change. At that point there is no mark left for `nodesBetween` to find, so the old change id is never queued for refresh/resolve and the comments sidebar can keep showing a stale tracked-change bubble. Capture the touched id from the pre-change range or resolve the matching comment when a queued id has no remaining marks.

Can you please check these? Thank you!

packages/superdoc/src/stores/comments-store.js
```diff
@@ -17,6 +17,14 @@ import useComment from '@superdoc/components/CommentsLayer/use-comment';
 import { groupChanges } from '../helpers/group-changes.js';
 import { buildFloatingCommentInstances } from './helpers/floating-comment-instances.js';
 
+const isShallowEqual = (a, b) => {
+  if (Object.is(a, b)) return true;
+  if (!a || !b || typeof a !== 'object' || typeof b !== 'object') return false;
+  const keysA = Object.keys(a);
+  const keysB = Object.keys(b);
+  return keysA.length === keysB.length && keysA.every((key) => Object.is(a[key], b[key]));
+};
```
luccas-harbour: shallowEqual already exists at packages/super-editor/src/ui/equality.ts:12, is re-exported from @superdoc/super-editor, and even has a public sub-entry at packages/superdoc/src/ui.js:13. the published one handles arrays explicitly and uses hasOwnProperty. mind swapping to import { shallowEqual } from '@superdoc/super-editor'; and dropping the local copy?

palmer-cl:
@luccas-harbour do you have any ideas why I wouldnt be able to repro? different envs maybe? Im using the same doc. 
Screen Cast 2026-05-14 at 8 59 46 PM

luccas-harbour:
> @luccas-harbour do you have any ideas why I wouldnt be able to repro? different envs maybe? Im using the same doc. 

You know what, my bad @palmer-cl. I started suspecting a browser extension was getting in the way. I disabled it and I can't reproduce the problem anymore. Sorry for that and thanks for making the other changes!

luccas-harbour (approved):

PR #3309, opened 2026-05-14
refactor(painters/dom): unify paragraph rendering across body and table cells (SD-2838)

luccas-harbour:
## Summary

Collapses two parallel paragraph-rendering implementations (body fragments in `renderer.ts` and table cells in `renderTableCell.ts`) into a single shared module, then reorganizes the painter so paragraph and run rendering live in dedicated directories instead of being interleaved with ~3k lines of `renderer.ts`.

### 1. Unified paragraph pipeline (`paragraph/`)

New `paragraph/renderParagraphContent.ts` owns the full per-paragraph pipeline — block styles, decoration layers, list markers, indentation, line walking, and drop caps — and is consumed by both the body-fragment path in `renderer.ts` and the table-cell path in `renderTableCell.ts`. The two sites had drifted into parallel ~500-line implementations with subtly different marker/indent/border behavior; one module makes them identical by construction.

Supporting modules:
- `paragraph/indentation.ts` — line indent and available-width resolution
- `paragraph/list-marker.ts` — list marker rendering (absorbs the deleted `utils/marker-helpers.ts`)
- `paragraph/styles.ts` — block-level paragraph styles
- `paragraph/borders/` — moved from `features/paragraph-borders/` so all paragraph pieces live together; feature registry updated
- `paragraph/renderParagraphContent.test.ts` — new coverage for the resolved-paragraph path

Table-cell side effect: borders and shading now use the same `.superdoc-paragraph-border` / `.superdoc-paragraph-shading` layers as the body path instead of being stamped onto the cell-paragraph wrapper. `renderTableCell.test.ts` was updated to assert against those layers.

### 2. Run pipeline extracted into `runs/`

Pulled all line- and run-level rendering (~2500 lines) out of `renderer.ts` into `runs/`, split by concern:

- Per-run-kind: `text-run`, `image-run`, `math-run`, `tab-run`, `field-annotation-run`
- Cross-cutting: `links`, `sdt`, `tracked-changes`, `formatting-marks`
- Orchestrators: `render-line`, `render-run`
- `hash` — run hashing helpers moved out of `paragraph-hash-utils.ts`
- `types`, `index`

Public re-exports (`RenderedLineInfo`, `sanitizeUrl`, `linkMetrics`, `applyRunDataAttributes`) now resolve through `runs/index.ts`. No behavior change in this step — pure colocation in preparation for the body/table unification.

### 3. Shared list-marker types

`shared/common/list-marker-utils.ts`:
- `MinimalMarkerRun` gains `color`, `letterSpacing`, `vanish` so callers don't redeclare a parallel marker-run type
- `MinimalMarker.justification` / `suffix` tightened from `string` to literal unions
- `MinimalWordLayout` gains `indentLeftPx`

### 4. Bug fixes folded in along the way

- `fix(painters/dom): keep list marker right indent` — right-indent was being dropped on the unified path
- `fix(painters/dom): keep split paragraph fragment height` — split fragments lost their measured height
- `fix(painters/dom): mark remeasured paragraph final lines` — final-line marking was missed after remeasure


packages/layout-engine/painters/dom/src/paragraph/renderParagraphContent.test.ts
```diff
@@ -0,0 +1,252 @@
...
+      frameEl,
+      block,
+      measure,
+      containerKind: 'body-fragment',
+      width: 200,
+      localStartLine: 0,
+      localEndLine: 1,
+      lineIndexOffset: 0,
+      linesOverride: measure.lines.slice(0, 1),
+      contextSection: 'body',
```
tupizz: `RenderParagraphContentParams` does not declare `contextSection`, but this new test passes it here and repeats it below. Since the painter-dom tsconfig includes `src/**/*.ts`, these object literals are part of the package type-check and should fail excess-property checking once workspace imports resolve. Please remove `contextSection` from these calls, or add it to the typed params if it is intended to be part of the rendering API.

packages/layout-engine/painters/dom/src/paragraph/list-marker.ts
```diff
@@ -0,0 +1,299 @@
...
+import type { ParagraphMeasure, ResolvedListMarkerItem, SourceAnchor } from '@superdoc/contracts';
+import {
+  computeTabWidth,
+  resolveListMarkerGeometry,
+  resolveListTextStartPx,
+  type MinimalMarker,
+  type MinimalWordLayout,
+  type ResolvedListMarkerGeometry,
+} from '@superdoc/common/list-marker-utils';
+import { applySourceAnchorDataset } from '../renderer.js';
```
tupizz: This import creates a runtime cycle through the new extraction: `renderer.ts` imports `renderParagraphFragment`, which reaches `renderParagraphContent`, which imports this module, which then imports back from `renderer.ts`. It also makes the paragraph helper depend on the renderer orchestration module for a small DOM utility. Please move `applySourceAnchorDataset` into a shared utility module (or pass it in) so the extracted paragraph/list-marker code remains leaf-level and avoids circular initialization risk.

tupizz (changes requested):
Re-reviewed current head. The RenderParagraphContentParams test type issue remains, and I added an inline comment for a new renderer -> paragraph helper circular import.

packages/layout-engine/painters/dom/src/paragraph/renderParagraphContent.test.ts
luccas-harbour: Fixed 👍

packages/layout-engine/painters/dom/src/paragraph/list-marker.ts
luccas-harbour: That makes sense. I fixed it.

tupizz (approved):
lgtm

VladaHarbour (approved):

PR #3312, opened 2026-05-14
fix(super-converter): fix table-cell shading export — strip # prefix and emit val="clear" (SD-3142)

nazlo90:
Partial fix for SD-3142. Addresses the export half of the bug — the import half is not fixed in this PR (details below).

## Root cause

ECMA-376 §17.4.32 requires `w:fill` to be a bare 6-char uppercase hex string and, for solid fills, `w:val="clear"`. SuperDoc was writing the color from the internal ProseMirror attr as-is, which could include a leading `#` (e.g. `#A1B2C3`), and was omitting `w:val` entirely. LibreOffice treats a missing `w:val` as `nil` (transparent) and ignores the fill value, rendering those cells black regardless of color.

## What changed

**`translate-table-cell.js`** — the export path for `<w:shd>`:
- Runs `normalizeHexColor` to strip the `#` prefix and uppercase the value before writing `w:fill`.
- Adds `val: 'clear'` to every solid-fill shading element.
- Guards against non-hex values (e.g. `"auto"`) via `isValidHexColor`: these now produce no `<w:shd>` rather than invalid markup.

**`helpers.js`** (`resolveShadingFillColor`) — used on the import/display path:
- Adds an `isValidHexColor` guard so `w:fill="auto"` (Word's sentinel for "no fill") returns `null` instead of the string `"AUTO"`, preventing it from being treated as a color downstream.

## Tests

11 new assertions in `helpers.test.js` and `translate-table-cell.test.js`. Also corrected an existing wrong assertion that expected `w:fill` as `'#FF00FF'`; OOXML requires `'FF00FF'`. All 12 827 existing tests pass.

```
pnpm --filter @superdoc/super-editor exec vitest run \
  src/editors/v1/core/super-converter/helpers.test.js \
  src/editors/v1/core/super-converter/v3/handlers/w/tc/helpers/translate-table-cell.test.js
```

## What is NOT fixed

The import side of SD-3142: cells whose background comes from a table style's conditional formatting (e.g. `tblStylePr` with `type="firstRow"`) currently render as white in SuperDoc's layout engine. The style cascade in `style-engine` resolves the conditional-format shading, but that value isn't threaded through to the painted cell yet — it requires changes in `style-engine` and `pm-adapter` and is a separate, larger task.

Related: SD-3142.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/tc/helpers/translate-table-cell.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/tc/helpers/translate-table-cell.js
@@ -1,4 +1,4 @@
-import { pixelsToTwips, inchesToTwips, twipsToPixels } from '@converter/helpers';
+import { pixelsToTwips, inchesToTwips, twipsToPixels, normalizeHexColor, isValidHexColor } from '@converter/helpers';
 import { translateChildNodes } from '@converter/v2/exporter/helpers/index';
 import { translator as tcPrTranslator } from '../../tcPr';
 import {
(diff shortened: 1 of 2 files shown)


caio-pizzol:
Thanks for this, @nazlo90. I refreshed the branch against current `main` and trimmed it to the remaining export fix.

The import-side shading fix is already on `main`, so this now only keeps the table-cell export change: write a bare hex `w:fill` and add `w:val="clear"`.

caio-pizzol:
Thanks for the fix, @nazlo90, and sorry for the delay. We verified that table-cell shading exports correctly and survives reopening in V2, so we’re closing this V1-specific PR.

For upgrading, see the [V2 migration guide](https://docs.superdoc.dev/editor/migrate-from-v1/overview/).

[v2-cell-shading.webm](https://github.com/user-attachments/assets/6cf969ce-fcae-4f9b-9efa-49d5d3654451)

PR #3319, opened 2026-05-15
fix: upload large images/resize images in headers/footers

VladaHarbour:


packages/super-editor/src/editors/v1/core/parts/adapters/relationships-mutation.ts
```diff
@@ -52,13 +57,55 @@ function getMaxIdInt(elements: RelElement[]): number {
...
+    rel.attributes.TargetMode = 'External';
+  }
+
+  return rel;
+}
+
+function createRelationshipsPart(elements: RelElement[] = []): RelsXml {
+  return {
+    type: 'element',
+    name: 'document',
```
artem-harbour: The name: 'document' wrapper is wrong. In OPC, <Relationships> is the document root — there is no outer element. If the serializer uses this name, the emitted .rels file will be <document><Relationships>…</Relationships></document>, which is invalid OPC and will break on re-import.

packages/super-editor/src/editors/v1/extensions/image/imageHelpers/startImageUpload.js
```diff
@@ -136,6 +165,21 @@ export async function uploadAndInsertImage({ editor, view, file, size, id }) {
 }
 
 export function addImageRelationship({ editor, path }) {
+  if (editor.options.isHeaderOrFooter) {
+    const parentEditor = editor.options.parentEditor;
+    const headerFooterRefId = editor.options.headerFooterRefId || editor.options.documentId;
```
artem-harbour: documentId is a document GUID, not a Word relationship ID (rId3). resolvePartIdFromRefId will never match it and returns null, causing the relationship write to silently fail. Replace with an explicit guard and a console.warn.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/decode-image-node-helpers.js
```diff
@@ -500,13 +505,26 @@ function addImageRelationshipForId(params, id, imagePath) {
...
+function getDocumentRelationships(params) {
+  const docx = params.converter?.convertedXml || {};
+  const rels = docx['word/_rels/document.xml.rels'];
+  return rels?.elements?.find((el) => el.name === 'Relationships')?.elements ?? [];
+}
+
+function findImageRelationship(relationships = [], { id, target }) {
+  return relationships.find((rel) => {
+    if (rel?.attributes?.Type !== IMAGE_REL_TYPE) return false;
+    return rel.attributes.Id === id || rel.attributes.Target === target;
```
artem-harbour: If Id matches one relationship and Target matches a different one, the wrong entry is returned and its rId gets reused. Match by id first; fall back to target only when id is absent.

artem-harbour (changes requested):
Added some comments.

packages/super-editor/src/editors/v1/core/parts/adapters/relationships-mutation.ts
VladaHarbour: It's historical parts-system convention taken from relsPartDescriptor. Mutation and export process will skip this wrapper and take internal relationship element so I will skip it in createRelationshipsPart as well to respect ooxml

packages/super-editor/src/editors/v1/extensions/image/imageHelpers/startImageUpload.js
VladaHarbour: I will remove this because it's definitely confusing but I want to mention that we pass headerFooterRefId as documentId when create header/footer editor:

`const headerFooterEditor = createStoryEditor(editor, data, {
    documentId: headerFooterRefId || 'headerFooterRefId',
    isHeaderOrFooter: true,
    currentPageNumber,
    totalPageCount,
    element: editorContainer,
    editorOptions: {
      headerFooterRefId,
      headerFooterType: type,
      onCreate: (evt) => setEditorToolbar(evt, editor),
      onBlur: (evt) => onHeaderFooterDataUpdate(evt, editor, headerFooterRefId, type),
    },
  });`

artem-harbour (approved):
LGTM

PR #3322, opened 2026-05-15
refactor(painters/dom): unify SDT container rendering (SD-2838)

luccas-harbour:
## Summary

Consolidates SDT (Structured Document Tag) container rendering across paragraphs, tables, and nested tables into a single shared module, then fixes a long tail of nested/split-fragment chrome bugs that the unified pipeline made tractable.

Previously, the duplicate-suppression contract was implicit: cells were handed a raw `tableSdt` reference and each renderer reimplemented its own metadata comparisons. With nested tables, idless SDT controls, and split fragments, this produced missing borders, duplicated chrome, clipped labels, and chrome that disappeared across page splits. The new pipeline uses a key-based ancestor chain and an upward chrome-rendered callback, so every renderer (paragraph, table fragment, table cell, embedded table) makes the same decisions from the same data.

## Changes

**New shared module — `painters/dom/src/sdt/container.ts`** (replaces `utils/sdt-helpers.ts`)
- Renames `applySdtContainerStyling` → `applySdtContainerChrome`
- Adds `shouldRenderSdtContainerChrome`, `getSdtContainerKeyForBlock`, `getSdtSiblingBoundaries`
- Adds `SdtAncestorOptions` type — `ancestorContainerKey(s)` + `ancestorContainerSdt(s)` chains for explicit suppression

**Contracts (`contracts/src/sdt-container.ts`)** — new file
- `getSdtContainerKey`, `getSdtContainerMetadata`, `hasExplicitSdtContainerKey`, `isSdtContainerMetadata`
- `layout-resolved` drops the local `sdtContainerKey.ts` and imports from contracts (single source of truth)
- `idlessSdtContainerKeys` WeakMap groups id-less sibling controls by identity, with an `AIDEV-NOTE` explaining why alias/title matching is unsafe

**Renderer threading**
- `renderTableFragment`, `renderTableRow`, `renderTableCell`, `renderEmbeddedTable`, and `renderParagraphContent` all accept the `ancestorContainerKey(s)`/`ancestorContainerSdt(s)` chain and an `onSdtContainerChrome` callback
- Cells/tables flip `overflow: visible` only when a descendant actually renders chrome, not preemptively based on metadata
- `sdtBoundary` start/end flags are now correctly narrowed at split points (paragraph mid-line splits, embedded table mid-row splits)

**Nested + split-fragment fixes** (each is its own commit)
- Continue SDT chrome across paragraph splits, partial embedded tables, and nested-table continuations
- Suppress duplicate chrome when a nested table or cell already inherits the same container
- Preserve the SDT ancestor chain through nested tables so id-less controls don't re-render
- Skip media (drawings) at SDT boundaries
- Use the *rendered* lock mode (not the resolved one) so the dataset matches what's visually shown
- Allow nested SDT labels to overflow rather than being clipped
- Merge idless SDT siblings into a single container instead of fragmenting

**Tests**
- New `contracts/src/sdt-container.test.ts` (42 lines) and `painters/dom/src/sdt/container.test.ts` (200 lines)
- `renderTableCell.test.ts` expanded with ~1.2k lines covering nested SDT cases
- `renderTableFragment.test.ts` adds ~195 lines for ancestor/chrome-callback wiring


packages/layout-engine/contracts/src/sdt-container.ts
```diff
@@ -0,0 +1,78 @@
...
+const idlessSdtContainerKeys = new WeakMap<SdtMetadata, string>();
+let nextIdlessSdtContainerKey = 0;
+
+function getIdlessSdtContainerKey(metadata: SdtMetadata): string {
+  const existingKey = idlessSdtContainerKeys.get(metadata);
+  if (existingKey) return existingKey;
+
+  // AIDEV-NOTE: Id-less SDT grouping relies on pm-adapter sharing the same
+  // SdtMetadata object across sibling blocks in one container. Do not replace
+  // this with alias/title matching; separate controls can share display text.
```
VladaHarbour, line 19: Would be good to add a one-line cross-link to this comment in pm-adapter or layout-bridge AGENTS.md so future imports do not clone metadata per block and break grouping silently.

VladaHarbour (approved):
Just a minor suggestion
Approve

packages/layout-engine/contracts/src/sdt-container.ts
luccas-harbour: Sounds good, added 👍

luccas-harbour:
The broken paths in AGENTS.md were fixed in the top-most PR in [this commit](https://github.com/superdoc-dev/superdoc/pull/3413/changes/adf7e516f8085420a4206e4fb096a74158f4430b).

PR #3328, opened 2026-05-15
refactor(painters/dom): unify image block rendering (SD-2838)

luccas-harbour:
## Summary

- Extracts image `` creation, clip-path resolution, and hyperlink-anchor wrapping into shared helpers under `painters/dom/src/images/`, and routes the renderer's block/drawing image paths plus `renderTableCell`'s flowing, drawing, and anchored image paths through them — eliminating five near-identical copies of the same setup (objectFit, cover→left-top, clip-path, filters, hyperlink wrap).
- Co-locates the remaining image utilities (`image-clip-path`, `image-selectors`) under `images/` alongside the new modules and shares the inline-style applier + clip-path reader.
- Repairs version-signature gaps so visual edits actually invalidate: inline image runs (including inside tables), table images, and image hyperlinks now contribute their full visual state to `deriveBlockVersion`, and raw inline clip paths are hashed rather than dropped.
- Removes the now-dead `deriveBlockVersion`-on-render machinery (paragraph/image/drawing/table version hashing, SDT version helpers, list-marker guards) from `renderer.ts` since image-block rendering no longer derives its own per-block versions.
- Preserves the srcRect-vs-geometry transform invariant with an `AIDEV-NOTE` (clip-path scaling stays on the ``; rotation/flip stays on the fragment wrapper) so future edits don't collapse them onto the same element.

## Changes by area

**`painters/dom/src/images/` (new)**
- `image-block.ts` — `createBlockImageContent` (single source of truth for sizing, objectFit, clip-path, filters, hyperlink wrap) + `resolveBlockImageClipPath`.
- `image-fragment.ts` — `renderImageFragment` + `buildImageGeometryTransform` / `applyImageGeometryTransform`, lifted out of `renderer.ts`.
- `drawing-image.ts` — `createDrawingImageElement`, `createShapeGroupImageElement`, `createShapeTextImageElement`.
- `hyperlink.ts` — shared `buildImageHyperlinkAnchor` (sanitized href, target/rel, tooltip, role/tabindex, block vs inline-block display).
- `types.ts` — shared `BuildImageHyperlinkAnchor` anchor type.
- `image-clip-path.ts`, `image-selectors.ts` — moved from `utils/`.

**`painters/dom/src/renderer.ts`**
- Block image, drawing image, shape-group image, and shape-text image creation delegated to the new modules; `renderImageFragment` is now a thin call into `renderImageFragmentElement`.
- Drops the no-longer-used `deriveBlockVersion`, paragraph/cell-border hashing, SDT version helpers, list-marker guards, `applyImageClipPath`, and `buildImageFilters` imports (~530 lines net reduction in this file).

**`painters/dom/src/table/renderTableCell.ts`**
- Flowing image, table-cell drawing image, and anchored image paths all go through `createBlockImageContent` with a per-cell `buildImageHyperlinkAnchor` bound to the cell's document.
- Forces flowing/table images to `display: block` regardless of `block.display` (matches prior renderer behavior).

**`layout-resolved/src/versionSignature.ts`**
- New `renderedBlockImageVersion` / `renderedInlineImageRunVersion` capturing src/dims/alt/title/objectFit/display/filters/lum/rotation/flip/hyperlink/clip-path.
- Image blocks, drawing-image blocks, and inline image runs now hash the full rendered version; raw `clipPath` on inline image runs is hashed instead of dropped; table cell signatures recurse into nested blocks via `deriveBlockVersion`.

**Tests**
- `images/image-block.test.ts` — `resolveBlockImageClipPath` precedence and unsupported-value handling; drawing image filters and hyperlink wrapping.
- `table/renderTableCell.test.ts` — flowing-image block display, clip-path container `overflow: hidden`, filters, and hyperlink anchors on both flowing and anchored images.
- `layout-resolved/versionSignature.test.ts` — coverage for the new image hashing paths.

## Why

Five copies of the img-setup logic had already drifted (e.g., `objectPosition: left top` for `cover` was missing from some paths, hyperlinks weren't wrapped in table cells, clip-path container `overflow: hidden` wasn't always set). Centralizing prevents future drift and lets inline visual edits (hyperlink, luminance, gain/blacklevel, rotation, flip) invalidate caches reliably so re-renders pick them up.


packages/layout-engine/layout-resolved/src/versionSignature.ts
VladaHarbour: This comment is now misleading since it says that `deriveBlockVersion` is duplicated from `painters/dom/src/renderer.ts` and that two copies must stay in sync but this branch removed the renderer copy

VladaHarbour (approved):
One minor thing which may be confusing

packages/layout-engine/layout-resolved/src/versionSignature.ts
luccas-harbour: Sounds good. I fixed the comment

PR #3333, opened 2026-05-15
SD-2663 - feature: support table of contents hovering

chittolinag:
Introducing a new hover state for TOCs. Like MS Word does, when you hover in a TOC, we provide a visual feedback to the user with a grey background.


packages/super-editor/src/editors/v1/core/layout-adapter/sdt/toc.ts
```diff
@@ -53,6 +54,9 @@ export function applyTocMetadata(
       if (metadata.instruction) {
         block.attrs.tocInstruction = metadata.instruction;
       }
+      if (metadata.tocId) {
```
luccas-harbour, line 57: the `tocId` comment promises a `docPartObj uniqueId or parent sdBlockId` fallback, but the SDT call sites only pass `docPartObjectId ?? undefined`. since `tableOfContentsHandler`/`genericDocPartHandler` default the id to `''` (empty string) when `w:id` is missing, the `if (metadata.tocId)` guard here drops it and group hover silently degrades to single-entry. fall back to the parent block's `sdBlockId` here, or stop defaulting the importer to empty string?

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -3093,6 +3093,10 @@ export class DomPainter {
       // Add TOC-specific styling class
       if (isTocEntry) {
         fragmentEl.classList.add('superdoc-toc-entry');
```
luccas-harbour: the painter still hardcodes `'superdoc-toc-entry'` while `DOM_CLASS_NAMES.TOC_ENTRY` is right there and used in PresentationEditor. swap so the class name has one source of truth. same thing at `EditorInputManager.ts:1365`.

packages/super-editor/src/editors/v1/core/presentation-editor/tests/EditorInputManager.anchorClick.test.ts
```diff
@@ -133,9 +152,27 @@ describe('EditorInputManager — anchor-href click routing (SD-2537)', () => {
     viewportHost.appendChild(tocWrapper);
 
     firePointerDown(a);
+    expect(goToAnchor).not.toHaveBeenCalled();
+    firePointerUp(a);
     expect(goToAnchor).toHaveBeenCalledWith('#_Toc123');
   });
 
+  it('does not navigate on TOC pointerdown alone — deferral guard', () => {
```
luccas-harbour, line 199: the whole reason for deferring nav to pointerup is to let drag-select work — but only the pointerdown-alone and clean-click paths are covered. can we add a pointerdown → pointermove past the drag threshold → pointerup case asserting `goToAnchor` is not called? otherwise a regression that ignores `#dragThresholdExceeded` slips through.

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
```diff
@@ -6739,6 +6750,117 @@ export class PresentationEditor extends EventEmitter {
     this.#lastHoveredStructuredContentBlock = { id, elements };
   }
 
+  /** TOC analogue of {@link #handleStructuredContentBlockMouseEnter}, keyed by `data-toc-id`. */
+  #handleTocEntryMouseEnter = (event: MouseEvent) => {
```
luccas-harbour: none of the new hover logic has tests — group highlight on enter, cross-group clear on leave, same-group re-entry guard, `#reapplyTocGroupHover` after paint, and the zoom-scaled gap-fill math are all uncovered. at minimum a unit test with two `.superdoc-toc-entry[data-toc-id]` fragments verifying `toc-group-hover` lands on both would catch the most likely regressions.

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
```diff
@@ -6739,6 +6750,117 @@ export class PresentationEditor extends EventEmitter {
...
+  #clearHoveredTocGroup() {
+    if (!this.#lastHoveredTocGroup) return;
+    this.#lastHoveredTocGroup.elements.forEach((element) => {
+      element.classList.remove(DOM_CLASS_NAMES.TOC_GROUP_HOVER);
+      element.style.removeProperty('--toc-gap-below');
+    });
+    this.#lastHoveredTocGroup = null;
+  }
+
+  #setHoveredTocGroup(id: string) {
```
luccas-harbour: the five new methods mirror the SDT pair almost line-for-line (`#handleX{MouseEnter,Leave}`, `#setHoveredX`, `#clearHoveredX`, `#reapplyXHover`) plus a second mouseover/mouseout pair on the same host. fine for now since TOC has the gap-fill side effect, but a parameterized `HoverGroupCoordinator` would collapse both and the next hover-group concept won't multiply by five again. worth extracting now or wait until the third one shows up?

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
```diff
@@ -6739,6 +6750,117 @@ export class PresentationEditor extends EventEmitter {
...
+      const gap = scaleY > 0 ? rawGap / scaleY : rawGap;
+      if (gap > 0) {
+        current.element.style.setProperty('--toc-gap-below', `${gap + 1}px`);
+      }
+    }
+  }
+
+  #queryTocEntryElementsById(id: string): HTMLElement[] {
+    if (!this.#painterHost) return [];
+    const escapedId = typeof CSS !== 'undefined' && CSS.escape ? CSS.escape(id) : id.replace(/"/g, '\\"');
```
luccas-harbour: the CSS.escape fallback ladder is inlined here for the third time in this file — there's already a file-scope escapeAttrValue helper around line 10287. can we reuse that here?

luccas-harbour (changes requested):
hey @chittolinag!
codex flagged this issue below which I had noticed while testing this manually: https://www.loom.com/share/8047b874a3894628bbc6d13936ee8f43

you can see as I move the mouse around the TOC, the background flashes.

> - [P2] Preserve TOC hover while crossing entry gaps — packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts:6773-6781
>  When the cursor moves from one TOC entry into the paragraph-spacing gap, `relatedTarget` is the underlying page/container rather than another `.superdoc-toc-entry` (the gap filler is a pseudo-element, not a DOM entry), so this immediately clears `toc-group-hover`. That makes the intended continuous grey hover block disappear/flicker in the exact spacing gap that `--toc-gap-below` is trying to cover; keep the group active while the pointer is inside the measured gap or otherwise make the gap participate in the hover target.

claude found some other things that I thought were relevant. I added them as inline comments, hopefully they should be easy fixes!

thanks!

packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
chittolinag: good idea!

PR #3334, opened 2026-05-16
refactor(painters/dom): unify drawing block rendering (SD-2838)

luccas-harbour:
## Summary

Builds on the image-block unification (parent branch `luccas/sd-2838-unify-image-block-rendering`) by collapsing the remaining drawing-rendering duplication between `DomPainter` and `renderTableCell` onto a single implementation. Non-image drawings (vector shapes, shape groups, charts, shape text) now render through the same code path in body content and table cells, and the table-cell drawing wrapper is no longer reinvented for in-flow vs. anchored cases.

## What changed

**New modules in `painters/dom/src/drawings/`:**

- `renderDrawingContent.ts` — extracts ~890 lines of vector-shape / chart / shape-group / shape-text rendering (preset + custom geometry, gradient/alpha fills, line ends, fallback styles, shape text layout) out of `DomPainter` into a pure function. Accepts `doc`, `block`, optional `geometry`, optional `clipContainer`, and a `buildImageHyperlinkAnchor` callback — no painter coupling.
- `renderDrawingFragment.ts` — moves `renderDrawingFragment` and `isHeaderWordArtWatermark` out of `DomPainter`; the painter now passes its frame/anchor helpers (`applyResolvedFragmentFrame`, `applyFragmentFrame`, `applyFragmentWrapperZIndex`, `createErrorPlaceholder`) in as parameters.
- `tableDrawingFrame.ts` — single helper that builds the table-cell drawing wrapper for both `position: relative` (in-flow) and `position: absolute` (anchored) drawings, deduplicating the wrapper/inner/SDT-dataset scaffolding that previously existed twice in `renderTableCell`.
- `placeholder.ts` — small shared factory for the diagonal-stripe drawing placeholder.

**`renderer.ts`** shrinks by ~1000 lines. Drawing-related private methods (`renderDrawingFragment`, `renderDrawingContent`, `createVectorShapeElement`, `createShapeGroupElement`, `createChartElement`, `createDrawingPlaceholder`, `isHeaderWordArtWatermark`, plus their effect-extent / SVG / transform helpers) all delegate to the new modules. SVG/WordArt constants and local types (`LineEnd`, `EffectExtent`, `VectorShapeDrawingWithEffects`) move with them.

**`renderTableCell.ts`** is slimmed by ~95 lines:
- Both in-flow and anchored drawing branches now call `renderTableDrawingFrame` instead of building the wrapper inline.
- The `renderDrawingContent` callback signature gains an optional `{ clipContainer }` so image clipping can flow through the shared path.
- When the caller doesn't supply a callback (e.g. tests), the cell falls back to `renderDrawingContent` directly with `buildSharedImageHyperlinkAnchor` — non-image drawings in tables no longer disappear.

**`images/drawing-image.ts`** — `createDrawingImageElement` accepts an optional `clipContainer` forwarded to `createBlockImageContent`, so the shared renderer can pipe the table cell's inner wrapper through for image clipping.

## Tests

- New `renderDrawingContent.test.ts` covers vector shapes, shape groups, charts, fallback placeholders, and image hyperlink/clip-container behavior through the shared path.
- `renderTableCell.test.ts` gains ~260 lines of new coverage: image drawings rendering through the shared renderer, placeholder for image without source, callback override behavior, fallback to the shared renderer when no callback is supplied, clip-container threading, and SDT metadata sourced from `attrs.sdt`.


packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -694,6 +695,17 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
     hyperlink: ImageHyperlink | undefined,
     display: 'block' | 'inline-block',
   ): HTMLElement => buildImageHyperlinkAnchor(doc, imageEl, hyperlink, display);
+  const renderTableCellDrawingContent =
```
tupizz: Check if this make sense or not
This widens the table-cell drawing callback to all drawing kinds, even though the documented contract above says the callback is for vector shapes / shape groups. If a caller supplies a callback that only knows how to render those non-image drawing types, image drawings now bypass the shared image renderer and can lose filters, clip paths, and hyperlinks that `renderSharedDrawingContent` already handles. Please keep `drawingKind === 'image'` on the shared image path even when a custom callback is provided, and reserve the callback for non-image drawings.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -570,7 +571,7 @@ type TableCellRenderDependencies = {
    * The returned element will have width: 100% and height: 100% styles applied automatically.
    * If undefined, a placeholder element with diagonal stripes pattern is rendered instead.
```
VladaHarbour: This comment says the optional renderDrawingContent callback is for “vectorShapes and shapeGroups” and that if it is undefined, a placeholder is rendered. But image drawings can now go through that callback too and if the callback is undefined, the shared drawing renderer is used, not just a placeholder

tupizz (approved):
LGTM after re-review.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
luccas-harbour: That makes sense. I fixed it.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
luccas-harbour: This was pointing at the same issue as Tadeu's comment. It's fixed now.

tupizz (approved):
lgtm

VladaHarbour (approved):

chittolinag:
after talking to @luccas-harbour we decided to close this PR that has been sitting here for too long. doesn't seem to be a priority and has many conflicts. if needed, a new PR must be created after v2.

PR #3367, opened 2026-05-18
refactor: unify nested tables rendering (SD-2838)

luccas-harbour:
## Summary

Continuation of the SD-2838 unification effort, focused on **nested table rendering**. The painter and the layout engine had grown two parallel implementations of cell-segment expansion, partial-row pricing, and table fragment height math — slightly out of sync, and both living inside large `renderer.ts` / `layout-table.ts` files. This PR collapses them onto a single source of truth in `@superdoc/contracts` and breaks the painter's table branch out into co-located modules.

### Shared cell-slice + fragment-height helpers (`contracts/`)
- New `table-cell-slice.ts` (moved from `layout-engine/`): owns `getCellLines`, `getEmbeddedRowLines`, `describeCellRenderBlocks`, `createCellSliceCursor`, `computeFullCellContentHeight`, and `computeCellSliceContentHeight`. Exposed via the contracts barrel and marked internal in JSDoc.
- New `table-fragment-height.ts`: `computeTableFragmentHeight()` is the single formula for adding header rows + body rows (with partial-row substitution) + cell-spacing gutters + outer borders when `border-collapse: separate`.
- `layout-table.ts` and the DOM painter both call into these helpers; the painter's even-split height approximation for embedded rows is gone — slice heights now come from the same recursive walk the layout engine uses.

### `DomPainter.renderFragment` (table branch) decomposition
Carved out of `renderer.ts` into `painters/dom/src/table/`:
- `renderResolvedTableFragment.ts` — the former `renderTableFragment` method plus its `resolveTableRenderData` helper. Takes `renderLine`, snapshot capture, frame application, and SDT helpers as injected deps.
- `fragmentKey.ts` — `tableFragmentKey()` with the partial-row-aware cache key.
- `snapshot.ts` — `getTableSnapshotFlags()` for the `inTableFragment` / `inTableParagraph` `.closest()` checks used during paint-snapshot capture.
- `embeddedTableFragment.ts` — `createEmbeddedTableFragment`, `mapEmbeddedTableRowSlice`, `computeRenderedTableFragmentHeight`, and `getEmbeddedTableSegmentCount`, replacing the duplicated segment counters, visible-height loops, and partial-row computation inside `renderTableCell`.

Each new module has a co-located test. `renderer.ts` shrinks by ~150 lines and no longer carries table-specific helper closures.

### Bug fixes uncovered along the way
- `fix(layout-engine): price embedded table slice height` — partial embedded table slices now report their actual visible height instead of the full table height when computing whether they fit on the page.
- `fix(layout-engine): price partial embedded table slices` — pricing now uses the shared cell-slice cursor so deeply nested partial rows are sized consistently with the splitter's decisions.
- Removed a redundant `table-cell-slice.ts` shim in `layout-engine/` now that `contracts/` owns the implementation.

### Tests
- New unit tests in `contracts/`: `table-cell-slice.test.ts` (132 lines), `table-fragment-height.test.ts` (51 lines).
- New painter tests: `renderResolvedTableFragment.test.ts`, `embeddedTableFragment.test.ts`, `fragmentKey.test.ts`, `snapshot.test.ts`.
- Expanded `renderTableCell.test.ts` covers embedded-row slice mapping.
- `renderer-dispatch.test.ts` updated for the new dispatch path.


luccas-harbour:
I deleted codex' comment as it related to the removal of rendering of `list-item` fragments, which was handled in another PR.

packages/layout-engine/contracts/src/table-cell-slice.ts
```diff
@@ -0,0 +1,491 @@
...
+
+    rowCount += 1;
+    const rowLocalStart = Math.max(0, localStart - row.localStartLine);
+    const rowLocalEnd = Math.min(row.lineHeights.length, localEnd - row.localStartLine);
+    const rendersFullRow = rowLocalStart === 0 && rowLocalEnd >= row.lineHeights.length;
+
+    if (rendersFullRow) {
+      height += row.height;
+    } else {
+      height += sumArray(row.lineHeights.slice(rowLocalStart, rowLocalEnd));
```
tupizz: Potential pagination regression: partial nested-table row slices use only `row.lineHeights` here, but those line heights can miss the nested row's cell padding and block spacing. The DOM-side partial-row mapper includes that padding when computing rendered height, so layout can fit more content than the painter can actually draw and the result may clip/overflow on recursive nested tables. Please make this shared table-row slice height use the same per-cell block/padding semantics as the DOM `buildPartialRowInfo` path, and add a contract test that slices through a table block whose nested row has padded or spacing-bearing cell content.

packages/layout-engine/painters/dom/src/table/embeddedTableFragment.ts
```diff
@@ -0,0 +1,164 @@
...
+    const rowStart = segmentOffset;
+    const rowEnd = segmentOffset + rowSegmentCount;
+    segmentOffset = rowEnd;
+
+    if (rowEnd <= localFrom || rowStart >= localTo) continue;
+
+    if (fromRow === -1) fromRow = r;
+    toRow = r + 1;
+
+    if (rowSegmentCount > 1 && (rowStart < localFrom || rowEnd > localTo)) {
```
tupizz, line 147: `partialRow` is still overwritten when the visible segment window clips more than one multi-segment embedded-table row. `renderTableFragment` can apply only one `fragment.partialRow`, so a slice containing the tail of one nested row plus the head of another will render/measure one clipped row correctly and the other as a full row, duplicating hidden nested-table content and overcounting height. Please split these windows into multiple fragments or extend the fragment contract to carry partial metadata per row before rendering.

tupizz (changes requested):
Re-reviewed current head. The nested-table partial-row issue still applies in the extracted embeddedTableFragment helper; I added an updated inline comment on the current code.

packages/layout-engine/contracts/src/table-cell-slice.ts
luccas-harbour: I fixed this one.

packages/layout-engine/painters/dom/src/table/embeddedTableFragment.ts
luccas-harbour: This was fixed too

PR #3382, opened 2026-05-19
fix: skip empty open when defaults are off and keep editor raw

artem-harbour:
Linear: SD-3153

Prevents opening an empty context menu when `includeDefaultItems: false` and no custom items are available, and avoids `PresentationEditor` private-field proxy errors by storing context editor as `markRaw`.


harbournick (approved):
LGTM

PR #3383, opened 2026-05-19
refactor(painter-dom): unify footnote/endnote story detection (SD-2838)

luccas-harbour:
## Summary

Consolidates the footnote/endnote story-block-id checks that were previously duplicated between the DOM painter and the super-editor's body DOM index. This is a pure refactor — no behavioral changes — and continues the SD-2838 effort to unify rendering helpers.

- Extracts note story detection into `painters/dom/src/notes/story.ts`:
  - `getNoteStoryKind(blockId)` — classifies a block id as `footnote` / `endnote` / `semantic-footnote` / `semantic-endnote`.
  - `isNonBodyStoryBlockId(blockId)` — boolean shortcut used by the renderer's position-mapping guard and by the editor's DOM index exclusion.
  - `shouldApplyPlainFootnotePainterReadOnly(blockId)` — encodes the existing rule that only plain `footnote-*` frames get `contenteditable="false"` at the painter layer (endnotes and semantic notes have dedicated editing sessions).
- Extracts the contenteditable-false toggle into `notes/frame.ts` (`applyNoteStoryFrameAttributes`) and routes both `paint` paths in `renderer.ts` through it.
- Exports `isNonBodyStoryBlockId` from `@superdoc/painter-dom` and replaces the duplicated prefix list in `super-editor/.../DomPositionIndex.ts` with a single call.
- Parameterizes the existing renderer position-mapping test and the `DomPositionIndex` exclusion test with `it.each` over all four block-id variants so the unified rule is exercised everywhere.
- Adds dedicated unit tests for the new `notes/story.ts` and `notes/frame.ts` helpers, plus a smoke test that `isNonBodyStoryBlockId` is re-exported from the painter-dom entry point.

### DIFF
--- a/packages/super-editor/src/editors/v1/dom-observer/DomPositionIndex.ts
+++ b/packages/super-editor/src/editors/v1/dom-observer/DomPositionIndex.ts
@@ -1,4 +1,5 @@
 import { DOM_CLASS_NAMES } from '@superdoc/dom-contract';
+import { isNonBodyStoryBlockId } from '@superdoc/painter-dom';
 import { sortedIndexBy } from 'lodash';
 import { debugLog, getSelectionDebugConfig } from '../core/presentation-editor/selection/SelectionDebug.js';
 
@@ -35,12 +36,7 @@ function isExcludedFromBodyDomIndex(node: HTMLElement): boolean {
   }
 
   const blockId = node.closest<HTMLElement>('[data-block-id]')?.dataset.blockId ?? '';
-  return (
-    blockId.startsWith('footnote-') ||
-    blockId.startsWith('__sd_semantic_footnote-') ||
-    blockId.startsWith('endnote-') ||
-    blockId.startsWith('__sd_semantic_endnote-')
-  );
+  return isNonBodyStoryBlockId(blockId);
 }
 
 /**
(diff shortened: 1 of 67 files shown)


packages/super-editor/src/editors/v1/dom-observer/DomPositionIndex.ts
```diff
@@ -1,4 +1,5 @@
 import { DOM_CLASS_NAMES } from '@superdoc/dom-contract';
+import { isNonBodyStoryBlockId } from '@superdoc/painter-dom';
```
tupizz: Potential blocker: this new runtime import is likely to break existing `vi.mock('@superdoc/painter-dom')` factories in super-editor tests. Several PresentationEditor tests mock only `createDomPainter`/`DOM_CLASS_NAMES`, then instantiate code that imports `DomPositionIndex`; Vitest will usually fail with a missing `isNonBodyStoryBlockId` export before assertions run. Please either update the affected mocks to include this export or keep this helper in a less heavily mocked/shared location.

tupizz (changes requested):
Re-reviewed current head. The existing inline comment about the new painter-dom runtime import still looks blocking because the super-editor Vitest mocks for @superdoc/painter-dom do not expose isNonBodyStoryBlockId.

packages/super-editor/src/editors/v1/dom-observer/DomPositionIndex.ts
luccas-harbour: Sounds good. It's fixed

tupizz (approved):
lgtm

VladaHarbour (approved):

PR #3388, opened 2026-05-19
fix(super-editor): select inline SDT on backspace before run start (SD-3165)

luccas-harbour:
## Summary

Fixes SD-3165: a content-locked inline SDT could not be deleted with Backspace when the cursor was placed at the start of the run immediately following it. The run-aware Backspace chain scanned into the SDT's content, so the locked wrapper never became the deletion target.

A new command, `selectInlineSdtBeforeRunStart`, runs early in the Backspace chain. When the cursor is at offset 0 of a `run` whose previous sibling is a `structuredContent` node, it promotes the selection to a `NodeSelection` over that wrapper (unless `lockMode` is `sdtLocked`/`sdtContentLocked`, in which case it consumes the keystroke to block deletion). A subsequent Backspace then deletes the now-selected SDT as a unit.

The command tags its transaction with a `SELECT_INLINE_SDT_BEFORE_RUN_START_META` meta so the existing inline-SDT auto-select plugin (`structured-content-select-plugin`) does not clobber the new `NodeSelection` back into a `TextSelection`.

## Changes

- Add `selectInlineSdtBeforeRunStart` command and register it in the core command map, index, and types.
- Insert the new command into the `handleBackspace` keymap chain between `deleteBlockSdtAtTextBlockStart` and `backspaceEmptyRunParagraph`.
- Teach `structured-content-select-plugin` to skip its normalization when the new meta is present on the dispatched transaction.

## Tests

- Unit tests for `selectInlineSdtBeforeRunStart` covering: `unlocked` / `contentLocked` (selects wrapper), `sdtLocked` / `sdtContentLocked` (consumes without selecting), cursor not at run start, and previous sibling not an inline SDT.
- `core-command-map` declaration test asserts the new entry in `CoreCommandNames`.
- `keymap-backspace-chain` test pins the new command's position in the Backspace chain.
- Integration test in `structured-content-lock-plugin.test.js` for the full two-step flow (Backspace selects, second Backspace deletes) on a `contentLocked` inline SDT.
- `structured-content-select-plugin` test verifies the meta escape keeps the Backspace-created `NodeSelection` intact.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/commands/core-command-map.d.ts
+++ b/packages/super-editor/src/editors/v1/core/commands/core-command-map.d.ts
@@ -62,6 +62,7 @@ type CoreCommandNames =
   | 'backspaceNextToRun'
   | 'backspaceAcrossRuns'
   | 'backspaceAtomBefore'
+  | 'selectInlineSdtBeforeRunStart'
   | 'deleteBlockSdtAtTextBlockStart'
   | 'deleteSkipEmptyRun'
   | 'deleteNextToRun'
--- a/packages/super-editor/src/editors/v1/core/commands/core-command-map.test.ts
+++ b/packages/super-editor/src/editors/v1/core/commands/core-command-map.test.ts
@@ -1,0 +1,14 @@
+import { readFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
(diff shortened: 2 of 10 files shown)


missysuperdoc:
Closing as this PR is no longer needed. This was already addressed as part of SD-3218.

PR #3389, opened 2026-05-19
feat(public-facade): re-curate root index.ts with three tiers (SD-3212 b)

caio-pizzol:
PR B of the SD-3212 (Phase 4b) sequence. Grows `packages/superdoc/src/public/index.ts` from 24 exports to 200, organized into three tiers per the classification artifact merged in #3380.

**Distribution**:

| Tier | Count |
|---|---|
| supported-root | 132 |
| legacy-root | 60 |
| internal-candidate | 8 |

**The three tiers**:

1. **Supported root** — documented public API; first-class root surface. Clean re-exports.
2. **Legacy root** — typed for backward compatibility; not the recommended path. Per-name `@deprecated` JSDoc only where a real replacement exists (e.g. `editor.commands.*` → `editor.doc.*`). Section-level framing carries "legacy compat" intent when no replacement exists.
3. **Internal candidate** — accidental implementation leak. Kept typed under compat re-export with `@internal` so a future major can remove it. Only at root because at least one supported/legacy export reaches it transitively (enforced by the SD-3212 a1b closure gate).

**Load-bearing architectural decision — please scrutinize**: `src/public/index.ts` intentionally uses deep `@superdoc/common/...` imports at the source level (e.g. `import { DOCX, PDF, HTML } from '@superdoc/common/document-types'`). At source, deep imports are needed because the package-root re-export form (`from '@superdoc/common'`) trips vite-plugin-dts into following the `export *` chain and landing on the WRONG file (`shared/common/comments-types.js`, which doesn't export these names).

Deep imports normally trip `audit-declarations.cjs` ("private `@superdoc/*` specifier in emitted .d.ts"). The existing postbuild inliner in `ensure-types.cjs` — which already handles this exact problem for `src/index.js` — has been extended to also patch `dist/superdoc/src/public/index.d.ts`. It strips the private specifiers from the emitted declarations and inlines local `declare const DOCX: 'application/...'` declarations.

Net effect:
- **Source side**: deep imports (TS compiles them, dts plugin resolves to the correct definitions).
- **Emitted .d.ts**: no `@superdoc/*` specifiers; values appear as local `declare const`s. Consumers see no behavior change vs `src/index.js`'s historical handling.
- **Audit clean**: `audit-declarations.cjs` no longer sees the private specifier (stripped postbuild).

If a reviewer is uncomfortable with the deep-imports + strip pattern, the alternative is filing a separate ticket against vite-plugin-dts for cross-package `export *` resolution. The strip-and-inline approach is the same one the codebase has been relying on for `src/index.js` since before SD-3175 — this PR just extends it to the new facade entry point.

**Out of scope**:

- No `package.json#exports` change. PR C owns the root types flip.
- No removal of any name. Path-as-contract principle: if reachable, must be typed. Removal of `internal-candidate` names happens in a future major after deprecation cycle.

**Verified locally**:

- `pnpm --filter superdoc build:es`: all build steps green.
  - `verify-public-facade-emit`: 10 entries clean (root: **200 exports**, was 24).
  - `audit-declarations`: no private `@superdoc/*` leak in emitted .d.ts.
  - `check-export-coverage`: 12 entries OK.
- `typecheck-matrix.mjs`: 57/57 scenarios pass.
- `snapshot-superdoc-root-exports.mjs --check`: matches the locked baseline.
- `check-root-classification-closure.mjs`: **0 violations** (the closure gate validates that no supported/legacy export references an internal-candidate type).

**`AGENTS.md` update**: new "Contributing to the public surface" section documents the three tiers, the four files every public-surface change must update, and the three CI gates that enforce consistency.

**Sequence**:

- ✓ A0 (#3375) — root inventory + drift gate
- ✓ A1 + A1b (#3380, #3381) — classification + closure gate
- ✅ **B (this PR)** — re-curate src/public/index.ts
- ⏭ C — mechanical `package.json#exports['.'].types` flip
- ⏭ SD-3213 — gate consolidation (post-flip)

**Related**:

- Parent: SD-3175 (path-as-contract umbrella)
- SD-3212 (Phase 4b)
- SD-3185 (original under-curated facade; this re-curates)


caio-pizzol:
The `superdoc/ui/react` references are npm subpath imports, not filesystem paths; `packages/superdoc/package.json#exports['./ui/react']` exists, so the audit's broken-path finding is a false positive. The line-count warning is pre-existing (261 on main, 283 here); this PR adds the SD-3212 contributor rules, and splitting AGENTS.md should be handled separately.

PR #3401, opened 2026-05-19
refactor(painters/dom): unify textbox shape rendering (SD-2838)

luccas-harbour:
## Summary

Continues the SD-2838 effort to unify rendering paths in `painters/dom`. This branch consolidates the two pre-existing textbox text renderers — the WordArt SVG path and the HTML fallback path — into a single shared module so vector shapes, shape groups, and table-cell shapes all go through one code path.

- **New module `textbox/renderTextboxContent.ts`** centralizes WordArt (`SVG` + `<text>`/`<tspan>`) and fallback HTML rendering for shape text. Both branches share line-break handling, field resolution (`PAGE`/`NUMPAGES`), hex-color validation, and formatting application.
- **Drawing content path** (`drawings/renderDrawingContent.ts`) now delegates to `renderTextboxContent`, dropping ~270 lines of duplicated WordArt/fallback logic.
- **Legacy `createTextElement` in `svg-utils.ts`** is removed; it was the older `foreignObject`-based renderer that diverged from the drawing path.
- **WordArt watermark predicate** moves out of `renderDrawingFragment.ts` into `textbox/wordArtWatermark.ts` as `isWordArtTextboxWatermarkBlock`, with its own focused tests. `renderer.ts` imports from the new module for the header-watermark suppression rule.

### DIFF
--- a/packages/layout-engine/painters/dom/src/drawings/renderDrawingContent.ts
+++ b/packages/layout-engine/painters/dom/src/drawings/renderDrawingContent.ts
@@ -17,7 +17,7 @@ import { createChartElement as renderChartToElement } from '../chart-renderer.js
 import { createDrawingImageElement, createShapeGroupImageElement } from '../images/drawing-image.js';
 import type { BuildImageHyperlinkAnchor } from '../images/types.js';
 import { applyAlphaToSVG, applyGradientToSVG } from '../svg-utils.js';
-import type { FragmentRenderContext } from '../renderer.js';
+import type { FragmentRenderContext } from '../fragment-context.js';
 import { hasShapeTextContent, renderTextboxContent } from '../textbox/renderTextboxContent.js';
 import { createDrawingPlaceholder } from './placeholder.js';
 
--- a/packages/layout-engine/painters/dom/src/drawings/renderDrawingContent.ts
+++ b/packages/layout-engine/painters/dom/src/drawings/renderDrawingContent.ts
@@ -14,18 +14,12 @@ import type {
 } from '@superdoc/contracts';
 import { getPresetShapeSvg } from '@superdoc/preset-geometry';
 import { createChartElement as renderChartToElement } from '../chart-renderer.js';
-import {
-  createDrawingImageElement,
-  createShapeGroupImageElement,
-  createShapeTextImageElement,
-} from '../images/drawing-image.js';
+import { createDrawingImageElement, createShapeGroupImageElement } from '../images/drawing-image.js';
 import type { BuildImageHyperlinkAnchor } from '../images/types.js';
-import { applyAlphaToSVG, applyGradientToSVG, validateHexColor } from '../svg-utils.js';
+import { applyAlphaToSVG, applyGradientToSVG } from '../svg-utils.js';
 import type { FragmentRenderContext } from '../renderer.js';
+import { hasShapeTextContent, renderTextboxContent } from '../textbox/renderTextboxContent.js';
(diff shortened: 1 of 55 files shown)


luccas-harbour:
> ### 💡 Codex Review
> https://github.com/superdoc-dev/superdoc/blob/cc9f268259dac402cdee8c133b9bc3ce8e6d9f7d/packages/layout-engine/painters/dom/src/drawings/tableDrawingFrame.ts#L60-L61
> 
> 
> **P2 Badge Keep image drawings on built-in table rendering path**
> This helper now routes every table drawing through `renderDrawingContent`, including `drawingKind: 'image'`. In the previous behavior, image drawings were rendered by the built-in image path (object-fit, clip-path, hyperlink handling) and the callback was only for non-image drawing kinds; callers that provide a callback for vector/group shapes can now unintentionally replace image drawings with custom placeholders or lose image-specific styling. Guard image drawings before invoking the callback so custom renderers don't break table images.
> 
> ℹ️ About Codex in GitHub

Fixed.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -509,17 +506,27 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
...
+      geometry: 'geometry' in block ? block.geometry : undefined,
+      context,
+      clipContainer: options?.clipContainer,
+      buildImageHyperlinkAnchor: buildTableImageHyperlinkAnchor,
+    });
+  const renderTableCellDrawingContent = (
+    block: DrawingBlock,
+    options?: { clipContainer?: HTMLElement },
+  ): HTMLElement => {
+    if (block.drawingKind === 'image' || !renderDrawingContent) {
```
tupizz: Coverage gap: this intentionally bypasses the supplied drawing callback for image drawings so table-image styling/hyperlinks are preserved. The added test covers an in-flow image, but the anchored drawing path also goes through this callback wrapper. Please add a regression for an anchored table image drawing with a callback present, so anchored positioning and hyperlink wrapping cannot regress independently.

packages/layout-engine/painters/dom/src/drawings/renderDrawingContent.ts
```diff
@@ -212,275 +202,17 @@ const createShapeTextElement = (
...
-      block.textInsets,
-      width,
-      height,
-      context,
-    );
-  }
-
-  return createFallbackTextElement(
-    renderer,
+  return renderTextboxContent({
```
tupizz, line 205: Coverage gap: this extraction preserves page/context-dependent textbox rendering, but the production table path supplies `renderDrawingContent` through DomPainter rather than leaving the table-cell callback absent. Please add an integration-style test that reaches this call through the table rendering path for a vector-shape textbox resolving a `PAGE` field, so the real callback/context path is protected.

tupizz (approved):
LGTM after re-review.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
luccas-harbour: Fixed

packages/layout-engine/painters/dom/src/drawings/renderDrawingContent.ts
luccas-harbour: Fixed

tupizz (approved):
lgtm

VladaHarbour (approved):

PR #3412, opened 2026-05-20
fix(editor): avoid copying paraId on split

### LINKED ISSUE #3405: create.table + Enter on trailing paragraph duplicates paraId, blocking subsequent Document API mutations
### What happened?

`editor.doc.create.table({ at: { kind: 'after', nodeId } })` produces a trailing paragraph with a non-null `w14:paraId`. Pressing Enter on that paragraph triggers ProseMirror's default split-block, which copies attrs onto the new sibling — including the `paraId`. The document is then in a state where two paragraphs share one `paraId`.

The next call to any Document API mutation (`editor.doc.delete`, `editor.doc.create.table`, `editor.doc.styles.paragraph.setStyle`, …) throws:

```
PlanError: DOCUMENT_IDENTITY_CONFLICT — Document contains blocks with duplicate identities.
```

from `assertNoDuplicateBlockIds` inside `compilePlan`. The error's `details.remediation` names `document.repair()`, but no such method appears on the Document API surface.

### Steps to reproduce

In the browser console :

```js
const doc = sd.activeEditor.doc;
doc.clearContent({});
// Click into the empty paragraph so the caret is set.
const seg = doc.selection.current().target.segments[0];
doc.create.table({
  rows: 3, columns: 3,
  at: { kind: 'after', nodeId: seg.blockId },
});
// Click into the trailing paragraph below the table, then press Enter once.
// Any further Document API mutation now throws DOCUMENT_IDENTITY_CONFLICT.
```

After the Enter keystroke, walking `editor.view.state.doc.descendants(...)` shows two paragraphs sharing one `paraId`:

```
paragraph at pos 46: paraId="78FF6DBD", sdBlockId="580b559b-…"
paragraph at pos 48: paraId="78FF6DBD", sdBlockId="449664ef-…"
```

### SuperDoc version

1.32.0

### Browser

Firefox 150 (via Zen Browser 1.19.13b)

### Additional context

Failure shape from a real throw:

```json
{
  "name": "PlanError",
  "code": "DOCUMENT_IDENTITY_CONFLICT",
  "details": {
    "duplicateBlockIds": ["08266E78"],
    "blockCount": 1,
    "remediation": "Re-import the document or call document.repair() to assign unique identities."
  }
}
```

wookieb:
Summary
- Prevent paragraph splits from carrying over DOCX paraId identity.
- Avoid duplicate block IDs after pressing Enter in table separator paragraphs.

Closes #3405

# Considerations

Regular `splitRunToParagraph` does not copy `paraId` on purpose. Split on empty paragraph is special, therefore I believe the best fix is to just mark attribute as ignored on `split`.

I was thinking about not having `paraId` at for table separator but I've noticed it was added on purpose in recent months. Unless... I'm wrong?

### DIFF
--- a/packages/super-editor/src/editors/v1/extensions/paragraph/paragraph.js
+++ b/packages/super-editor/src/editors/v1/extensions/paragraph/paragraph.js
@@ -132,7 +132,7 @@ export const Paragraph = OxmlNode.create({
   addAttributes() {
     return {
       paraId: { rendered: false, keepOnSplit: false },
-      textId: { rendered: false },
+      textId: { rendered: false, keepOnSplit: false },
       rsidR: { rendered: false },
       rsidRDefault: { rendered: false },
       rsidP: { rendered: false },
--- a/packages/super-editor/src/editors/v1/extensions/paragraph/paragraph.js
+++ b/packages/super-editor/src/editors/v1/extensions/paragraph/paragraph.js
@@ -131,7 +131,7 @@ export const Paragraph = OxmlNode.create({
 
   addAttributes() {
     return {
-      paraId: { rendered: false },
+      paraId: { rendered: false, keepOnSplit: false },
       textId: { rendered: false },
       rsidR: { rendered: false },
       rsidRDefault: { rendered: false },
--- a/packages/super-editor/src/editors/v1/document-api-adapters/tables-adapter.regressions.test.ts
+++ b/packages/super-editor/src/editors/v1/document-api-adapters/tables-adapter.regressions.test.ts
@@ -208,7 +208,6 @@ function makeTableEditor(options: TableEditorOptions = {}): Editor {
     attrs: {
       sdBlockId: 'table-1',
       tableProperties: {},
-      tableGrid: [5000, 5000],
       grid: [{ col: 1200 }, { col: 3000 }],
     },
     isBlock: true,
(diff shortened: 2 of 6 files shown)


packages/super-editor/src/editors/v1/extensions/paragraph/paragraph.js
```diff
@@ -131,7 +131,7 @@ export const Paragraph = OxmlNode.create({
 
   addAttributes() {
     return {
-      paraId: { rendered: false },
+      paraId: { rendered: false, keepOnSplit: false },
       textId: { rendered: false },
```
caio-pizzol: `textId` has the same split problem as `paraId`: if Enter copies it, export can write duplicate `w14:textId`. The table split path already drops both (`tables-adapter.ts:2122` -- "Avoid duplicate w14:textId after split."), so can we opt this one out too?

```suggestion
      textId: { rendered: false, keepOnSplit: false },
```

caio-pizzol (approved):
@wookieb thanks for the PR. this matches the editor split path, and the regression test covers the reported duplicate paraId case. lgtm once the inline is handled.

gpardhivvarma:
after this + the `textId` change land, three sites will be nulling `paraId`/`textId` on split — `keepOnSplit: false` here, the spread in `split-run.js:97-105`, and the `delete`s in `tables-adapter.ts:2119-2122`. easy to update one and miss the others. worth routing `splitBlockPatch` through `Attribute.getSplittedAttributes` so this file is the only declaration site? rough sketch:

```js
// split-run.js, replacing the spread at 97-105
const extensionAttrs = editor?.extensionService?.attributes ?? [];
paragraphAttrs = Attribute.getSplittedAttributes(extensionAttrs, node.type.name, node.attrs);
```

one wrinkle: `getSplittedAttributes` would also drop `listRendering` (also `keepOnSplit: false`), which the hand-rolled spread currently preserves — not sure if that's intentional.

packages/super-editor/src/editors/v1/extensions/paragraph/paragraph.js
wookieb: Done

wookieb:
That was a good tip @gpardhivvarma - Thank You :)

caio-pizzol:
@cubic-dev-ai

caiopizzol (approved):
@wookieb thanks for the follow-up. the `textId` handling and shared split-attr cleanup look right to me.

lgtm.

PR #3413, opened 2026-05-20
refactor(painter-dom): consolidate rendering helpers and harden table edge cases (SD-2838)

luccas-harbour:
## Summary

Builds on `luccas/sd-2838-unify-textbox-shape-rendering` with a wave of follow-up work that (1) formalizes the `painters/dom` package boundary, (2) collapses duplication between body and table-cell rendering paths, and (3) fixes a cluster of correctness issues around embedded tables, anchored/zero-height media, and SDT continuation chrome.

## Renderer organization (docs)

- New `packages/layout-engine/painters/dom/AGENTS.md` makes `renderer.ts` an explicit page-level orchestration layer and enumerates the concern directories (`paragraph/`, `runs/`, `table/`, `images/`, `drawings/`, `sdt/`, `notes/`, `textbox/`, `ruler/`, `utils/`, `features/`) that own content rendering. Cross-linked from `layout-engine/AGENTS.md` and `painters/dom/README.md`; the DomPainter JSDoc was trimmed to point at the new guide.

## Helper consolidation

- Move `FragmentRenderContext` into its own `painters/dom/src/fragment-context.ts` so concern modules can depend on it without pulling in `renderer.ts`. Updated all importers (`runs/`, `drawings/`, `textbox/`, `paragraph/`, `images/`, `table/`).
- Inline previously-injected helpers: `runs/render-line.ts`, `runs/text-run.ts`, etc. now call `applyTrackedChangeDecorations`, `resolveTrackedChangesConfig`, `resolveRunSdtId`, `syncInlineSdtWrapperTypography`, and `expandSdtWrapperPmRange` directly instead of going through `RunRenderContext` plumbing. `applySdtDataset` / `applyContainerSdtDataset` are imported in the modules that need them rather than threaded through every dep object.
- Extract `renderListMarkerFrame` so `renderLegacyListMarker` and `renderResolvedListMarker` share the marker DOM/style logic.
- Extract `drawings/drawingFrame.ts` (`renderDrawingFrame`) shared by `tableDrawingFrame`; cell-only `images/table-image-frame.ts` carved out of `renderTableCell.ts`.
- Fold SDT ancestor singletons (`ancestorContainerKey` / `ancestorContainerSdt`) into the `ancestorContainerKeys[]` / `ancestorContainerSdts[]` arrays everywhere — same data, one shape.
- Hoist a pure `computeBetweenBorderContext(entries)` out of `computeBetweenBorderFlags`; the latter is now a thin adapter that maps resolved items to entries.
- Drop the unused drawing-body frame code path.
- Reuse the shared `utils/apply-styles.ts` helper instead of redefining `applyInlineStyles` in `renderTableCell.ts` and `runs/render-line.ts`.

## Correctness fixes

- **Split embedded tables**: `mapEmbeddedTableRowSlice` becomes `mapEmbeddedTableRowSlices` (plural). Embedded tables that straddle page splits now render each row slice with the correct `continuesFromPrev` / `continuesOnNext` flags, suppress chrome/label on continuation slices, and coalesce contiguous full rows. Height computation and the wrapper stack lay slices end-to-end.
- **Separate-table borders without `cellSpacing`**: `renderTableFragment` now draws top/right/bottom/left borders for any `borderCollapse === 'separate'` table, not just ones with explicit cell spacing. Continuation slices skip the top/bottom border (and subtract the avoided spacing) so heights stay consistent.
- **Stale SDT boundary chrome**: `shouldRebuildForSdtBoundary` now also detects label-visibility flips and `paddingBottomOverride` changes; new tests cover both.
- **Anchored & zero-height media in cells**:
  - Contracts: `table-cell-slice.ts` exposes a `getVisibleCellBlockIndexes` helper so anchored out-of-flow and zero-height object blocks no longer steal the "first/last visible block" designation that drives final-paragraph spacing.
  - Painter: anchored media is ignored for cell spacing, kept passive in row layout, clamped to cell bounds, and excluded from border-grouping decisions. Zero-height media is excluded from spacing and segment counts but stays in border groups.
- **Table cell paragraphs**: split paragraphs in cells are now flagged so border continuation logic sees them as continuations.
- **Contracts**: `table-cell-slice.ts` adds typing for visible cell blocks (no behavior change beyond the spacing fix above).

## Performance & test hygiene

- Defer table-cell border hashing until needed (`perf(painter-dom)`).
- Stabilize table-border selectors in `renderTableFragment.test.ts` and tighten `embeddedTableFragment.test.ts` coverage for the new continuation flags.
- Add `ImageInteractionLayer.test.ts` regression covering anchored-image interaction.

### DIFF
--- a/packages/layout-engine/contracts/src/table-cell-slice.test.ts
+++ b/packages/layout-engine/contracts/src/table-cell-slice.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import type { ParagraphMeasure, TableBlock, TableCellMeasure, TableMeasure } from './index.js';
+import type { ParagraphMeasure, TableBlock, TableCell, TableCellMeasure, TableMeasure } from './index.js';
 import {
   computeCellSliceContentHeight,
   computeFullCellContentHeight,
@@ -30,11 +30,20 @@ describe('table cell segment mapping', () => {
     height,
   });
 
-  const makeParagraphBlock = (id: string, spacing?: { before?: number; after?: number }) => ({
+  const makeParagraphBlock = (
+    id: string,
+    spacing?: { before?: number; after?: number },
+  ): TableCell['blocks'][number] => ({
     kind: 'paragraph' as const,
     id,
     runs: [],
     attrs: spacing ? { spacing } : undefined,
+  });
+
+  const makeImageBlock = (): TableCell['blocks'][number] => ({
+    kind: 'image',
+    id: 'zero-height-image',
+    src: 'data:image/png;base64,AAA',
   });
 
   it('counts paragraph and positive-height object segments', () => {
@@ -45,6 +54,24 @@ describe('table cell segment mapping', () => {
     };
 
     expect(getCellLines(cell)).toHaveLength(6);
+  });
+
+  it('ignores zero-height object blocks for final paragraph spacing', () => {
+    const cell: TableCellMeasure = {
+      blocks: [makeParagraph(1), makeImage(0)],
+      width: 200,
+      height: 20,
+    };
+    const block: TableCell = {
+      id: 'cell-zero-height-tail',
+      blocks: [makeParagraphBlock('paragraph-after', { after: 12 }), makeImageBlock()],
+    };
+    const blocks = describeCellRenderBlocks(cell, block, { top: 0, bottom: 5 });
+
+    expect(blocks).toHaveLength(1);
+    expect(blocks[0].isLastBlock).toBe(true);
+    expect(blocks[0].spacingAfter).toBe(0);
+    expect(computeFullCellContentHeight(cell, block, { top: 0, bottom: 5 })).toBe(27);
   });
 
   it('falls back to legacy single-paragraph cells', () => {
--- a/packages/layout-engine/contracts/src/table-cell-slice.test.ts
+++ b/packages/layout-engine/contracts/src/table-cell-slice.test.ts
@@ -46,6 +46,14 @@ describe('table cell segment mapping', () => {
     src: 'data:image/png;base64,AAA',
   });
 
+  const makeAnchoredImageBlock = (): TableCell['blocks'][number] => ({
+    kind: 'image',
+    id: 'anchored-image',
+    src: 'data:image/png;base64,AAA',
+    anchor: { isAnchored: true },
+    wrap: { type: 'None' },
+  });
+
   it('counts paragraph and positive-height object segments', () => {
     const cell: TableCellMeasure = {
       blocks: [makeParagraph(2), makeImage(50), makeImage(0), makeParagraph(3)],
@@ -69,6 +77,24 @@ describe('table cell segment mapping', () => {
     const blocks = describeCellRenderBlocks(cell, block, { top: 0, bottom: 5 });
 
     expect(blocks).toHaveLength(1);
+    expect(blocks[0].isLastBlock).toBe(true);
+    expect(blocks[0].spacingAfter).toBe(0);
+    expect(computeFullCellContentHeight(cell, block, { top: 0, bottom: 5 })).toBe(27);
+  });
+
+  it('ignores anchored out-of-flow object blocks for final paragraph spacing', () => {
+    const cell: TableCellMeasure = {
+      blocks: [makeParagraph(1), makeImage(20)],
+      width: 200,
+      height: 20,
+    };
+    const block: TableCell = {
+      id: 'cell-anchored-tail',
+      blocks: [makeParagraphBlock(12), makeAnchoredImageBlock()],
+    };
+    const blocks = describeCellRenderBlocks(cell, block, { top: 0, bottom: 5 });
+
+    expect(blocks).toHaveLength(2);
     expect(blocks[0].isLastBlock).toBe(true);
     expect(blocks[0].spacingAfter).toBe(0);
     expect(computeFullCellContentHeight(cell, block, { top: 0, bottom: 5 })).toBe(27);
--- a/packages/layout-engine/contracts/src/table-cell-slice.test.ts
+++ b/packages/layout-engine/contracts/src/table-cell-slice.test.ts
@@ -90,7 +90,7 @@ describe('table cell segment mapping', () => {
     };
     const block: TableCell = {
       id: 'cell-anchored-tail',
-      blocks: [makeParagraphBlock(12), makeAnchoredImageBlock()],
+      blocks: [makeParagraphBlock('paragraph-before-anchor', { after: 12 }), makeAnchoredImageBlock()],
     };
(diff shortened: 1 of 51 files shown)


packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -310,33 +290,59 @@ function renderPartialEmbeddedTable(params: {
...
+  let hasSdtContainerChrome = false;
+  rowSlices.forEach((rowSlice, index) => {
+    const sliceHeight = computeRenderedTableFragmentHeight({
+      block,
+      measure: tableMeasure,
+      fromRow: rowSlice.fromRow,
+      toRow: rowSlice.toRow,
+      partialRow: rowSlice.partialRow,
+      continuesFromPrev: localFrom > 0 || index > 0,
+      continuesOnNext: localTo < totalTableSegments || index < rowSlices.length - 1,
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -622,12 +801,54 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
     const effectiveCellWidth = cellWidth ?? cellMeasure.width;
     const contentWidthPx = Math.max(0, effectiveCellWidth - paddingLeft - paddingRight);
     const contentHeightPx = Math.max(0, rowHeight - paddingTop - paddingBottom);
+    let paragraphContextY = 0;
+    let borderContextSegmentStart = 0;
+    const betweenEntryBlockIndexes: number[] = [];
+    const betweenInfoByBlockIndex = computeBetweenBorderContext(
```
tupizz, line 845: This computes between-border context over every cell block before the render loop filters by `globalFromLine/globalToLine`. If a partial row slice starts at the second paragraph, a hidden previous paragraph with matching borders can still form a group and set `suppressTopBorder` on the first visible paragraph. Please add/fix coverage for two adjacent table-cell paragraphs with identical borders rendered with `fromLine` starting at the second paragraph, and ensure the visible paragraph keeps its top border.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
```diff
@@ -622,12 +801,54 @@ export const renderTableCell = (deps: TableCellRenderDependencies): TableCellRen
...
+      cellBlocks.slice(0, rawBlockCount).flatMap((block, index) => {
+        const measure = blockMeasures[index];
+        const blockStartGlobal = borderContextSegmentStart;
+        const blockLineCount = blockLineCounts[index] ?? 0;
+        borderContextSegmentStart += blockLineCount;
+        if (isAnchoredMediaBlock(block, measure) || isZeroHeightMediaBlock(block, measure)) {
+          return [];
+        }
+        const y = paragraphContextY;
```
tupizz, line 858: This synthetic `y` advances by measured block height only, but `renderParagraphContent` later includes paragraph spacing (for example spacing-after) in the rendered total height. That means grouped paragraph borders inside table cells may not extend through the paragraph spacing gap, unlike body fragments where the actual fragment `y` includes the gap. Please add a regression with two same-border paragraphs in a cell and `spacing.after > 0`, asserting the expected gap/border extension metadata.

packages/super-editor/src/editors/v1/core/presentation-editor/layout/FootnotesBuilder.ts
```diff
@@ -265,6 +272,64 @@ function resolveNoteDocJson(
...
+  const paragraphProperties =
+    typeof node.attrs?.paragraphProperties === 'object' && node.attrs.paragraphProperties !== null
+      ? (node.attrs.paragraphProperties as Record<string, unknown>)
+      : {};
+
+  if (!converterContext) {
+    return paragraphProperties;
+  }
+
+  return resolveParagraphProperties(
```
tupizz: This resolves paragraph properties and writes `listRendering` from inside `presentation-editor/layout`. That crosses the current rendering boundary: `PresentationEditor` should bridge editor state into layout/paint, while OOXML/list/style resolution belongs upstream in conversion/style resolution. It also mutates note PM JSON before `toFlowBlocks`, baking visual list metadata into the presentation builder. Please move this footnote list-resolution step into the upstream conversion/adapter path, or pass already-resolved footnote FlowBlocks into the builder.

tupizz (changes requested):
Re-reviewed current head. The existing table-cell between-border slice/gap findings still apply, and I added a new inline finding around resolving footnote list rendering inside PresentationEditor layout.

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
luccas-harbour: This was fixed

packages/layout-engine/painters/dom/src/table/renderTableCell.ts
luccas-harbour: This was fixed too

packages/super-editor/src/editors/v1/core/presentation-editor/layout/FootnotesBuilder.ts
luccas-harbour: This was a good call. I refactored it using a different strategy and I think it's a lot better. Thanks!

tupizz (approved):
lgtm

packages/layout-engine/contracts/src/table-cell-slice.test.ts
```diff
@@ -47,6 +64,42 @@ describe('table cell segment mapping', () => {
...
+    const block: TableCell = {
+      id: 'cell-anchored-tail',
+      blocks: [makeParagraphBlock(12), makeAnchoredImageBlock()],
+    };
+    const blocks = describeCellRenderBlocks(cell, block, { top: 0, bottom: 5 });
+
+    expect(blocks).toHaveLength(2);
+    expect(blocks[0].isLastBlock).toBe(true);
+    expect(blocks[0].spacingAfter).toBe(0);
+    expect(computeFullCellContentHeight(cell, block, { top: 0, bottom: 5 })).toBe(27);
```
VladaHarbour, line 100: Current test is failing here. It calls makeParagraphBlock(12), so 12 becomes the paragraph id and no { after: 12 } spacing is set. The assertion at line 100 expects 27, but the test actually computes 20

VladaHarbour (changes requested):

packages/layout-engine/contracts/src/table-cell-slice.test.ts
luccas-harbour: Oops, you're right. I just pushed a fix!

PR #3429, opened 2026-05-21
fix: preserve tab underline via runProperties fallback in collab

artem-harbour:
Linear: SD-3103

This PR fixes a collaboration rendering regression where underlined tab runs could lose underline styling after Yjs sync (especially for the second participant joining a session).

#### What changed
- Updated tab conversion in `pm-adapter` to hydrate tab formatting from `runProperties` (same resilience path used by text runs).
- Kept mark precedence intact: explicit tab marks still override fallback formatting.

#### Why
In Yjs collaboration flows, marks on element/atom nodes like `tab` may not always round-trip as reliably as text marks. This fallback ensures tab underline remains stable without changing importer or Yjs behavior.


harbournick (approved):
LGTM

PR #3444, opened 2026-05-22
fix(sdt): correct cursor placement and label interactions for structured content (SD-3237)

luccas-harbour:
## Summary

Fixes [SD-3237](https://linear.app/superdocworkspace/issue/SD-3237/bug-sdt-hover-and-click-to-place-cursor-interactions-are-unreliable): SDT hover and click-to-place cursor interactions were unreliable. Clicking inside a structured content block highlighted the entire content on the first click (SD-1584's "first-click select-all"), repeated clicks didn't reliably reposition the caret, and the label/tab chrome appeared on hover before the user had actually focused the SDT. Nested SDTs amplified the jank.

This PR restores Word-like behavior: hover only greys the background, clicking places the caret at the clicked position (even inside nested SDTs), and label clicks select the whole control while still preserving the native drag-handle.

### Pointer/selection behavior (`EditorInputManager.ts`)
- Removed the SD-1584 click-to-select-node behavior on block SDT bodies, and removed the `select-all-on-first-click` behavior on inline SDTs. Body clicks now place a collapsed caret like normal content.
- Rewrote label click handling: capture-phase `pointerdown` / `pointerup` / `click` on the owner document (visible host + viewport host + fallback through `elementsFromPoint`), with a movement threshold so a label drag still triggers a native `dragstart`.
- Label selection is now deferred to `mouseup`/`click` (not `pointerdown`), scoped to the editor that owns the hit label, and routed through `NodeSelection` with a `pm-start`-driven fallback. Gesture state is cleared on `pointercancel`.
- Re-focuses the editor after applying a label-driven selection so the toolbar/selection state stays in sync.

### DOM contract (`dom-contract`, painter, mapping)
- Promoted SDT label class names to the shared DOM contract (`INLINE_SDT_LABEL`, `BLOCK_SDT_LABEL`, plus a `STRUCTURED_CONTENT_CHROME_LABEL_CLASS_NAMES` set). Painter, `sdt-helpers`, layout-bridge `dom-mapping`, and `DomPointerMapping` all use the contract instead of string literals.
- Excluded SDT label chrome elements from text-position mapping in both `layout-bridge/dom-mapping.ts` and `DomPointerMapping.ts` so their `pm-start`/`pm-end` ranges no longer bleed into body caret resolution.
- Dropped the `sdt-group-hover` style that revealed the block label on hover — hover only greys the background; the label appears on selection.

### Structured-content plugins
- `structured-content-select-plugin.js`: removed the click-into-inline-SDT select-all behavior; now only handles boundary/exit navigation. Also tightened the "selects whole content" check so a parent block selection that *contains* a nested inline SDT no longer collapses to that inline.
- `structured-content-lock-plugin.js`: updated comments to reflect that the select-plugin no longer produces first-click select-all.

### Resolution
- `findStructuredContentBlockAtPos` now does a `doc.nodeAt(pos)` check first so block labels resolve at the node boundary (where the click lands on the chrome rather than inside the content).

### Tests
- New `sd-3237-sdt-interactions.spec.ts` behavior spec + `sd-3237-nested-sdt-lorem-ipsum.docx` fixture covering:
  - block SDT body clicks place and move a collapsed caret
  - hover greys background but does not show the label; click activates label
  - block/inline label clicks select the wrapping control (`NodeSelection`)
  - nested inline SDT clicks reposition caret while both wrappers stay active
  - block label click on an outer block containing a nested inline SDT keeps selecting the outer block
- Updated `sdt-drag-drop.spec.ts` and `structured-content.spec.ts` for the new behavior.
- Extensive unit coverage in `EditorInputManager.structuredContent.test.ts`, `DomPointerMapping.test.ts`, `structured-content-resolution.test.ts`, and the lock/select plugin tests.


packages/super-editor/src/editors/v1/core/presentation-editor/pointer-events/EditorInputManager.ts
```diff
@@ -2059,38 +2118,120 @@ export class EditorInputManager {
...
+
+        try {
+          applySelection();
+        } catch {
+          const state = editor.view?.state ?? editor.state;
+          const tr = state.tr.setSelection(TextSelection.create(state.doc, resolved.start, resolved.end));
+          editor.view?.dispatch(tr);
         }
+
+        this.#focusEditor();
```
caio-pizzol, line 2173: clicking a block SDT label selects the block and focuses the editor, but it never calls `scheduleSelectionUpdate`. That leaves selection listeners out of date until something else updates them. The inline label path already does this.

```suggestion
        this.#callbacks.scheduleSelectionUpdate?.();
        this.#focusEditor();
```

caio-pizzol (commented):
the inline label path looks good. two label-click issues to fix before merge - left both inline.

packages/super-editor/src/editors/v1/core/presentation-editor/pointer-events/EditorInputManager.ts
```diff
@@ -2059,38 +2118,120 @@ export class EditorInputManager {
...
+        rect.width > 0 &&
+        rect.height > 0 &&
+        event.clientX >= rect.left &&
+        event.clientX <= rect.right &&
+        event.clientY >= rect.top &&
+        event.clientY <= rect.bottom
+      ) {
+        return label;
+      }
     }
```
caio-pizzol: clicking UI placed over a label can still select the covered label underneath. I reproduced this with a fixed button over a block SDT label: the button click was blocked, and the SDT was selected instead.

`elementsFromPoint()` includes elements behind the button, and the rect fallback has the same issue. Please only accept labels from the real click path, or check that the label is the top element with `document.elementFromPoint(...)` before handling it.

packages/super-editor/src/editors/v1/core/presentation-editor/pointer-events/EditorInputManager.ts
luccas-harbour: fixed

packages/super-editor/src/editors/v1/core/presentation-editor/pointer-events/EditorInputManager.ts
luccas-harbour: fixed

caio-pizzol (approved):
lgtm

PR #3465, opened 2026-05-23
fix(style-engine): keep list marker visible when paragraph-mark is vanished (SD-3269)

tupizz:
## Demo

CleanShot 2026-05-23 at 09 06 30@2x

## Summary

Numbered Heading2 paragraphs in [SD-3269](https://linear.app/superdocworkspace/issue/SD-3269) lost their auto-generated marker (e.g. `Section 2.01`) on render. Root cause: `w:vanish` / `w:specVanish` on the paragraph-mark rPr (`w:pPr/w:rPr`) was leaking into the list marker's resolved run properties, and the renderer drops markers when `marker.run.vanish` is true.

Per ECMA-376 §17.3.2.36 / §17.3.2.41 those properties are scoped to the paragraph-mark glyph (¶); the spec explicitly says "if this element is applied to any other run, it can be ignored." Word renders the section number because the marker's formatting comes from the numbering definition's rPr, not the paragraph-mark rPr.

## Fix

`packages/layout-engine/style-engine/src/ooxml/index.ts` — in `resolveRunProperties`, when `isListNumber=true`, strip `vanish` and `specVanish` from the inline rPr before composing the cascade. Vanish set on the numbering definition's own `w:lvl/w:rPr` still wins, so authors can still hide a marker the documented way.

## Why this seam

- Single source of truth: per project conventions, OOXML cascade lives in style-engine (`CLAUDE.md` "Style Resolution Boundary"); a fix in `pm-adapter` or the painter would silo the rule.
- Covers both renderer code paths in one place (`renderer.ts:3328` `resolvedMarker.vanish` and `:3553` `marker.run.vanish`).
- Does not alter the regular inline run path, where `w:vanish` legitimately hides actual text content.

## Tests

Two regression tests in `packages/layout-engine/pm-adapter/src/attributes/paragraph.test.ts`:

1. Paragraph-mark `vanish`/`specVanish` must not leak into the marker run — fails on `main`, passes after the fix.
2. Numbering-definition `vanish` is still honoured for the marker — positive control to prevent over-stripping.

All pm-adapter (1840), style-engine (142), and layout-engine (652) unit tests pass.

## Test plan

- [x] `pnpm --filter @superdoc/style-engine test`
- [x] `pnpm --filter @superdoc/pm-adapter test`
- [x] `pnpm --filter @superdoc/layout-engine test`
- [x] Manual: load the fixture from SD-3269 in `pnpm dev`, confirm `Section 1.01 Assets Purchased.`, `Section 2.01 Purchase Price.`, etc. now render
- [x] Visual / layout regression (`pnpm test:layout`) before merge

## Linear

Closes [SD-3269](https://linear.app/superdocworkspace/issue/SD-3269)


VladaHarbour (approved):

caio-pizzol (commented):
hey @tupizz! good call keeping pPr/rPr out of marker styling - that matches the OOXML boundary cleanly :)

before merge, one thing on the fuse trigger. i made four small Word-native fixtures and compared Word 16.0 against this branch:

- `w:vanish` + `w:specVanish` on the paragraph mark: both fuse
- `w:vanish` only on the paragraph mark: Word fuses, branch doesn't
- `w:specVanish` only on the paragraph mark: Word doesn't fuse, branch does
- `w:vanish` on the numbering def (`w:lvl/w:rPr`): neither fuses

in this matrix, Word follows paragraph-mark `w:vanish`, not `w:specVanish`. so vanish-only docs miss the fuse, and specVanish-only docs get one Word doesn't do. looks like a new Word-fidelity divergence from this PR.

the spec text points the other way (§17.3.2.36 makes specVanish the "never break" property), so this isn't a misread; Word's behavior just disagrees with the text.

as a sanity check, i swapped the local trigger to paragraph-mark `w:vanish`. all four fixtures matched Word, and pm-adapter / style-engine / layout tests stayed green. existing tests don't catch this because every fixture sets both flags.

marker visibility i also checked end-to-end - that part matches Word in all four cases.

caveat: only Word 16.0 tested. happy to share the fixtures + probe if useful.

adjust the trigger here, or explicitly accept/split as a follow-up?

tupizz:
@caio-pizzol thanks for the matrix, the fixtures were spot on. I reproduced each one through Word on my Mac via `pdftotext` to be sure, and the four cases line up exactly with what you saw on the Windows VM.

| Fixture | OOXML | Word | Branch before | Branch now |
|---|---|---|---|---|
| F1 | vanish + specVanish | fused | fused | fused |
| F2 | vanish only | fused | not fused | fused |
| F3 | specVanish only | not fused | fused | not fused |
| F4 | lvl/rPr vanish | not fused, marker hidden | not fused, marker hidden | not fused, marker hidden |

Pushed `eec05f29e` on `tadeu/sd-3269-vanish-marker`:

- Trigger flipped from `runProperties.specVanish === true` to `runProperties.vanish === true` on the paragraph-mark rPr.
- Renamed `ParagraphAttrs.specVanish` to `suppressParagraphBreak` so the field describes the resolved effect, not the historical trigger guess.
- Renamed the post-process from `mergeSpecVanishParagraphs` to `mergeFusedParagraphs`.
- Dropped the prior three integration tests, replaced them with a six-test matrix: F1, F2, F3, F4 plus chain and trailing-flag edge cases. F2 and F3 are the regression catchers.

I also built a 44-page Word vs SuperDoc side-by-side PDF over all four probe fixtures plus the three real-world ones (HVY-45 Farnon, PURCHASE AND SALE OF ASSETS 2, PURCHASE PRICE 3). Page counts match Word across the board, including the 37-page Farnon doc. Happy to share the PDF if useful.

All pm-adapter (1846), style-engine (142), and layout-engine (652) tests pass.

caio-pizzol (approved):
@tupizz nice turnaround. i re-ran the four Word fixtures on `eec05f29e`, and they all match Word now: F1/F2 join, F3/F4 stay separate, and marker visibility is correct. pm-adapter, style-engine, and layout-tests are all green.

i also pushed `5de107eb9` onto this branch with a Playwright import spec at `tests/behavior/tests/importing/sd-3269-vanish-numbered-heading.spec.ts` covering the same four fixtures through the browser render path. validated locally that reverting either the trigger or the marker-strip makes the matching case fail. drop the commit if you'd rather ship without it.

lgtm.

PR #3470, opened 2026-05-23
fix: improve vml watermarks rendering

VladaHarbour:
Improve parsing watermark opacity and font.
Word/Before comparison:
image

Fix applied:
Screenshot 2026-05-23 at 23 13 39


tupizz (approved):

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/pict/helpers/handle-shape-text-watermark-import.js
```diff
@@ -323,8 +374,10 @@ function generateTextWatermarkSVG({ text, width, height, rotation, fill, textSty
     dominant-baseline="middle"
     font-family="${fontFamily}"
     font-size="${sanitizedFontSize}px"
+    textLength="${sanitizedWidth}"
+    lengthAdjust="spacingAndGlyphs"
```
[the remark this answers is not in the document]
VladaHarbour: fixed

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/pict/helpers/handle-shape-text-watermark-import.js
```diff
@@ -314,27 +395,51 @@ function generateTextWatermarkSVG({ text, width, height, rotation, fill, textSty
...
   const centerY = svgHeight / 2;
+  const textAttributes = [
+    `x="${centerX}"`,
+    `y="${centerY}"`,
+    'text-anchor="middle"',
+    'dominant-baseline="middle"',
+    `font-family="${fontFamily}"`,
+    `font-size="${sanitizedFontSize}px"`,
+    ...(fitShape ? [`textLength="${sanitizedWidth}"`, 'lengthAdjust="spacingAndGlyphs"'] : []),
+    `fill="${color}"`,
```
caio-pizzol, line 473: repro: open `03-solid.docx` (watermark inserted with "semi-transparent: off"). Word leaves out `<v:fill>` for solid marks - the VML spec says missing opacity means fully opaque (`1.0`). line 71 treats it as `.5`, then `resolveRenderedTextWatermarkOpacity` halves it to `.25`, so a solid Word watermark shows up as faint as a semi-transparent one.

i tested nine Word-generated fixtures and the same halving pulls the `.5` cases visibly lighter than Word too (the red one comes out pink). could we treat missing opacity as `1.0` and revisit the halving against the fixtures? changing only the default still cuts the solid case to `.5`.

caio-pizzol (commented):
hey @VladaHarbour! good direction - the fixture matches Word much closer than main :)

one thing on the transparency. left inline.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/pict/helpers/handle-shape-text-watermark-import.js
VladaHarbour: Removed transparency halving but mimic Word approach instead: render opacity halved on body edit but make it 1 on header/footer edit(since watermarks belong to the header)

caio-pizzol (approved):
hey @VladaHarbour! thanks for addressing last round - the opacity and positioning fix is much closer to Word now :)

i pushed a small test-only follow-up commit with three additions:

1. a unit test for the Word-native t136 no-fill shape form Word emits for "semi-transparent: off" watermarks.
2. a unit test asserting explicit `opacity="0"` is preserved end-to-end.
3. a behavior import smoke test using a Word-native solid watermark fixture from the shared corpus, asserting the rendered watermark SVG keeps `fill-opacity="1"`.

broader VML spec gaps (full shapetype inheritance, `on`/`textpathok` gating, `fitshape="true"` alt form, and `v:fill/@color` vs `fillcolor` precedence) should be tracked separately - those are out of scope for this customer fix.

lgtm.

PR #3488, opened 2026-05-25
fix(presentation): keep virtualized pages in sync with host scroll

tupizz:
## Summary
- Fix SD-3230 by observing PresentationEditor virtualization scroll events during capture, so embedded host scroll containers cannot starve DomPainter virtual-window updates with stopImmediatePropagation().
- Add behavior harness options for fixed-height preview scroll containers and host-level blocked scroll events.
- Add a Playwright regression that reproduces the Hebbia failure mode with real SuperDoc/DomPainter rendering.

## Root cause
SuperDoc virtualizes pages by updating DomPainter when the detected scroll owner emits a scroll event. In embedded preview containers, a host app can install a capture-phase scroll listener that calls stopImmediatePropagation(). With SuperDoc listening only in the normal phase, the preview container scrollTop moves but DomPainter keeps the initial mounted page window, leaving the visible area blank.

## Test plan
- pnpm --dir tests/behavior exec playwright test tests/virtualization/blocked-scroll-event.spec.ts
- Red-checked locally by temporarily restoring the old normal-phase listeners; the new test failed with mountedPages still starting at 0 after scrolling.
- Manual visual before/after captured locally:
  - /tmp/sd-3230-visual-before-fix.png
  - /tmp/sd-3230-visual-after-fix.png

Fixes SD-3230

### DIFF
--- a/packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
+++ b/packages/super-editor/src/editors/v1/core/presentation-editor/PresentationEditor.ts
@@ -499,7 +499,8 @@ export class PresentationEditor extends EventEmitter {
   #semanticResizeDebounce: number | null = null;
   #lastSemanticContainerWidth: number | null = null;
   #editorListeners: Array<{ event: string; handler: (...args: unknown[]) => void }> = [];
-  #scrollHandler: (() => void) | null = null;
+  #scrollHandler: ((event?: Event) => void) | null = null;
+  #handledScrollEvents = new WeakSet<Event>();
   #scrollContainer: Element | Window | null = null;
   #scrollContainerValidated = false;
   #sectionMetadata: SectionMetadata[] = [];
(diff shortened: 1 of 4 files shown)


tupizz:
Addressed the duplicate-scroll concern in the latest push.

I kept both capture listener paths because the window capture fallback is what preserves the SD-3230 case when the host preview container has its own target capture listener that calls stopImmediatePropagation(). To avoid doing duplicate virtualization work for normal embedded element scrolls, #scrollHandler now dedupes the same native scroll Event with a WeakSet before calling painterAdapter.onScroll().

Verified after the change:
- pnpm --dir tests/behavior exec playwright test --project=chromium tests/virtualization/blocked-scroll-event.spec.ts tests/virtualization/scroll-acceleration.spec.ts tests/virtualization/zoom-scroll-visibility.spec.ts
- pnpm --dir tests/behavior exec playwright test tests/virtualization/blocked-scroll-event.spec.ts

harbournick (approved):
LGTM

PR #3489, opened 2026-05-25
fix(sdt): SDT field deletion and editing contract (SD-3218, SD-3165 and SD-3257)

luccas-harbour:
> [!NOTE]
> This PR fixes SD-3218, SD-3165 and SD-3257

## Summary

Hardens the deletion / typing / caret contract for inline structured-content (SDT) fields so editing across the wrapper boundary behaves consistently with Word and stops leaking into locked content.

Stacked on top of `luccas/sd-3237-bug-sdt-hover-and-click-to-place-cursor-interactions-are`; review against that branch.

### Inline SDT deletion across the wrapper boundary

- New commands `selectInlineSdtBeforeRunStart` / `selectInlineSdtAfterRunEnd` (`packages/super-editor/src/editors/v1/core/commands/selectInlineSdtBeforeRunStart.js`). When the caret sits at the start of the run after an inline SDT (Backspace) or the end of the run before one (Delete), the command installs a `TextSelection` over the SDT's inner content instead of letting run-aware deletion scan into the locked field.
- Wired into the Backspace and Delete keymap chains (`keymap.js`).
- For unlocked / `contentLocked` modes the next keystroke deletes the selected content; for `sdtLocked` / `sdtContentLocked` the command consumes the keystroke and the boundary stays put.
- Selection is applied uniformly across all lock modes as a `TextSelection` (not a `NodeSelection` over the chrome), so labels/handles never get caught in the selection. A `SELECT_INLINE_SDT_BEFORE_RUN_START_META` meta flag tells the structured-content select plugin not to collapse this selection back to a `TextSelection` over the wrapper. (SD-3165)

### contentLocked wrapper deletion

The lock plugin previously promoted exact-content selections covering a `contentLocked` SDT to a `NodeSelection`, requiring a second keystroke to delete. It now deletes the wrapper directly on Backspace/Delete. Cut still promotes to `NodeSelection` so PM's clipboard handler can serialize the wrapper in one keystroke (`structured-content-lock-plugin.js`).

### Empty inline SDT placeholder

Empty inline SDTs were filtered out of layout entirely, leaving a zero-width invisible wrapper with no caret target. Introduces a layout-only placeholder run that flows end-to-end:

- **contracts**: `TextRun.visualPlaceholder = 'emptyInlineSdt'` and an `isEmptyInlineSdtPlaceholderRun` guard. `sliceRunsForLine` preserves placeholders that have no chars.
- **pm-adapter**: emit a placeholder run for inline `structuredContent` nodes with empty/missing content. `mergeAdjacentRuns` refuses to merge placeholders into neighbors.
- **measuring/dom**: reserve an 8px inline box (0px for `appearance='hidden'`) without taking the empty-paragraph code path.
- **painters/dom**: render `<span class="superdoc-empty-inline-sdt-placeholder" aria-hidden="true">` inside the SDT wrapper and tag the wrapper `data-empty="true"`. CSS gives the wrapper a visible border without inflating line-box height (`styles.ts`).
- **DomSelectionGeometry**: anchor the caret to the line's Y (the placeholder is `height: 0`) and always to its left edge.
- **lock plugin**: collapsed Backspace/Delete inside an empty inline SDT deletes the wrapper when the lock mode allows.

### beforeinput insertText at inline SDT boundaries

When the caret sits directly before or after an inline SDT, native `beforeinput` handling can drift the inserted character into the wrapper. `editable.js` now forces the manual `insertText` path for collapsed selections at those boundaries so the character lands on the outer side.

### `sdBlockRev` on ancestors of inline edits

Text changes inside an inline SDT weren't bumping the containing paragraph's `sdBlockRev` because `nodesBetween` over the replace range didn't always visit the ancestor block. `block-node.js` now walks the ancestor chain at each changed range's boundaries.

The plugin's own metadata transaction is tagged with `BLOCK_NODE_METADATA_UPDATE_META` so it doesn't re-trigger `appendTransaction` and the lock plugin doesn't filter it out.


caio-pizzol (commented):
One question on the beforeinput boundary path.

packages/super-editor/src/editors/v1/core/extensions/editable.js
```diff
@@ -62,6 +62,20 @@ const recordStoryInputDebug = (view, event, editor, phase, extra = {}) => {
...
+  if (typeof pos !== 'number' || pos < 0 || pos > doc.content.size) {
+    return false;
+  }
+
+  const $pos = doc.resolve(pos);
+  const before = $pos.parent.childBefore($pos.parentOffset).node;
+  const after = $pos.parent.childAfter($pos.parentOffset).node;
+
+  return isInlineStructuredContentNode(before) || isInlineStructuredContentNode(after);
+};
```
caio-pizzol, line 77: This boundary check only looks at siblings of `$pos.parent`. The Backspace command added in this PR also handles the caret when it is inside a `run` at the start or end, but the `beforeinput` tests only seem to cover the outer SDT boundary.

Should this helper handle the same run-level positions too, or should we add a test showing `beforeinput` never reaches this path from inside a run?

PR #3515, opened 2026-05-26
SD-3229 - fix: enhance TOC entry processing and bookmark handling

chittolinag:
### Issue
Clicking *Update table of contents* on a mixed-source TOC (one that combines `\t` custom-style mappings with `\f` TC-field identifiers) destructively rebuilt the entries: auto-numbered heading prefixes disappeared, hyperlink anchors switched from the existing body `_Toc…` bookmarks to synthetic UUID names, multi-run Section rows collapsed into a single text node, the `pageReference` field downgraded to a plain text run with a `tocPageNumber` mark, the TOC2 left tab stop was dropped, and heading character marks (bold/underline/Times New Roman Bold) leaked into Article rows. In the worst case the rebuild produced the "No table of contents entries found." placeholder when sources weren't being collected at all. ([SD-3229](https://linear.app/superdocworkspace/issue/SD-3229/bug-mixed-source-toc-update-replaces-entries-with-no-table-of-contents))

### Proposed solution
- Fix the TOC `SWITCH_PATTERN` so unquoted switch arguments (`\f C`) parse with their identifier instead of as a bare flag.
- Hoist embedded `w:bookmarkStart` / `w:bookmarkEnd` out of the synthesized `sd:tableOfContentsEntry` atom so the entry stays valid and the bookmark remains visible to the bookmark map.
- Bridge the v3 `sd:tableOfContentsEntry` translator into the v2 importer pipeline so TC fields actually materialize as PM `tableOfContentsEntry` nodes.
- Teach `collectTocSources` to honor `\t` custom-style mappings, capture the source paragraph's existing `_Toc…` body bookmark (so the rebuilt entry reuses it), and pick up the rendered list-marker prefix + the body title marks for TC entries.
- Rewrite `buildTocEntryParagraphs` to emit the importer-equivalent shape — multi-run entries, real `pageReference` PM node, level-2 left tab stop alongside the right tab stop, and the existing body bookmark as anchor.
- Match Word's mark-propagation rules per ECMA-376 §17.16.5.68: heading-driven entries (`\o`/`\u`/`\t`) get plain text and inherit typography from the linked TOC{n} paragraph style; TC-field titles inherit only bold/italic/underline (no `textStyle`) so the TOC{n} font isn't overridden.
- Update `toc.update` mode `pageNumbers` to walk `pageReference` subtrees, and `syncTocBookmarks` to skip sources with an existing body anchor so the rebuild doesn't litter the document with duplicate synthetic bookmarks.
- Added unit + end-to-end regression tests covering the new entry shape and the mark-propagation rules.


caio-pizzol:
Thanks for the work on this, Gabriel! We confirmed that mixed-source TOC updates still need attention in V2 and have retained the reproduction and findings for that work. We’re closing this V1 PR as we focus the implementation on V2. Your fixtures and investigation are helping guide that effort.

PR #3525, opened 2026-05-27
fix(painters/dom, super-editor): set explicit document text color floor (SD-3456)

### LINKED ISSUE #3456: Auto-numbered list markers render as white text (invisible) — .superdoc-page sets color: rgb(255,255,255)
## Description
  
  When rendering any `.docx` containing auto-numbered lists, the list-number markers
  (`1.`, `1.1`, `2.2.1`, etc.) render as white text on the white page background — i.e.
  invisible. The same affects any other text that relies on inheriting its color rather
  than setting `<w:color>` explicitly.

  ## Root cause
  
  `.superdoc-page` sets `color: rgb(255, 255, 255)` as the default text colour for the
  page container:

  ```css
  .superdoc-page { color: rgb(255, 255, 255); }

  Document content with an explicit <w:color w:val="000000"/> renders black correctly,
  but auto-generated numbering (which has no rPr) inherits the white default.

  

### Steps to reproduce

Reproduction

  1. Open any .docx with multi-level numbered lists (e.g. a legal MSA or any Word
  doc using numPr).
  2. Observe: clause numbers 1., 1.1, 1.1.1 are invisible against the page.
  3. Inspect via devtools — confirm getComputedStyle(...).color === 'rgb(255, 255, 255)'
  on the marker spans, traced up to .superdoc-page.

  Tested with @superdoc-dev/react@latest (installed 2026-05-22) via a minimal Vite +
  React app. Document used was a standard MSA with the typical Word numbering definition.

  Expected behaviour

  Page text colour should default to black (or inherit), matching Word, Google Docs,
  LibreOffice, and every other .docx renderer. Auto-numbered list markers should be
  visible without consumers needing to add CSS overrides.

  Workaround

  Consumers of @superdoc-dev/react can patch their own stylesheet with:

  .superdoc-page,
  .superdoc-fragment,
  .presentation-editor__pages { color: #000 !important; }

  This restores list markers without affecting SuperDoc's own UI chrome (the white-on-blue
  "Inline content" / "Structured content" labels for content controls render correctly
  because they have explicit colour set).

  Environment
  
  - @superdoc-dev/react latest as of 2026-05-22
  - Chrome on macOS
  - Minimal Vite + React app, single <SuperDocEditor document={file}/> component

  ---


### SuperDoc version

- @superdoc-dev/react latest as of 2026-05-22

### Browser

Chrome

### Additional context

_No response_

gpardhivvarma:
Closes #3456.

## Summary

`.superdoc-page` (layout-engine page surface) and `.sd-editor-scoped .ProseMirror` (editor isolation surface) both relied on the browser default `color`. On dark-themed OSes that resolves to the `canvastext` system color (white), so any document text without an explicit `<w:color>` — auto-numbered list markers in particular — rendered invisible on the white page. There is no source CSS rule that sets `color: rgb(255,255,255)` in the repo; the reporter's devtools trace pointed at `.superdoc-page` because that was the highest-up element in the cascade where the color happened to resolve (via the user-agent default), not the rule that set it.

This PR sets an explicit `color: var(--sd-layout-page-text, #000)` floor on both surfaces. OOXML-explicit colors still win (inline `style.color` on runs has higher specificity than the inherited page color). Consumers customize via the `--sd-layout-page-text` token, documented in `variables.css` alongside `--sd-layout-page-bg`/`--sd-layout-page-shadow`.

## Why two surfaces in one PR

`.superdoc-page` covers presentation-mode rendering (where the reported bug manifests). `.sd-editor-scoped .ProseMirror` covers the editor isolation surface for super-editor direct mounts (header/footer section editors, popovers, rich-text Field editors, etc.). Both are reverted to `canvastext` by `isolation.css` (`all: revert`) and both need the same floor. Bundling them avoids shipping the symptom-fix while leaving a sibling surface broken for the same reason.

## Files

| File | Change |
|---|---|
| `packages/layout-engine/painters/dom/src/styles.ts` | `PageStyles.color?: string`; `DEFAULT_PAGE_STYLES.color = 'var(--sd-layout-page-text, #000)'`; `pageStyles()` returns `color: merged.color` in the inline-style payload. |
| `packages/layout-engine/painters/dom/src/styles.test.ts` | 3 new tests: default contract, consumer override path, `DEFAULT_PAGE_STYLES` exposure. |
| `packages/super-editor/src/editors/v1/assets/styles/elements/prosemirror.css` | Adds `color: var(--sd-layout-page-text, #000)` to `.sd-editor-scoped .ProseMirror` with an inline comment linking it to the page-painter fix. |
| `packages/superdoc/src/assets/styles/elements/editor-scoped-color-floor.test.js` | 4 new tests guarding the editor-scoped rule (modeled on existing `search-match-precedence.test.js`). |
| `packages/superdoc/src/assets/styles/helpers/variables.css` | Adds `--sd-layout-page-text: #000;` to the "Styles: layout" section. |

## Test plan

- [x] `pnpm --filter @superdoc/painter-dom test -- styles.test.ts` — 8/8 (3 new + 5 existing)
- [x] `pnpm --filter superdoc test` — 1030/1030 (4 new + 1026 existing)
- [ ] Manual verification on macOS Chrome with system dark mode: open any DOCX with multi-level numbered lists (e.g. a legal MSA). Confirm clause numbers (`1.`, `1.1`, `1.1.1`) are visible against the white page.
- [ ] Manual verification on light mode: confirm no visible change (text was black before via canvastext-on-light, still black after via the explicit floor).
- [ ] Layout-comparison run against published baseline — diff should show only the new `color: ...` property on the page element's inline style, no pixel deltas on Linux CI.

## Reviewer notes / consumer impact

- **Consumers using `.superdoc-page { color: white }` without `!important`** to force dark text on the page will now lose to the inline style. The intended way to customize is `--sd-layout-page-text` (or `pageStyles.color` via the SuperDoc constructor). `!important` overrides still win.
- **Token naming**: `--sd-layout-page-text` is named after the layout page but is reused inside the editor isolation surface. Considered introducing a separate `--sd-editor-text` token but kept one shared token so theme authors set it once for both surfaces.
- **Visual regression test not added**: visual tests run on light-mode Linux CI where `canvastext` is already black, so a pixel-diff baseline would never catch this class. The CSS-content tests added here are the right safety net.

## Follow-ups (not in this PR)

- Consider adding an architecture guard (sibling to Guards D/E in `architecture-boundaries.test.ts`) enforcing "every container hosting inherited document text must declare `color` explicitly in `DEFAULT_*_STYLES`."

### DIFF
--- a/packages/layout-engine/painters/dom/src/styles.test.ts
+++ b/packages/layout-engine/painters/dom/src/styles.test.ts
@@ -1,5 +1,11 @@
 import { describe, expect, it } from 'vitest';
-import { ensureSdtContainerStyles, ensureTrackChangeStyles, lineStyles } from './styles.js';
+import {
+  DEFAULT_PAGE_STYLES,
+  ensureSdtContainerStyles,
+  ensureTrackChangeStyles,
+  lineStyles,
+  pageStyles,
+} from './styles.js';
 
 describe('lineStyles', () => {
   it('sets height and lineHeight from the argument', () => {
@@ -11,6 +17,27 @@ describe('lineStyles', () => {
   it('sets fontSize to 0 to eliminate the CSS strut', () => {
     const styles = lineStyles(20);
     expect(styles.fontSize).toBe('0');
+  });
+});
+
+describe('pageStyles', () => {
+  // SD-3456 / IT-1102: auto-numbered list markers (and any other text that
+  // inherits its color rather than carrying an explicit `<w:color>`) render
+  // invisible on dark-themed OSes because the browser default `canvastext`
+  // system color resolves to white on the white page. The page element must
+  // therefore set an explicit text color floor.
+  it('sets an explicit color on the page so document text does not inherit the OS canvastext system color', () => {
+    const styles = pageStyles(816, 1056);
+    expect(styles.color).toBe('var(--sd-layout-page-text, #000)');
+  });
+
+  it('lets consumers override the page color via the pageStyles option', () => {
+    const styles = pageStyles(816, 1056, { color: '#222' });
+    expect(styles.color).toBe('#222');
+  });
+
+  it('exposes the color default through DEFAULT_PAGE_STYLES so consumers and themes can read it', () => {
+    expect(DEFAULT_PAGE_STYLES.color).toBe('var(--sd-layout-page-text, #000)');
   });
 });
 
--- a/packages/layout-engine/painters/dom/src/styles.ts
+++ b/packages/layout-engine/painters/dom/src/styles.ts
@@ -22,6 +22,7 @@ export type PageStyles = {
   boxShadow?: string;
   border?: string;
   margin?: string;
+  color?: string;
 };
 
 export const DEFAULT_PAGE_STYLES: Required<PageStyles> = {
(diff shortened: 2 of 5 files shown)


gpardhivvarma:
@caio-pizzol

gpardhivvarma:
hey @caio-pizzol just wanted to remind you

gpardhivvarma:
@caio-pizzol Just wanted to remind you

caio-pizzol:
Thanks for the PR and for looking into this.

This issue has already been fixed on `main`, so this PR is no longer needed. Keeping it open would duplicate work and conflict with the current code.

I'm going to close this one as already fixed.

Thanks again for the contribution.

PR #3541, opened 2026-05-27
llm tools presets

andrii-harbour:
SD-3128


harbournick:
@andrii-harbour please investigate these: 
Findings

  1. [Medium] Exported Node ToolCatalog lost useful structure
     File: packages/sdk/langs/node/src/presets.ts:51
     Evidence: ToolCatalog.tools is now unknown[]; on main it was ToolCatalogEntry[] with toolName, description, inputSchema, mutates, and operations.
      Impact: TypeScript consumers inspecting getToolCatalog().tools lose property access and can fail to compile even though runtime shape is unchanged.
     Recommended fix: preserve/export the structured catalog entry type, or make ToolCatalog generic with the legacy structured entry as the default.
     Source: Codex reviewer + parent investigation.

  2. [Medium] Malformed provider bundles now fail open as empty tool lists
     File: packages/sdk/langs/node/src/presets/legacy.ts:263, packages/sdk/langs/python/superdoc/tools_api.py:45, packages/sdk/langs/python/superdoc/presets/legacy.py:145
     Evidence: both preset implementations coerce a missing/non-array tools field to []; the previous public listTools / list_tools paths threw TOOLS_ASSET_INVALID.
     Impact: a broken generated/package artifact can look like a valid empty tool set, leaving agents with no tools instead of failing fast.
     Recommended fix: validate provider bundle shape in the preset loader or list wrappers, and add malformed-bundle coverage.
     Source: Codex reviewer + parent investigation.

  3. [Low] Python choose_tools() silently treats preset: "" as default
     File: packages/sdk/langs/python/superdoc/tools_api.py:80
     Evidence: input.get('preset') or DEFAULT_PRESET turns an explicit empty string into legacy; Node chooseTools({ preset: "" }) and Python get_tool_catalog("") both raise PRESET_NOT_FOUND.
     Impact: Python can hide preset misconfiguration and diverges from Node/MCP fail-fast behavior.
     Recommended fix: default only when the preset key is absent/None; pass "" through to get_preset(""), and add parity coverage.
     Source: Codex reviewer + Claude reviewer + parent investigation.

harbournick (approved):
LGTM

PR #3547, opened 2026-05-28
fix: make toc toolbar icon configurable

VladaHarbour:
Toolbar configuration to re-enable:
```
modules: {
  toolbar: {
    showTableOfContentsButton: true,
  },
}
```


harbournick (approved):
LGTM

PR #3550, opened 2026-05-28
fix: sdt-related issues

luccas-harbour:
Fixes SD-3303, SD-3302 and SD-3304.

## Summary

Block structured-content (SDT) fields in the Template Builder were rendering a grey background highlight directly on the content element, which bled in front of content and persisted into output/print modes. This moves the block SDT fill onto the dedicated `::before` chrome layer so it sits *behind* content and is suppressed where it shouldn't appear, and folds in two related SDT interaction fixes uncovered alongside it.

### Block SDT background → chrome layer (primary fix)
- `styles.ts`: the `.superdoc-structured-content-block` element is now  `transparent` with `z-index: 0`; the fill (`--sd-content-controls-block-bg`)  is painted on its `::before` pseudo-element at `z-index: -1`, keeping the  background behind content while the border layer sits at `z-index: 1`.
- Viewing mode no longer relies on hover to show the chrome fill, and the  `::before` fill is cleared (`background: none`) in output/print modes so the  highlight doesn't leak into rendered documents.

### Inline SDT wrappers containing images
- `render-line.ts`: inline SDT wrappers that contain an image run are tagged  with `data-contains-inline-image` (in both the explicitly-positioned and  inline render paths).
- `styles.ts`: a scoped rule promotes only those wrappers to  `display: inline-block` / `vertical-align: top`, so image geometry and SDT  chrome size correctly while plain text inline SDTs stay `display: inline`.

### Block SDT wrapper deletion follows lock rules
- `structured-content-lock-plugin.js`: adds  `isAtBlockSdtWrapperDeletePosition`, which detects a caret at the very start  of a block SDT's first paragraph. Backspace/Delete there is now let through so  deletion is governed by the SDT's own lock mode (blocked when content-locked,  allowed otherwise) instead of being unconditionally swallowed.

## Tests
- `styles.test.ts`: asserts only image-bearing inline SDT wrappers get  inline-block geometry.
- `index.test.ts`: verifies the `data-contains-inline-image` marker is set when  an inline SDT contains an image and absent otherwise.
- `structured-content-lock-plugin.test.js`: parameterized coverage that  `contentLocked` block wrappers survive Backspace/Delete while  `sdtContentLocked` ones are deleted.
- `template-builder-block-field-styling.spec.ts`: updated to read the block  fill from the `::before` pseudo-element (element itself is now transparent).

### DIFF
--- a/packages/layout-engine/painters/dom/src/index.test.ts
+++ b/packages/layout-engine/painters/dom/src/index.test.ts
@@ -2738,6 +2738,7 @@ describe('DomPainter', () => {
     expect(wrapper.dataset.sdtScope).toBe('inline');
     expect(wrapper.dataset.sdtId).toBe('sc-inline-1');
     expect(wrapper.dataset.sdtTag).toBe('dropdown');
+    expect(wrapper.dataset.containsInlineImage).toBeUndefined();
 
     // The wrapper should span all contained runs (pmStart=7 to pmEnd=22)
     expect(wrapper.dataset.pmStart).toBe('7');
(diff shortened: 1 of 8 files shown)


caio-pizzol (approved):
lgtm

PR #3563, opened 2026-05-29
feat: add layered style export

artem-harbour:
Linear: SD-2559

Adds optional layered CSS support via `superdoc/style.layered.css` while keeping `superdoc/style.css` unchanged.

This PR introduces a build-time Vite plugin that emits `dist/style.layered.css` from the final bundled `style.css` by wrapping it in `@layer superdoc`, exposes the new entry in package exports, and updates theming/docs guidance for layered cascade usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,14 @@ const superdoc = new SuperDoc({
 });
 ```
 
+Optional layered CSS mode:
+
+```css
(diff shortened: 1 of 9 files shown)


harbournick (approved):
LGTM

PR #3566, opened 2026-05-29
fix(pm-adapter): render bibliography/index/TOA inside documentPartObject (SD-3066)

tupizz:
> [!NOTE]
> ### 📚 Stacked PRs — review bottom-up
>
> | Order | PR | Scope | |
> |:--:|---|---|---|
> | 1️⃣ | #3538 | Single-paragraph BIBLIOGRAPHY/INDEX/TOA crash (SD-3005) | |
> | 2️⃣ | #3565 | super-editor — field-code import fidelity + DRY | |
> | 3️⃣ | #3566 | pm-adapter — render fields inside `documentPartObject` | 👈 **this PR** |
>
> **Review order:** #3538 → #3565 → #3566. Each PR's base auto-retargets to `main` as the one below it merges.

---

## Summary

Word wraps a generated bibliography (and other block fields) in a `docPartObject` SDT, sometimes via a nested content control. The pm-adapter handler only converted `paragraph` and `tableOfContents` children to flow blocks, so the field's **entry paragraphs were silently dropped** — the heading rendered, the entries didn't.

> **Stacked on #3565** (which fixes the import side so these fields parse correctly). Base retargets to `main` as the stack merges.

Linear: [SD-3066](https://linear.app/superdocworkspace/issue/SD-3066)

## Fix
- `documentPartObject` now renders `bibliography`/`index`/`tableOfAuthorities` children via the shared `handleParagraphContainerNode`, and `structuredContentBlock` children via their handler.
- `structuredContentBlock` recurses block-field children (transparent wrapper).

## Section-counting invariant (the real hazard)
`findParagraphsWithSectPr` recurses `bibliography` but **not** `structuredContentBlock`. So:
- direct bibliography → handler advances `currentParagraphIndex` per entry (matches analysis);
- scb-wrapped → renders **without** advancing (matches analysis not recursing scb).

The prior code both dropped entries **and** under-counted, which could drift section breaks. Both paths are validated against the invariant.

Also extracts the shared `handleParagraphContainerNode` (previously three byte-identical copies in the bibliography/index/toa handlers).

## Test plan
- [x] Regression tests for both nesting shapes + the counter behavior
- [x] `pnpm --filter @superdoc/pm-adapter exec vitest run src/sdt src/sections src/internal.test.ts` — 302 pass
- [x] **Layout corpus (474 docs)**: 439 unchanged; the 35 changed are a `verticalAlign` schema-evolution + 7 image/table/section docs that contain **zero** docPartObj/bibliography — provably untouched by this change. No regressions attributable to the fix.
- [x] Browser-verified: `bibliography+document.docx` now renders the entries (matches Word)


harbournick (approved):
LGTM

PR #3567, opened 2026-05-29
feat(converter): preserve row-level content control wrappers in table round-trip (SD-3291)

luccas-harbour:
### Summary

Word lets a user wrap a whole table row in a content control (a `<w:sdt>` around a `<w:tr>`) — for example, a repeating section that controls a row. Before this change, the table importer read rows directly and ignored any row wrapped in `<w:sdt>`, so the wrapped row never entered the editable document data: it was missing from the editor, absent from text search, and dropped on export back to DOCX.

This is the row-level analog of the cell-level content control bug fixed in **SD-3289 / IT-1119**. This PR extends that fix to rows and factors the now-shared logic into common helpers so the two scopes stay in lockstep.

### What changed

**New shared helpers** — `super-converter/v3/handlers/w/sdt/helpers/sdt-envelope.js`
- `getSdtEnvelopeParts` — pulls `sdtPr` / `sdtEndPr` / `sdtContent` out of a `<w:sdt>` (replaces inline `find(w:sdtContent)` across five SDT handlers).
- `normalizeSdtContentChildren` — on import, unwraps a single child (`w:tr` or `w:tc`) from its `<w:sdt>` and preserves the envelope (`sdtPr`, `sdtEndPr`, and any `contentBefore` / `contentAfter` siblings) as opaque metadata.
- `wrapSdtContentChildren` — on export, reconstructs the `<w:sdt>` envelope around the exported child from that metadata.

**Row-level wiring**
- `tbl-translator.js` attaches `rowSdt` metadata on import and re-wraps rows on export; row enumeration (incl. the first-row width helper) now goes through the normalizer so wrapped rows are included.
- `table-row.js` registers a `rowSdt` schema attribute (`default: null`, `rendered: false`).
- `legacy-handle-table-cell-node.js` now finds vMerge continuations inside row-level SDT wrappers.

**Cell-level path** now routes through the same shared helpers (`tr-translator.js`, `row-cell-children.js`), and additionally gained `contentBefore` / `contentAfter` sibling preservation as a side effect of the extraction.

**Types** — `node-attributes.ts` introduces a shared `SdtMetadata<Scope>` with `CellSdtMetadata` / `RowSdtMetadata` aliases; adds `rowSdt` to `TableRowAttrs`.

### Tests
- `table-row-children.test.js` — normalizer unit tests (direct rows, single-row unwrap + metadata, content siblings, multi-row wrappers, empty SDTs).
- `tbl-translator.row-sdt.integration.test.js` — full encode → export round-trip with no mocks; wrapped row renders, text is findable, `<w:sdt>` reconstructed.
- `tbl-translator.test.js`, `tr-translator.cell-sdt.test.js` — encode/decode coverage for both scopes.
- `legacy-handle-table-cell-node.test.js` — vMerge across cell- and row-level SDT wrappers.

### Known limitations
- **Multi-row wrappers**: a single `<w:sdt>` wrapping more than one `<w:tr>` imports defensively — the rows survive but the wrapper metadata is dropped (SuperDoc doesn't yet model a single control spanning rows). Documented and tested.
- **Repeating sections**: the fix handles a `<w:sdt>` directly wrapping one `<w:tr>`. If a given Word version emits the nested `repeatingSection > repeatingSectionItem > w:tr` shape, that row is not yet unwrapped — worth confirming against a real fixture.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/handle-annotation-node.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/w/sdt/helpers/handle-annotation-node.js
@@ -1,6 +1,7 @@
 import { parseTagValueJSON } from './parse-tag-value-json';
 import { parseMarks } from '@converter/v2/importer/markImporter';
 import { generateDocxRandomId } from '@core/helpers/generateDocxRandomId';
+import { getSdtEnvelopeParts } from './sdt-envelope';
 
 /**
  * @param {Object} params
@@ -14,8 +15,7 @@ export function handleAnnotationNode(params) {
   }
 
   const node = nodes[0];
-  const sdtPr = node.elements.find((el) => el.name === 'w:sdtPr');
-  const sdtContent = node.elements.find((el) => el.name === 'w:sdtContent');
+  const { sdtPr, sdtContent } = getSdtEnvelopeParts(node);
 
   const sdtId = sdtPr?.elements?.find((el) => el.name === 'w:id');
   const alias = sdtPr?.elements.find((el) => el.name === 'w:alias');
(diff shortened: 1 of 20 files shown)


caio-pizzol:
V2 preserves the wrapped rows through editing and export in our checks. The separate row-height issues are tracked in SD-4979. Closing this V1 PR as we focus the remaining work on V2.

PR #3577, opened 2026-05-29
refactor(layout): centralize header/footer ref resolution in a shared contract (SD-2989)

luccas-harbour:
### Summary

Header/footer reference resolution — the OOXML inheritance logic that decides *which* variant (`default` / `first` / `even` / `odd`) applies to a page and *which* `rId` backs it — was duplicated across four call sites, each with subtly different fallback rules. This drift caused mismatches between the margins computed during pagination and the content actually painted, especially around `titlePg`, `evenAndOddHeaders`, and cross-section inheritance.

This PR extracts that logic into a single shared module in `@superdoc/contracts` and routes every consumer through it, so pagination, rendering, the presentation editor, and the document-api adapters all resolve refs identically.

### What changed

**New shared contract** (`packages/layout-engine/contracts/src/header-footer-resolution.ts`):

- `selectHeaderFooterVariantForPage(...)` — pure variant selection from `titlePg`, `alternateHeaders`, and the document/section page numbers. First page of a section with `titlePg` → `first`; alternating headers → `even`/`odd` by document page parity; otherwise `default`.
- `resolveEffectiveHeaderFooterRef(...)` — walks sections backward (toward section 0) to find the effective `rId` for a `(kind, variant)`, returning the matched section index and the *matched* variant. Encodes the OOXML rule that an `odd` request may fall back to `default` content, while `even` must not borrow `default`.

**Consumers now delegate to the shared resolver** (removing their bespoke inheritance/fallback blocks):

- `layout-bridge/headerFooterUtils.ts` — `getHeaderFooterTypeForSection`, `getHeaderFooterIdForPage`, and `resolveHeaderFooterForPageAndSection`. The identifier now carries an ordered `sections` array (built in `buildMultiSectionIdentifier`) for the resolver to consume.
- `layout-engine/index.ts` — per-page margin calculation during pagination. The old inline `getVariantTypeForPage` helper and the multi-step manual inheritance are replaced with calls to the shared functions; heights resolve to `0` when no ref applies.
- `presentation-editor/HeaderFooterSessionManager.ts` — drops its duplicated header/footer rId resolution and uses the matched variant for layout-slot lookup.
- `document-api-adapters/header-footer-refs-mutation.ts` — `resolveEffectiveRef` is now a thin wrapper over the shared resolver (and no longer needs the `editor` argument, simplifying its three call sites in the adapter, slot materialization, and story runtime).

### Notable fixes folded in

- **Sparse section metadata resolved by index, not array position** (`layout-engine/index.ts`): `getSectionMetadata` now matches on `section.sectionIndex` rather than the array slot, so documents with non-contiguous section metadata resolve correctly. Runtime section refs are tracked in a `Map` keyed by section index.
- **Converter-provided header refs preserved**: section 0 now merges fallback refs from the converter (`mergeSection0FallbackRefs`) so explicitly-imported refs aren't dropped during resolution.

### Tests

- New unit suite for the shared resolver (`header-footer-resolution.test.ts`, +121).
- Expanded coverage in `headerFooterUtils.test.ts` (+105), `layout-engine/index.test.ts` (+159), and `header-footer-refs-mutation.test.ts` (+46) covering variant selection, odd→default fallback, even-without-default, cross-section inheritance, and sparse metadata.

### DIFF
--- a/packages/layout-engine/layout-bridge/src/headerFooterUtils.ts
+++ b/packages/layout-engine/layout-bridge/src/headerFooterUtils.ts
@@ -388,10 +388,16 @@ export function getHeaderFooterTypeForSection(
   }
 
   if (identifier.alternateHeaders) {
-    // Keep parity-based variant selection even when this section doesn't
-    // explicitly define that variant. Resolution/inheritance happens later.
     if (!hasAny) return null;
-    return parityPageNumber % 2 === 0 ? 'even' : 'odd';
+    const parityVariant = parityPageNumber % 2 === 0 ? 'even' : 'odd';
+    return resolveInheritedHeaderFooterRef({
+      identifier,
+      sectionIndex,
+      kind,
+      variantType: parityVariant,
+    })
+      ? parityVariant
+      : null;
   }
 
   if (hasDefault) {
--- a/packages/layout-engine/layout-bridge/src/headerFooterUtils.ts
+++ b/packages/layout-engine/layout-bridge/src/headerFooterUtils.ts
@@ -356,7 +356,7 @@ export function buildMultiSectionIdentifier(
  * This function determines which header/footer variant (default, first, even, odd)
  * should be used for a given page number within a specific section. It respects:
  * - Per-section titlePg (first page of section uses 'first' variant)
- * - Alternate headers (even/odd pages based on section-aware page numbering)
+ * - Alternate headers (even/odd pages based on the effective Word page number)
  * - Fallback to default variant
  *
  * **Important**: When `titlePg` is enabled, this function returns 'first' even if the
@@ -365,7 +365,7 @@ export function buildMultiSectionIdentifier(
  * sections. The rendering layer is responsible for resolving the actual content ID
  * through inheritance fallback logic.
  *
- * @param pageNumber - Physical page number (1-indexed)
+ * @param pageNumber - Effective Word page number (1-indexed), after section page numbering settings
  * @param sectionIndex - Index of the section this page belongs to
  * @param identifier - Multi-section identifier with per-section mappings
  * @param options - Optional settings (kind, sectionPageNumber, parityPageNumber)
@@ -442,10 +442,10 @@ export function getHeaderFooterIdForPage(
   const kind = options?.kind ?? 'header';
   const sectionIndex = page.sectionIndex ?? 0;
   const sectionPageNumber = options?.sectionPageNumber ?? page.number;
-  const parityPageNumber = options?.parityPageNumber ?? page.displayNumber ?? page.number;
+  const effectivePageNumber = options?.parityPageNumber ?? page.effectivePageNumber ?? page.displayNumber ?? page.number;
   const sectionTitlePg = getSectionTitlePg(identifier, sectionIndex);
   const variantType = selectHeaderFooterVariantForPage({
-    documentPageNumber: parityPageNumber,
+    documentPageNumber: effectivePageNumber,
     sectionPageNumber,
     titlePg: sectionTitlePg,
     alternateHeaders: identifier.alternateHeaders,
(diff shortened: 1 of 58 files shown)


packages/layout-engine/layout-engine/src/index.ts
```diff
@@ -1383,77 +1386,42 @@ export function layoutDocument(blocks: FlowBlock[], measures: Measure[], options
...
-        // Use effectiveVariantType for header height lookup to match the fallback
-        const headerHeight = getHeaderHeightForPage(effectiveVariantType, headerRef, activeSectionIndex);
-        const footerHeight = getFooterHeightForPage(
-          variantType !== 'default' && !activeSectionRefs?.footerRefs?.[variantType] ? 'default' : variantType,
-          footerRef,
-          activeSectionIndex,
-        );
+        const headerHeight = headerResolved
+          ? getHeaderHeightForPage(headerResolved.matchedVariant, headerResolved.refId, activeSectionIndex)
+          : 0;
```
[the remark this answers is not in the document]
luccas-harbour: Fixed

packages/layout-engine/layout-engine/src/index.ts
```diff
@@ -1383,77 +1390,48 @@ export function layoutDocument(blocks: FlowBlock[], measures: Measure[], options
...
+        const sectionMetadata = getSectionMetadata(activeSectionIndex);
         const titlePgEnabled = sectionMetadata?.titlePg ?? false;
         const alternateHeaders = options.alternateHeaders ?? false;
 
-        // Determine which header/footer variant applies to this page
-        const variantType = getVariantTypeForPage({
+        // Determine which header/footer variant applies to this page.
+        const variantType = selectHeaderFooterVariantForPage({
           sectionPageNumber,
           documentPageNumber: newPageNumber,
```
tupizz: blocking on this one: this should use the effective Word page number, not the physical page index. `evenAndOddHeaders` follows section page numbering, including `w:pgNumType/@start`; so a section on physical page 3 that restarts at 2 should pick `even`, but this passes 3 and picks `odd`. layout-engine already tracks `activePageCounter`, so let's pass that through and add a restart-at-2 test. Otherwise we resolve the wrong rId and reserve the wrong header/footer height.

packages/super-editor/src/editors/v1/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
```diff
@@ -2379,40 +2389,25 @@ export class HeaderFooterSessionManager {
         ? getHeaderFooterTypeForSection(pageNumber, sectionIndex, multiSectionId, { kind, sectionPageNumber })
```
tupizz: same issue leaks through here: this is still passing the physical `pageNumber` into the section resolver. For restarted numbering it can render/materialize a different slot than Word. Once the shared selector takes the effective numbered page, this caller should feed that value too.

packages/layout-engine/layout-bridge/src/headerFooterUtils.ts
```diff
@@ -167,8 +176,44 @@ export const defaultMultiSectionIdentifier = (): MultiSectionHeaderFooterIdentif
...
+    ...Array.from(identifier.sectionFooterIds.keys()),
+    ...Array.from(identifier.sectionTitlePg.keys()),
+  );
+
+  const sections: HeaderFooterResolutionSection[] = [];
+  for (let sectionIndex = 0; sectionIndex <= maxIndex; sectionIndex += 1) {
+    sections.push({
+      sectionIndex,
+      titlePg: identifier.sectionTitlePg.get(sectionIndex) ?? false,
+      headerRefs: identifier.sectionHeaderIds.get(sectionIndex),
```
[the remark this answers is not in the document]
luccas-harbour: I don’t think we should copy converter refs into `identifier.sections`; that reintroduces the section-aware leak for explicit empty section refs. Converter refs remain on the legacy fields and are used only by the decoration fallback when no section refs object exists for that kind. Explicit empty refs should resolve blank.

packages/layout-engine/layout-engine/src/index.ts
luccas-harbour: good call! I fixed it.

packages/super-editor/src/editors/v1/core/presentation-editor/header-footer/HeaderFooterSessionManager.ts
luccas-harbour: fixed

tupizz (approved):
Lgtm! Great work as always

Issue #3586, opened 2026-05-30
MCP: expose block-node deletion (blocks.delete / blocks.deleteRange) via superdoc_edit


jacobjove:
### What problem does this solve?

The `apps/mcp` server can delete text (`superdoc_edit action:"delete"`, `superdoc_mutations text.delete`) but has no way to remove an entire block node. Deleting a heading's or paragraph's text leaves the empty `<w:p>` container behind — it still renders as block/heading spacing — so an MCP client (e.g. an AI agent running a redline/cleanup pass) can't remove a clause, heading, or stray empty paragraph.

The capability already exists in `document-api`: `blocks.delete` and `blocks.deleteRange` are implemented, tested, and wired through the super-editor adapter (they physically remove the container via `deleteBlockNodeById` / `tr.delete()`), merged in #3444. They're just not exposed on the MCP tool surface: their `operation-definitions.ts` entries carry no `intentGroup`/`intentAction`, so codegen skips them.

### Proposed solution

Expose both as actions on the existing `superdoc_edit` tool (`intentGroup: 'edit'`):
- `blocks.delete` → `delete_block`, input `{ target: { kind:'block', nodeType, nodeId } }`
- `blocks.deleteRange` → `delete_block_range`, input `{ start, end }`

This is the smallest additive change: two field-pairs on the contract plus regenerated catalog/dispatch, no new tool, no plan-engine work. The `delete_block` name is forced by a collision, since the existing text-delete op already owns `intentAction:'delete'` in the `edit` group.

A draft PR implementing exactly this is up (#3587): regenerated MCP catalog plus Node/browser/Python dispatch, and end-to-end `apps/mcp` protocol tests proving physical block removal (block-count decrement + nodeId absence). Opened as draft to confirm the surface choice before marking ready.

### Alternatives considered

- **Dedicated `superdoc_blocks` tool** (`action:"delete"` / `"delete_range"`). Matches the structural-delete precedent (`lists.delete`→`superdoc_list action:"delete"`, `tables.delete`→`superdoc_table action:"delete"`) and avoids the `delete` collision, but `blocks.list` already lives under `superdoc_get_content`, so this yields a 2-op tool that can't list; a whole new tool for two ops is heavier.
- **`superdoc_mutations` `block.delete` step.** Enables atomic batching of block removal within a multi-step redline, but requires extending the plan-engine step framework. A sound follow-up, not v1.

Happy to rework toward either if you'd prefer.

### Additional context / limitations

[...]

PR #3587, opened 2026-05-30
feat(mcp): expose blocks.delete/deleteRange via superdoc_edit

jacobjove:
## Summary

Wires the existing `document-api` `blocks.delete` / `blocks.deleteRange` ops into the MCP tool surface as two new `superdoc_edit` actions — `delete_block` and `delete_block_range` — so an MCP client can remove an entire block node, not just its text. Closes the gap where deleting a heading/paragraph's text leaves an empty container that still renders as block spacing.

Proposed in #3586 (surface options A/B/C discussed there). This PR implements option A.

## What changed

- **Contract** (`packages/document-api/src/contract/operation-definitions.ts`): `intentGroup:'edit'` + `intentAction:'delete_block'` on `blocks.delete`; `intentGroup:'edit'` + `intentAction:'delete_block_range'` on `blocks.deleteRange`; one disambiguation sentence and one `inputExample` in `INTENT_GROUP_META.edit`.
- **Regenerated** via `pnpm run generate:all`: `apps/mcp/src/generated/{catalog,intent-dispatch.generated}.ts`, `packages/sdk/langs/browser/src/intent-dispatch.ts`, `packages/sdk/tools/intent_dispatch_generated.py`.
- **Tests** (`apps/mcp/src/__tests__/protocol.test.ts`): action-enum assertion plus three end-to-end cases asserting physical block removal (block-count decrement + nodeId absence) — one demonstrating the text-delete-leaves-empty-container bug vs `delete_block` removing it, and one asserting `delete_block_range` removes a contiguous top-level span (`deletedCount` + spanned nodeIds absent, with a neighbor outside the range surviving so the delete is proven bounded).

## Addressing

`delete` (existing) takes a text `ref`; `delete_block` takes `target:{kind:'block',nodeType,nodeId}`; `delete_block_range` takes `start`/`end` (each a `{kind:'block',nodeType,nodeId}` top-level block, inclusive). The `delete_block` name is forced by the existing `delete` collision in the `edit` group.

## Verification

`apps/mcp` 41 pass (4 new) · `document-api` 1480 pass (blocks + contract-parity) · conformance 1198 pass · codegen incl. cross-lang-parity (Node↔Python) 57 pass · typecheck 0 new errors vs baseline.

## Limitations

- `supportsTrackedMode: false`, so block-node removal isn't recordable as a tracked change (cf. #3343 for tables).
- Does not add a `superdoc_mutations` `block.delete` step (option C); a sound follow-up.
- `blocks.delete` overlaps `tables.delete`/`lists.delete` as the type-agnostic primitive; the type-specific tools remain.
- The `delete_block_range` contract entry carries the disambiguation sentence but not its own `inputExample` (the `edit` group example covers `delete_block`); deferred until the maintainer surface decision in #3586 (option A vs a dedicated `superdoc_blocks` tool vs a `superdoc_mutations` `block.delete` step) is settled, to avoid regenerating prose that a surface change would rewrite.

### DIFF
--- a/apps/mcp/src/__tests__/protocol.test.ts
+++ b/apps/mcp/src/__tests__/protocol.test.ts
@@ -203,4 +203,231 @@ describe('MCP protocol integration', () => {
     const body = JSON.parse(textContent(result));
     expect(body).toHaveProperty('session_id');
   });
+
+  // ---------------------------------------------------------------------------
+  // Block-node deletion via superdoc_edit (delete_block / delete_block_range)
+  //
+  // These tests prove the *reported bug* is fixed end-to-end through the MCP
+  // layer: a text `delete` (by ref) empties a block but leaves the container
+  // behind, whereas `delete_block` (by {kind:'block', nodeType, nodeId} target)
+  // physically removes the whole block node.
+  //
+  // Addressing model (must be respected or the test fails):
+  //   - get_content action:"blocks" returns { total, blocks: [...], revision };
+  //     each entry exposes BOTH a `nodeId` and a `ref`.
+  //   - text `delete` takes a `ref`.
+  //   - `delete_block` takes `target: {kind:'block', nodeType, nodeId}`.
+  //   - refs and block handles EXPIRE after ANY mutation; we re-fetch blocks
+  //     before every call that consumes a handle.
+  // ---------------------------------------------------------------------------
+
+  interface BlockEntry {
+    nodeId: string;
+    nodeType: string;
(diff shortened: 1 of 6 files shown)


caio-pizzol:
Thanks for the contribution, Jacob! We reproduced the block-deletion gap in the released MCP package. The newer core path handles selected paragraph deletion on main, but it hasn’t shipped, and we haven’t established equivalent range support. We’re keeping this PR open while we work through those remaining gaps.

PR #3595, opened 2026-06-01
fix(pickled): correct toolbar scenario, add 3 external legibility scenarios

caio-pizzol:
Follow-up to #3494. Fixes a scoring bug in the merged toolbar scenario and grows the external suite from 1 to 4 scenarios.

The toolbar scenario shipped with two false-negatives, both confirmed against a real run: `expected.excludes` banned `createHeadlessToolbar` even though the prompt asks "what should I avoid?" (so a correct answer naming it as the thing to avoid failed), and it required `useSuperDocUI`, which is a real export but absent from the docs bundle (so docs-grounded answers failed). Now scored positives-only: `SuperDocUIProvider` + `superdoc/ui/react`.

Three new scenarios, each with terms verified present in `docs.superdoc.dev/llms-full.txt` before locking:
- Programmatic edits: Document API vs UI (`Document API` + `editor.doc`)
- Enable real-time collaboration (`Yjs` + `modules.collaboration`)
- Is `createHeadlessToolbar` right for new React UI? (positive check on `superdoc/ui/react`, no traps)

Two candidates were dropped because the docs bundle contains no crisp deterministic term for them (built-in toolbar `customButtons`, an export API name) - requiring absent terms would false-negative the docs cells.

Validated before opening: config loads, each scenario's docs cell scores YES on real output, and two sampled paid passes (both interfaces, every toolset) score ~66 with tool-use provenance verified on every web/MCP cell and no false-fires.

One product finding worth a separate task: on web/MCP discovery (no injected docs), agents reliably reach the `superdoc/ui/react` import path but often not the `SuperDocUIProvider` component, and at least one cell recommended the Document API for a React toolbar - conflating the mutation surface with the UI surface. That points at a docs/MCP discoverability gap, not a config issue.

No CI wiring or secrets in this PR; runs are manual for now.

### DIFF
--- a/pickled.yml
+++ b/pickled.yml
@@ -2,9 +2,8 @@
 #
 # Pickled runs the scenarios below across a matrix of interfaces,
 # sources, and toolsets, then scores answers with deterministic checks.
-# Each cell isolates one context-delivery path (docs link / web tools /
-# MCP server) so the report shows where agents do well and where they
-# do not.
+# Each cell isolates one context-delivery path so the report shows where
+# agents do well and where they do not.
 #
 # Quick start:
 #   bunx @pickled-dev/cli check .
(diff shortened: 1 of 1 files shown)


caio-pizzol:
Closing in favor of #3601, which removes the root pickled.yml. Pickled's config schema changed in the latest CLI (new product/sources/agents/access/questions/checks model; traps removed), so these scenario corrections would land on a schema that no longer loads. The suite will be reintroduced deliberately later, and this pass's learnings fold into that.

PR #3596, opened 2026-06-01
fix: update marker on font change

VladaHarbour:


packages/super-editor/src/editors/v1/core/layout-adapter/list-marker-font.ts
```diff
@@ -0,0 +1,133 @@
...
+  const contentFont = resolveContentFont(block, para, contentFontSource);
+  if (!contentFont) return;
+
+  const preserveMarkerFontFamily = numberingDefinesMarkerFontFamily(block.attrs?.numberingProperties, converterContext);
+
+  if (!preserveMarkerFontFamily && contentFont.fontFamily) {
+    markerRun.fontFamily = contentFont.fontFamily;
+  }
+  if (contentFont.fontSize) {
+    markerRun.fontSize = contentFont.fontSize;
```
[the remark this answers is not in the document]
VladaHarbour: The same concern as in codex review above. addressed

packages/super-editor/src/editors/v1/core/layout-adapter/converters/paragraph.ts
```diff
@@ -604,6 +605,19 @@ export function paragraphToFlowBlocks({
   const defaultSize =
     usePreviousFont && previousParagraphFont.fontSize ? previousParagraphFont.fontSize : extracted.defaultSize;
 
+  const finalizeParagraphBlocks = (outputBlocks: FlowBlock[]): FlowBlock[] => {
+    outputBlocks.forEach((block) => {
+      if (block.kind === 'paragraph') {
+        syncListMarkerFontFromParagraphRuns({
```
luccas-harbour, line 611: the fresh-path sync overwrites the marker font from body text whenever numbering doesn't define it — it drops the original `!hasExplicitParagraphRunProperties` guard from `attributes/paragraph.ts`. for a doc whose paragraph-mark pPr/rPr sets a marker font different from the body text (and not via numbering), this clobbers the imported marker on every render with no user edit. can we gate the family overwrite on `!hasExplicitParagraphRunProperties` too, and add a fixture for pPr/rPr marker font ≠ body?

packages/super-editor/src/editors/v1/core/layout-adapter/list-marker-font.ts
```diff
@@ -0,0 +1,155 @@
...
+): { fontFamily: boolean; fontSize: boolean } => {
+  const numId = numberingProperties?.numId;
+  if (numId == null || numId === 0 || !converterContext) {
+    return { fontFamily: false, fontSize: false };
+  }
+  const ilvl = numberingProperties?.ilvl ?? 0;
+  const numberingRunProps = getNumberingProperties<RunProperties>('runProperties', converterContext, ilvl, numId);
+  return {
+    fontFamily: numberingRunProps.fontFamily != null,
+    fontSize: numberingRunProps.fontSize != null || numberingRunProps.fontSizeCs != null,
```
luccas-harbour: the `fontSize` half of this guard is stricter than the family guard needs to be — Symbol/Wingdings glyphs care about family, not size, and Word still scales the marker glyph with the run size. for an imported list whose level rPr carries a `w:sz`/`w:szCs`, this makes "restyle marker on font-size change" (a core SD-3238 AC) a no-op. was guarding size intentional, or should only family be protected?

packages/super-editor/src/editors/v1/core/layout-adapter/list-marker-font.ts
```diff
@@ -0,0 +1,155 @@
...
+  if (typeof fontFamily === 'string' && fontFamily.trim().length > 0) {
+    partial.fontFamily = fontFamily.trim();
+  }
+  if (typeof fontSize === 'number' && Number.isFinite(fontSize) && fontSize > 0) {
+    partial.fontSize = fontSize;
+  }
+  return Object.keys(partial).length > 0 ? partial : undefined;
+};
+
+const getFontFromRuns = (runs: ReadonlyArray<Run>): Partial<ParagraphFont> | undefined => {
```
luccas-harbour, line 37: `getLastParagraphFont` was patched to skip empty (`text === ''`) first runs but this one wasn't — `mergeAdjacentRuns` doesn't drop empty runs, so a leading zero-length run with a stale font can drive the marker sync. worth skipping empty text runs here too for consistency.

packages/super-editor/src/editors/v1/core/layout-adapter/converters/paragraph.ts
```diff
@@ -1133,15 +1150,26 @@ export function handleParagraphNode(node: PMNode, context: NodeHandlerContext):
...
+      // previousParagraphFont instead of falling back to stale cached runs.
       const delta = pmStart - cached.pmStart;
       const reusedBlocks = shiftCachedBlocks(cached.blocks, delta);
+      const paragraphProps = node.attrs?.paragraphProperties as ParagraphProperties | undefined;
+      const previousParagraphFont = !hasExplicitParagraphRunProperties(paragraphProps)
+        ? getLastParagraphFont(blocks)
+        : undefined;
+      reusedBlocks.forEach((block) => {
+        if (block.kind === 'paragraph') {
+          syncListMarkerFontFromParagraphRuns({
```
luccas-harbour, line 1164: on cache hits `resolveContentFont` ignores the converted runs and uses `fromPara` (textStyle marks only) ?? `previousParagraphFont`. for a non-empty item whose font comes from the paragraph/style default rather than a run-level mark, `fromPara` is undefined, so the marker inherits the *previous* paragraph's font — which diverges from the fresh path that reads the resolved run font. narrow, but worth a guard or a test for list items with differing style-default fonts and no textStyle mark.

packages/super-editor/src/editors/v1/core/layout-adapter/list-marker-font.ts
```diff
@@ -0,0 +1,155 @@
...
+  const fromPara = para ? getFontFromParagraphContent(para) : undefined;
+  if (source === 'paragraph') {
+    // Cache hits must not fall back to stale cached runs. Empty list items have no
+    // textStyle marks, so inherit from the preceding paragraph when available.
+    return mergeContentFont(fromPara, previousParagraphFont);
+  }
+  return fromRuns ?? fromPara;
+};
+
+const getNumberingMarkerFontOverrides = (
```
luccas-harbour: this duplicates the numbering Symbol/Wingdings guard in `attributes/paragraph.ts:424-441`, and the two have already drifted — that one checks family only, this one also checks `fontSize`/`fontSizeCs`. two sources of truth for "is the marker font pinned by numbering." can we extract one shared helper both consume?

luccas-harbour:
hey @VladaHarbour! I had a look and found a few things that I left as inline comments. thanks!

packages/super-editor/src/editors/v1/core/layout-adapter/list-marker-font.ts
VladaHarbour: created shared helper

packages/super-editor/src/editors/v1/core/layout-adapter/converters/paragraph.ts
VladaHarbour: must be fixed

packages/super-editor/src/editors/v1/core/layout-adapter/list-marker-font.ts
VladaHarbour: fixed

packages/super-editor/src/editors/v1/core/layout-adapter/list-marker-font.ts
VladaHarbour: Size guard was over-broad, changing it to font-family only

packages/super-editor/src/editors/v1/core/layout-adapter/converters/paragraph.ts
VladaHarbour: Restored the explicit-pPr guard for fresh-path sync

luccas-harbour (approved):
Thanks for addressing the comments!
LGTM

PR #3606, opened 2026-06-02
feat(painter-dom): Register chart renderer as a feature module

ggdaltoso:
I've been studying the codebase and noticed that `chart-renderer.ts`  was already living in its own file but wasn't registered in `feature-registry.ts`, unlike `inline-direction` and `math`, which follow the feature module pattern described in `CLAUDE.md`.

This PR wires it up. No logic was changed; just the import path in `renderer.ts` now goes through the feature barrel, and charts are finally discoverable via the registry.

(Genuinely impressed by how well-documented and thought-through the architecture is, btw. Learned a lot just reading through it!)


caio-pizzol:
Thanks for the thoughtful contribution, Gabriel! Making the chart renderer easier to discover is a useful DX improvement. We verified that this organization gap still exists in V2, and we’d like to carry your work forward on main through the internal mirror.

PR #3609, opened 2026-06-02
SD-3357 - fix: dedupe tracked changes from merged activity feed

chittolinag:
Tracked changes were being displayed twice in the sidebar.


demos/editor/custom-ui/src/components/ActivitySidebar.tsx
```diff
@@ -70,9 +70,31 @@ export function ActivitySidebar({ composeOpen, onCloseComposer, decided }: Props
...
+  // suggestion shows twice (one comment card + one change card).
+  //
+  // We dedupe by id, NOT by the `trackedChange` flag: a real comment
+  // thread also gets `trackedChange: true` when its anchor overlaps a
+  // suggestion (`comments.list()` links it via `assignTrackedChangeLink`),
+  // so a flag filter would wrongly hide those discussions. Only the
+  // synthetic rows reuse a tracked-change id; real comments have their
+  // own. `pairedWithChangeId` covers replacement pairs that
+  // `trackChanges.list()` collapses into one row (this demo runs
+  // `replacements: 'paired'`) but which still surface as two comments.
```
andrii-harbour: The id-dedupe approach is definitely the right call here (verified that real comment threads overlapping a suggestion do get `trackedChange: true` via `assignTrackedChangeLink` — `comments-wrappers.ts:262`, `:661` — so the flag filter would have hidden real discussions 👍), but this last paragraph has two inaccuracies a future reader might copy:

1. *"this demo runs `replacements: 'paired'`"* — it actually runs `'independent'` (`EditorMount.tsx:26`). The dedupe works here via the plain `tc.id` branch: independent mode emits both sides of a replacement as separate change rows, so both synthetic comment ids land in the set.
2. *"`pairedWithChangeId` covers replacement pairs"* — the engine never populates that field (see comment below).

Suggested rewrite of this paragraph:

```
// own. Note: this demo runs `replacements: 'independent'`, so both
// sides of a replacement surface as separate change rows and both
// synthetic comment ids land in the dedupe set. Under the default
// `'paired'` mode the collapsed row drops the delete-side id, which
// `trackChanges.list()` does not currently expose — the delete-side
// synthetic comment would leak as a duplicate. Engine follow-up needed
// before this pattern is safe in paired mode.
```

demos/editor/custom-ui/src/components/ActivitySidebar.tsx
```diff
@@ -70,9 +70,31 @@ export function ActivitySidebar({ composeOpen, onCloseComposer, decided }: Props
...
   const feed = useMemo<ActivityItem[]>(() => {
+    const changeIds = new Set<string>();
+    for (const tc of trackChanges.items) {
+      changeIds.add(tc.id);
+      if (tc.change.pairedWithChangeId) changeIds.add(tc.change.pairedWithChangeId);
```
andrii-harbour: Small one: this branch is currently dead code — `pairedWithChangeId` is never populated by the engine. Every projection path passes `null` (`track-changes-wrappers.ts:164`, `:191`, `:243` — including the collapsed paired row), and the list handler at `:466`/`:483` just passes that through. The engine's own tests pin it: collapsed paired row → `pairedWithChangeId: undefined` (`track-changes-wrappers.test.ts:879`), independent rows → `undefined` (`:934-937`).

Harmless at runtime, but it reads as paired-mode safety that doesn't exist yet. Suggest dropping the line, or keeping it with an explicit "currently never populated; forward-compat for when list() exposes the collapsed delete-side id" note.

andrii-harbour (changes requested):
Nice job! Minor comments here.

P.S.  PR title typo: title says SD-3557; the ticket is SD-3357

chittolinag:
@andrii-harbour thanks for the review! updated the ticket title and also pushed a commit removing dead code + adding comments.

harbournick (approved):
LGTM

PR #3610, opened 2026-06-02
SD-3358 - tracked deletion is captured char by char in suggesting mode

chittolinag:
### Issue

In suggesting mode, deleting a contiguous run keystroke-by-keystroke (Backspace/Delete) created one tracked change per character instead of one logical deletion. It also re-fragmented at run seams - e.g. Google-Docs multi-run paragraphs where comment-anchor markers sit between runs. (SD-3358)

### Proposed solution

When a tracked deletion is adjacent to the same author's existing unresolved deletion — directly, or separated only by zero-width markers (run boundaries, comment/bookmark anchors) with no live text between - reuse that change's id so the run coalesces into a single logical deletion. Mirrors the existing same-author insertion refinement. Covered by a new behavior test.


packages/super-editor/src/editors/v1/extensions/track-changes/review-model/overlap-compiler.js
```diff
@@ -274,6 +274,65 @@ const findAdjacentInsertedSegment = (ctx, pos) => {
...
+ *
+ * Adjacency is checked exactly first (contiguous within a single run), then
+ * across an empty structural gap. Multi-run paragraphs — e.g. Google Docs
+ * exports that split "Open comment " and "from Google Docs." into separate
+ * runs — separate the prior deletion from this range by run open/close tokens
+ * with no intervening text; without the gap-tolerant pass, deleting the space
+ * at the run seam would mint a new change. The gap is bounded and gated on
+ * `isEmptyStructuralGap`, so a live character between two deletions still
+ * splits them. Mirrors the insertion refinement in `compileTextInsert`
+ * (`findAdjacentInsertedSegment` + `findSegmentAcrossEmptyStructuralGap`).
```
andrii-harbour: Two claims here drifted from the code, and one will mislead future readers:

1. *"The gap is bounded and gated on `isEmptyStructuralGap`"* — it's gated on `isCoalescibleDeletionGap`, which is deliberately *more permissive* (tolerates zero-width anchors).
2. *"Mirrors the insertion refinement"* — it intentionally diverges from it: `findSegmentAcrossEmptyStructuralGap` still rejects **any** inline leaf, so the insertion side splits at exactly the GD comment-anchor seams this PR fixes for deletions. Worth noting here that TC-EDIT-018 (labs PR #6) covers *"deleted or inserted"* — so the insertion side is a known conformance gap, not a mirror. Fine to leave for a follow-up ticket, but the comment shouldn't claim symmetry that isn't there.

packages/super-editor/src/editors/v1/extensions/track-changes/review-model/overlap-compiler.js
```diff
@@ -291,6 +350,49 @@ const isEmptyStructuralGap = (ctx, from, to) => {
...
+ * them. They must not block deletion coalescing — e.g. Google Docs anchors
+ * comments with inline marker nodes that sit between runs.
+ */
+const ZERO_WIDTH_ANCHOR_NODE_NAMES = new Set([
+  'commentRangeStart',
+  'commentRangeEnd',
+  'commentReference',
+  'bookmarkStart',
+  'bookmarkEnd',
+]);
```
andrii-harbour, line 391: Three notes on the allowlist:

1. **`bookmarkStart` is inert** — it's defined with `content: 'inline*'` (`bookmark-start.js:21`), so it's not a leaf and never reaches the `.has()` check. Either remove it or note why it's listed.
2. **`permStart` / `permEnd` are missing** — they're zero-width inline atoms exactly like the comment/bookmark markers (`perm-start.js:36-37`, `perm-end.js:30-31`), so a deletion spanning a permission-range seam will still split. Same question for `fieldAnnotation`.
3. Whatever the decision, worth a line documenting why those types are in/out — the jsdoc currently implies generic "review/anchor markers" coverage.

packages/super-editor/src/editors/v1/extensions/track-changes/review-model/overlap-compiler.js
```diff
@@ -274,6 +274,65 @@ const findAdjacentInsertedSegment = (ctx, pos) => {
...
+  const sameUserDeleted = (segment) =>
+    segment.side === SegmentSide.Deleted &&
+    !segment.attrs?.replacementGroupId &&
+    !segment.attrs?.overlapParentId &&
+    isSameUserForRefinement(ctx, segment);
+
+  const exactLeft = ctx.graph.segments.find((segment) => sameUserDeleted(segment) && segment.to === from);
+  if (exactLeft) return exactLeft;
+  const exactRight = ctx.graph.segments.find((segment) => sameUserDeleted(segment) && segment.from === to);
+  if (exactRight) return exactRight;
```
andrii-harbour, line 329: Known limitation worth documenting (or a follow-up ticket): deleting the single live char **between two of your own deletions** hits `exactLeft` first, reuses the left id, and the right deletion is never consulted. Since `mergeAdjacentSpans` only merges same-id spans, the result is two touching logical deletions where Word shows one. Strictly better than pre-fix (which minted a third id), but short of TC-EDIT-018's "one logical change."

packages/super-editor/src/editors/v1/extensions/track-changes/review-model/overlap-compiler.js
```diff
@@ -543,16 +645,37 @@ const compileTextDelete = (ctx, intent) => {
...
+  // (e.g. holding Backspace) surfaces as ONE logical tracked deletion rather
+  // than one review object per keystroke — mirroring the same-user insertion
+  // refinement in compileTextInsert (TC-EDIT-018). Replacement-driven deletes
+  // keep their caller-provided pairing id; preserve-review-state edits never
+  // fold into an existing change.
+  const coalesceTarget =
+    intent.replacementGroupHint || intent.preserveExistingReviewState
+      ? null
+      : findAdjacentDeletedSegment(ctx, intent.from, intent.to);
+  const sharedDeletionId = intent.replacementGroupHint || coalesceTarget?.changeId || null;
```
andrii-harbour, line 697: Date question to verify before merge: the coalesced run's mark is written with the **current** edit's date while the existing runs keep theirs — one changeId, mixed dates. The graph keys by id so it stays one change, but (a) which `w:date` wins on export when one logical deletion spans `w:del` runs with different dates, and (b) does the side panel read a stable date for the change? This also applies when coalescing into an *imported* same-author deletion from an older session (which is correct Word behavior to allow — just makes the date spread wider).

andrii-harbour (commented):
Well done! I have only optional comments, up to you to address them or merge as is

packages/super-editor/src/editors/v1/extensions/track-changes/review-model/overlap-compiler.js
chittolinag: thanks! just took a look.

1. export -> each run becomes a `w:del` with the same `w:id`, and `mergeConsecutiveTrackedChanges` merges them by `w:id `and keeps the first wrapper's attrs → one `w:del` with the first run's `w:date`.

2. panel: `buildLogicalChange` pulls the date from `primary = segments[0]` (first segment), which is the same first-run date export uses - so they always agree.

also not really new: same-author insert refinement already does the "one changeId, mixed run dates" thing and round-trips fine through this same merge. The imported older-session case is the same thing - both export and panel just take the first segment's date, so the wider spread is internal only.

added a comment at the coalescing spot so it's documented anyways!

harbournick (approved):
LGTM

PR #3611, opened 2026-06-02
fix(layout-engine): render underlined tabs flush with text underline (SD-3330)

tupizz:
## What

Underlined tab characters now render their underline **flush with the text underline**, producing one continuous line as Word does. Fixes **SD-3330** and delivers the core of **SD-3347** (Word-style underlined fill-in / signature lines).

Linear: SD-3347 (parent feature), SD-3330 (child bug). Related: SD-1289, SD-1499, SD-3103.

## The bug

In a DOCX where a run carries `<w:u w:val="single"/>` and contains `<w:tab/>`, Word draws a single continuous underline under the text and the tab advance. SuperDoc drew the tab's underline **~8px below** the text underline, so the line looked broken where text met tabs.

**Root cause:** the tab underline is a `border-bottom` on the tab box. The box was the full line height and `vertical-align: bottom`, so the border sat at the **bottom of the line box (including leading)** — not at the text baseline where `text-decoration` underlines render. ECMA-376 §17.3.2.40 specifies the underline appears "directly below the character height (less all spacing above and below the characters on the line)", i.e. at the baseline, not the line-box bottom.

## The fix

`painters/dom/src/runs/tab-run.ts`: for **underlined** tabs, pin the box top to the line-box top and end it at a baseline offset derived from the resolved line metrics (`ascent`/`descent`/`lineHeight`), so the `border-bottom` lands flush with the adjacent text underline. No DOM measurement is added (honors the painter's SD-2957 no-measure invariant).

The change is **gated to underlined tabs only** — non-underlined tabs keep their previous geometry, so tab-stop positioning and nearby line layout are unchanged.

## Validation (Layout Engine mode)

Measured the vertical step between the tab underline and the text underline on `underlined tab stops (1).docx`: **+8px → −1px** (visually continuous).

| Fixture | Result |
|---|---|
| `underlined tab stops (1)` | continuous flush line (SD-3330 repro) ✅ |
| `tab-underlines` | clean blank signature fill-in lines ✅ |
| `sd-1289-tabs-with-lines` | `By:/Name:/Title:` lines flush with labels (were floating low) ✅ |
| `tab-stops-test-signer-area` | tab-stop positioning unchanged ✅ |
| Asset Purchase Agreement, NVCA SPA ×2 (57K–128K chars) | load without crash, tabs render normally ✅ |

- New regression test `runs/tab-run.test.ts` (inline + positioned paths): RED on `main`, GREEN with the fix.
- Full `@superdoc/painter-dom` suite: **1183 passed**. `tsc -b` typecheck clean.

## Notes / follow-ups

- **SD-3330 part 2** ("can't underline tabbed spaces in the editor") does **not** reproduce on current `main`: `toggleUnderline`, the toolbar, mark application, and `isActive('underline')` all handle tab nodes correctly. It appears to have been fixed since the reported v1.38.0; only the rendering remained.
- None of the supplied fixtures use right/center/decimal tab stops, so the positioned-tab path is covered by the unit test rather than a real fixture. Worth a spot-check if such a doc surfaces.


harbournick (approved):
LGTM

PR #3615, opened 2026-06-03
chore(deps): bump uuid to ^11.1.1 (CVE-2026-41907)

mattConnHarbour:
Updates SuperDoc's direct uuid dependency from ^9.0.1 to ^11.1.1 in the workspace catalog to address the npm audit warning for SNYK-JS-UUID-16133035 (CVE-2026-41907). SuperDoc was not actually vulnerable: we only use the 2-param signature, which returns a string directly. The vulnerability only affects the 4-param signature that writes to a caller-provided buffer.

Scope: this clears the advisory for consumers of the published superdoc package, effective with the next release. Transitive uuid copies elsewhere in the workspace (dev and demo tooling) are out of scope here and would need a separate cleanup.

Also drops @types/uuid: uuid@11 ships its own type definitions and the DefinitelyTyped package is now a deprecation stub. The lockfile was regenerated from a clean checkout of current main.


pnpm-lock.yaml
```diff
@@ -2868,6 +3075,9 @@ importers:
       '@superdoc/painter-dom':
         specifier: workspace:*
         version: link:../layout-engine/painters/dom
+      '@superdoc/pm-adapter':
```
caio-pizzol: this lockfile was generated from a working tree that doesn't match the committed manifests, so the frozen install fails here first: the super-editor entry adds `@superdoc/pm-adapter` and `fast-xml-parser` and drops the `@superdoc/layout-engine` devDep. hand-fixing this block won't be enough:

- the superdoc importer has the same problem: `@tiptap/core`/`@tiptap/pm`/`@tiptap/starter-kit` (lines 3323-3331) that are in no manifest. pnpm reports one project at a time, so the error just moves there next.
- the file records importers for 8 directories that don't exist in the repo (apps/codebase-agent, demos/document-api-repl, packages/layout-engine/pm-adapter, ...). those are untracked local dirs the workspace globs picked up, so rerunning `pnpm install` in the same tree reproduces this.
- the regen split yjs: apps/cli, super-editor, and collaboration-yjs resolve the catalog to 13.6.30 but superdoc resolves 13.6.31 (on main these four agree on 13.6.30). frozen install can't catch that - it checks the `catalog:` specifier, not the resolved version - and two yjs copies break `instanceof` across the superdoc/super-editor boundary and both get inlined into the CDN bundle.

fix from a clean tree on top of current main (main has since added `@superdoc/font-system` to super-editor):

```
git fetch origin && git rebase origin/main
git status --porcelain        # stash or remove the untracked dirs first
git checkout origin/main -- pnpm-lock.yaml
pnpm install
pnpm install --frozen-lockfile --ignore-scripts   # must pass
```

then check the yjs entries under the catalogs section and the apps/cli, collaboration-yjs, super-editor, and superdoc importers all show one version. the expected diff is ~16 lines.

pnpm-workspace.yaml
```diff
@@ -121,7 +121,7 @@ catalog:
   typescript-eslint: ^8.49.0
   unified: 11.0.5
   utif2: ^4.1.0
-  uuid: ^9.0.1
+  uuid: ^11.1.1
```
caio-pizzol, line 123: while you're in here: `@types/uuid: ^9.0.8` (line 46) can go. uuid@11 ships its own types and @types/uuid is now a deprecation stub, so drop it from the catalog and from the superdoc + super-editor devDeps.

caio-pizzol (changes requested):
the bump itself is safe - every call site is a named `v4`/`v5` import, and uuid@11 keeps both exports. but the committed lockfile can't install; details inline on the first mismatched entry.

one note for the description rather than the diff: the published package still declares `^9.0.1`, so the consumer-side npm audit warning clears only once a release ships this. repo-side `pnpm audit` will keep flagging GHSA-w5hq-g745-h8pq either way: uuid 8.3.2, 9.0.1, 11.1.0, and 13.0.0 all still resolve via transitive deps (verdaccio, @google-cloud/storage in two demos, @hocuspocus/server, and an exact 11.1.0 pin in mintlify). either scope the claim to the published package or add a `pnpm.overrides` entry for uuid.

pnpm-lock.yaml
```diff
@@ -2868,6 +3075,9 @@ importers:
...
+        specifier: workspace:*
+        version: link:../layout-engine/pm-adapter
```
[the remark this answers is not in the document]
caio-pizzol: @mattConnHarbour valid, but please rebase on current main and regenerate pnpm-lock.yaml from a clean tree, then verify frozen install plus matching yjs catalog resolutions.

caio-pizzol (approved):
lgtm

Issue #3242, opened 2026-05-12
Bug: `superdoc_open` fails with "Invalid content for node type table" on Google Docs DOCX exports


law-and-code:
## Summary

Opening a `.docx` file exported from Google Docs fails with the error:

```
Failed to open document: Failed to load document: Invalid content for node type table
```

The root cause is a row-level `<w:sdt>` (Structured Document Tag / Content Control) wrapping `<w:tr>` elements directly inside `<w:tbl>`. Google Docs emits this pattern on some DOCX exports. SuperDoc's table parser appears to expect `<w:tr>` as a direct child of `<w:tbl>` and does not handle the `<w:sdt>` wrapper at the row level.

## Steps to Reproduce

**Reliable reproduction using the attached fixture:**

1. Use `fixture-01-sdt-row.docx` (attached below) — it contains a table whose rows are confirmed to be wrapped in `<w:sdt>`.
2. Call `superdoc_open({ path: "/path/to/fixture-01-sdt-row.docx" })`.
3. Observe the error: `Failed to load document: Invalid content for node type table`.

**Reproduction via Google Docs export (not guaranteed every export):**

1. Create a document in Google Docs that contains a table.
2. Export it as `.docx` (File → Download → Microsoft Word (.docx)).
3. Call `superdoc_open({ path: "/path/to/exported.docx" })`.
4. If Google Docs included the `goog_rdk_*` SDT wrapper in this export, the error will occur. If not, the file opens normally. Use the fixture above for a guaranteed reproduction.

## Error Message

```
Failed to open document: Failed to load document: Invalid content for node type table
```

### Expected OOXML structure

SuperDoc appears to expect table rows as direct children of the table element:

```xml
<w:tbl>
  <w:tblPr>...</w:tblPr>
  <w:tblGrid>
    <w:gridCol w:w="4680"/>
    <w:gridCol w:w="4680"/>
  </w:tblGrid>
  <w:tr>           <!-- direct child -->
    <w:tc>...</w:tc>
    <w:tc>...</w:tc>
  </w:tr>
</w:tbl>
```

### Actual OOXML structure (Google Docs export)

Google Docs wraps the `<w:tr>` in a `<w:sdt>` (Structured Document Tag) at the row level:

```xml
<w:tbl>
  <w:tblPr>...</w:tblPr>
  <w:tblGrid>
    <w:gridCol w:w="4680"/>
    <w:gridCol w:w="4680"/>
  </w:tblGrid>
[...]

PR #3616, opened 2026-06-03
fix(super-editor): support nested content controls

xy200303:
## Summary
- Fix DOCX import failures for legal nested Word content controls (`w:sdt`).
- Classify imported SDTs using translated ProseMirror JSON content and import context instead of only inspecting direct `w:sdtContent` XML child names.
- Safely normalize bare inline content inside block SDTs by wrapping inline runs into paragraphs.
- Preserve SDT metadata including `id`, `tag`, `alias`, `lockMode`, `controlType`, and raw `sdtPr`.
- Add regression coverage for nested block SDTs and mixed block/inline SDT content.

## Root Cause
The importer previously decided whether to emit `structuredContent` or `structuredContentBlock` by checking whether the direct children of `w:sdtContent` included `w:p`, `w:tbl`, or a known block field node.

That logic fails for valid Word documents where `w:sdtContent` directly contains another `w:sdt`. In those cases, the real content shape is only visible after translating the nested SDT. As a result, the importer could emit:
- `structuredContent` with block children, violating the inline-only schema.
- `structuredContentBlock` with bare inline children, violating the block-only schema.

This produced errors such as:
- `RangeError: Invalid content for node structuredContentBlock`
- `RangeError: Invalid content for node structuredContent`

## Changes
- Updated `handleStructuredContentNode` to translate SDT content first, then classify the SDT based on the translated node shape.
- Kept the existing direct XML block detection as a fallback for known block signals such as paragraphs, tables, and block fields.
- Added context-aware handling so nested inline SDTs can still be emitted as inline content when appropriate.
- Added block normalization that groups contiguous inline translated nodes into generated paragraphs when a block SDT needs block-safe content.
- Added `image` to the inline-node fallback list for schema-unavailable converter paths.
- Added regression tests covering:
  - `w:sdtContent -> w:sdt -> w:sdtContent -> w:p`
  - An outer block SDT containing a nested inline SDT, a paragraph, and a table
- Confirmed nested SDT metadata remains readable after import.

## Verification
- `pnpm --filter @superdoc/super-editor test src/editors/v1/core/super-converter/v3/handlers/w/sdt --run`
- Commit hooks passed:
  - format
  - lint
  - commitlint

## Notes
- No schema change was required. The existing schema distinction is correct:
  - `structuredContent` accepts inline content.
  - `structuredContentBlock` accepts block content.
- The fix is scoped to the DOCX import/converter path and does not modify generated dist output.


caio-pizzol:
@xy200303 thanks for this. The approach looks right: decide whether a content control is block or inline from where it appears in the document, not only from its direct child nodes.

I pushed commit `4848caa02` with real `.docx` tests for:
- nested block controls
- nested inline controls
- mixed block content, including the malformed defensive case
- inline picture controls, where the image and original control metadata are preserved

The tests check the source XML, the imported document shape, and that export keeps the content-control wrappers.

Scope note: this covers nested block and inline content controls. Picture controls are preserved, but not fully supported as their own control type yet. Table-row wrappers (`w:tbl > w:sdt > w:tr`) are still separate work tracked in #3242.

caio-pizzol (approved):
lgtm

xy200303:
> @xy200303 thanks for this. The approach looks right: decide whether a content control is block or inline from where it appears in the document, not only from its direct child nodes.
> 
> I pushed commit `4848caa02` with real `.docx` tests for:
> 
> * nested block controls
> * nested inline controls
> * mixed block content, including the malformed defensive case
> * inline picture controls, where the image and original control metadata are preserved
> 
> The tests check the source XML, the imported document shape, and that export keeps the content-control wrappers.
> 
> Scope note: this covers nested block and inline content controls. Picture controls are preserved, but not fully supported as their own control type yet. Table-row wrappers (`w:tbl > w:sdt > w:tr`) are still separate work tracked in #3242.

Thanks for the review.

PR #3617, opened 2026-06-03
fix: table rendering fidelity (SD-3345, SD-2969)

tupizz:
## Summary

Table rendering fidelity for the SD-3345 documents and SD-2969, under the SD-3028 epic. Four independent fixes, one per commit:

1. **Style-based cell shading (pct/auto)** `feat(super-converter)`
   `resolveShadingFillColor` now implements the full CT_Shd model (ECMA-376 §17.3.5): `nil`/`none` yields no background, a pattern (`pctNN`/`solid`) blends the foreground over the fill, and the `auto` sentinel resolves to white (fill) and black (foreground). `parseTableCell` resolves style-based shading through it, so a table style's `pct10`/`auto` band renders as the gray Word paints instead of being dropped. (SD-2969)

2. **keepNext table pagination** `fix(layout-engine)`
   In a keepNext chain anchored to a table, reserve only the first row's height, so the table starts on the page and splits, matching Word. Previously a heading plus a tall splittable table advanced together and left a large gap. (SD-3345)

3. **§17.4.66 cell-border conflict resolution** `feat(painter)`
   New `resolveBorderConflict` helper that collapses two cells' shared-edge borders to the single displayed border: a `nil`/`none` side yields to the opposing border; otherwise greater weight wins, then style precedence, then color brightness, then reading order. Pure helper with unit tests.

4. **Collapsed cell-border rendering** `fix(table)`
   Renders `w:tblPrEx` row-level borders (merged per edge over the table borders) and uses single-owner ownership plus the §17.4.66 winner, so adjacent cell borders draw exactly once (no doubling), asymmetric and borderless edges keep their line (no drop), and trailing `w:gridAfter` columns are handled: the rightmost real cell owns the right border, the resize overlay skips the degenerate spacer, and a cell spanning past the next row's coverage owns its full-width bottom while that row suppresses its top. (SD-3345, SD-2969)

## Test plan

- [x] painter-dom unit tests (1195)
- [x] layout-engine tests (658)
- [x] super-editor tests (15779)
- [x] Browser-verified on the fixtures: SD-2969 shading and clause-header borders, 23_notification callout corner, 15_Form_F3 form grid, M&A checklist (no doubled borders)
- [ ] `pnpm test:visual` corpus gate (high blast radius: keepNext pagination and the shared border resolver)
- [ ] `pnpm check:public` (new `contracts` `TableRowAttrs.borders` field)

## Notes

- `resolveShadingFillColor` has a second caller (`legacy-handle-table-cell-node.js`); behavior shifts for `nil`/`none`/`solid` toward spec correctness, full suite green.
- Commits are dependency-ordered and each builds, so the branch is bisect-safe whether rebased or squashed.
- `resolveBorderConflict` weighting is style-based, not `sz`-width-based; this does not affect any fixture here (all are identical-border or present-vs-none cases).

Linear: SD-3028 (epic), SD-3345, SD-2969


luccas-harbour (approved):
LGTM

PR #3618, opened 2026-06-03
fix: render multi-column sections with explicit column definitions (SD-2324)

tupizz:
<table>
<th>before</th>
<th>after</th>
<tbody>
<tr>
<td>
CleanShot 2026-06-03 at 09 18 11@2x
</td>
<td>
CleanShot 2026-06-03 at 09 18 00@2x
</td>
</tr>

<tr>
<td>
CleanShot 2026-06-03 at 09 19 32@2x
</td>

<td>
CleanShot 2026-06-03 at 09 19 49@2x

</td>
</tr>
</tbody>
</table>

## Summary

DOCX sections that define explicit columns (`<w:cols w:num="N" w:equalWidth="0">` with `<w:col>` children) rendered all content in the left column instead of flowing across the columns like Word. The reported 2002 ISDA Master Agreement and the targeted fixtures are affected. Two fixes, one per layer:

1. **Balance explicit equal-width continuous columns** `fix(layout-engine)`
   `balanceSectionOnPage` skipped every `equalWidth=false` section that had explicit widths, so continuous newspaper sections whose `<w:col w:w>` children are all equal (the common case, e.g. 4×2340) never balanced and stayed single-column. The skip is now narrowed to GENUINELY-unequal widths; explicit-but-equal widths balance like implicit equal columns. Genuinely-unequal widths still fill column-by-column (Word parity, unchanged).

2. **Honor per-column `w:space` for unequal columns (ECMA-376 §17.6.4)** `fix(layout-adapter)`
   Per §17.6.4, when columns are not equal width the section-level `w:cols/@w:space` is ignored and the inter-column gap comes from each `<w:col w:space>`. `extractColumns` used the section space, over-spacing explicit columns so their widths scaled down to fit and diverged from Word. It now uses the per-column space for unequal columns; equal-width columns keep the section space. (Advances SD-2629 for the uniform-spacing case.)

Builds on SD-2452 (continuous-break balancing, already shipped). Part of the SD-2549 multi-column epic. Both call sites that reach `balanceSectionOnPage` are already continuous-scoped, so nextPage/body unequal sections are unaffected.

## Verification

OOXML-spec-grounded (§17.6.4 column definitions; §17.18.77 continuous balancing). Verified in the browser across the issue's fixtures (distinct fragment left-x = columns):

| Fixture | Before | After |
|---|---|---|
| 2002 ISDA full customer doc (num=2/3/4) | single column | multi-column, matches Word |
| ISDA problem-area isolated (num=4) | single column | 4 columns, 0 gap |
| Sections with columns, unequal widths (num=4, child space 720) | single column | 4 columns, 48px gap (per child space) |
| spec-test-1 (num=2 equal), regression | 2 columns | 2 columns (unchanged) |
| spec-test-3 (num=3 equal), regression | 3 columns | 3 columns (unchanged) |

## Test plan

- [x] layout-engine unit tests (658, incl. a new explicit-equal-width balancing test)
- [x] super-editor unit tests (15788, incl. a new §17.6.4 gap test)
- [x] Browser-verified across all fixtures above
- [ ] `pnpm test:layout` (562-doc corpus): could not run locally (the R2/Wrangler corpus token is expired); needs CI or `npx wrangler login`. This is the key regression gate for column/pagination changes.

## Notes

- Genuinely-unequal column widths (different `<w:col w:w>`) still fill column-by-column rather than rebalancing: the height-balancer measures each fragment at a single width and cannot reflow per column. No fixture exercises that case, so it is out of scope here.

Linear: SD-2324 (parent SD-2549; related SD-2629)


caio-pizzol:
hey @tupizz, I checked this more deeply before adding to the branch.

Your fix for `equalWidth="0"` is right. While testing it against Word, we found a few nearby cases that also needed to be covered:

- when `equalWidth` is `1` or missing, Word ignores the child columns
- in that mode, the gap comes from `w:cols/@w:space`, or defaults to 720 twips
- when `equalWidth="0"` but `w:num` is bigger than the usable child columns, we cap the count so extra tiny columns are not rendered
- different gaps between each column are still a bigger follow-up, so we left that out

I verified this with small Word fixtures, then traced how SuperDoc reads and lays out the columns.

caio-pizzol:
FYI, PR #3618 is fixing the reported explicit column spacing issue.

While validating it against Word, we found a few related column cases that need a broader follow-up. I opened SD-2629 for those: mixed column gaps, explicit widths staying as written, and making all column placement use the same source of truth.

luccas-harbour (approved):
LGTM

PR #3619, opened 2026-06-03
refactor: centralizing anchored graphic placement

VladaHarbour:
**This PR does:**
- Centralizing anchored graphic placement (graphic-placement.ts) - small step towards bigger graphic refactor
- Fixes wrapping issue

**What is not included:**
- Vertical image misplacement (going to be in the next PR in order to separate things)

### DIFF
--- a/packages/layout-engine/contracts/src/graphic-placement.test.ts
+++ b/packages/layout-engine/contracts/src/graphic-placement.test.ts
@@ -1,0 +1,98 @@
+import { describe, expect, it } from 'vitest';
+import {
+  resolveAnchoredGraphicY,
+  computeParagraphContentStartY,
+  computeParagraphLayoutStartY,
+} from './graphic-placement.js';
+
+const base = {
+  objectHeight: 100,
+  contentTop: 72,
+  contentBottom: 720,
+  pageBottomMargin: 72,
+};
+
+describe('resolveAnchoredGraphicY', () => {
+  it('positions margin-relative top with offset', () => {
+    expect(
+      resolveAnchoredGraphicY({
+        ...base,
+        anchor: { vRelativeFrom: 'margin', alignV: 'top', offsetV: 10 },
+      }),
+    ).toBe(82);
+  });
+
+  it('positions page-relative bottom with page margin', () => {
+    expect(
+      resolveAnchoredGraphicY({
+        ...base,
+        anchor: { vRelativeFrom: 'page', alignV: 'bottom', offsetV: 5 },
+      }),
+    ).toBe(720 + 72 - 100 + 5);
+  });
+
+  it('positions paragraph-relative center on first line', () => {
+    expect(
+      resolveAnchoredGraphicY({
+        ...base,
+        anchor: { vRelativeFrom: 'paragraph', alignV: 'center', offsetV: 0 },
+        anchorParagraphY: 200,
+        firstLineHeight: 24,
+      }),
+    ).toBe(200 + (24 - 100) / 2);
+  });
+
+  it('uses pre-registered fallback when vRelativeFrom is paragraph without paragraph context', () => {
+    expect(
+      resolveAnchoredGraphicY({
+        ...base,
+        anchor: { vRelativeFrom: 'paragraph', offsetV: 20 },
+        preRegisteredFallbackToContentTop: true,
+      }),
+    ).toBe(92);
+  });
+
+  it('legacy undefined vRelativeFrom uses anchor paragraph Y', () => {
+    expect(
+      resolveAnchoredGraphicY({
+        ...base,
+        anchor: { offsetV: 15 },
+        anchorParagraphY: 300,
+      }),
+    ).toBe(315);
+  });
+});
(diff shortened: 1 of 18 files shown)


packages/layout-engine/contracts/src/graphic-placement.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  if (vRelativeFrom === 'paragraph') {
+    const baseAnchorY = anchorParagraphY;
+    if (alignV === 'bottom') {
+      return baseAnchorY + firstLineHeight - objectHeight + offsetV;
+    }
+    if (alignV === 'center') {
+      return baseAnchorY + (firstLineHeight - objectHeight) / 2 + offsetV;
+    }
+    return baseAnchorY + offsetV;
```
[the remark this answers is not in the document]
VladaHarbour: added condition for preRegisteredFallbackToContentTop case

packages/layout-engine/contracts/src/graphic-placement.ts
```diff
@@ -0,0 +1,196 @@
...
+    return cursorY;
+  }
+  const prevTrailing = trailingSpacing ?? 0;
+  return cursorY + Math.max(spacingBefore - prevTrailing, 0);
+}
+
+/**
+ * Paragraph text start Y including contextual-spacing rewind from the previous paragraph.
+ */
+export function computeParagraphLayoutStartY(input: {
```
luccas-harbour: heads up that this helper ends up being a second copy of the spacing-before/trailing-collapse logic rather than the single source of truth. it's only used to compute the preview start Y for the float scan — the value the paragraph actually lays out at is still computed inline in PHASE 2 of layout-paragraph.ts. so if that inline math ever changes, this has to change with it (the line 137 comment is basically tracking that). is it worth having PHASE 2 call this too so they can't drift? not blocking.

separate minor thing: it's pure paragraph-spacing math with no anchor/graphic concepts, so graphic-placement.ts feels like an odd home — layout-utils maybe?

packages/layout-engine/contracts/src/graphic-placement.test.ts
```diff
@@ -0,0 +1,116 @@
...
+} from './graphic-placement.js';
+
+const base = {
+  objectHeight: 100,
+  contentTop: 72,
+  contentBottom: 720,
+  pageBottomMargin: 72,
+};
+
+describe('resolveAnchoredGraphicY', () => {
```
luccas-harbour, line 16: the Y resolver is well covered, but a few gaps on the new contract:
- resolveAnchoredGraphicX has no direct tests at all (only exercised indirectly via computeAnchorX in floating-objects.test.ts, and only for column-relative). the page/margin relativeFrom × left/center/right permutations all live here now.
- the legacy (undefined vRelativeFrom) + preRegisteredFallbackToContentTop branch is untested — that's exactly the path the 97dcc3ee0 fixup commit touched, so it'd be good to lock it down.
worth adding a few direct cases here?

packages/super-editor/src/editors/v1/core/layout-adapter/converters/shapes.ts
```diff
@@ -338,6 +339,12 @@ export const buildDrawingBlock = (
   },
 ): ShapeDrawingBlock => {
   const normalizedWrap = normalizeWrap(rawAttrs.wrap);
+  if (normalizedWrap) {
+    mergeWrapDistancesFromPadding(
```
luccas-harbour, line 343: image.ts got a test for the mergeWrapDistancesFromPadding fallback but shapes.ts didn't — the drawing path is uncovered. quick test passing a shape with a wrap that omits dist attrs + non-empty padding would cover it.

packages/layout-engine/contracts/src/graphic-placement.test.ts
VladaHarbour: added coverage for resolveAnchoredGraphicX and relativeFrom + preRegisteredFallbackToContentTop case

packages/super-editor/src/editors/v1/core/layout-adapter/converters/shapes.ts
VladaHarbour: added test case

packages/layout-engine/contracts/src/graphic-placement.ts
VladaHarbour: Moved computeParagraphLayoutStartY / computeParagraphContentStartY out of graphic-placement.ts into layout-utils.ts
Extracted shared primitives collapseSpacingBefore and rewindPreviousParagraphTrailing that PHASE 2 and the float-scan preview both use. Phase 2 can't call computeParagraphLayoutStartY end-to-end because this helper misses pagination and post-keepLines state

VladaHarbour:
Hi @luccas-harbour! Visual testing regression must be addressed as well

luccas-harbour (approved):
Thank you for addressing the comments!

LGTM

PR #3625, opened 2026-06-03
fix: prevent focus stealing from external contenteditable editors

mattConnHarbour:
SuperDoc's PresentationInputBridge was intercepting keyboard events from external editors on the same page. Now checks if events originate from a contenteditable outside SuperDoc's DOM tree before intercepting.


mattConnHarbour:
closing, moved

PR #3629, opened 2026-06-04
SD-3355 - bug: dangling comment after resolving a thread exporting

chittolinag:
### Issue

Resolving a comment thread and exporting to DOCX produced a document Word mishandles: the export marked only the root `w15:done="1"` (replies stayed `"0"`), so Word dropped the resolved root and left a dangling reply. The resolved thread also kept its in-document highlight, undoing a resolve took two steps, and resolved replies lost their `w:commentReference` in `document.xml` — Word silently drops an unreferenced `comments.xml` entry.

### Proposed solution

- **Export thread-wide resolved state**: `commentsExporter.js` inherits `done="1"` from any resolved ancestor, so the whole thread is consistently marked resolved (`w:done` + `w15:done`).
- **Thread-wide resolve in one transaction**: new `resolveCommentThread` command + `resolveCommentsInTr` convert the root mark to anchor nodes and normalize reply marks away in a single transaction — highlight clears for every comment and one undo restores the full thread.
- **Keep every comment bound on export**: `prepareCommentsForExport` re-anchors markless resolved-thread replies inside the root's preserved range, so each reply keeps its `w:commentReference` (per `CMTS-EXPORT-010`) and Word retains the entire resolved thread.


luccas-harbour (approved):
LGTM

PR #3633, opened 2026-06-04
refactor(columns): resolved column-geometry foundation (SD-2629)

caio-pizzol:
Unifies column handling onto one resolved model so every consumer agrees on how many columns exist and where each sits, instead of a fill path and a width path that could disagree. This is the behavior-preserving foundation for SD-2629 and does not change rendering; the visual fixes for unequal explicit widths and per-column gaps stack on top.

- One count authority (`resolveColumnCount` = `min(num, usable widths)`): fill, column breaks, balancing, and page/region metadata all resolve it the same way, so a section declaring more columns than it supplies widths no longer advances into phantom columns or advertises them.
- Column x/width come from one state-aware geometry derived per page from its own columns, size, and margins, replacing the separate x math in the paginator, renderer separators, and placement.
- Render metadata (`page.columns`, `layout.columns`, `columnRegions`) is resolved so it never reports more columns than render.
- Change-detection compares the resolved render form, so configs that render identically don't split into spurious regions.
- Per-column gaps are parsed and stored, but not used for positioning yet.

**Verified:** `check:types` clean repo-wide; a 484-doc layout-corpus comparison against the base shows zero geometry changes (only nondeterministic snapshot IDs and the new `gaps` metadata differ).


tupizz (approved):

harbournick (approved):
LGTM

PR #3638, opened 2026-06-04
fix: balance multi-column sections at continuous section breaks (SD-3359)

tupizz:
## Summary

<table>

<td>
Before
</td>
<td>
After
</td>

<tbody>
<td>
CleanShot 2026-06-05 at 09 36 11@2x
</td>
<td>
CleanShot 2026-06-05 at 09 36 34@2x
</td>
</tbody>

</table>

A two-column section ending at a continuous section break rendered with lumpy columns. On the IT-1150 repro NDA the left column ran 169px deeper than the right, and the trailing single-column paragraph started below the unbalanced overhang. Word balances the section: per the ECMA-376 17.18.77 note, "a continuous section break balances the content of the previous section", defined as "starting the next section at the minimum section height such that all content constraints are met".

Linear: [SD-3359](https://linear.app/superdocworkspace/issue/SD-3359/bug-two-column-section-is-not-balanced-before-a-continuous-single)

## Root cause

Mid-page balancing infrastructure already existed and fired for this repro. Tracing showed three defects in the balancing path itself:

1. `balanceSectionOnPage` fed the balancer atomic fragments (`canBreak: false`, no `lineHeights`). The core algorithm already computes line-level break points with widow/orphan control (`blockBreakPoints`), but no caller ever consumed them. A 273px paragraph straddling the column boundary could not split, forcing a 169px imbalance.
2. The binary search floored its target at the tallest block's full height. For a breakable paragraph the indivisible unit is its tallest line, so the search was pinned above the balanced height and packed the overflow lines into column 0.
3. The post-layout gate keyed "mid-doc continuous" off the section's own begin type (its sectPr `w:type`, 17.6.22) instead of the break that ends it (the next section's begin type). A 2-col section that merely started continuous balanced even when ended by a nextPage break.

## Fix

- Paragraph fragments opt into line-level breaking: per-line heights sliced to the fragment's `fromLine..toLine`. A chosen break is applied as fragment surgery: the first half keeps the leading lines, a cloned second half carries the rest to the top of the next column, the same `fromLine`/`toLine` and continuation-flag model pagination uses.
- `w:keepLines` paragraphs (17.3.1.14) and sectPr marker paragraphs stay atomic. `keepLinesBlockIds` is threaded from the layout walk into both balancing call sites.
- The search floor uses each block's tallest indivisible chunk.
- The gate reads the next section's begin type, and the misleading comment on `sectionEndBreakType` that caused the off-by-one is corrected.

## Spec conformance matrix

Verified in the browser against the repro plus eleven mutations derived from ECMA-376:

| Document | Spec basis | Result |
|---|---|---|
| Repro: 2-col then continuous 1-col | 17.18.77 note | 169px -> 14px, trailing tucked below |
| 3 columns | 17.6.4 | balanced, 13px |
| Explicit equal w:col | 17.6.3 | balanced, 14px |
| Genuinely unequal w:col | Word column-flow | not balanced (guard kept) |
| keepLines on the straddling paragraph | 17.3.1.14 | balanced without splitting it |
| Explicit column break | author intent | not balanced |
| End break nextPage | 17.18.77 | not balanced, trailing page 2 (gate fix) |
| w:sep="1" | 17.6.4 | balancing unaffected |
| End break nextColumn | 17.18.77 | not balanced |
| End break evenPage | 17.18.77 | not balanced, next even page |
| 2-col on implicit body sectPr | sd-1655 rule | not balanced (679px column-flow) |
| 2-col on explicit-continuous body | mixed-columns-tabs-tnr precedent | not balanced |

Known pre-existing gap surfaced by the sweep, out of scope here: a `nextColumn` break starts the next section on a new page instead of the following column.

## Test plan

- [x] 4 new unit tests for `balanceSectionOnPage` (straddle split with line partition, keepLines, single tall paragraph, fromLine offset for pagination-split fragments), written failing first
- [x] 2 new `layoutDocument` integration tests (straddle split end-to-end geometry; nextPage end break does not balance)
- [x] Full layout-engine suite: 0 failures
- [x] Browser verification of the repro and all 11 mutations on the dev server


luccas-harbour (approved):
LGTM

PR #3641, opened 2026-06-05
fix(super-editor): rethrow export docx errors

### LINKED ISSUE #3579: Editor.exportDocx() swallows export errors — catch returns undefined instead of rethrowing
### What happens

`Editor.exportDocx()` wraps its whole body in a `try { … } catch`. The catch logs and emits, but has **no** `return` **or** `throw`, so on any export error the method resolves to `undefined` and the real error only reaches stderr:

[https://github.com/superdoc-dev/superdoc/blob/main/packages/super-editor/src/editors/v1/core/Editor.ts#L3916-L3920](<https://github.com/superdoc-dev/superdoc/blob/main/packages/super-editor/src/editors/v1/core/Editor.ts#L3916-L3920>)

```ts
      return result;
    } catch (error) {
      const err = error instanceof Error ? error : new Error(String(error));
      this.emit('exception', { error: err, editor: this });
      console.error(err);
      // ← no return / no rethrow: the function resolves to `undefined`
    }
  }
```

### Why it's a problem

A caller that does `const bytes = await editor.exportDocx()` gets `undefined` with no thrown error to catch. The downstream failure surfaces far from the cause — e.g. `… is not binary (got undefined)` when the bytes are later written — and the actual `TypeError` is buried in `console.error`. Anyone hitting an export-path failure has to reverse-engineer it from stderr instead of catching a rejected promise.

### How it was hit

Building a headless/collab editor with no docx source: `converter.convertedXml` lacks the base OOXML parts, and `exportDocx`'s header/footer export dereferences them, throwing a `TypeError`. The catch swallowed it to `undefined`. (Context: superdoc-dev/superdoc#3568 / [#3569](<https://github.com/superdoc-dev/superdoc/issues/3569>), where seeding a blank-docx template avoids the throw — but that's a workaround for *that* path, not for the swallowing itself, which affects every export error.)

### Suggested fix

Rethrow (or `return`/reject) from the catch after emitting, so the error propagates to the awaiting caller:

```ts
    } catch (error) {
      const err = error instanceof Error ? error : new Error(String(error));
      this.emit('exception', { error: err, editor: this });
      throw err; // propagate instead of resolving undefined
    }
```

Keeping the `exception` emit preserves existing observers; the change is that callers can now `try/catch` (or `.catch`) a real rejection. Same pattern appears in the sibling catch a few lines down (\~L3932) and could get the same treatment. Worth checking existing consumers that may rely on the current undefined-on-error behavior before changing the contract.

xy200303:
## Summary
- Re-throw normalized `Editor.exportDocx()` errors after preserving the existing exception event and console logging.
- Remove `undefined` from the internal implementation return union.
- Add a regression test that verifies failed `converter.exportToDocx` calls reject to the caller.

## Verification
- `pnpm test:editor -- src/editors/v1/tests/export/exportDocx.errors.test.js`
- `pnpm test:editor -- src/editors/v1/tests/export/exportDocx.errors.test.js src/editors/v1/tests/export/exportDocx.commentsFallback.test.js`
- `pnpm --prefix packages/super-editor run types:check` (fails: existing unresolved `@superdoc/font-system` declarations and broad pre-existing JS/TS errors)

Closes #3579


caio-pizzol (approved):
@xy200303 thanks for the fix. I pushed a follow-up commit that covers the SuperDoc wrapper path and avoids duplicate export exception events.

The added coverage checks that a rejected editor export falls back to the original DOCX data, sibling documents still export, and a bridged editor exception does not cause a second wrapper exception.

I only ran `git diff --check`; local unit tests were not run because this repo asks us not to run them locally. CI is queued.

lgtm.

PR #3654, opened 2026-06-05
fix(painter): render double table borders as two rules (SD-3308)

tupizz:
## Summary

Word renders w:val="double" table borders as two parallel rules. SuperDoc emitted the authored width verbatim, and CSS double collapses to a single solid-looking line below 3px (1px rule + 1px gap + 1px rule), so a typical sz12 double border (~2px) rendered as one thin line.

The fix clamps the rendered width up to 3px when the resolved CSS style is double, without shrinking authored widths that are already 3px or wider. All border paint paths (cell borders, outer table edges, continuation rows) funnel through applyBorder, so the one change covers them all.

Linear: [SD-3308](https://linear.app/superdocworkspace/issue/SD-3308/feature-table-borders)

## Verification

- Mutation fixture double_or_dotted_borders.docx: the double table now shows two rules on every outer and interior edge; dotted and dashed tables render unchanged (2px dotted/dashed before and after).
- The prior unit test asserted the collapsed 2px double output and locked in the bug; the expectation flip to 3px is deliberate and documented in the test.
- painter-dom suite: 1243 passed.

## Notes

One step of the SD-3308 plan. Border precedence, nil/none suppression, inside gridlines, and cell-spacing borders were verified at Word parity against the SD-3308 fixture set and are untouched here.

### DIFF
--- a/packages/layout-engine/painters/dom/src/table/border-utils.test.ts
+++ b/packages/layout-engine/painters/dom/src/table/border-utils.test.ts
@@ -39,10 +39,19 @@ describe('applyBorder', () => {
     expect(element.style.borderTop).toMatch(/2px solid (#FF0000|rgb\(255,\s*0,\s*0\))/i);
   });
 
-  it('should apply border with double style', () => {
+  // SD-3308: CSS `double` only renders two distinct rules at >= 3px (1px rule + 1px gap +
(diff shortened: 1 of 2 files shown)


tupizz:
Superseded by #3658, which carries this commit unchanged alongside the SD-3035 style shading fix.

PR #3655, opened 2026-06-05
SD-3352 - fix: refine deletion behavior for anonymous users in overlap-compiler

chittolinag:
### Issue

In suggesting mode, deleting part of your **own** pending insertion should simply shrink the
suggestion - the way Word handles it. Instead, SuperDoc stacked a tracked deletion on top of your own insertion, as if another reviewer had crossed out your text: two overlapping review items for one intent.

This only happened in sessions without an identified user (no `user` id/email passed by the
integrator). The editor handled differently each action: typing into your insertion used a permissive author check that recognizes an anonymous session as itself, but deleting inside it used the strict identity check. So the same user was "you" while typing and "someone else" the moment they deleted.
### Repro

Comment out this line from `SuperdocDev.vue` to simulate the default user.

Screenshot 2026-06-05 at 16 16 15

### Proposed solution

Make deletion answer "is this my insertion?" exactly the way typing already does - one shared rule
(`isSameUserForRefinement` in `overlap-compiler.js`) for all same-user refinement: typing into,
backspacing through, and now deleting inside your own pending insertion.

**User-visible result**: deleting inside your own suggestion shrinks it in place (deleting all of it
removes the suggestion entirely), with or without an identified user.

**Review protection is untouched**: other authors' suggestions - including named authors from imported Word files - still produce an independent, reviewable deletion, and accept/reject decisions still require strong identity.


caio-pizzol:
Verified this behavior in V2 2.13.0: deleting inside your own pending insertion now refines the existing suggestion without creating a separate deletion. Closing this V1 PR because the reported behavior is already addressed in V2.

PR #3657, opened 2026-06-05
fix(style-engine): table style base shading as wholeTable layer (SD-3035)

tupizz:
## Summary

A table style's base-level w:tcPr shading (e.g. fill F2F2F2) is the wholeTable conditional layer per ECMA-376 17.7.6: Word paints it on every cell of a table that references the style. SuperDoc's cascade only read tableStyleProperties[region] and never the style definition's base-level tableCellProperties, so style-only cell fills were dropped and such tables rendered with no background.

The fix collects the base-level tableCellProperties into the wholeTable chain while walking the basedOn hierarchy. Ordering preserved: explicit wholeTable region entries beat the base level within one definition, leaf styles beat ancestors, bands and conditional regions sit above wholeTable, inline cell shading wins over everything.

Linear: [SD-3035](https://linear.app/superdocworkspace/issue/SD-3035/feature-table-styles)

## Verification

- Fixture gates (SD-3035 mutation set): tblStyle_applied, style_plus_direct_border_overrides, style_plus_width_interactions now paint F2F2F2 on all six cells each, matching Word render pixel values exactly.
- Regression locks: first_row_styling and banded_rows_or_columns render byte-identical fills before and after (bands mask wholeTable, as in Word).
- 4 new unit tests (base-level resolution, basedOn precedence, explicit-region precedence, inline wins).
- Suites: style-engine 146, painter-dom 1242, super-editor 16136 passing.
- Layout corpus compare: 476 docs, zero changes.

## Notes

Part of the SD-3035 plan. The plan's banding-gate step was dropped after pixel-probing Word's renders proved the four "spurious band" verdicts were misreads: Word paints the same band fills SuperDoc does, so the existing banding behavior is correct and untouched.

### DIFF
--- a/packages/layout-engine/style-engine/src/ooxml/index.test.ts
+++ b/packages/layout-engine/style-engine/src/ooxml/index.test.ts
@@ -887,6 +887,92 @@ describe('ooxml - resolveTableCellProperties basedOn tblStylePr inheritance', ()
 });
 
 // ──────────────────────────────────────────────────────────────────────────────
+// Style base-level tcPr as the wholeTable layer (ECMA-376 17.7.6, SD-3035)
+// A table style's base-level <w:tcPr><w:shd/></w:tcPr> is stored on the style
+// def's own tableCellProperties (sibling of tableStyleProperties) and IS the
+// wholeTable conditional layer. Word paints it on every cell.
+// ──────────────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


tupizz:
Superseded by #3658, which carries this commit unchanged alongside the SD-3308 double-border fix.

PR #3663, opened 2026-06-05
fix(document-api): subsume inline cell changes inside a whole-table tracked change

andrii-harbour:
trackChanges.list returned the structural whole-table change PLUS the inline trackInsert/trackDelete marks on its cell content as separate review items, breaking the "one whole table = one logical change" contract. groupTrackedChanges now drops inline grouped entries fully contained in a decidable whole-table structural range before appending the structural change, mirroring the comments-store / review-UI containment rule. Only decidable whole-table ranges suppress; a partial or undecidable structural shape keeps its inline marks.


harbournick:
merged in https://github.com/superdoc-dev/superdoc/commit/211b4c3af1b946255db35c958629b142f64579bb

PR #3664, opened 2026-06-05
SD-3386 - fix: deletion handling across paragraph boundaries

chittolinag:
### Issue

In suggesting mode, deleting a selection that spans two paragraphs pushed the remainder of the second paragraph onto a new line, as if Enter was pressed. On the real Backspace/Delete keymap path the tracked interception could crash position mapping (`RangeError: Position NaN`), and the dispatch fallback then deleted the text untracked. Accepting or rejecting the change failed silently, leaving the inline decoration painted, and the hover bubble reported "Accepted" even after a reject.

### Proposed solution

- ProseMirror describes a cross-paragraph deletion as a ReplaceStep carrying a structural paragraph re-join shell; the tracked-changes compiler misread that shell as replacement content, inserted it at the selection end (splitting the paragraph), and minted a malformed replacement the decision engine refuses. `replaceStep.js` now compiles slices without inline leaf content as plain tracked deletions, skips the invert/mirror mapping pairing for that mark-only compile (the source of the NaN), and emits an explicit caret hint at the deletion's left edge.
- Decision resolve events now carry which decision was taken; the comment model stores it and the resolved badge shows "Rejected" vs "Accepted". The bubble's accept/reject no longer mark the thread resolved when the underlying editor decision did not apply.
- Regression tests cover the command path, the real keymap path (Backspace and Delete), the accept/reject lifecycle, and the badge/resolve behavior.


luccas-harbour (approved):
LGTM

PR #3675, opened 2026-06-07
wait for Y.js sync before disconnecting collaborative sessions

mattConnHarbour:
## Summary

- **Wait for Y.js sync before disconnecting**: When closing a collaborative document, the SDK was immediately disconnecting without waiting for pending updates to flush to the server. This caused edits to be lost if `close()` was called too quickly after editing. The fix adds `waitForFinalFlush()` which waits for the websocket buffer to drain before disconnecting.

- **Allow `close()` without `--discard` for collab-only sessions**: Previously, `close()` on a collab-only session (no source file) threw "unsaved changes" errors, forcing users to use `close({ discard: true })`. Since collab sessions persist changes to the server (not a local file), the "dirty" flag doesn't apply. Now `close()` works without `--discard` for collab-only sessions.


mattConnHarbour:
closing, moved

PR #3677, opened 2026-06-08
fix: caret positioning for tabs and soft-break lines

garhm:
- anchor tab caret Y to line top for trailing tab (was bottom-aligned tab box)
 - position soft-break continuation caret at line start, alignment-aware
   (left/center/right), instead of the page right edge


caio-pizzol (approved):
@garhm the code path looks right, and I verified this against the real painter DOM.

i pushed a follow-up commit with the missing coverage: behavior tests for soft-break continuation caret geometry and trailing tab Y, plus unit assertions for the painter contracts and zoom/RTL cases.

lgtm once CI is green.

PR #3681, opened 2026-06-08
fix(export): sanitize DOCX relationships and numbering namespaces (SD-3399)

mattConnHarbour:
## Summary
- Prune dangling relationships in `word/_rels/document.xml.rels` that reference non-existent parts (e.g., `stylesWithEffects.xml` from legacy Word 2010 documents)
- Ensure `xmlns:w15` namespace is declared in `numbering.xml` when `w15:restartNumberingAfterBreak` attributes are present
- Fixes exported DOCX files requiring Word repair before opening


mattConnHarbour:
Closed, moved

PR #3688, opened 2026-06-09
fix(selection): table cell and multiline selection rendering (SD-3328)

tupizz:
## What

Fixes SD-3328. The selection highlight disappeared over empty space inside table cells. While tracking that down I found two more bugs on the same selection rendering path, so this PR groups the three related fixes.

## Demo

https://github.com/user-attachments/assets/812e2ba6-6072-4ed3-a9e0-6ad8384f57a7

## Background

Presentation mode draws its own selection overlay. It does not use the browser selection. The overlay rects come from `computeSelectionRectsFromDom`, which maps the ProseMirror selection to the painted DOM and reads `getClientRects()`.

```mermaid
flowchart LR
  PM["PM selection (from, to)"] --> C[computeSelectionRectsFromDom]
  C --> R["DOM Range + getClientRects()"]
  R --> O[overlay rects]
```

## 1. Dragging into an empty paragraph in a table collapsed the selection (the reported bug)

prosemirror-tables normalizes a text selection whose endpoints resolve to different cells when the head sits at the start of its block (`parentOffset === 0`). An empty paragraph in a cell only ever has offset 0, so dragging a body selection onto one rewrote `[44, 2026]` to `[44, 49]` on dispatch. Text positions in a cell escape because their offset is greater than 0, which is why dragging through cell text worked but an empty paragraph killed it.

```mermaid
sequenceDiagram
  participant Drag
  participant PM as TextSelection.create
  participant Tables as prosemirror-tables
  Drag->>PM: create [44, 2026]
  PM-->>Drag: [44, 2026]
  Drag->>Tables: dispatch
  Tables-->>Drag: normalizes to [44, 49] (collapse)
```

Fix: a detector mirrors that exact condition. When a drag frame would collapse, the handler keeps the last good selection and resumes extending as soon as the head moves into cell text. The detector fails open on any unexpected document shape so it can never block a normal selection.

## 2. Interior lines of a multiline selection rendered blank

A multiline selection built one DOM Range across every line and trusted `getClientRects()`. Each `.superdoc-line` is absolutely positioned, and some Chrome builds return incomplete rects for a Range crossing those boxes: the first and last lines render, the interior lines collapse to a few slivers. `intersectsNode` does not catch it because the entries still report as intersected.

Fix: when a selection spans more than one line, compute rects per line. Interior lines use the line element box, which is reliable across browsers and gives the full width highlight normal selection shows. The first and last lines keep the precise Range so they respect the selection offset and first line indent.

## 3. Empty cell paragraphs resolved to no hit position

The geometry hit test returned null for an empty cell paragraph (no runs), which aborted the active drag and froze the selection. Now it falls back to the paragraph PM start so every empty cell resolves to a valid position.

## Testing

- New regression tests for each fix. The table detector test pins the behavior against the real prosemirror-tables plugin (it must flag exactly the selections that actually collapse).
- All presentation editor tests pass (1139).
- Verified each fix with a real drag against the dev server.


packages/super-editor/src/editors/v1/core/presentation-editor/selection/CellSelectionOverlay.ts
```diff
@@ -120,15 +120,28 @@ export function renderCellSelectionOverlay({
...
+      if (pos === tableStart) {
+        selectedTableIndex = seenTables;
+        return false;
+      }
+      seenTables += 1;
+      return true;
+    });
+    if (selectedTableIndex !== -1) {
+      const tableBlocks = blocks.filter((block) => block.kind === 'table') as TableBlock[];
+      tableBlock = tableBlocks[selectedTableIndex];
```
luccas-harbour, line 144: the `descendants` walk counts every table including nested ones (returning `true` descends into the node's children), but `blocks.filter(kind === 'table')` only holds top-level tables — nested tables live under their parent cell's `blocks`. so a nested table earlier in the doc desyncs `selectedTableIndex` and you index the wrong block, or get `undefined` and the overlay silently disappears. returning `false` after counting a table would keep both counts top-level.

also, this is the third copy of the "Nth table in doc order ↔ Nth table block" mapping (`getCellPosFromTableHit` and `getTablePosFromHit` do the inverse) — might be worth extracting one shared helper in TableSelectionUtilities, which would also be a natural single place for the nested-table fix. and since the existing overlay tests always pass a matching `cellAnchorTableBlockId`, this fallback path never runs in any test — a two-table doc with `cellAnchorTableBlockId: null` would cover it.

packages/layout-engine/layout-bridge/src/index.ts
```diff
@@ -582,6 +582,28 @@ const sumLineHeights = (measure: ParagraphMeasure, fromLine: number, toLine: num
...
+): void {
+  rects.push({
+    x: opts.x,
+    y: opts.y,
+    width: Math.max(1, opts.width),
+    height: opts.height,
+    pageIndex: opts.pageIndex,
+  });
+}
+
```
luccas-harbour, line 606: `pushEmptyLineSelectionBand` got inserted between the `selectionToRects` jsdoc and its signature, so the `@param` block is now orphaned from the exported function. moving the helper (and its doc) above the `selectionToRects` doc block fixes it.

while you're there: the helper only clamps width and pushes, but the `emptyLineOffset` math — the part most likely to drift — is copy-pasted at both call sites. consider moving that into the helper (pass `measure`/`index`/`startLine`), or just inlining the push.

packages/super-editor/src/editors/v1/core/presentation-editor/tables/TableSelectionUtilities.ts
```diff
@@ -229,6 +229,72 @@ export function getTablePosFromHit(
...
+  headPos: number,
+): { anchorCellPos: number; headCellPos: number } | null {
+  const anchor = resolveCellContext(doc, anchorPos);
+  const head = resolveCellContext(doc, headPos);
+  if (!anchor || !head) return null;
+  if (anchor.tablePos !== head.tablePos) return null; // different tables
+  if (anchor.cellPos === head.cellPos) return null; // same cell -> text selection
+  return { anchorCellPos: anchor.cellPos, headCellPos: head.cellPos };
+}
+
```
luccas-harbour, line 297: `resolveCellContext` is a hand-rolled version of `cellAround($pos)` from prosemirror-tables, and the same-table check duplicates its `inSameTable` — both are public exports, and `cellAround`/`cellWrapping` are already vendored in `extensions/table/tableHelpers/`. building on those would drop ~40 lines of traversal. same goes for the inline `cellAncestor` closure in `SelectionHelpers.ts:204` (that one is `cellWrapping`). worth reusing, or was there a reason to avoid the prosemirror-tables helpers here?

packages/super-editor/src/editors/v1/core/presentation-editor/selection/table-cell-selection-collapse.test.js
```diff
@@ -0,0 +1,125 @@
...
+
+    return { bodyPos, cellTextPos, emptyCellParaPos };
+  }
+
+  /** Whether dispatching TextSelection(anchor -> target) actually keeps the head at target. */
+  function dispatchKeepsHead(anchor, target) {
+    const doc = editor.state.doc;
+    const applied = editor.state.apply(editor.state.tr.setSelection(TextSelection.create(doc, anchor, target)));
+    return applied.selection.to === target;
+  }
```
luccas-harbour, line 95: `dispatchKeepsHead` compares `selection.to` to the target, which only holds for forward selections. all current cases are forward so this is fine today, but a backward-drag case added later would report a false collapse — a `Math.max(anchor, target)` or a short comment pinning the forward-only assumption would prevent that.

luccas-harbour:
hey @tupizz! nice work here, the selections are looking a lot better!

codex found this issue below and I can reproduce it in the browser. I can see that empty lines don't generate any selection rectangles and so you can't really tell that they are selected.

> - [P2] Preserve zero-width lines in DOM selection fallback — packages/super-editor/src/editors/v1/dom-observer/DomSelectionGeometry.ts:290-291
>  When a mounted selection spans an empty paragraph/blank table-cell line, the DOM index entry for that line has `pmStart === pmEnd`. This new `spansMultipleLines` branch now forces such selections through `collectClientRectsByLine`, but that helper skips line groups where `linePmEnd <= linePmStart`, so the empty line still produces no rect and `computeSelectionRectsFromDom` returns a non-null, incomplete result instead of falling back to the layout rects that were just fixed. This affects the normal mounted-page path for selections crossing blank lines.

also, claude found a few more things that I added as inline comments.

PR #3690, opened 2026-06-09
fix: anchored textbox in header

VladaHarbour:


artem-harbour:
FYI @VladaHarbour - moving PR to the draft for now.

PR #3691, opened 2026-06-09
fix: process alphaModFix for images

VladaHarbour:


packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
```diff
@@ -596,6 +603,7 @@ export function handleImageNode(node, params, isAnchor) {
     ...(originalChildren.length ? { originalDrawingChildren: originalChildren } : {}),
     ...(hasGrayscale ? { grayscale: true } : {}),
     ...(lum ? { lum } : {}),
+    ...(alphaModFix ? { alphaModFix } : {}),
```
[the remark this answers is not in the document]
VladaHarbour: added alpha on export

packages/layout-engine/painters/dom/src/images/drawing-image.ts
```diff
@@ -29,13 +30,18 @@ export const createShapeGroupImageElement = (doc: Document, child: ShapeGroupChi
     src: string;
     alt?: string;
     clipPath?: string;
+    alphaModFix?: { amt: number };
```
luccas-harbour: the shape-group path reads `alphaModFix` through this local cast, but `ShapeGroupImageChild.attrs` in `packages/layout-engine/contracts/src/index.ts` was never extended — so the field the encoder writes isn't type-checked anywhere. can you add `alphaModFix?: ImageAlphaModFix` to the contract and use it here instead of the inline `{ amt: number }`? that matches what you did for `ImageRun`/`ImageBlock`.

packages/super-editor/src/editors/v1/core/layout-adapter/converters/inline-converters/image.ts
```diff
@@ -165,6 +165,11 @@ export function imageNodeToRun({ node, positions, sdtMetadata }: InlineConverter
       ...(contrast != null ? { contrast } : {}),
     };
   }
+  const alphaModFix = isPlainObject(attrs.alphaModFix) ? attrs.alphaModFix : undefined;
```
luccas-harbour, line 168: `imageNodeToRun` is the only changed converter without an `alphaModFix` test — the describe block in `paragraph.test.ts` covers `lum` and `grayscale` but not this. one quick case asserting `result.alphaModFix` would close the parity gap with the block converter.

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
```diff
@@ -892,6 +900,7 @@ const handleShapeGroup = (params, node, graphicData, size, padding, marginOffset
       const blipFill = pic.elements?.find((el) => el.name === 'pic:blipFill');
       const blip = findChildByLocalName(blipFill?.elements, 'blip');
       if (!blip) return null;
+      const alphaModFix = extractAlphaModFix(blip);
```
luccas-harbour, line 941: the shape-group extraction here (and `createShapeGroupImageElement` on the painter side) has no test, and that's the path the customer's watermark actually goes through. worth a case each since a regression here wouldn't be caught by the existing tests.

luccas-harbour:
hey @VladaHarbour! it's all looking good here. Claude/codex found a few things that I left as inline comments.
I noticed that there's a unit test that's failing but it doesn't seem to be related to your changes, perhaps merging the latest changes from `main` into this branch might fix it?

also, not sure if you did this already but it would be good to upload the test document to the corpus (with `pnpm corpus:upload`) to lock this fix.

luccas-harbour (approved):
LGTM

PR #3693, opened 2026-06-09
fix(presentation): prevent input blocking in multi-instance scenarios

palmer-cl:
- Fixes a bug where mounting two SuperDoc instances on the same page blocked typing and could throw `RangeError: Maximum call stack size exceeded`
- Each PresentationEditor registers window-level capture listeners via PresentationInputBridge to reroute input from stale hidden ProseMirror editors
- With multiple instances, each bridge treated the other's editor as "stale," suppressed the real keystroke, and re-dispatched a synthetic event that the other bridge intercepted in turn — causing infinite recursion and blocked input
- Adds an instance-ownership check so stale rerouting only runs when the event originated from a hidden editor owned by the same SuperDoc instance
- Story-session editors (footnotes, headers/footers) within one instance still reroute correctly


caio-pizzol:
Verified with two SuperDoc instances on the same page: typing is blocked in V1 1.46.3, while both editors work independently in V2 2.13.0. Closing this V1 PR because the reported scenario is already addressed in V2.

PR #3696, opened 2026-06-09
feat(footnote): word-like footnote interactions (SD-3400)

tupizz:
Implements [SD-3400 Feature: Footnote Interactions](https://linear.app/superdocworkspace/issue/SD-3400/feature-footnote-interactions): create, navigate, select, edit, and delete footnotes with Word-like behavior from both the body and the footnote area. All ticket Must-dos are in scope; nothing is deferred.

## Prerequisite fix

Footnotes were silently dropped when the body filled a terminal page: the SD-2656 bodyMaxY-anchored reserve collapses to ~0 on the last page (no continuation target), the planner places nothing, and the body never yields. Reproduced on the ticket's own fixture (0 of 6 footnotes rendered). Fixed with a terminal-page reserve bump mirroring the existing carry-forward bump, guarded so already-placed footnotes are untouched. `Footnote tests.docx` now renders all 6 notes. Validated against the layout corpus footnote set (7 docs, no layout changes).

## Interactions delivered

- **Staged delete from the body**: first Backspace selects the footnote marker, second Backspace deletes it AND prunes the `w:footnote` element when it was the note's last reference (no orphaned elements in the exported file); renumbering cascades. Forward Delete mirrors it. Handles the marker-wrapped-in-its-own-run caret boundaries.
- **Delete from the note area**: clearing all note content removes the footnote everywhere: every body reference (multi-reference notes lose all markers) plus the OOXML element, in one compound mutation, committed immediately when the note empties. Resolution is type-aware, so emptying endnote N never touches footnote N. Freshly inserted empty notes are exempt until they have held content.
- **Insert footnote**: `FootnoteInsertInput.at` is now optional (omitted = insert at the caret), and `editor.commands.insertFootnote()` inserts an empty note (creating the footnotes part with separators when absent) and moves focus into it. Inserting from inside a footnote, endnote, header, or footer story is rejected (ECMA-376 §17.11.14: a footnoteReference inside a note is non-conformant), and reference nodes that reach a note via paste are stripped at commit. Built for custom toolbars and intentionally not registered in the default toolbar; the dev app header has a demo button.
- **Double-click navigation**: double-clicking a body reference opens its note for editing, resolving through the elementsFromPoint hit chain (real pointer events land on the selection overlay).
- **Cross-reference navigation**: double-clicking a REF/NOTEREF cross-reference that points at a footnote or endnote opens the referenced note. The resolver follows the field's target bookmark (flat start/end marker pairs as real imports emit them, container shape too) and scans its range for the note reference. Live-verified on the NVCA fixture. Cross-references to non-note targets fall through to default behavior.
- **Right-to-left selection**: backward drags in note content were verified symmetric with forward drags (single-line, multi-line, past-margin, and escaping-the-note cases) and the direction-agnostic drag path is pinned with a regression test.
- **Resize/font-change smoothing**: the note marker font now follows a fallback chain (explicit first-run value, then the document default run properties, then a constant), so markers resize predictably with the document. Keystroke perf was measured with 94 footnotes (NVCA): `buildFootnotesInput` costs 7.3ms median / 9.2ms max per keystroke, inside the 16ms frame budget, so no footnote measure cache is needed.
- **Visible affordance and focus feedback**: reference markers get a pointer cursor, hover pill, and an enlarged invisible hit halo (the painted digit is ~6x11px). The active note is highlighted at the page bottom (tint + accent bar + activation pulse) while its session is open.
- **Smart scroll**: opening a note session scrolls the note into view (no-op when already visible; inserted notes scroll once they paint).
- **Text cursor**: note content shows an I-beam instead of the default arrow.

Endnotes share the marker/navigation/removal plumbing (`endnoteReference` is handled by the same resolvers, type-aware removal, and CSS).

## Word fidelity

- Generated note paragraphs carry `w:pStyle FootnoteText`/`EndnoteText` (Word always stamps it; without it new footnotes export at the Normal size).
- Bootstrapping a missing notes part also writes the §17.11.9 special-footnote list (`w:footnotePr` with ids -1 and 0) to settings.xml; imported documents keep their settings untouched.
- Separator-variant roundtrip (explicit separators + special list + reference order) is pinned with a dedicated fixture test.

## Architecture

Logic lives outside the ProseMirror extension: the extension keeps schema, NodeView, and thin command shims. New modules:

- `presentation-editor/notes/note-target.ts`: single source of truth for note-target parsing.
- `presentation-editor/notes/NoteSessionCoordinator.ts`: highlight + smart scroll + emptied-note commit, unit-tested in jsdom.
- `pointer-events/note-reference-hit.ts`: pure pointer-to-note-target resolver (body markers + cross-references).
- `extensions/footnote/insert-footnote.js` / `delete-note-marker.js`: insert/staged-delete orchestration as plain functions.
- `plan-engine/footnote-wrappers.ts`: `removeNoteReferenceAt` (position-addressed single-reference removal with element pruning, shared by the document API and the keymap) and `removeNoteEverywhere` (type-aware full removal for emptied notes).

## Tests

- Unit: dense-page render regression (layout-bridge), staged-delete commands and keymap chain order, double-click + cross-reference resolution (incl. overlay path and flat bookmark pairs), NoteSessionCoordinator, §17.11.14 guard + commit sanitizer, type-aware remove-everywhere, position-addressed single removal, marker font fallback chain, bootstrap special list, backward-drag symmetry pin, unmocked area-delete integration (real removal pipeline against a real part).
- Behavior (Playwright): double-click body marker navigation, staged Backspace end-to-end with element pruning + renumbering, area-delete both sides, insertFootnote on a document without footnotes. Existing footnote behavior suite stays green (13 tests).
- Roundtrip: separator-variant fixture (footnote-tests-B) import → export preserves ids, separator types, the special list, and body reference order.
- Layout corpus: footnote set (7 docs) compared against npm@next — no layout changes.


caio-pizzol:
Closing this V1 PR as superseded by the merged footnote work. During verification, we also confirmed a separate V2 issue: the footnote API can return internal IDs instead of displayed numbers. We’re tracking that separately for a V2 fix.

PR #3700, opened 2026-06-10
feat: textbox content editing - rendering and interaction (phase 0 and 1)

artem-harbour:
Main task: SD-3395

Sub-tasks (phases): SD-3402, SD-3404

### Summary

This PR adds real text editing inside text boxes.

Text boxes imported from DOCX can now behave much more like normal document
content: you can click inside them, place the caret, type, delete text, and
move the caret in body content, headers/footers, and table cells.

It also improves import/export so textbox content and positioning survive
round-trip more reliably for both DrawingML and VML textboxes.

#### Editable textbox content
Text inside supported text boxes is now laid out and rendered as addressable
document content rather than as a flat text blob. That makes caret placement,
click-to-position, and text editing work inside the textbox.

#### Better import/export round-trip
DrawingML textboxes are now imported into the same editable structure we already
use for VML textboxes, and exported back from live editor state. VML textbox
positioning is also preserved more reliably across save/reload.

#### Works in more contexts
Textbox editing now works not only in the main document flow, but also inside:

- headers and footers
- table cells
- `behindDoc=1` header/footer textboxes

### Implementation notes

- Normalized DrawingML textbox import to the same PM structure as VML textboxes
- Added textbox paragraph layout/measuring in the layout engine
- Rendered textbox lines with PM position metadata for DOM-based hit testing
- Routed clicks and caret geometry through textbox DOM content
- Added layout invalidation for textbox content changes
- Fixed header/footer interaction issues for `behindDoc` textboxes


packages/layout-engine/layout-bridge/src/diff.ts
```diff
@@ -514,13 +515,26 @@ const drawingBlocksEqual = (a: DrawingBlock, b: DrawingBlock): boolean => {
...
+    (a.drawingKind === 'vectorShape' || a.drawingKind === 'textboxShape') &&
+    (b.drawingKind === 'vectorShape' || b.drawingKind === 'textboxShape')
+  ) {
+    const textboxContentEqual =
+      a.drawingKind !== 'textboxShape' ||
+      b.drawingKind !== 'textboxShape' ||
+      jsonEqual((a as TextboxDrawing).contentBlocks, (b as TextboxDrawing).contentBlocks);
+
     return (
       drawingGeometryEqual(a.geometry, b.geometry) &&
```
luccas-harbour, line 528: a textbox/vector shape whose only change is `customGeometry`, `lineEnds`, or `effectExtent` gets put in `stableBlockIds` here, so incrementalLayout reuses the old measure and the paint version signature (versionSignature.ts:453) also skips repaint — the shape keeps stale path geometry / arrowheads. note `hashDrawingBlock` in cache.ts already folds these three in; can we add them here to keep the three mechanisms consistent?

packages/layout-engine/layout-bridge/src/incrementalLayout.ts
```diff
@@ -1359,6 +1368,8 @@ export async function incrementalLayout(
     }
   }
 
+  hydrateTableTextboxMeasures(currentBlocks, (block, maxWidth) => remeasureParagraph(block, maxWidth));
```
luccas-harbour, line 1371: `hydrateTableTextboxMeasures` runs only on body `currentBlocks`, but header/footer content goes through the separate `layoutHeaderFooterWithCache` path, which doesn't hydrate textbox measures. a textbox in a header/footer table cell therefore never gets `contentMeasures`, so it renders as static text and isn't editable. should we hydrate the header/footer block batches too?

packages/layout-engine/layout-bridge/src/position-hit.ts
```diff
@@ -700,6 +723,111 @@ export const hitTestTableFragment = (
...
+      ? fragmentWithContent.contentMeasures
+      : Array.isArray(block.contentMeasures)
+        ? block.contentMeasures
+        : [];
+    const contentBlocks = Array.isArray(block.contentBlocks) ? block.contentBlocks : [];
+    if (contentMeasures.length === 0 || contentBlocks.length === 0) continue;
+
+    const insets = block.textInsets ?? { top: 0, right: 0, bottom: 0, left: 0 };
+    const localX = Math.max(0, point.x - fragment.x - insets.left);
+    const localY = point.y - fragment.y - insets.top;
```
luccas-harbour: for `textVerticalAlign: 'center' | 'bottom'` the painter offsets the lines downward via flex `justifyContent` (renderer.ts:3071), but hit testing only subtracts `insets.top` and walks paragraphs from y=0 — so clicks in centered/bottom-aligned textboxes can map to the wrong paragraph. should we apply the same vertical content offset before comparing against `paragraphStartY`?

packages/super-editor/src/editors/v1/core/presentation-editor/textboxes/TextboxCaretDomGeometry.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (!fragmentEl) return null;
+
+  const lineEls = Array.from(fragmentEl.querySelectorAll<HTMLElement>('.superdoc-line[data-pm-start][data-pm-end]'));
+  if (lineEls.length === 0) return null;
+
+  for (const lineEl of lineEls) {
+    const pmStart = Number(lineEl.dataset.pmStart ?? 'NaN');
+    const pmEnd = Number(lineEl.dataset.pmEnd ?? 'NaN');
+    if (!Number.isFinite(pmStart) || !Number.isFinite(pmEnd)) continue;
+    if (pos < pmStart || pos > pmEnd) continue;
```
luccas-harbour: at a soft-wrap boundary line N's `pmEnd` equals line N+1's `pmStart` (computeLinePmRange), so `pos === pmEnd` matches the previous line first and the caret sticks to the end of the wrapped line instead of the start of the next. should this prefer the later line, i.e. exclusive `pmEnd` except on the last line?

packages/layout-engine/layout-bridge/src/position-hit.ts
```diff
@@ -902,6 +1030,73 @@ export function clickToPositionGeometry(
       };
     }
 
+    if (
+      fragment.kind === 'drawing' &&
+      fragment.drawingKind === 'textboxShape' &&
+      block.kind === 'drawing' &&
+      block.drawingKind === 'textboxShape'
+    ) {
+      const textboxHit = hitTestTextboxFragment(pageHit, blocks, measures, pageRelativePoint);
```
luccas-harbour: we already resolved this specific textbox into `fragment`/`block` above, but `hitTestTextboxFragment` re-scans every page fragment and returns the first textbox containing the point in array order — for overlapping textboxes that can resolve a different one than the user clicked. could we pass the resolved fragment directly instead of re-discovering it by point?

packages/super-editor/src/editors/v1/core/super-converter/v3/handlers/wp/helpers/encode-image-node-helpers.js
```diff
@@ -1351,6 +1389,105 @@ const buildShapePlaceholder = (node, size, padding, marginOffset, shapeType) =>
...
+  }
+  if (node?.name === 'sd:sectionPageCount') {
+    const cachedText = node?.attributes?.resolvedText ?? node?.attributes?.importedCachedText ?? '';
+    if (!cachedText) return [];
+    return [{ type: 'text', text: cachedText }];
+  }
+  return [];
+}
+
+function extractInlineNodesFromRun(run, paragraphProperties, params) {
```
luccas-harbour: `extractInlineNodesFromRun` never reads `paragraphProperties` — looks like a dead param, can we drop it from the signature and the call site?

packages/layout-engine/painters/dom/src/renderer.ts
```diff
@@ -3043,7 +3040,60 @@ export class DomPainter {
...
+    contentRoot.style.height = '100%';
+    contentRoot.style.display = 'flex';
+    contentRoot.style.flexDirection = 'column';
+    contentRoot.style.boxSizing = 'border-box';
+    contentRoot.style.overflow = 'hidden';
+
+    const insets = block.textInsets ?? { top: 0, right: 0, bottom: 0, left: 0 };
+    contentRoot.style.padding = `${insets.top}px ${insets.right}px ${insets.bottom}px ${insets.left}px`;
+
+    const verticalAlign = block.textVerticalAlign ?? 'top';
```
luccas-harbour, line 3071: none of the painter tests cover `textVerticalAlign: 'center' | 'bottom'`, and the hit-test units only use single-paragraph top-aligned textboxes — worth adding a centered multi-paragraph case since that's exactly where the alignment offset bug hides?

luccas-harbour:
hey @artem-harbour! I had a look, tested everything manually in the browser and it's looking good.

claude/codex found a few things that I left as inline comments.

artem-harbour:
Hi @luccas-harbour, 

All comments were addressed.

luccas-harbour (approved):
LGTM

PR #3702, opened 2026-06-10
fix(super-editor): emit OOXML property children in canonical schema order

dbrogan-OSI:
## Problem

OOXML property containers — `w:rPr`, `w:pPr`, `w:numPr`, `w:trPr`, `w:tcPr`, `w:tblPr`, `w:tblPrEx`, `w:sectPr` — are defined by ECMA-376 as `xsd:sequence` complex types (`CT_RPr`, `CT_PPr`, `CT_NumPr`, …). Their child elements **must** appear in a fixed schema order.

On export, SuperDoc emits these children in JavaScript insertion order:

- `decodeProperties()` (`v3/handlers/utils.js`) iterates `Object.keys(attrs)`, i.e. ProseMirror attribute insertion order.
- `generateRunProps()` (`exporter.js`) emits run marks in mark-array order.

Neither matches the schema sequence, so children are emitted in arbitrary order: `w:numId` before `w:ilvl` on every list item, `w:color` before `w:rFonts`, `w:trPr` `w:hidden` before `w:trHeight`, `w:tblPr` `w:tblLook` before `w:tblCellMar`, etc.

### Why this is easy to miss

**Desktop Word silently repairs** out-of-order children on open, so the document looks fine there. **Word for the web rejects the package** and opens the document **read-only as "corrupt."** We hit this repeatedly against real exports — it is the cause of "document opens read-only in Word for the web."

### Before / after (a list item)

Before (rejected by Word online):
```xml
<w:numPr>
  <w:numId w:val="3"/>
  <w:ilvl w:val="0"/>
</w:numPr>
```

After (canonical `CT_NumPr` order):
```xml
<w:numPr>
  <w:ilvl w:val="0"/>
  <w:numId w:val="3"/>
</w:numPr>
```

## Fix

Add a shared canonical child-order table keyed by container local name (`v3/handlers/ooxml-property-order.js`) and stable-sort the emitted children at the two emit sites:

- `createNestedPropertiesTranslator.decode` in `utils.js` — covers `w:rPr`, `w:pPr`, `w:numPr`, `w:trPr`, `w:tcPr`, `w:tblPr`.
- `generateRunProps` in `exporter.js` — the run-mark `w:rPr` path that bypasses the v3 translator.

Unknown/extension elements (`w14:*`, `mc:AlternateContent`) sort after all known children with their relative order preserved (stable). Containers without a canonical sequence are returned unchanged.

This generalizes the existing `CT_TC_MAR_CHILD_ORDER` pattern (`tcMar`/`tblCellMar`) into one reusable table rather than per-container wrappers.

`w:sectPr` and `w:tblPrEx` sequences are included in the table for completeness; their emission currently runs through separate v2 paths, so wiring those is a ready follow-up.

## Tests

- New `ooxml-property-order.test.js`: utility unit tests for all wired containers + integration through real `w:numPr`, `w:trPr`, and `w:rPr` translators (13 tests).
- All existing converter/exporter suites pass.

## Test plan

- [x] `pnpm --filter super-editor test` for the changed converter/exporter tests
- [ ] Open an exported list/table doc in **Word for the web** and confirm it is no longer read-only / flagged corrupt

### DIFF
--- a/packages/super-editor/src/editors/v1/core/super-converter/exporter.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/exporter.js
@@ -2,6 +2,7 @@ import { SuperConverter } from './SuperConverter.js';
 import { inchesToTwips, linesToTwips, rgbToHex } from './helpers.js';
 import { DEFAULT_DOCX_DEFS } from './exporter-docx-defs.js';
 import { translateChildNodes } from './v2/exporter/helpers/index.js';
+import { sortPropertyChildElements } from './v3/handlers/ooxml-property-order.js';
 import { translator as wBrNodeTranslator } from './v3/handlers/w/br/br-translator.js';
 import { translator as wHighlightTranslator } from './v3/handlers/w/highlight/highlight-translator.js';
 import { translator as wTabNodeTranslator } from './v3/handlers/w/tab/tab-translator.js';
@@ -508,7 +509,12 @@ export function wrapTextInRun(nodeOrNodes, marks) {
 export function generateRunProps(marks = []) {
   return {
     name: 'w:rPr',
-    elements: marks.filter((mark) => !!Object.keys(mark).length),
+    // w:rPr is an ECMA-376 xsd:sequence (CT_RPr); emit children in canonical
+    // schema order rather than mark-array order so Word for the web accepts it.
+    elements: sortPropertyChildElements(
+      'w:rPr',
+      marks.filter((mark) => !!Object.keys(mark).length),
+    ),
   };
 }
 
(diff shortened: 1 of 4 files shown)


caio-pizzol:
Thanks for the contribution and for investigating the export compatibility issues. We verified that V2 exports the nested-list properties in the correct order. We also confirmed a separate table-property ordering issue in V2 and are tracking that for a fix.

We’re closing this V1 PR as we’re addressing the remaining export work in V2. If you’re upgrading, here’s our [migration guide](https://docs.superdoc.dev/editor/migrate-from-v1/overview/).

PR #3711, opened 2026-06-11
fix(super-editor): defer layout rerender and track-changes rewriting during IME composition (SD-2368)

luccas-harbour:
### Summary

CJK (and other IME-based) input was broken in the presentation editor, especially with track changes enabled: Chrome aborted the native composition on nearly every keystroke, dropping or mangling preedit text. This PR makes IME composition survive end-to-end by deferring everything that restructures the composing DOM or repaints layout until `compositionend`, then reconciling afterwards.

### Root causes fixed

1. **Track-changes rewriting killed the composition.** `trackedTransaction` rewrites each preedit update into tracked-insert mark spans plus decorations, which restructures the DOM text node Chrome is composing into — aborting the composition on every keystroke.
2. **Layout repaints detached the composition anchor.** `PresentationEditor` rerendered the visible layout on every doc change, including mid-composition preedit updates.
3. **Blink's empty-inline cleanup destroyed the hidden editor's contentDOM.** In an empty paragraph, Chrome's preedit rewrite (delete + reinsert) empties the inline `span.sd-paragraph-content`; Blink removes the "redundant" empty span, ProseMirror loses its contentDOM and redraws, and the first composed keystrokes vanish.
4. **Synthetic key events corrupted IME commits.** When a commit replaces a longer preedit ("ni h") with shorter text ("你好"), prosemirror-view's DOM-diff heuristic synthesizes a Backspace. Our run-aware Backspace chain succeeded (vanilla PM's would fail and fall through), so ProseMirror discarded the committed text and deleted a preedit character instead.

### How it works

**Deferred composition tracking (`Editor.ts`)**
- While a composition is in flight, composition transactions are applied raw (no `trackedTransaction` rewrite). The inserted range is tracked in `#deferredCompositionRange` and mapped through every subsequently applied transaction.
- On `compositionend` (microtask-deferred, with a forced `domObserver` flush so the trailing DOM read lands first), `#flushDeferredCompositionTracking` converts the composed range into a tracked insertion in one transaction. Text composed inside an existing suggestion is left alone (mark inheritance already covers it).
- Deletions outside the composed range (e.g. composing over a selection) are captured as slices, restored at flush time, and marked as tracked deletions — paired with the insertion as a replacement group (`replacementGroupId`) unless `trackedChanges.replacements: 'independent'` is set. Deferral falls back to immediate tracking when the deleted content can't be faithfully restored (tables, non-text leaves).
- `blur` force-flushes so compositions abandoned without `compositionend` don't leak raw untracked text; chained compositions (common with CJK IMEs) keep deferring until the last one ends.
- The flush transaction sets `compositionTrackingFlush` meta, which `trackedTransaction` now skips to avoid re-rewriting the marks it carries.

**Deferred layout repaint (`PresentationEditor.ts`)**
- `compositionstart`/`compositionend` listeners on the visible host and the active hidden editor target set an `#isComposing` flag; `#flushRerenderQueue` keeps `#pendingDocChange`/`#pendingMapping` intact while composing and a single rerender is scheduled when the composition ends. Listeners are re-pointed when the active hidden target changes (body ↔ header/footer ↔ story), with `blur`/`focusout`/non-composing input as escape hatches.

**Hidden-host DOM survival (`HiddenHost.ts`)**
- Injects a per-document stylesheet forcing `.sd-paragraph-content` to `display: block` inside hidden hosts only, so Blink gives the emptied container a placeholder `<br>` instead of removing it. Visible editors are unaffected.

**Keymap guards (`keymap.js`)**
- `handleEnter`/`handleBackspace`/`handleDelete` decline while `view.composing`, restoring vanilla ProseMirror's fall-through for synthesized mid-composition key events.

### Tests

- New Playwright spec `tests/comments/chinese-ime-composition.spec.ts` simulating real Chrome composition event sequences (preedit updates, commit, empty-paragraph composition, tracked-changes mode).
- Unit coverage for the deferral guardrails in `track-changes-extension.test.js` (deferred range mapping, replacement pairing, blur flush, skip-meta) and `PresentationEditor.test.ts` (composition deferral lifecycle, target swapping, rerender resume).
- `HiddenHost.test.ts` covers the injected stylesheet; `keymap-backspace-chain.test.js` covers the composing decline.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/Editor.ts
+++ b/packages/super-editor/src/editors/v1/core/Editor.ts
@@ -3351,7 +3351,8 @@ export class Editor extends EventEmitter<EditorEventMap> {
    * DOM compositionend/blur handler on the editor view. Stable reference so
    * unmount() can remove it.
    */
-  #handleDomCompositionEnd = (): void => {
+  #handleDomCompositionEnd = (event?: Event): void => {
+    const forceFlush = event?.type === 'blur';
     queueMicrotask(() => {
       // Nothing deferred — common for blur events outside composition. A
       // pending trailing composition read will reschedule via the dispatch
@@ -3359,7 +3360,7 @@ export class Editor extends EventEmitter<EditorEventMap> {
       if (!this.#deferredCompositionRange) return;
       // A new composition can chain immediately after the previous commit
       // (common with CJK IMEs); keep deferring until that one ends too.
-      if (this.view?.composing) return;
+      if (this.view?.composing && !forceFlush) return;
       // The final composition DOM read can still be pending in ProseMirror's
       // observer; force it so the commit transaction (deferred, extends the
       // range) applies before the range is converted to a tracked insert.
@@ -3368,7 +3369,7 @@ export class Editor extends EventEmitter<EditorEventMap> {
       } catch {
         // A failed forced flush only risks marking the range one tick early.
       }
-      if (this.view?.composing) return;
+      if (this.view?.composing && !forceFlush) return;
       this.#flushDeferredCompositionTracking();
     });
   };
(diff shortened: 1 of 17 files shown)


harbournick:
This change is now present on main in commit 406d5ab7. Closing this PR as landed.

PR #3712, opened 2026-06-11
fix(layout): suppress pageBreakBefore directly after explicit break (SD-3366)

tupizz:
## Summary

A paragraph whose style resolves `pageBreakBefore`, placed directly after an explicit page break (`w:br w:type="page"`), rendered an extra blank page. Word collapses the redundant break.

Fixes [SD-3366](https://linear.app/superdocworkspace/issue/SD-3366/bug-pagebreakbefore-inside-style-preceded-by-pagebreak-results-in-an). Also the root cause behind IT-1152 (the SDT wrapper is irrelevant, confirmed in the ticket).

## Root cause

The layout engine already had a fresh-page guard (`shouldSkipRedundantPageBreakBefore`, added for page-forcing section breaks), but it only fires when the fresh page has zero fragments. An explicit page break paragraph leaves its empty remnant (the paragraph mark) on the fresh page, so the guard never fired and the style break added a blank page.

## Word semantics (fixture-verified)

Word's rule is structural, not geometric. Verified by rendering a 5-shape fixture matrix in Word:

| Shape | Structure before the pageBreakBefore paragraph | Word |
|---|---|---|
| A | paragraph ending in explicit break (empty remnant) | suppressed |
| B | plain empty paragraph, no break | fires |
| C | explicit break + text after it | fires |
| D | explicit break + one extra empty paragraph | fires |
| E | explicit break at end of a text paragraph | suppressed |

Only the break paragraph's own remnant is forgiven; any other content re-arms the break.

## Fix

`isPageBreakBeforeSatisfiedByExplicitBreak` in `layout-engine/src/index.ts`: suppress a `source=pageBreakBefore` break when the preceding blocks are `[explicit pageBreak][empty remnant paragraph]`. The directly-adjacent case (shape E, no remnant emitted) was already covered by the existing geometric guard.

## Tests

- 6 new layout-engine tests covering shapes A–E plus the explicit+explicit double-break case (intentional blank page still honored).
- 3 new adapter contract tests pinning the FlowBlock emission shapes the window check relies on.
- Full suites green: layout-engine 755 tests, layout-adapter 1972 tests.

## Verification

- All 3 ticket fixtures + 5 mutation fixtures match Word page counts exactly (v05: 2 pages, was 3; IT-1152 repro: 4 pages; minimal repro: 2 pages).
- Corpus scan (488 docs): exactly 1 doc contains the trigger adjacency (`sd-1962-nested-tables-and-sections.docx`, where Word also suppresses); 10 others have both features but never adjacent, provably unchanged.

### DIFF
--- a/packages/layout-engine/layout-engine/src/index.test.ts
+++ b/packages/layout-engine/layout-engine/src/index.test.ts
@@ -2560,6 +2560,64 @@ describe('layoutDocument', () => {
         expect(pageContainsBlock(layout.pages[2], 'p2')).toBe(true);
       });
 
+      it('still fires pageBreakBefore when the remnant paragraph carries a list marker (numberingProperties)', () => {
+        // An empty list item renders a visible marker ("1.", "•") painted from
+        // paragraph attrs, not runs — that is page content, so the break re-arms.
+        const blocks: FlowBlock[] = [
+          textParagraph('p1'),
+          explicitBreak('pb-explicit'),
+          {
+            kind: 'paragraph',
+            id: 'list-remnant',
+            runs: [{ text: '', fontFamily: 'Arial', fontSize: 16 }],
+            attrs: { numberingProperties: { numId: 1, ilvl: 0 } },
+          },
+          styleBreak('pb-style'),
+          textParagraph('p2'),
+        ];
+        const measures: Measure[] = [
+          makeMeasure([40]),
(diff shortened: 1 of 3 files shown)


luccas-harbour:
hey @tupizz! it looks good to me. can you address the comment by codex? then I think we're good to go

luccas-harbour (approved):

PR #3714, opened 2026-06-11
fix(direction): auto-detect paragraph base direction via dir="auto"

shri-scale:
auto-detect paragraph base direction via dir="auto"

### Problem
Paragraphs with no explicit `w:bidi` render without a `dir` attribute. An absent `dir` makes the element **inherit** its container's base direction rather than auto-detecting from content — so a new/empty paragraph cannot pick up direction from the first strong character typed (e.g. typing Arabic on a fresh line stays LTR), and an editor embedded in an RTL/LTR shell takes the wrong base direction.

### Change
- Render `dir="auto"` for direction-unset paragraphs in both the editing node view (`ParagraphNodeView`) and the viewing painter (`inline-direction/rtl-styles`), so the browser resolves base direction per UAX #9 from the first strong character.
- Explicit `rightToLeft` true/false remain hard overrides → `dir="rtl"` / `dir="ltr"`.
- `setParagraphDirection({ direction: "ltr" })` now writes an explicit `false` (previously it deleted the flag) so a deliberate LTR choice survives auto-detection; reverting to auto stays `clearParagraphDirection()`.
- Updates the direction README to reflect the absent-`dir` vs `dir="auto"` distinction.

### Tests
- Unit: `ParagraphNodeView`, painter `rtl-styles`, `paragraphDirection`.
- Behaviour: `tests/behavior/tests/formatting/auto-paragraph-direction.spec.ts` — auto-detect RTL/LTR on typing, and manual override persistence.

### Note for reviewers
This changes the rendered `dir` for previously-undirected paragraphs (absent → `auto`), so **visual snapshots may shift** for documents containing such paragraphs. Those diffs should be reviewed and re-blessed.

### DIFF
--- a/packages/layout-engine/painters/dom/src/features/inline-direction/rtl-styles.test.ts
+++ b/packages/layout-engine/painters/dom/src/features/inline-direction/rtl-styles.test.ts
@@ -21,11 +21,27 @@ describe('applyRtlStyles direction mapping', () => {
     expect(element.style.direction).toBe('ltr');
   });
 
-  it('sets dir="auto" for unset paragraphs', () => {
+  it('sets dir="auto" for unset paragraphs (wrapper)', () => {
     const element = el();
     expect(applyRtlStyles(element, unsetAttrs as any)).toBe(false);
     expect(element.getAttribute('dir')).toBe('auto');
     expect(element.style.direction).toBe('');
+  });
+
+  it('inheritAuto: leaves dir unset for unset paragraphs (line inherits wrapper)', () => {
+    const element = el();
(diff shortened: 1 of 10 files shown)


shri-scale:
@caio-pizzol kindly review

caio-pizzol:
@shri-scale Thanks for the contribution, including the update addressing wrapped-line direction, and sorry for the delayed review.

We confirmed that explicitly LTR paragraphs can still inherit the surrounding app’s RTL direction in V2, and we’ve recorded a focused bug with reproduction evidence. We’re closing this V1 PR as we move that work to V2. Your automatic direction proposal goes beyond the toggle discussed in #3219 and still needs separate evaluation.

PR #3724, opened 2026-06-12
fix(release): stage recovered SDK Python wheels into workspace for PyPI publish

caio-pizzol:
The stable release recovery path writes the SDK Python wheels to a temp dir under the runner's `/tmp`, then passes that absolute path to the `pypa/gh-action-pypi-publish` action. That action runs in a Docker container that only mounts the workspace and `runner_temp`, not `/tmp`, so the recovery publish fails with `FileNotFoundError` and the Python SDK never ships on a re-run after a partial release.

This stages the recovered wheels into the workspace and publishes from a workspace-relative path, exactly like the non-recovery publish steps already do. Fresh releases are unaffected: the staging step is skipped when no recovery snapshot exists.

This path only runs during recovery (a re-run after npm published but the PyPI step failed), so CI on this PR can't exercise it; it will be validated on the next real recovery run.

### DIFF
--- a/.github/workflows/release-stable.yml
+++ b/.github/workflows/release-stable.yml
@@ -142,18 +142,44 @@ jobs:
       # `git tag --points-at HEAD`. After CLI -> SDK -> MCP, HEAD lands on
       # MCP's version commit and the SDK tag is no longer at HEAD.
       #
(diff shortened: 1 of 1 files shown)


caio-pizzol:
Closing — landing this through our standard internal release-management flow rather than a direct PR here.

PR #3735, opened 2026-06-14
docs(fonts): show the multi-face fonts.families API for custom fonts (SD-3441)

caio-pizzol:
Custom fonts can already register multiple faces through `fonts.families`, but the docs only mentioned the option without showing its shape, so it wasn't clear how to make bold and italic resolve to the right files.

Add a four-face example (regular / italic / bold / bold italic) to "Load your own fonts", and note that SuperDoc hands each `source` URL to the browser (one file per face, WOFF2 recommended) and that registering a family renders it but does not add it to the toolbar (that needs `modules.toolbar.fonts`).

Docs-only; no code change.

### DIFF
--- a/apps/docs/getting-started/fonts.mdx
+++ b/apps/docs/getting-started/fonts.mdx
@@ -91,7 +91,33 @@ For a brand font, a licensed font, or any family SuperDoc does not ship a fallba
 }
 ```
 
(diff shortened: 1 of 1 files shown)


caio-pizzol:
Closing - this docs change is going through our canonical docs pipeline instead.

PR #3741, opened 2026-06-16
fix(super-converter): keep user comments anchored to tracked changes on export

dheerajiiitv:
## Summary
`exportToDocx` drops genuine user comments that are anchored to (or overlap) a tracked change. Only SuperDoc's *synthetic* tracked-change projection rows should be excluded from `comments.xml`.

## Root cause
The export filter `comments.filter((c) => !c.trackedChange)` excludes **every** comment flagged `trackedChange: true`. That flag is set both on synthetic projection rows (no authored body) **and** on real user comments linked to / overlapping a tracked change. Excluded comments also lose their `commentRangeStart/End` + `commentReference` anchors, because the body translator skips comments missing from the export list. Result: any comment placed on a redline is silently lost on export, while plain comments survive.

## Fix
Add `isSyntheticTrackedChangeComment` (mirrors the `isSyntheticTrackedChangeProjection` discriminator in the document-api comment store) and narrow the filter to drop only synthetic rows: `trackedChange === true` **and** no body (`commentText`/`text`/`commentJSON`/`elements`) **and** no `parentCommentId`. Genuine comments stay in the export list and their anchors are re-emitted.

## Tests
Adds unit coverage in `commentsExporter.test.js` for `isSyntheticTrackedChangeComment` (keeps body-bearing tracked-change comments and replies; drops body-less projections).

### DIFF
--- a/packages/super-editor/src/editors/v1/core/super-converter/SuperConverter.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/SuperConverter.js
@@ -17,6 +17,7 @@ import { baseNumbering } from './v2/exporter/helpers/base-list.definitions.js';
 import { DEFAULT_CUSTOM_XML, DEFAULT_DOCX_DEFS } from './exporter-docx-defs.js';
 import {
   getCommentDefinition,
+  isSyntheticTrackedChangeComment,
   prepareCommentParaIds,
   prepareCommentsXmlFilesForExport,
 } from './v2/exporter/commentsExporter.js';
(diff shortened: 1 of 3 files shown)


dheerajiiitv:
@caio-pizzol  @harbournick Please review (its small change)

dheerajiiitv:
@codex review

caio-pizzol (changes requested):
Thanks for working on this. The goal is right: real comments on tracked changes should export, but the automatic tracked-change sidebar rows should not become Word comments.

I'm requesting changes because this version still lets some automatic rows through during export. That can create empty comment entries in the DOCX and make Word repair the file.

I'm going to open a follow-up PR based on this branch with a safer fix. It keeps this contribution in the history, but changes the approach so SuperDoc clearly marks automatic tracked-change rows when they are created and filters them out before writing the Word comment files.

dheerajiiitv:
Sure @caio-pizzol. Thanks for the prompt update

PR #3742, opened 2026-06-16
fix(document-api): make tables tracked-mode capability honest

### LINKED ISSUE #3594: Document API: capabilities().tracked is unreliable — tables.insertRow reports tracked:true but applies directly with no tracked change under changeMode:'tracked'
## Summary

In the Document API, some structural ops report `capabilities().operations[op].tracked === true` but, when invoked with `{ changeMode: 'tracked' }`, **apply the change directly and produce no tracked change** — no redline, nothing for a reviewer to accept/reject, and no error. The `tracked` capability flag is therefore unreliable as a signal of whether an op will actually record a tracked change.

This is distinct from superdoc-dev/superdoc#3592 (ops that legitimately *can't* be tracked and are flagged `tracked:false`). Here the capability flag says `true` but tracking silently no-ops.

## Reproduction (v1.38.0, programmatic, Suggesting mode)

```js
const doc = editor.doc;
doc.capabilities().operations['tables.insertRow'].tracked; // => true

// before: table has 4 rows; trackChanges.list({}).items.length === 10
doc.tables.insertRow({ nodeId: '<tableNodeId>' }, { changeMode: 'tracked' });
// after:  table has 5 rows  (row WAS inserted)
//         trackChanges.list({}).items.length === 10  (0 NEW tracked changes)
// no error thrown
```

Result observed: `rowAdded: true`, `newTrackedChanges: 0`, `callResult: NO-THROW`, with `capability.tracked === true`.

## Expected

One of (in order of preference):

1. The row insertion is recorded as a tracked change (Word records row insert/delete as reviewable revisions), **or**
2. If structural row/column tracking isn't supported, `capabilities()` reports `tracked:false` for these ops (so callers can branch correctly), **or**
3. At minimum, calling a `tracked:false`-in-reality op with `changeMode:'tracked'` errors consistently instead of silently applying directly.

## Actual

`tables.insertRow` reports `tracked:true`, accepts `changeMode:'tracked'` without error, and applies directly with zero tracked changes.

## Inconsistency across the surface (same version)

Calling with `{ changeMode: 'tracked' }`, three different behaviors occur:

| op (example) | `capabilities().tracked` | behavior with `changeMode:'tracked'` |
| -- | -- | -- |
| `format.apply` (inline bold/color) | true | **tracked** — records a tracked change ✅ |
| `styles.paragraph.setStyle`, `format.paragraph.setAlignment` | false | **throws** `"... does not support tracked mode"` |
| `tables.insertRow` | **true** | **silently applies directly**, 0 tracked changes ❌ |

So a caller cannot rely on `capabilities().tracked` alone, nor on a consistent error contract, to decide how an op will behave under tracked mode.

## Impact

For Document-API / AI-assisted review workflows, a tool that calls `tables.insertRow` (or other structural row/column ops) under tracked mode silently bypasses the review gate: the structural change lands with no redline and no way to accept/reject it, while the API reports the op as track-capable.

## Environment

* `@harbour-enterprises/superdoc` **1.38.0**
* Document loaded from `.docx`; body story; `documentMode: 'suggesting'`.
* Related: superdoc-dev/superdoc#3592 (paragraph split/style not tracked — the `tracked:false` side of this).

marekgit200:
## What

`capabilities().operations[op].tracked` reported `true` for four table operations, but only `tables.insertRow` could actually produce a tracked change. The other three silently applied **direct** edits when called with `changeMode: 'tracked'` — and `tables.deleteRow` physically removed the row with no `w:del` marker, i.e. **data loss in suggesting mode**.

## Why it happened

The inline overlap compiler can't represent `tableRow` insert/delete steps, and the review model (`structuralRowChanges.js`) only treats a **whole-table** insert/delete as a decidable change. So a row/column-level operation has no decidable tracked representation, and the tracked path failed closed to a direct apply.

## Changes

- **`tables.insertRow`** keeps tracked support — it stamps a structural `rowInsert` revision matching the importer shape (`row-track-change.js`), which exports as `w:ins` inside `w:trPr` so Word can accept/reject it. Rows inserted in one call now share a single `revisionGroupId`.
- **`tables.deleteRow` / `tables.insertColumn` / `tables.deleteColumn`** now reject `changeMode: 'tracked'` with `CAPABILITY_UNAVAILABLE` (the existing `rejectTrackedMode` convention) and report `supportsTrackedMode: false`.
- Regenerated Document API reference docs for the flipped flags.
- Registers `plan.execute` as a plan-engine meta-op in the adapter conformance test — it was added to the contract without being exempted from per-operation vector coverage, which had left the suite red on `main`.

## Behavior change (worth a reviewer's eye)

Calling those three ops in tracked mode now **throws** instead of silently succeeding, and their capability flag flips to `false`. This trades a silent-wrong / data-loss path for a loud, honest one.

## Test plan

- New tests: `insertRow` tracked stamps `rowInsert` and exports `w:ins`; a multi-row insert shares one `revisionGroupId`; direct mode does not stamp; the three sibling ops reject tracked mode with `CAPABILITY_UNAVAILABLE`.
- `pnpm check:public` (superdoc 14/14 + docapi 4/4), `pnpm check:types`, the adapter conformance + convenience suites, prettier, and eslint all pass.
- No visual changes — API / capability behavior only.

Closes #3594


caio-pizzol:
Thanks for this, @marekgit200. I refreshed the branch against current `main` and regenerated the Document API docs so we can review it.

I also dropped one stale hunk because `plan.execute` is already handled on `main`. The table tracked-mode fix itself is unchanged.

PR #3746, opened 2026-06-16
Fixes #3745

### LINKED ISSUE #3745: Nested block content controls are dropped during layout conversion
### What happened?

  SuperDoc drops nested block `w:sdt` content controls when converting to flow
  blocks.

### Steps to reproduce

Minimal repro shape:

  ```xml
  <w:sdt>
    <w:sdtPr>
      <w:alias w:val="Outer Content Control"/>
      <w:tag w:val="outer_content_control"/>
      <w:id w:val="1001"/>
    </w:sdtPr>
    <w:sdtContent>
      <w:p><w:r><w:t>Outer paragraph before.</w:t></w:r></w:p>

      <w:sdt>
        <w:sdtPr>
          <w:alias w:val="Inner Content Control"/>
          <w:tag w:val="inner_content_control"/>
          <w:id w:val="1002"/>
        </w:sdtPr>
        <w:sdtContent>
          <w:p><w:r><w:t>Inner paragraph.</w:t></w:r></w:p>
        </w:sdtContent>
      </w:sdt>

      <w:p><w:r><w:t>Outer paragraph after.</w:t></w:r></w:p>
    </w:sdtContent>
  </w:sdt>

  Expected: the inner content control content and metadata are preserved/
  rendered.

  Actual: the outer content control renders, but the nested block SDT subtree
  is skipped.

  The loss appears to happen in handleStructuredContentBlockNode, which
  handles paragraphs, tables, and transparent wrappers, but not direct child
  structuredContentBlock nodes.

[minimal-nested-block-sdt.docx](https://github.com/user-attachments/files/29022517/minimal-nested-block-sdt.docx)

### SuperDoc version

master

### Browser

None

### Additional context

_No response_

tom-material:
## Summary
  - Preserve direct child `structuredContentBlock` nodes during block SDT
  layout conversion
  - Add a regression test for nested block SDT content and metadata

  ## Test plan
  - `pnpm exec vitest run --config packages/layout-engine/tests/
  vitest.config.mjs packages/layout-engine/tests/src/sdt-metadata.test.ts`
  - `pnpm exec prettier --check packages/super-editor/src/editors/v1/core/
  layout-adapter/sdt/structured-content-block.ts packages/layout-engine/tests/
  src/sdt-metadata.test.ts`


caio-pizzol:
Thanks for this, Tom. You found the right failing path: a block `w:sdt` nested directly inside another block `w:sdt` was skipped during layout conversion, so valid content was dropped.

We checked it against the OOXML schema and a Word-authored fixture. Nested block SDTs are valid, and the full fix needs to preserve the outer and inner controls as a hierarchy, not just recurse into the inner one. That crosses more of the rendering stack than this PR touches, so we're going to take it over rather than ask you to keep expanding it.

We're tracking the full fix in #3752 and will open the first follow-up PR linked there, then close this once that's up. We'll keep coverage for the repro shape and credit your work. Thanks again for the clear repro.

caio-pizzol:
@tom-material Thanks again for identifying the failing path and proposing this fix. Following up on our earlier message: we verified that the reported nested-control structure now renders correctly in both V1 and V2, with both controls’ content, metadata, and nesting preserved through export and reopening.

We’re closing this PR as superseded by the work taken over internally. Sorry we didn’t close the loop sooner, and thanks for helping us catch this.

PR #3749, opened 2026-06-17
refactor: graphics placement normalization

VladaHarbour:


caio-pizzol:
Closing this as an obsolete V1 refactor. V2 uses the shared graphic-placement contract and its own layout adapter, so we won’t port this normalization layer. Remaining placement defects are being tracked separately; this closure doesn’t mean every graphics-placement case is resolved.

PR #3750, opened 2026-06-17
fix(context-menu): open editor menu on right-click over resize overlays

dawidbudaszewski:
## Summary

Right-clicking on a table (or image) sometimes shows the **browser's native context menu** instead of SuperDoc's editor menu — reproducible whenever the blue **resize handles** are under the cursor. This makes actions like **Edit table** unreachable in that spot.
Before:
Screenshot 2026-06-18 at 00 31 43
After:
Screenshot 2026-06-18 at 00 26 21

## Root cause

The table/image resize overlays (`.superdoc-table-resize-overlay` / `.superdoc-image-resize-overlay`, rendered by `TableResizeOverlay.vue` / `ImageResizeOverlay.vue`) are mounted as **siblings of the editor surface**, not inside it:

```
<div class="super-editor">
  <div class="editor-element …">         ← editor surface (getEditorSurfaceElement)
  <ContextMenu />
  <div class="superdoc-table-resize-overlay">  ← sibling, holds the resize handles
```

`ContextMenu.vue`'s `isEventWithinContextMenuTargets()` only treats the editor surface and the ProseMirror view DOM as in-editor. So when a right-click lands on a resize handle, `handleRightClick` bails early, never calls `event.preventDefault()`, and the native menu wins. (The ProseMirror `handleDOMEvents.contextmenu` and the `PresentationInputBridge` forwarder also skip it for the same reason.)

## Fix

Treat targets within the resize overlays as in-editor. Added a small pure helper `isWithinResizeOverlay(target)` in `contextmenu-helpers.js` and used it in `isEventWithinContextMenuTargets()`. Because the menu context (position, `isInTable`, etc.) is resolved from the cursor **coordinates** (`getEditorContext` → `posAtCoords`), the correct items (e.g. **Edit table**) are shown even though the DOM target is the overlay handle.

## Testing

- `pnpm exec vitest run` for the context-menu suite — all pass (165), including 3 new unit tests for `isWithinResizeOverlay`.
- Prettier clean; ESLint clean.
- Manually verified in an editor build: right-clicking directly on the blue resize handles now opens the SuperDoc context menu with the table actions.

### DIFF
--- a/packages/super-editor/src/editors/v1/components/context-menu/ContextMenu.vue
+++ b/packages/super-editor/src/editors/v1/components/context-menu/ContextMenu.vue
@@ -4,7 +4,7 @@ import { Selection } from 'prosemirror-state';
 import { isCellSelection } from '@extensions/table/tableHelpers/isCellSelection.js';
 import { ContextMenuPluginKey } from '../../extensions/context-menu/context-menu.js';
 import { getPropsByItemId, resolveContextMenuCommandEditor } from './utils.js';
-import { shouldBypassContextMenu } from '../../utils/contextmenu-helpers.js';
+import { shouldBypassContextMenu, isWithinResizeOverlay } from '../../utils/contextmenu-helpers.js';
 import { moveCursorToMouseEvent } from '../cursor-helpers.js';
 import { getEditorSurfaceElement } from '../../core/helpers/editorSurface.js';
 import { getItems } from './menuItems.js';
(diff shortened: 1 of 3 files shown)


shri-scale (approved):
Looks good

caio-pizzol:
Thanks for the contribution and for identifying the resize-overlay issue. We verified that V2 opens the editor menu, but found a remaining case where right-clicking an image shows actions for an unrelated table. We’ve captured a reproduction and are tracking that separately for V2. We’re closing this V1 PR as development is now focused on V2.

PR #3757, opened 2026-06-18
refactor: centralize anchored graphic placement

VladaHarbour:
- Added a shared resolveGraphicPlacement() in layout-engine to centralize anchored graphic placement for images, drawings, and tables.
- The resolver now produces one placement result for paint coordinates, text-wrap exclusion coordinates, layer data (behindDoc, zIndex), and measurement impact flags.
- Refactored paragraph anchoring, pre-registered/page-relative anchors, paragraphless anchored fallback, and float registration to consume the same resolved placement instead of recomputing X/Y separately.
- Removed duplicated table anchor X logic from the floating object manager.

### DIFF
--- a/packages/layout-engine/layout-engine/src/floating-objects.ts
+++ b/packages/layout-engine/layout-engine/src/floating-objects.ts
@@ -20,11 +20,11 @@ import type {
   DrawingMeasure,
   TableBlock,
   TableMeasure,
-  TableAnchor,
   TableWrap,
   ColumnLayoutForAnchor,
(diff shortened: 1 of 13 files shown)


caio-pizzol:
Closing this stacked refactor following the retirement of SD-3420 and closure of #3749. During validation, we found that a behind-text image inside a table cell still covers paragraph text in V2. We’ve retained the fixtures and reproduction for a focused V2 follow-up; this closure doesn’t mean the rendering issue is fixed.

PR #3758, opened 2026-06-19
feat(linked-styles): redefine named paragraph styles ("Update to match")

shri-scale:
## What

Adds a supported way to **redefine a named paragraph style's run-level look** (font family, size, bold/italic/underline, color) so it updates live and survives `.docx` export — the editor equivalent of Google Docs' "Update Heading 1 to match selection".

Previously the linked-styles extension could only *apply* an existing style (`setLinkedStyle`, `setStyleById`); there was no supported way to *redefine* what a style means.

## API

- **Command** — `editor.commands.updateLinkedStyle(styleId, formatting)` rewrites the style across all three converter representations (`linkedStyles[].definition.styles`, `translatedLinkedStyles.styles[id]`, `word/styles.xml`) under a single snapshot/rollback; never throws.
- **Helpers**
  - `editor.helpers.linkedStyles.getLinkedStyleFormatting(styleId)` — read a style's current run formatting.
  - `editor.helpers.linkedStyles.getEffectiveFormattingAtSelection()` — read the run formatting at the cursor (the source for "update to match").
- `formatting` shape: `{ bold, italic, underline, fontSizePt, fontFamily, colorHex }`.

## Rendering

A redefinition doesn't change the document, so both render paths are refreshed explicitly on every call:
- **ProseMirror mode** — a linked-styles plugin meta signal regenerates decorations from the updated styles.
- **Layout/presentation mode** — `PresentationEditor.refreshLinkedStyles()` clears the `FlowBlockCache` (keyed on node identity) and re-lays out, mirroring other non-edit re-render paths (e.g. `setShowBookmarks`).

## UI

The styles dropdown gains a per-row **"Update to match"** (✎) action. Clicking the style name still just applies the style as before.

## Tests
- Pure mappers (`style-formatting.test.js`) — formatting ⇄ OOXML conversions.
- Integration (`update-linked-style.test.js`) — engine, command, read helpers, repaint-on-every-redefinition (not just the first), and `word/styles.xml` + `translatedLinkedStyles` export round-trip.
- Dropdown emit (`LinkedStyle.test.js`).

`pnpm test` (linked-styles suites) green · `pnpm check:types` clean · eslint clean.

## Notes
- Scope is **run-level** formatting; paragraph-level layout (spacing/indent) round-trips to export, but live layout-repaint for those is a follow-up.
- Pure shape conversions are isolated in `style-formatting.js` (no editor/DOM deps).

### DIFF
--- a/packages/super-editor/src/editors/v1/components/toolbar/LinkedStyle.test.js
+++ b/packages/super-editor/src/editors/v1/components/toolbar/LinkedStyle.test.js
@@ -1,0 +1,38 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { mount } from '@vue/test-utils';
+
+// Mock the extension helpers so the component can mount without a real editor.
+vi.mock('@extensions/linked-styles/index.js', () => ({
+  getQuickFormatList: () => [
+    { id: 'Normal', definition: { attrs: { name: 'Normal' }, styles: {} } },
+    { id: 'Heading1', definition: { attrs: { name: 'Heading 1' }, styles: {} } },
+  ],
+  generateLinkedStyleString: () => '',
+}));
+
+import LinkedStyle from './LinkedStyle.vue';
+
+describe('LinkedStyle dropdown', () => {
+  let wrapper;
+  beforeEach(() => {
+    wrapper = mount(LinkedStyle, { props: { editor: {}, selectedOption: 'Normal' } });
+  });
+
+  it('emits "select" with the style when a row name is clicked', async () => {
+    await wrapper.findAll('.style-name')[1].trigger('click');
+    const events = wrapper.emitted('select');
+    expect(events).toBeTruthy();
+    expect(events[0][0].id).toBe('Heading1');
+  });
+
+  it('emits "update" with the style when the row update action is clicked', async () => {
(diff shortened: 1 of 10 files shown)


caio-pizzol:
Thanks @shri-scale. This is a useful feature and I agree it fills a real gap.

I do not think we should merge it in this shape though. The change adds a new `editor.commands.updateLinkedStyle(...)` API, but new document changes should go through the Document API (`editor.doc.*`). `editor.commands.*` is the older path we are moving away from.

Could you move this into the Document API styles area instead?

A possible shape:

```ts
editor.doc.styles.apply({
  target: {
    scope: 'style',
    type: 'paragraph',
    styleId,
    channel: 'run',
  },
  patch: { bold, italic, underline, fontSize, fontFamily, color },
});
```

The feature, UI, tests, and style-reading logic are worth keeping. The main request is to put the style update behind `editor.doc.*` so it works through the same public API as the rest of our document changes.

caio-pizzol:
Thanks again, @shri-scale. This is a useful feature, especially the “Update to match” workflow and attention to live updates and export. Following my earlier feedback about using the Document API, we’ve retained your PR, tests, and a reproducible example for a focused V2 follow-up. Named-style redefinition is still missing in V2; we’re closing this V1 PR so the work can continue internally through the Document API.

PR #3776, opened 2026-06-26
feat(mcp): make tracked-change author configurable, clarify save/close

caio-pizzol:
The MCP server opened every document as the fixed user `MCP Server`, so every tracked change and comment an agent made was attributed to `MCP Server` with no way to change it. An agent acting as a named user, or any tool that needs edits attributed to a specific author, had no knob for it.

`SessionManager` now takes the author as a constructor argument, and the server reads it from `SUPERDOC_AUTHOR_NAME` (plus optional `SUPERDOC_AUTHOR_ID`), defaulting to the previous `MCP Server` so existing behaviour is unchanged. The configured author flows to `Editor.open`'s `user`, so saved `.docx` tracked changes and comments carry that `w:author`.

Also tightens two tool descriptions to match the real lifecycle: edits live in memory until `superdoc_save`, and `superdoc_close` discards anything unsaved. Agents frequently edited and then closed without saving, so the edits never reached the output file; the sharper wording steers them to save first.

- `SUPERDOC_AUTHOR_NAME=Agent npx @superdoc-dev/mcp` -> tracked changes authored `Agent`.
- Default (no env) is unchanged: author `MCP Server`.

**Review**: confirm the default path is untouched (no env -> `MCP Server`) and the author flows to `editor.options.user`. 
**Verified**: `bun test` in apps/mcp -> 39 pass / 0 fail (adds 2 SessionManager author cases); end-to-end, driving the built server with `SUPERDOC_AUTHOR_NAME=Agent` through open -> tracked replace -> save produces `w:author="Agent"` while preserving a pre-existing reviewer revision.

### DIFF
--- a/apps/mcp/src/__tests__/session-manager.test.ts
+++ b/apps/mcp/src/__tests__/session-manager.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, afterEach } from 'bun:test';
 import { resolve } from 'node:path';
-import { SessionManager } from '../session-manager.js';
+import { SessionManager, DEFAULT_MCP_USER } from '../session-manager.js';
 
 const BLANK_DOCX = resolve(import.meta.dir, '../../../../shared/common/data/blank.docx');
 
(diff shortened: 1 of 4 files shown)


caio-pizzol:
Closing — this belongs in the orbit-canonical source, not the OSS mirror. The superdoc/public subtree is exported from orbit to this repo (per scripts/superdoc-export-public.mjs), so the change is re-opened against orbit: superdoc/orbit#217. It will reach this repo and npm via the normal export.

PR #3781, opened 2026-06-28
fix(vscode-ext): prevent autosave from opening a Save File dialog on every edit

### LINKED ISSUE #3768: VS Code extension: autosave fires a "Save File" dialog on every edit (export() called without triggerDownload: false)
### Summary

Editing any `.docx` opened with the SuperDoc VS Code extension pops a browser **"Save File" dialog** (default name `SuperDoc.docx`) roughly once per second of typing. Edits *do* persist to the underlying file (the `postMessage` save path works), but the recurring dialog makes real editing unusable.

### Environment

- `superdoc-vscode-ext` **2.15.0**
- macOS (VS Code custom-editor webview)

### Steps to reproduce

1. Open any `.docx` with the SuperDoc editor.
2. Type a few characters and pause.
3. ~1s later a "Save File" dialog appears (suggested name `SuperDoc.docx`). It recurs on essentially every edit.

### Root cause

In `dist/webview/main.js`, the debounced autosave calls:

```js
const blob = await editor.export({ format: "docx" });
```

SuperDoc's `export()` defaults `triggerDownload` to `true`:

```js
async export({ exportType = ["docx"], /* ... */, triggerDownload = true, /* ... */ } = {}) {
  // ...
  if (blobsToZip.length === 1) {
    if (triggerDownload) return createDownload(blobsToZip[0], baseFileName, exportType[0]); // <-- showSaveFilePicker
    return blobsToZip[0];
  }
  // ...
}
```

With `triggerDownload` left at its default, `export()` returns `createDownload(...)`, which calls `window.showSaveFilePicker` inside the webview. Because both the debounced autosave (`AUTO_SAVE_DELAY = 1000`) and the `Cmd/Ctrl+S` handler call `saveDocument()`, the picker fires on virtually every edit.

(Minor: `format` is not a recognized `export()` option; the intended key is `exportType`.)

### Fix

Pass `triggerDownload: false` so `export()` returns the blob without triggering a download:

```js
const blob = await editor.export({ exportType: ["docx"], triggerDownload: false });
```

Verified locally against 2.15.0: the dialog no longer appears, and the silent autosave-to-disk continues to work (edits persist across tab switches).

claudiusararu:
## Problem
Fixes #3768. Editing any `.docx` in the SuperDoc VS Code extension pops a browser "Save File" dialog on essentially every edit (the ~1s debounced autosave and Cmd/Ctrl+S both trigger it), which makes real editing unusable.

## Root cause
The webview autosave calls `editor.export({ format: 'docx' })`. SuperDoc's `export()` defaults `triggerDownload: true`, and for a single docx blob that path calls `createDownload(...)` → `window.showSaveFilePicker`, opening the dialog. (`format` is also not a recognized `export()` option — the correct key is `exportType`.)

## Fix
Pass `triggerDownload: false` (and the correct `exportType: ['docx']`) so `export()` returns the Blob directly — exactly what the `blob.arrayBuffer()` → `postMessage` save path consumes. No dialog; silent autosave-to-disk keeps working.

```js
// apps/vscode-ext/webview/main.js
- const blob = await editor.export({ format: 'docx' });
+ const blob = await editor.export({ exportType: ['docx'], triggerDownload: false });

### DIFF
--- a/apps/vscode-ext/__tests__/superdoc-contract.test.ts
+++ b/apps/vscode-ext/__tests__/superdoc-contract.test.ts
@@ -45,7 +45,7 @@ describe('SuperDoc API contract', () => {
     });
 
     it('has export() method', () => {
-      // webview/main.js: editor.export({ format: 'docx' })
+      // webview/main.js: editor.export({ exportType: ['docx'], triggerDownload: false })
       expect(superdocClassSrc).toMatch(/async\s+export\s*\(/);
     });
 
--- a/apps/vscode-ext/__tests__/webview-autosave.test.ts
+++ b/apps/vscode-ext/__tests__/webview-autosave.test.ts
@@ -1,0 +1,35 @@
+/**
+ * Regression test for #3768.
+ *
+ * The webview autosave must export the document *silently*. SuperDoc's
+ * `export()` defaults `triggerDownload: true`, which opens a browser
+ * "Save File" dialog (via `showSaveFilePicker`). Because the debounced
+ * autosave and the Cmd/Ctrl+S handler both call `saveDocument()`, the dialog
+ * popped on essentially every edit. Passing `triggerDownload: false` makes
+ * `export()` return the Blob directly — which is what the `postMessage` save
+ * path consumes.
+ *
+ * Uses static analysis of the source (no webview/DOM runtime needed), matching
+ * the approach in superdoc-contract.test.ts.
+ */
+import { describe, it, expect } from 'vitest';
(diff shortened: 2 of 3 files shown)


caio-pizzol:
Thanks, Claudiu—your change addresses the unwanted downloads. Our V2 testing also uncovered an initialization issue that prevents us from verifying autosave end to end. We’ve retained your approach and the reproduction evidence in our V2 follow-up, and we’re closing this V1 PR in favor of that work. The fix hasn’t shipped yet.

PR #3792, opened 2026-07-03
fix(super-editor): declare full namespace set on freshly-created numbering.xml

### LINKED ISSUE #3773: Adding numbered list to an imported docx causes export to not load in Word
### What happened?

When importing a docx made in Word that does not contain a numbered list, then adding the first numbered list in the super-editor, and exporting the docx, the docx cannot be opened in Word and cannot be repaired. The docx renders fine in super editor. This has to do with how super editor is creating that numbered list without the `w16cid` namespace for `w:numbering`.

### Steps to reproduce

1. Create a blank Word document
2. Import it into Super Editor
3. Add a numbered list
4. Export to DOCX
5. Open in Word - Fails to Open

### SuperDoc version

1.43.1

### Browser

Chrome

### Additional context

This happens because the XML namespace for `w:numbering` does not contain the `w16cid` and `w16cex` namespaces in the `<w:numbering>` declaration, but uses at least `w16cid:durableId`.

Adding those namespaces to the `NUMBERING_ROOT_ATTRS` in `packages/super-editor/src/editors/v1/core/parts/adapters/numbering-part-descriptor.ts` makes things export properly.

```typescript
  'xmlns:w16cid': 'http://schemas.microsoft.com/office/word/2016/wordml/cid',
  'xmlns:w16cex': 'http://schemas.microsoft.com/office/word/2018/wordml/cex',
```

MIt9:
## Summary
- Fixes #3773 — importing a docx with no `word/numbering.xml`, then adding the first numbered list, produced a numbering part missing `xmlns:w16cid` (and other `w16*` namespaces). Word refused to open/repair the exported file.
- `numbering-part-descriptor.ts`'s `ensurePart()` hand-maintained a 3-key namespace map (`xmlns:w`/`w15`/`mc`) instead of reusing `DEFAULT_DOCX_DEFS`, the full namespace + `mc:Ignorable` map that `document.xml`, `comments.xml`, `footnotes.xml`, and `people.xml` already apply unconditionally on every export.
- Every numbering-part creation goes through this `ensurePart()` (`mutateNumbering` → `mutatePart` → `executeMutate`), so the legacy `SuperConverter#exportNumberingFile` `baseNumbering` fallback (which does have the full namespace set) never runs for this path — the part descriptor's own map has to be complete on its own.
- Fix: reuse `DEFAULT_DOCX_DEFS` for `NUMBERING_ROOT_ATTRS` instead of hand-picking namespaces one bug report at a time.

## Test plan
- [x] New unit test on `numberingPartDescriptor.ensurePart()` asserting `xmlns:w16cid`/`w14`/`r` are present.
- [x] New end-to-end regression test: import `blank-doc.docx` (no numbering.xml) → `toggleOrderedList()` → export → assert the exported `<w:numbering>` root is namespace-complete. Verified this test fails without the fix (`xmlns:w16cid` → `undefined`) and passes with it.
- [x] Relaxed one `templates-adapter.integration.test.ts` assertion that hardcoded the old narrow `mc:Ignorable` merge result (`"w15 w14 w16cid"`) to the same flexible pattern the sibling `styles.xml` test in the same file already uses — the merged value is now a superset (still contains `w16cid`), not a regression.
- [x] Full `@superdoc/super-editor` test suite: 1215 files / 17165 tests passed.
- [x] `pnpm check:types` (repo-wide `tsc -b`): clean.


caio-pizzol:
Thanks @MIt9, this looks right.

I traced the broken case: when a doc starts without `word/numbering.xml`, SuperDoc creates that file when the first numbered list is added. That path was using a smaller namespace list, so the exported file could contain `w16cid:durableId` without declaring `w16cid`. Reusing `DEFAULT_DOCX_DEFS` fixes that.

I also pushed a tiny comment-only edit to clarify that `w16cid` comes from the seeded numbering definitions on `w:num`, not from list-level overrides.

Verified locally:
- the new regression test fails without the fix and passes with it
- focused touched tests pass
- broader numbering/import-export/parts tests pass
- `pnpm run check:types` passes

caio-pizzol (approved):
Verified locally end-to-end (regression test fails without the fix, full touched-test + numbering/import-export/parts sweep + repo-wide typecheck green). Namespace handling matches ECMA-376 Part 3 MCE requirements. Approving.

PR #3807, opened 2026-07-07
Respect localized built-in DOCX styles

replicant026:
## Summary

Updates DOCX default-style injection so SuperDoc recognizes localized built-in style equivalents before adding the built-in defaults.

## Why

`addDefaultStylesIfMissing` previously checked only for exact canonical style IDs such as `Title` and `Heading1`. Some Word-authored DOCX files, especially localized documents, can contain built-in title/heading styles with localized or sanitized IDs/names, for example:

- `Ttulo` / `Título`
- `Ttulo1` / `Título 1`
- `Kop1`

Those styles should count as existing built-in title/heading styles. Without that recognition, SuperDoc appends its default `Title`/`Heading1` definitions and consumers can see the default styling take precedence over the imported template's own heading/title styling.

## What changed

- Adds built-in style alias detection for the default style insertion path.
- Normalizes style IDs and `w:name` values by stripping diacritics, spacing, and punctuation before matching.
- Recognizes Portuguese title aliases (`Título`, `Título 1`) and existing localized heading examples such as `Kop1`.
- Adds regression coverage for localized title/heading styles and for the normal missing-default case.

## Validation

Passed:

```bash
node_modules/.bin/vitest.cmd run packages/super-editor/src/editors/v1/core/super-converter/v2/importer/docxImporter.test.js --root packages/super-editor
```

Result: `1 passed`, `41 tests passed`.

Note: running through `pnpm --filter @superdoc/super-editor test -- ...` on Windows triggered the monorepo `prepare` lifecycle with Unix shell syntax (`if [ -z "$CI" ]`), so the targeted test was run directly with the local Vitest binary after dependencies were installed.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/super-converter/v2/importer/docxImporter.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/v2/importer/docxImporter.js
@@ -786,7 +786,7 @@ export function addDefaultStylesIfMissing(styles) {
   const { elements } = updatedStyles.elements[0];
 
   Object.keys(DEFAULT_LINKED_STYLES).forEach((styleId) => {
-    const existsOnDoc = elements.some((el) => el.attributes?.['w:styleId'] === styleId);
+    const existsOnDoc = elements.some((el) => isEquivalentBuiltInStyle(el, styleId));
     if (!existsOnDoc) {
       const missingStyle = DEFAULT_LINKED_STYLES[styleId];
       updatedStyles.elements[0].elements.push(missingStyle);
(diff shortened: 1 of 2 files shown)


caio-pizzol:
Thanks @replicant026 for looking into this and sending the PR. It helped us confirm the issue and understand that it affects documents in many languages.

We’re planning a broader fix for the v2 editor, so we won’t merge this approach. We’ll track the work publicly in #3823 and close this PR for now.

Thanks again for helping us uncover this!

PR #3820, opened 2026-07-13
fix(super-editor): use fontFamily key for markdown code block monospace font

### LINKED ISSUE #3819: Fenced markdown code blocks render in the body font instead of monospace
### What happened?

When I open a markdown file with a code block, the code is rendered in the body font, not the monospace font Courier New.

Here is my simple markdown file:
[lorem-epsom-dense-monolith.md](https://github.com/user-attachments/files/29987263/lorem-epsom-dense-monolith.md)

### Steps to reproduce

1. Open lorem-epsom-dense-monolith.md
2. View the bash code block

### SuperDoc version

1.44.1

### Browser

None

### Additional context

I see the code block rendered in the body font instead of Courier New: Image

I expected to see it rendered in Courier New: 

Image

smxmdrissel:
## What changed

Fixes a key mismatch that silently dropped the monospace font override on fenced markdown code blocks.

`convertCodeBlock()` in the markdown → ProseMirror converter set the monospace run-property override under the key `rFonts`. Every downstream consumer — the `w:rFonts` OOXML translator (`sdNodeOrKeyName: 'fontFamily'`) and the style-engine's cascade resolution — reads run font overrides under the key `fontFamily`. Because the keys never matched, the Courier New override was dropped before reaching the style-engine/DomPainter, so fenced markdown code blocks rendered in the default body font instead of monospace.

This also fixes the same key mismatch in the reverse direction: `proseMirrorToMdast.ts`'s `isMonospaceRun()` checked `runProps.rFonts` instead of `runProps.fontFamily`, which would have prevented a docx-imported monospace run from round-tripping back to markdown inline-code spans on export.

Additionally, each code-block run now gets a `textStyle` mark (`{ fontFamily: 'Courier New' }`) alongside the direct `runProperties.fontFamily` attrs. The direct attrs remain as a fallback for callers that inspect the converter's raw JSON output without dispatching through a live `Editor`; once dispatched, `calculateInlineRunPropertiesPlugin`'s mark-based recalculation recomputes `runProperties.fontFamily` from the mark anyway (mirroring how `inlineCode` already works).

A shared `MARKDOWN_MONOSPACE_FONT` constant now backs both conversion directions so they can't drift apart again.

**Known limitation (not addressed here):** `proseMirrorToMdast.ts` has no path that emits an mdast fenced `code` block today, so a multi-line monospace run round-trips to a paragraph of `inlineCode` spans joined by line breaks, not to a fenced code block. Genuine fenced-block round-tripping is a separate, larger change.

## Test plan

- Added `mdastToProseMirror.test.ts` covering fenced code block conversion, asserting both the `textStyle` mark and the direct `runProperties.fontFamily` attrs are set (and that the incorrect `rFonts` key is never used).
- Updated `proseMirrorToMdast.test.ts` to exercise the corrected `fontFamily` key in `isMonospaceRun()`.
- Updated `calculateInlineRunPropertiesPlugin.test.js` to cover mark-based recalculation of code-block runs.
- Verified end-to-end via the CLI binary and the superdoc-docx demo service's `/convert` endpoint.
- `pnpm test` (scoped to the touched test files) — 76/76 passing.

Closes #3819

### DIFF
--- a/packages/super-editor/src/editors/v1/core/helpers/markdown/constants.ts
+++ b/packages/super-editor/src/editors/v1/core/helpers/markdown/constants.ts
@@ -1,0 +1,18 @@
+/**
+ * Shared constants for the Markdown ↔ ProseMirror conversion pipeline.
+ */
+
+/**
+ * The font family used to mark monospace ("code") text in both directions of
+ * the markdown ↔ ProseMirror conversion:
+ *  - `mdastToProseMirror.ts` applies this font (via a `textStyle` mark and/or
+ *    direct `runProperties.fontFamily`) to fenced code blocks (`code`) and
+ *    inline code spans (`inlineCode`).
+ *  - `proseMirrorToMdast.ts` detects this font (on either the mark or the
+ *    direct run properties) to convert monospace runs back into `inlineCode`
+ *    mdast nodes.
+ *
+ * Keeping this as a single shared constant avoids the two directions
+ * silently drifting out of sync if the monospace font ever changes.
+ */
+export const MARKDOWN_MONOSPACE_FONT = 'Courier New';
--- a/packages/super-editor/src/editors/v1/core/helpers/markdown/mdastToProseMirror.test.ts
+++ b/packages/super-editor/src/editors/v1/core/helpers/markdown/mdastToProseMirror.test.ts
@@ -1,0 +1,163 @@
+import { describe, expect, it } from 'vitest';
+import { Schema } from 'prosemirror-model';
+import { convertMdastToBlocks } from './mdastToProseMirror.js';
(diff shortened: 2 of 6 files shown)


caio-pizzol:
Thanks again for digging into this and putting together such a thoughtful PR. Your diagnosis of the `rFonts` and `fontFamily` mismatch is correct.

We have decided not to expand our Markdown support right now, so we will not be merging this. We shared more context in [the related issue](https://github.com/superdoc-dev/superdoc/issues/3819#issuecomment-4968660653).

This is a product-scope decision, not an issue with your work. We really appreciate the time and care you put into the contribution.

PR #3829, opened 2026-07-15
fix(comments): don't drop DOCX comments containing bookmark nodes (#3828)

### LINKED ISSUE #3828: DOCX comment bookmarks cause `Unknown node type: bookmarkStart` and drop comments
### Summary

DOCX comments containing `w:bookmarkStart` / `w:bookmarkEnd` nodes fail during comment HTML conversion with:

```
RangeError: Unknown node type: bookmarkStart
Failed to convert comment
```

The affected comment is then omitted from the loaded comments list.

### Minimal shape

A comment containing an invisible bookmark pair around visible mention text is sufficient:

```xml
<w:comment w:id="1">
  <w:p>
    <w:r><w:t>Before mention </w:t></w:r>
    <w:bookmarkStart w:id="42" w:name="_mention"/>
    <w:r><w:t>@Person</w:t></w:r>
    <w:bookmarkEnd w:id="42"/>
  </w:p>
</w:comment>
```

### Root cause

The full DOCX editor schema includes `bookmarkStart` and `bookmarkEnd`, so the comment XML is converted to valid DOCX-schema JSON.

During `processLoadedDocxComments`, `getHtmlFromComment` normalizes that JSON and creates a headless text editor using `getRichTextExtensions()`. That smaller schema does not include bookmark nodes, but `normalizeCommentForEditor` currently passes them through. `schema.nodeFromJSON` throws, the catch returns no HTML, and this guard skips the non-tracked comment:

```js
if (!htmlContent && !comment.trackedChange) {
  return;
}
```

### Expected behavior

Invisible bookmark boundary nodes should not prevent the visible comment content from importing.

### Suggested narrow fix

Ignore only bookmark boundary nodes while producing the comment's display HTML:

```js
if (node.type === 'bookmarkStart' || node.type === 'bookmarkEnd') {
  return null;
}
```

This leaves the original `docxCommentJSON` unchanged, so the imported DOCX representation still retains the bookmark metadata. In a local reproduction with two affected comments, both comments imported successfully, visible text and mentions remained intact, the original bookmark JSON was preserved, and the console errors disappeared.

### Versions checked

- `@harbour-enterprises/superdoc@1.44.1`
- `@harbour-enterprises/superdoc@1.45.0`
- Current `main` implementation of `packages/superdoc/src/stores/comments-store.js`

gpardhivvarma:
## Summary

Fixes #3828. DOCX comments containing `w:bookmarkStart` / `w:bookmarkEnd` nodes were silently dropped from the loaded comments list, with `RangeError: Unknown node type: bookmarkStart` / `Failed to convert comment` logged to the console.

## Root cause

A comment's imported ProseMirror JSON (`comment.elements`) is produced against the **full DOCX schema**, which registers `bookmarkStart`/`bookmarkEnd`. But `getHtmlFromComment` renders the comment's display HTML with a throwaway headless `Editor` built from the **reduced rich-text schema** (`getRichTextExtensions()`), which does **not** register those nodes.

`normalizeCommentForEditor` forwarded the bookmark nodes verbatim, so ProseMirror's `nodeFromJSON` threw `Unknown node type: bookmarkStart`. The `try/catch` in `getHtmlFromComment` swallowed the error and returned `undefined`, and the guard `if (!htmlContent && !comment.trackedChange) return;` in `processLoadedDocxComments` then discarded the entire comment.

## Fix

Normalize defensively: drop or unwrap **any** node type the rich-text schema does not register — not just bookmarks — so this whole class of crash is prevented (stray `lineBreak`, `tab`, etc. would fail the same way). The supported-node set is derived directly from the extension objects' `.type`/`.name` (ProseMirror adds no implicit nodes, so no schema construction is needed).

- Invisible leaf nodes (e.g. bookmark boundaries) are dropped.
- Nodes wrapping inline content are unwrapped so the **visible text survives**.
- The original `docxCommentJSON` is untouched, so exported bookmark metadata is preserved on round-trip.

The normalizer is extracted into a pure `comment-normalize.js` module so it can be unit-tested directly.

## Tests

- **`comment-normalize.test.js`** (new) — fast, pure unit tests: bookmark stripping, unknown-node unwrap/drop, supported-node passthrough, `textStyle` font-attr stripping, and the empty-set fallback (strip bookmarks only).
- **`comments-store.bookmark.test.js`** (new) — end-to-end regression test using the **real** headless `Editor` (the existing `comments-store.test.js` mocks `Editor`, so it could not reproduce the schema throw). Verified to **fail before** the fix (comment dropped) and **pass after**.
- All **136** existing `comments-store` tests still pass.

## Files

- `packages/superdoc/src/stores/comment-normalize.js` *(new)* — defensive normalizer + `getRichTextSupportedNodeNames`.
- `packages/superdoc/src/stores/comments-store.js` — import the helpers, remove the inline normalizer, thread the supported-node set through `getHtmlFromComment`.
- `packages/superdoc/src/stores/comment-normalize.test.js`, `packages/superdoc/src/stores/comments-store.bookmark.test.js` *(new tests)*.

### DIFF
--- a/packages/superdoc/src/stores/comment-normalize.js
+++ b/packages/superdoc/src/stores/comment-normalize.js
@@ -1,0 +1,106 @@
+/**
+ * Node names the reduced rich-text schema registers, derived from the extension
+ * objects themselves (each carries `.type === 'node'` and `.name`). ProseMirror adds
+ * no implicit nodes, so this set equals the schema's node set. Used to strip nodes
+ * that would make `nodeFromJSON` throw when rendering comment HTML. (#3828)
+ *
+ * @param {Array<{ type?: string, name?: string }>} [extensions] Editor extensions.
+ * @returns {Set<string>} Registered node type names.
+ */
+export const getRichTextSupportedNodeNames = (extensions = []) =>
+  new Set(
+    (Array.isArray(extensions) ? extensions : [])
+      .filter((ext) => ext?.type === 'node')
+      .map((ext) => ext?.name)
+      .filter(Boolean),
+  );
+
+/**
+ * Normalize imported DOCX comment JSON into content the reduced rich-text schema can
+ * load. Unwraps `run` nodes, strips font attrs from `textStyle` marks, and drops any
+ * node the rich-text schema does not register (e.g. `bookmarkStart`/`bookmarkEnd`),
+ * preserving inline content those nodes wrap so the visible text survives. The caller's
+ * original `docxCommentJSON` is untouched, so exported metadata (bookmarks, etc.) is
+ * retained.
+ *
+ * When `supportedNodeNames` is empty/undefined (schema unavailable) it falls back to
+ * stripping only the invisible bookmark boundary nodes. (#3828)
+ *
+ * @param {*} node A ProseMirror JSON node, array of nodes, or primitive.
+ * @param {Set<string>} [supportedNodeNames] Node names the target schema supports.
+ * @returns {*} Normalized node(s), or `null` for a dropped leaf node.
+ */
+export const normalizeCommentForEditor = (node, supportedNodeNames) => {
+  if (Array.isArray(node)) {
+    return node
+      .map((child) => normalizeCommentForEditor(child, supportedNodeNames))
+      .flat()
+      .filter(Boolean);
+  }
+
+  if (!node || typeof node !== 'object') return node;
+
+  // Drop invisible bookmark boundary nodes and any node absent from the rich-text
+  // schema used to render comment HTML, preserving inline content they wrap so the
+  // visible text survives. Falls back to stripping only bookmark boundary nodes when
+  // the supported-node set can't be determined (size 0). (#3828)
+  const isBoundaryNode = node.type === 'bookmarkStart' || node.type === 'bookmarkEnd';
+  const isUnsupported = supportedNodeNames && supportedNodeNames.size > 0 && !supportedNodeNames.has(node.type);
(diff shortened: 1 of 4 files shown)


gpardhivvarma:
hey @caio-pizzol . Can you please take a look at this. Coudn't contribute for some time because of some personal reasons. Good to be back

gpardhivvarma:
Addressed the automated review feedback in 547d56006:

- **Tabs / line breaks dropped** (Codex & Qodo): `normalizeCommentForEditor` no longer maps *every* unsupported leaf to `null`. Visible whitespace leaves (`lineBreak`/`hardBreak` → `\n`, `tab` → `\t`) are now mapped to text so word separation survives in the rendered HTML; truly invisible nodes (bookmark boundaries, etc.) are still dropped. Added a unit test covering `One<br/>Two<tab>Three` → `One\nTwo\tThree`. (Note: before this PR, such comments crashed and were dropped entirely, so this is strictly an improvement either way.)
- **Fake timers not restored** (Qodo): added `afterEach(() => vi.useRealTimers())` to `comments-store.bookmark.test.js`, matching the pattern in `comments-store.test.js`.

All 145 local tests pass (8 unit + 1 e2e + 136 existing).

Re the two red CI checks (`collaboration (liveblocks)` / `CI Examples / validate`): those fail at the "Create .env for cloud providers" step because `VITE_LIVEBLOCKS_PUBLIC_KEY` isn't available to fork PRs — unrelated to this change.

gpardhivvarma:
@caio-pizzol

Issue #3824, opened 2026-07-14
Security advisory: malicious AsyncAPI package versions published (Miasma RAT)


kunalsingh-safedep:
**Security advisory from [SafeDep](https://safedep.io).** On 14 July 2026 an attacker published malicious, credential-stealing versions of four `@asyncapi` npm packages. Your dependency graph pulls in one of them. You are most likely safe. You are only exposed if your lockfile pins a malicious version below.

**How it reaches you** (from your GitHub dependency graph):

- `superdoc-dev/superdoc` → `mintlify@4.2.624` → `@mintlify/cli@4.0.1227` → `@mintlify/common@1.0.947` → `@asyncapi/specs@6.8.1`

[Verify it yourself](https://github.com/superdoc-dev/superdoc/network/dependencies)

**Malicious → safe versions**

- `@asyncapi/generator` `3.3.1` → `3.3.0`
- `@asyncapi/generator-helpers` `1.1.1` → `1.1.0`
- `@asyncapi/generator-components` `0.7.1` → `0.7.0`
- `@asyncapi/specs` `6.11.2` / `6.11.2-alpha.1` → `6.11.1`

**What to do:** grep your lockfile for the malicious versions. If none appear, pin the safe versions and you are done. If one appears, the payload dropped a `sync.js` backdoor and stole credentials, so rotate the build machine's npm, GitHub, SSH, and AWS tokens plus browser passwords.

Full analysis: https://safedep.io/asyncapi-generator-supply-chain-attack-miasma-rat/

PR #3830, opened 2026-07-15
chore(security): pin safe @asyncapi versions (Miasma RAT advisory, #3824)

gpardhivvarma:
## Summary

Addresses #3824 — the 2026-07-14 **Miasma RAT** supply-chain attack that published malicious, credential-stealing versions of four `@asyncapi` packages. superdoc reaches `@asyncapi/specs` via the `apps/docs` devDependency `mintlify@4.2.624` → `@mintlify/common` / `@asyncapi/parser` → `@asyncapi/specs`.

## Exposure assessment

**Not currently exposed** — verified across every lockfile in the repo:

- Root `pnpm-lock.yaml` resolves only `@asyncapi/specs@6.8.1` (safe) and `@asyncapi/parser@3.4.0` (not on the advisory list). None of the malicious versions appear; the three `@asyncapi/generator*` packages aren't in the tree at all.
- `demos/grading-papers/pnpm-lock.yaml`, `.github/scripts/package-lock.json`, and `tests/consumer-typecheck/package-lock.json` contain no `@asyncapi`/`mintlify`.

**But the bad version is genuinely reachable:** `@asyncapi/parser@3.4.0` declares `"@asyncapi/specs": "^6.8.0"` (`>=6.8.0 <7.0.0`), which includes the malicious `6.11.2`. An unfrozen re-resolution (e.g. `visual-baseline.yml` runs `pnpm install --no-frozen-lockfile`) could adopt it.

**No credential rotation is required** — no malicious version was ever installed. (That step only applies if a malicious version had been pinned.)

## Change

Surgical `pnpm.overrides` (root `package.json`) that rewrite **only** the exact malicious versions to their safe counterparts, for all four advisory packages:

| Package | Malicious → Safe |
|---|---|
| `@asyncapi/specs` | `6.11.2`, `6.11.2-alpha.1` → `6.11.1` |
| `@asyncapi/generator` | `3.3.1` → `3.3.0` |
| `@asyncapi/generator-helpers` | `1.1.1` → `1.1.0` |
| `@asyncapi/generator-components` | `0.7.1` → `0.7.0` |

- Exact-version selectors keep the blast radius minimal: the currently-resolved `@asyncapi/specs@6.8.1` is untouched (no forced upgrade), and a legitimate future `6.11.3` would still be allowed. `6.11.1` satisfies parser's `^6.8.0`.
- The three generator entries are defensive (absent from the tree today) and complete the advisory's list so any future dependency pulling them is auto-remediated.
- Matches the repo's existing version-pinned override style (`protobufjs@7`/`protobufjs@8`) and pnpm's documented security-override pattern.

## Verification

- `git diff` is exactly **10 insertions** — five in `package.json`, five in the `pnpm-lock.yaml` `overrides:` section. **No package resolution changes / no lockfile churn.**
- `pnpm install --frozen-lockfile` succeeds (lockfile internally consistent — matters because most CI jobs run frozen-by-default).
- Post-change scan confirms no malicious version resolves and `@asyncapi/specs@6.8.1` is unchanged.


gpardhivvarma:
@caio-pizzol

caio-pizzol:
Thanks, Pardhiv, for investigating this and putting together the safeguards, and sorry for the delayed response. We checked the current V2 manifests and lockfiles: the Mintlify → AsyncAPI dependency chain is no longer present, so these overrides are no longer needed for V2. With our development focus now on V2, we’re closing this V1 PR. We appreciate the contribution.

PR #3838, opened 2026-07-21
fix(super-editor): flip toolbar tooltip below the trigger when it clips the viewport top

### LINKED ISSUE #3786: Toolbar tooltips are cut off by edge of the window
### What happened?

When toolbar is at the top of the page the tooltips on buttons are cut off by the edge of the window.

### Steps to reproduce

1. Open demo https://codepen.io/jcubic/pen/XJpeEZW
2. Hover over the toolbar buttons

### SuperDoc version

1.43.2

### Browser

Helium (Chromium based)

### Additional context

The tooltips should appear below the toolbar if the top edge is outside the page (the top edge is negative).

Image

hitpopdimestop:
## What & why

Toolbar tooltips were clipped when the toolbar sits flush with the window top. The tooltip was always placed _above_ the trigger, so `triggerRect.top - contentHeight - 10` went negative and rendered above the viewport edge (issue #3786).

It now:

- **flips below** the trigger (arrow pointing up, transition origin adjusted) when there is no room above, and
- clamps horizontally to `document.documentElement.clientWidth`, which excludes the window scrollbar.

## Screenshots

**Before** — the "Insert table" tooltip is clipped at the window top:
Знімок екрана 2026-07-21 о 14 59 10

**After** — the tooltip flips below the button:
Знімок екрана 2026-07-21 о 14 59 22

## Test plan

- `SdTooltip.test.js`: flip-below, stay-above, a realistic-height flip (proves the tooltip's height is used, not just when the trigger touches the top), and a `clientWidth < innerWidth` horizontal-clamp test.
- Toolbar test suite passing; Prettier clean.

Closes #3786

### DIFF
--- a/packages/super-editor/src/editors/v1/components/toolbar/SdTooltip.test.js
+++ b/packages/super-editor/src/editors/v1/components/toolbar/SdTooltip.test.js
@@ -1,12 +1,55 @@
 import { afterEach, describe, expect, it, vi } from 'vitest';
-import { mount } from '@vue/test-utils';
+import { flushPromises, mount } from '@vue/test-utils';
 import { nextTick } from 'vue';
 import SdTooltip from './SdTooltip.vue';
+
+async function openTooltipWithTriggerRect(rect) {
+  const wrapper = mount(SdTooltip, {
+    attachTo: document.body,
+    props: { delay: 0, duration: 0 },
+    slots: { trigger: '<button type="button">Bold</button>', default: 'Bold' },
+  });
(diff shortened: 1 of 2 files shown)


caio-pizzol:
Thanks for the contribution, @hitpopdimestop! I confirmed the issue, but this PR targets v1 and we’re focusing development on v2.

The same issue still exists in v2, so we’d be happy to review an updated PR against `main`. I’m going to close this one for now. Thanks again!

PR #3840, opened 2026-07-21
perf(super-editor): materialize editor commands lazily

hitpopdimestop:
## What & why

`CommandService.get commands()` rebuilt a wrapper method for **every** registered command (276 of them) on **every** `editor.commands` read. That object is read on hot paths — the toolbar checks command state on each selection change, and `.commands` / `.can()` is read across ~1,400 sites — so each read allocated hundreds of throwaway closures, adding avoidable GC pressure during editing.

The getter now returns a `Proxy` that materializes only the command actually accessed (cached for that read). No call site iterates the commands object (no `Object.keys` / spread), so the lazy path is never forced to build them all.

## Impact

This is a small, low-risk efficiency cleanup — **not** a user-facing latency fix; the per-read saving is sub-millisecond in isolation. The point is removing obviously-wasteful repeated allocation on a very hot path.

A quick microbenchmark of the getter construction (50,000 single-command accesses):

| | time | wrapper closures |
| --- | --- | --- |
| before (eager, builds all) | ~2600 ms | ~13.8M |
| after (lazy proxy) | ~10 ms | ~50k |

## Behavior

Behavior-preserving: the proxy emulates the plain object for `get` / `has` / `ownKeys` / `delete` / `defineProperty` / `preventExtensions`. Covered by 33 `CommandService` tests.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/CommandService.js
+++ b/packages/super-editor/src/editors/v1/core/CommandService.js
@@ -55,14 +55,18 @@ export class CommandService {
    * editor.commands.insertText('hello'); // falls back to editor.dispatch
    */
   get commands() {
-    const { editor, state } = this;
+    const { editor, state, rawCommands } = this;
     const { view } = editor;
     const { tr } = state;
     const props = this.createProps(tr);
 
-    const entries = Object.entries(this.rawCommands).map(([name, command]) => {
-      /** @type {(...args: any[]) => boolean} */
-      const method = (...args) => {
+    /**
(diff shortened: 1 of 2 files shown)


caio-pizzol:
Thanks @hitpopdimestop! I confirmed the speed improvement is real.

V2 already avoids this issue by keeping command methods stable instead of rebuilding them on every access. So we’re going to focus our performance work there and close this v1 PR.

Thanks again for the thoughtful contribution!

PR #3845, opened 2026-07-23
fix(super-editor): preserve rows without cells when merging

### LINKED ISSUE #3782: Merging multiple rows in a table causes formatting issues
### What happened?

Merging multiple rows in a table causes formatting issues。

Image

Image

### Steps to reproduce

_No response_

### SuperDoc version

_No response_

### Browser

_No response_

### Additional context

_No response_

bogdanyluk:
Fixes [#3782](https://github.com/superdoc-dev/superdoc/issues/3782)

## Summary

Fixes table formatting corruption after merging multiple full-width rows in an imported DOCX.

## Root cause

ProseMirror represents a full-width vertical merge as:

- A merged cell with `colspan` and `rowspan`
- An empty continuation row
- The following normal rows

When serialized, the empty row has no `content` property. The layout adapter discarded this row, causing the next row to be processed while the merged columns were still occupied. AutoFit then placed its cells in additional columns, widening the table and rendering cells outside its right boundary.

## Fix

- Preserve structurally empty table rows in the layout adapter.
- Normalize missing row content to an empty array.
- Keep empty rows in the table’s row timeline so rowspans advance correctly.
- Continue omitting tables where every row is empty.
- Add regression coverage for a full-width, multi-row merge followed by a normal row.

#### Before
зображення

#### After
зображення

## Test Plan
- Import `.docx` with a table (from the issue).
- Confirm the table initially renders with three aligned columns.
- Select every cell across two or more adjacent blank rows and merge them.
- Verify the merged cell spans the full table width and both selected rows.
- Verify all subsequent rows remain aligned to the original three-column grid.
- Verify no cells protrude beyond the table’s right edge.
- Undo and redo the merge and confirm the table remains correctly formatted.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/layout-adapter/converters/table.test.ts
+++ b/packages/super-editor/src/editors/v1/core/layout-adapter/converters/table.test.ts
@@ -985,6 +985,55 @@ describe('table converter', () => {
       expect(result.rows[0].cells[0].colSpan).toBe(3);
     });
 
+    it('preserves content-less rows covered by a full-width rowspan', () => {
+      const cell = (text: string): PMNode => ({
+        type: 'tableCell',
+        content: [{ type: 'paragraph', content: [{ type: 'text', text }] }],
+      });
+      const node: PMNode = {
+        type: 'table',
+        attrs: {
+          grid: [{ col: 1000 }, { col: 1000 }, { col: 1000 }],
+        },
+        content: [
+          {
+            type: 'tableRow',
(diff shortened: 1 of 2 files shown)


caio-pizzol (approved):
Thanks, @bogdanyluk! I reproduced the issue and confirmed that this fix preserves the empty rows covered by the vertical merge. The following rows now stay aligned with the original three-column grid, including after undo and redo. The fix looks good. Approved!

PR #3851, opened 2026-07-26
Fix undo and redo after collaboration upgrade

jelkes1:
## Summary

- detect the editor's active history backend whenever a history snapshot is read
- switch existing local history adapters to Yjs after `upgradeToCollaboration()` attaches collaboration in place
- add regression coverage for the late collaboration upgrade path

## Root cause

`EditorHistorySnapshotAdapter` cached whether the editor was collaborative in its constructor. The unified history coordinator creates the adapter while the editor is still local, so an instance later upgraded with `upgradeToCollaboration()` continued reading the removed ProseMirror history stack. Undo and redo depths therefore remained zero even when the Yjs UndoManager contained entries, leaving both controls disabled.

## Impact

Editors upgraded from local mode to collaboration now expose the Yjs undo/redo depths to the unified history coordinator and toolbar.

## Validation

- `vitest` history adapter, coordinator, and PresentationEditor suites: 199 passed, 5 skipped
- `vitest` `upgrade-collaboration.test.js`: 14 passed
- Prettier check for both changed files: passed
- `git diff --check`: passed

The package-wide `types:check` remains red on existing unrelated JS/TS migration errors across the package; no reported error points to either changed file.

### DIFF
--- a/packages/super-editor/src/editors/v1/core/presentation-editor/history/editor-history-snapshot-adapter.integration.test.ts
+++ b/packages/super-editor/src/editors/v1/core/presentation-editor/history/editor-history-snapshot-adapter.integration.test.ts
@@ -1,0 +1,88 @@
+import { describe, expect, it } from 'vitest';
+import { Awareness } from 'y-protocols/awareness';
+import * as Y from 'yjs';
+import type { Editor } from '../../Editor.js';
+import { seedEditorStateToYDoc } from '../../../extensions/collaboration/seed-editor-to-ydoc.js';
+import { initTestEditor } from '../../../tests/helpers/helpers.js';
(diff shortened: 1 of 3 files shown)


caio-pizzol:
Thanks for the contribution, @jelkes1! This only affects v1, and Redo still doesn’t work after Undo. Since collaboration works differently in v2, we’re going to close this and focus there instead.

caio-pizzol:
Thanks, @jelkes1! I reproduced the V1 bug and confirmed the fix works. I pushed a test covering the full edit → upgrade → edit → undo → redo flow. All related tests and the full editor suite pass.

One limitation: edits made before the upgrade still can’t be undone afterward.

caio-pizzol (approved):

PR #3864, opened 2026-08-04
fix(super-converter): tolerate empty style properties on table styles (#3861)

### LINKED ISSUE #3861: Empty `<w:rPr />` (and similar) in `styles.xml` table styles aborts DOCX import
### What happened?

Opening certain valid DOCX files in SuperDoc results in an empty document body (import appears to abort). The same files open correctly in Microsoft Word.

We traced this to empty OOXML property containers in `word/styles.xml` (commonly emitted by LibreOffice and some Word templates), for example a self-closing `<w:rPr />` on a table style such as Table Grid.

In SuperDoc 1.43, `_getReferencedTableStyles` appears to dereference these empty property nodes unsafely and throw. The exception aborts import, so the editor mounts with little or no body content.

We cannot share the original customer document, but the trigger is reproducible from a minimal `styles.xml` fragment (below). Stripping empty `w:rPr` / `w:pPr` / `w:tblPr` / `w:tcPr` elements from `styles.xml` before load makes the same document import successfully.

### Expected behavior

Empty / self-closing style property containers should be treated as no-ops. The document body should import and render.

### Actual behavior

Import fails / aborts when `styles.xml` contains empty property containers on referenced table styles. The editor can show a blank body.

### Steps to reproduce

1. Create (or edit) a DOCX whose `word/styles.xml` includes a table style with an empty run-properties element, e.g.:

```xml
<w:style w:type="table" w:styleId="TableGrid" w:styleName="Table Grid">
  <w:basedOn w:val="TableNormal"/>
  <w:rPr/>
</w:style>
```

(Equivalent forms also fail: `<w:rPr />`, and empty paired tags like `<w:rPr></w:rPr>`. Same class of issue for empty `w:pPr`, `w:tblPr`, and `w:tcPr`.)

2. Ensure the document body references that table style (any simple table using Table Grid is enough).

3. Open the DOCX in SuperDoc.

4. Observe blank / incomplete import.

5. Remove the empty property containers from `styles.xml`, re-zip, and reopen — import succeeds.

### SuperDoc version

1.43.x (also observed via `@superdoc-dev/react`)

### Browser

Chrome (macOS)

### Additional context

- This is valid OOXML; Word and other viewers tolerate empty property containers.
- Suggested fix: null-check / skip empty property nodes in `_getReferencedTableStyles` (and any similar style-resolution paths) rather than assuming child props exist.

### Steps to reproduce

_No response_

### SuperDoc version

1.43.x

### Browser

Chrome

### Additional context

_No response_

caio-pizzol:
Keeps tables in documents whose `word/styles.xml` carries empty property containers, such as a self-closing `<w:rPr/>` on Table Grid. Word opens these files; SuperDoc was dropping the tables. Issue #3861 reports LibreOffice and some Word templates as the source.

xml-js omits the `elements` key rather than emitting an empty one, so `<w:rPr/>`, `<w:rPr />` and `<w:rPr></w:rPr>` all arrive with no children, and the style reader assumed that key was there. The importer catches per node, so the throw was swallowed and only the table vanished, which is why an affected document looked blank rather than raising an error.

- Reading through shared accessors closes five unguarded spots, not just the reported `w:rPr`. A `w:basedOn` pointing at a childless style throws first, before execution reaches the `w:rPr`, and with an identical message, so fixing only the reported line could look like no fix at all.
- Conditional formatting entries with no `w:type` are skipped rather than keyed under `undefined`. ECMA-376 marks that attribute required.
- Empty `w:tblPr`, `w:tcPr` and `w:trPr` were already safe. The issue lists them as broken, so tests record that they are not.
- Fixtures parse XML strings deliberately. Hand-authored nodes using `elements: []` do not reproduce the parser shape and pass against this bug.

Byte-identical across v1.43.2, v1.44.2, v1.45.2 and main.

Fixes superdoc/docx-editor#3861

### DIFF
--- a/packages/super-editor/src/editors/v1/core/super-converter/docx-helpers/index.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/docx-helpers/index.js
@@ -2,3 +2,4 @@
 export * from './docx-helpers.js';
 export * from './docx-constants.js';
 export * from './get-default-style-definition.js';
+export * from './xml-node-access.js';
--- a/packages/super-editor/src/editors/v1/core/super-converter/docx-helpers/xml-node-access.js
+++ b/packages/super-editor/src/editors/v1/core/super-converter/docx-helpers/xml-node-access.js
@@ -1,0 +1,54 @@
+// @ts-check
+
+/**
+ * Safe reads over xml-js element trees.
+ *
+ * AIDEV-NOTE: xml-js omits `elements` and `attributes` entirely rather than
+ * emitting empty ones, so `<w:rPr/>`, `<w:rPr />` and `<w:rPr></w:rPr>` all
+ * parse to `{ type: 'element', name: 'w:rPr' }` with no `elements` key. Every
+ * child of `CT_Style` is optional in ECMA-376, so empty property containers are
+ * schema-valid, but readers that assume `node.elements` exists throw on them
+ * (issue #3861). Reach into parsed OOXML through these helpers rather than
+ * dereferencing `.elements` or `.attributes` directly.
+ */
+
+/**
+ * A node in a parsed xml-js tree.
+ * @typedef {object} XmlNode
+ * @property {string} [type] Node kind, e.g. `element` or `text`.
+ * @property {string} [name] Qualified element name, e.g. `w:rPr`.
+ * @property {Record<string, string>} [attributes] Absent when the element has no attributes.
+ * @property {XmlNode[]} [elements] Absent when the element has no children.
+ * @property {string} [text] Character data, for text nodes.
+ */
+
+/**
+ * Child elements of a parsed OOXML node, or an empty array when it has none.
+ * @param {XmlNode | null | undefined} node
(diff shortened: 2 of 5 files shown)


caio-pizzol:
Ran the schema lookups the review could not, against the ECMA-376 transitional XSD graph.

`w:type` on `CT_TblStylePr` is confirmed **required**:

| name | use | type |
| --- | --- | --- |
| type | required | `ST_TblStyleOverrideType` |

So a `w:tblStylePr` without it is schema-invalid and cannot be addressed by the cascade. Dropping it is correct rather than a spec violation, which matches the review.

Also confirmed while here, since the same reasoning underpins the rest of the diff:

- Every child of `CT_Style` is `minOccurs=0`, so an empty `<w:rPr/>`, `<w:pPr/>` or bare `<w:style/>` is schema-valid input. That is the premise of the whole fix.
- `w:rFonts` carries only optional attributes, so the `fonts?.attributes` guard the review called out is load-bearing for real files, not defensive noise.

Issue #3507, opened 2026-05-26
Caret rendered below text baseline after Tab (cosmetic only)


nathan-ahs:
### What happened?

When typing text in the editor and pressing **Tab** to insert indentation, the caret is **rendered slightly below the current line** rather than aligned with the text baseline.

Visually, the cursor appears dropped down below the line it should be on. Typing any character realigns the caret back to the current line, so the actual insertion position is correct — only the caret's rendered position between the Tab and the next keystroke is wrong.

The bug is purely visual:
- The caret *looks* dropped below the line after pressing Tab.
- If you type a character, it is inserted on the correct line, and the caret then snaps back to its expected vertical position.

**Expected:** the caret should render aligned with the current line's text baseline immediately after the Tab.

**Actual:** the caret renders slightly below the current line until the next keystroke.

### Steps to reproduce

1. Open SuperDoc (reproduced on the public demo and in a local integration on v1.35.0).
2. Click into a paragraph and type any text, e.g. `hello`.
3. Press **Tab** to insert indentation.
4. Observe the caret — it is rendered slightly below the current line.
5. Type any character — it is inserted on the correct line and the caret snaps back to its expected position.

### SuperDoc version

`1.35.0` (also reproduced on the live demo)

### Browser

Firefox

vivienne-sd:
@nathan-ahs thanks for reporting this cursor issue as well. It's also tracked, and I'll update you on both next Wednesday (06/03). Please let us know if you have any questions!

vivienne-sd:
@nathan-ahs Thanks for your patience; this is set to be prioritized within the coming week. I'll follow up with you then as soon as it's in review.

vivienne-sd:
@nathan-ahs , a fix for the caret placement discrepancy is now live in [v1.40.0-next.10](https://github.com/superdoc-dev/superdoc/releases/tag/v1.40.0-next.10) and laster versions 🙌 

Thank you again for the detailed write-up! We'll be moving to a stable release by EOD Friday if you want to wait to upgrade until then.

PR #3870, opened 2026-08-05
fix(super-editor): anchor header/footer caret to the line box after a tab

PopaSerban:
PR description:

Typing in a header or footer and pressing Tab renders the caret about 5.5px below its own line, hanging past the bottom of the row. Typing any character snaps it back, so the insertion point was always correct — only the caret's painted position between the Tab and the next keystroke was wrong.

This is the same symptom as #3507, which #3677 fixed for body text in June. That fix only touched `DomSelectionGeometry.ts`, so the header/footer caret path never got it.

Refs #3507, #3677

<table>
<thead>
<tr><th>Before</th><th>After</th></tr>
</thead>
<tbody>
<tr>
<td>

https://github.com/user-attachments/assets/c919d8a1-e15a-406d-917b-17edf50b7f89

</td>
<td>

https://github.com/user-attachments/assets/b110c662-1759-4db9-aff3-f5a9516f72c8

</td>
</tr>
</tbody>
</table>

<a href="https://cubic.dev/pr/superdoc/docx-editor/pull/3870?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


caio-pizzol (approved):
Thanks for the fix, @PopaSerban! I reproduced the bug and confirmed this PR fixes it. The code and tests look good.

Approving - the failing CI check is unrelated.

PR #3872, opened 2026-08-05
chore(pnpm): consolidate workspace settings into pnpm-workspace.yaml

caio-pizzol:
Moves the pnpm settings still living in `package.json#pnpm` into `pnpm-workspace.yaml`, which is where pnpm has read them since 10.6.2 and the only place pnpm 11 will read them at all.

Keeping both locations had already let them drift, which is the real reason to do this now. `better-sqlite3` was allowed to run build scripts only via `package.json`, and the `openapi-types` patch was declared only there, so the workspace file was quietly incomplete.

Resolution is deliberately untouched, and that is the thing worth checking:

- `pnpm-lock.yaml` is byte-identical before and after
- `pnpm install --frozen-lockfile` succeeds, and both patches are applied on disk afterwards
- no dependency version moves; only two files change

Landing it separately also shrinks the diff of the Vite and Vitest upgrades stacked behind it, since a Vite+ migration would relocate these same keys anyway.

<a href="https://cubic.dev/pr/superdoc/docx-editor/pull/3872?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,56 +123,5 @@
   },
   "optionalDependencies": {
     "canvas": "catalog:"
-  },
-  "pnpm": {
-    "overrides": {
-      "canvas": "3.2.3",
-      "happy-dom": "20.4.0",
-      "jsdom": "27.3.0",
-      "vue": "3.5.32",
-      "@vue/compiler-core": "3.5.32",
-      "@vue/compiler-dom": "3.5.32",
-      "@vue/compiler-sfc": "3.5.32",
-      "@vue/runtime-core": "3.5.32",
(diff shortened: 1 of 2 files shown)


caio-pizzol:

PR #3873, opened 2026-08-05
chore(deps): move from the rolldown-vite 7 alias to Vite 8

caio-pizzol:
Vite 8 ships Rolldown as its own bundler, so the `npm:rolldown-vite@7.3.1` override was pointing at a preview of something Vite now provides directly. Swapping it for a plain Vite 8 range removes the indirection and puts the workspace on the supported release line.

The override itself stays, because it is what keeps every workspace member on one Vite runtime. Examples and demos pin older Vite ranges on purpose, since they are copy-pasteable consumer fixtures, so those declarations are left exactly as they are rather than rewritten to a catalog reference they could not resolve outside this workspace. All 51 Vite consumers resolve to 8.2.0.

Three plugin bumps came with it:

- `@vitejs/plugin-vue` 6.0.2 to 6.0.8 and `@vitejs/plugin-react` 5.x to 6.0.5, both of which capped at Vite 7. plugin-react 6 adds `@rolldown/plugin-babel` and `babel-plugin-react-compiler` as peers; both are optional and neither gets installed.
- `vite-plugin-node-polyfills` 0.25 to 0.28, worth calling out because the main SuperDoc build uses it directly.
- `vite-plugin-dts` deliberately stays at 4.5.4 (peer `vite: *`) so the declaration pipeline does not take a second major bump in the same change.

Verified on the cumulative branch: `build:superdoc`, `check:types`, `pnpm test` (1531 files, 22972 tests) and `check:public` (83 stages) all pass. Rolldown now arrives as `rolldown` 1.x plus `lightningcss`, matching the Vite 8 dependency change.

Two things left for follow-up rather than fixed here: six configs still use the compatibility-mapped `build.rollupOptions` name, and several configs warn that they are incompatible with the `native` config loader that Vite plans to default to later.

<a href="https://cubic.dev/pr/superdoc/docx-editor/pull/3873?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -76,11 +76,11 @@ catalogs:
       specifier: ^8.49.0
       version: 8.57.2
     '@vitejs/plugin-react':
-      specifier: ^5.1.1
-      version: 5.2.0
+      specifier: ^6.0.5
+      version: 6.0.5
     '@vitejs/plugin-vue':
-      specifier: 6.0.2
-      version: 6.0.2
+      specifier: 6.0.8
+      version: 6.0.8
     '@vitest/coverage-v8':
       specifier: ^3.2.4
       version: 3.2.4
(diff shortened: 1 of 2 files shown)


caio-pizzol:

PR #3874, opened 2026-08-05
test(vitest): upgrade Vitest 3.2 to 4.1

caio-pizzol:
Takes Vitest to 4.1.10, which the rest of the toolchain now wants: Vite 8 pairs with it, and `@angular/build` in the Angular example already demanded `vitest@^4.0.8`.

Most of the diff is one mechanical change. Vitest 4 stopped wrapping mock implementations in a constructable function, so any mock that production code calls with `new` had to move from an arrow to a `function`. That was applied to 58 sites via a codemod scoped to class-like mock targets only, so arrows stay everywhere a mock is merely called. It is kept in the first commit so it stays separable from the behavioural fixes in the second.

Four behavioural differences needed real fixes:

- Mock history is no longer cleared between retry attempts. Both packages set `retry: 2`, so call-count assertions accumulated across attempts and reported 2, then 3, then 4 calls for a single expected call. `clearMocks: true` keeps attempts independent.
- `restoreAllMocks` now only restores `vi.spyOn` spies, so a `mockReturnValue` set on a module mock survived into later tests and sent one list wrapper down the clone-on-write path. That default is re-established explicitly alongside the ones the suite already re-established.
- Test support files were leaking vitest types into declaration emit. Excluding `*.test.*` was never enough because helpers sit beside the tests they serve, so `@vitest/spy` reached inferred return types and left stale outputs behind. Both super-editor and superdoc needed it, since the superdoc project also includes `../super-editor/src`. Nothing in production imports from those directories.
- `vi.unmock` in shared-doc was paired with `vi.doMock`. Being hoisted, it ran before any test and never released those registrations, so `vi.doUnmock` is the correct counterpart. Vitest warned about this and will make it an error.

`environmentMatchGlobs` was removed outright, so the rules routing about 630 pure-logic tests to the node environment are gone. Only the two headless command suites depended on node for correctness, and they now declare it with a `@vitest-environment` docblock. What is left is the lost startup saving, which needs node and DOM projects; that is a topology change, since this package is itself a project of the root config, and is deliberately not bundled in here. The dead config block is removed rather than left as the record.

Verified from a clean detached checkout in this order: `pnpm install --frozen-lockfile`, `pnpm run build:superdoc`, `pnpm run check:types`, `pnpm test` (1531 files, 22972 tests, 0 failures), `pnpm run check:public` (83 stages, 0 failed). No vitest warnings remain.

<a href="https://cubic.dev/pr/superdoc/docx-editor/pull/3874?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ai/src/ai-actions/__tests__/core/ai-planner.test.ts
+++ b/packages/ai/src/ai-actions/__tests__/core/ai-planner.test.ts
@@ -20,7 +20,7 @@ const createActionInstance = () => ({
 });
 
 vi.mock('../../services/ai-actions-service', () => {
-  const factory = vi.fn(() => {
+  const factory = vi.fn(function () {
     const instance = createActionInstance();
     actionInstances.push(instance);
     return instance;
--- a/packages/collaboration-yjs/src/shared-doc/shared-doc.test.ts
+++ b/packages/collaboration-yjs/src/shared-doc/shared-doc.test.ts
@@ -269,9 +269,14 @@ beforeEach(() => {
 });
 
 afterEach(() => {
-  vi.unmock('../shared-doc/constants.js');
-  vi.unmock('../shared-doc/callback.js');
-  vi.unmock('../shared-doc/utils.js');
+  // AIDEV-NOTE: These modules are registered with `vi.doMock`, so they must be
+  // released with the runtime counterpart. `vi.unmock` is hoisted to the top of
+  // the module regardless of where it appears, so it ran before any test and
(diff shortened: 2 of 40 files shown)


caio-pizzol:

PR #3879, opened 2026-08-07
fix(super-editor): insert mentions on first click

### LINKED ISSUE #3878: Mention popover requires two clicks to insert a user
### What happened?

When selecting a user from the mention popover in the v1 editor, the first click on a user row does not insert the mention. The popover appears to remount during that first interaction. Clicking the same user a second time inserts the mention successfully.

**Expected behavior:** clicking a user once should immediately replace the typed `@mention` text with the selected mention node.

**Actual behavior:** the first click has no visible effect, and a second click is required.

### Steps to reproduce

1. Configure SuperDoc with at least one mentionable user (or open ).
2. Place the caret in an editable paragraph.
3. Type `@` to open the mention popover.
4. Click a user in the popover.
5. Observe that the mention is not inserted.
6. Click the same user again.
7. Observe that the mention is inserted on the second click.

### SuperDoc version

`1.46.0` / current `main` branch at `2a9df40bf`

### Browser

Chrome

### Additional context

https://github.com/user-attachments/assets/981afa8e-065a-40d3-95f2-1de7aa07f6b8

malizhev:
## Summary

- Prevent mention user rows from transferring focus or triggering the document-level selection handler on `mousedown`.
- Keep the existing mention popover mounted until its `click` handler inserts the selected user.
- Correct the internal `inserMention` prop typo to `insertMention` in the mention component and its caller.
- Add regression coverage for inserting a mention on the first click.

## Why

Clicking a user in the mention popover currently requires two clicks. The first `mousedown` reaches the document-level custom-selection handler, which reapplies the editor state. Because the popover is still marked for an update, this remounts `Mentions.vue` before the browser can dispatch `click` to the original row.

Closes #3878

<a href="https://cubic.dev/pr/superdoc/docx-editor/pull/3879?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/super-editor/src/editors/v1/components/popovers/Mentions.test.ts
+++ b/packages/super-editor/src/editors/v1/components/popovers/Mentions.test.ts
@@ -1,0 +1,38 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { mount } from '@vue/test-utils';
+import Mentions from './Mentions.vue';
+
+describe('Mentions.vue rendering', () => {
+  let wrapper: ReturnType<typeof mount> | undefined;
+  let documentMousedown: ReturnType<typeof vi.fn> | undefined;
+
+  afterEach(() => {
+    wrapper?.unmount();
+    if (documentMousedown) document.removeEventListener('mousedown', documentMousedown);
+  });
+
+  it('keeps the user row mounted through the first click', () => {
+    const user = { name: 'Alice', email: 'alice@example.com', role: 'editor' };
+    const insertMention = vi.fn();
+    documentMousedown = vi.fn();
+    document.addEventListener('mousedown', documentMousedown);
+
+    wrapper = mount(Mentions, {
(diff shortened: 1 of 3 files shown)


caio-pizzol:
Hey @malizhev, thanks for the fix! We reproduced the issue and confirmed that the implementation fixes first-click mention insertion.

I pushed a small follow-up to your branch so the regression test mounts into `document.body`. This ensures the test actually verifies that `mousedown` does not reach the document listener. Without it, the test stayed green even if `.stop` was removed.

The focused mention and popover tests are passing. Thanks again!

PR #3932, opened 2026-08-25
fix(toolbar): give each toolbar separator its own item

PopaSerban:
## Summary
- Fix a Vue "Duplicate keys found during update" warning on `ButtonGroup` by giving
  each separator in the built-in toolbar its own item, instead of reusing one
  shared item six times (`packages/superdoc/src/internal/toolbar/built-in/default-items.js`).
  UI layer only — no pipeline/document-data involvement.

Screenshot 2026-08-25 at 20 07 25

## Context
`ButtonGroup.vue` keys its `v-for` on `item.id.value`. `default-items.js` created a
single `separator` toolbar item and spliced that same object into the toolbar list
six times; since separators default to the `center` group, all six ended up in one
`ButtonGroup` sharing one `id`, so Vue re-render passes for that group patched a
list with six duplicate keys.

The warning only surfaces on a re-render after the initial mount (Vue's duplicate-key
check runs in `patchKeyedChildren`, not on first mount), which is why it doesn't show
up right after page load — it needs a `toolbar-state-change` event (e.g. clicking
into the document, changing selection, typing) to trigger the re-render.

Fix: `default-items.js` now calls `useToolbarItem(...)` fresh for each separator
occurrence, giving each its own `id`.

| File | Change | Risk |
|---|---|---|
| `packages/superdoc/src/internal/toolbar/built-in/default-items.js` | shared `separator` item → `makeSeparator()` factory, called at each of its 6 use sites | LOW — internal to the module, no exported signature change |

### DIFF
--- a/packages/superdoc/src/internal/toolbar/built-in/default-items-identity.test.js
+++ b/packages/superdoc/src/internal/toolbar/built-in/default-items-identity.test.js
@@ -1,0 +1,21 @@
+import { describe, expect, it } from 'vite-plus/test';
+import { makeDefaultItems } from './default-items.js';
+import { toolbarIcons } from './toolbarIcons.js';
+import { toolbarTexts } from './toolbarTexts.js';
+
+describe('built-in toolbar item identity', () => {
+  it('gives every rendered item a unique id', () => {
+    const { defaultItems } = makeDefaultItems({
+      superToolbar: { config: { mode: 'docx' }, ui: {} },
+      toolbarIcons,
+      toolbarTexts,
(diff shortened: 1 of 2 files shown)


caio-pizzol (approved):
@PopaSerban, thanks for the PR—this was a precise diagnosis and a clean fix. I added a focused regression test to the branch; it fails with the previous shared-separator implementation and passes with your change. The full SuperDoc package suite also passes locally with 2,493 tests.

Approving—thanks again for the contribution!

PR #3935, opened 2026-08-26
feat(page-numbering): support hebrew1 and hebrew2 page number formats

Nathaniel-260:
Relates to #3933. Sibling of #3934 — the two are independent and can be reviewed and merged in either order.

Two commits:

- `feat(page-numbering): support hebrew1 and hebrew2 page number formats`
- `feat(document-api): allow hebrew page numbering formats`

## What changed

`PageNumberFieldFormat['format']` had no Hebrew member, so a section carrying
`<w:pgNumType w:fmt="hebrew1"/>` fell through `formatPageNumber`'s `default:`
arm and rendered **Latin digits**. Note this is a different symptom from the
list-marker gap in #3934 — there an unmapped format produced an *empty* marker;
here it produces the *wrong number*.

The format is preserved in OOXML, but the shipped v2 runtime currently filters it out before the public formatter and `sections.list`. This PR adds the public formatter and API support. A companion DOCX Engine change is still required for rendered output and API reads; we implemented that internally and verified the pair against Word-authored fixtures. The `tests/consumer-typecheck` fixture also checks that both values are assignable from outside the package.

## The one judgement call

**Word does not wrap on this path — it errors.** Both paths stop representing
Hebrew numerals above 392, but where a list marker restarts from `א`, a `PAGE`
field substitutes a localized error string. Measured directly, with
`<w:pgNumType w:fmt="hebrew1" w:start="388"/>` and ten pages:

| page value | Word |
|---|---|
| 390 | `שצ` |
| 391 | `שצא` |
| 392 | `שצב` |
| **393** | **`שגיאה! אין אפשרות לייצג את המספר בתבנית שצוינה.`** |
| 394+ | same error |

`hebrew2` was measured the same way in a second section of the same document and
has the identical bound — `‏תתתתתתתתתתתתתתתתתצ` at 392, the same error string at
393 — so this is a property of Word's Hebrew converter, not of one format.

That string is written in whatever language the Word UI runs in, so reproducing it faithfully would mean guessing a locale. **This PR falls back to decimal above 392 instead.** We agree with that choice: it keeps the page readable, matches the existing `default:` behavior of `formatPageNumber`, and avoids inventing a locale-specific error message.

This is also why the page formatters deliberately do **not** reuse the
list-numbering formatters in #3934, despite identical glyph rules in range.

## Also verified, no change needed

`isDigitBucketCompatiblePageNumberFormat` in `layoutHeaderFooter` already
returns `false` for anything but `decimal`/`numberInDash`, so both Hebrew
formats correctly take the per-page layout path as-is.

`formatPageNumber` already clamps its input to `>= 1`, so unlike the list path
these formatters need no zero/negative guard.

## Pre-existing gap left alone

`PageNumberFieldFormat` carries `'ordinal'`, which `SectionPageNumberingFormat`
still does not. That predates this change; widening it here would be unrelated
scope.

## Checks

- `packages/layout-engine/contracts` — 15 tests pass (10 existing + 5 new)
- `@superdoc/document-api` — pass
- `pnpm run check:types` (`tsc -b`) — pass
- `pnpm run docapi:sync` regenerates cleanly; contract parity and contract
  output checks pass when run directly
- `vp lint`, `vp fmt --check` — clean

## End-to-end verification

We reproduced the decimal fallback and confirmed the missing filter lives in the DOCX Engine ahead of this public formatter. With this PR plus the companion engine change, two Microsoft Word-created fixtures render `Footer א` / `Footer ב` for `hebrew1` and the matching U+200F values for `hebrew2`. After an unrelated edit, export, and reopen, both files keep `w:pgNumType/@w:fmt` and still render correctly. Removing the engine change makes both fixtures fail back to `Footer 1` / `Footer 2`. The visible fix therefore needs both changes in the same DOCX Engine release.

<a href="https://cubic.dev/pr/superdoc/docx-editor/pull/3935?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/document-api/src/contract/contract.test.ts
+++ b/packages/document-api/src/contract/contract.test.ts
@@ -128,6 +128,26 @@ describe('document-api contract catalog', () => {
     expect(input.required).not.toContain('isLgl');
   });
 
+  it('publishes Hebrew page-numbering formats for section reads and writes', () => {
+    const schemas = buildInternalContractSchemas();
+    const formats = [
+      'decimal',
+      'lowerLetter',
+      'upperLetter',
+      'lowerRoman',
+      'upperRoman',
+      'numberInDash',
+      'hebrew1',
+      'hebrew2',
+    ];
+    const setInput = schemas.operations['sections.setPageNumbering'].input as ContractTestSchemaShape;
+    const listOutput = schemas.operations['sections.list'].output as ContractTestSchemaShape;
+    const listedFormat = listOutput.properties?.items?.items?.properties?.pageNumbering?.properties?.format;
+
+    expect(setInput.properties?.format?.enum).toEqual(formats);
+    expect(listedFormat?.enum).toEqual(formats);
+  });
+
   it('publishes review-aware blocks.list input and effective-mode output', () => {
     const schemas = buildInternalContractSchemas();
     const operation = schemas.operations['blocks.list'];
--- a/packages/document-api/src/contract/schemas.ts
+++ b/packages/document-api/src/contract/schemas.ts
@@ -1725,7 +1725,7 @@ const sectionHeaderFooterKindSchema: JsonSchema = { enum: ['header', 'footer'] }
 const sectionHeaderFooterVariantSchema: JsonSchema = { enum: ['default', 'first', 'even'] };
 const sectionLineNumberRestartSchema: JsonSchema = { enum: ['continuous', 'newPage', 'newSection'] };
 const sectionPageNumberFormatSchema: JsonSchema = {
-  enum: ['decimal', 'lowerLetter', 'upperLetter', 'lowerRoman', 'upperRoman', 'numberInDash'],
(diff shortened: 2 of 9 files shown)


caio-pizzol (approved):
Thanks for the detailed work here — approved! We verified this with two Microsoft Word-created fixtures, not just unit tests. With the companion DOCX Engine change, both formats render correctly in Chromium and survive an unrelated edit, export, and reopen without losing `w:fmt`. Removing that engine change reproduces the decimal fallback. We also agree that decimal is the safest fallback above 392. I updated `Fixes` to `Relates` so #3933 stays open until both numbering paths ship.

